PR #40, opened 2016-01-05
View / Mutable View support in rblas integration

bluss:
Only supporting c-order array views so far (f-order support was dropped from this PR). The concern with f-order is that the rblas library that we integrate with doesn't check the ordering for each used matrix (only one of them), so it's easy to make an error if matrices are mixed.

### DIFF
--- a/benches/bench1.rs
+++ b/benches/bench1.rs
@@ -415,9 +415,9 @@ fn bench_mat_mul_rblas_large(bench: &mut test::Bencher)
     bench.iter(|| {
         // C ← α AB + β C
         f32::gemm(&1.,
-                  Transpose::NoTrans, &a.blas(),
-                  Transpose::NoTrans, &b.blas(),
-                  &1., &mut c.blas());
+                  Transpose::NoTrans, &a.bv(),
+                  Transpose::NoTrans, &b.bv(),
+                  &1., &mut c.bvm());
     });
 }
 
--- a/src/blas.rs
+++ b/src/blas.rs
@@ -58,6 +58,7 @@ use self::rblas::{
 };
 use super::{
     ArrayBase,
+    ArrayView,
     ArrayViewMut,
     Ix, Ixs,
     ShapeError,
@@ -69,7 +70,6 @@ use super::{
 };
 
 
-/*
 /// ***Requires `features = "rblas"`***
 pub struct BlasArrayView<'a, A: 'a, D>(ArrayView<'a, A, D>);
 impl<'a, A, D: Copy> Copy for BlasArrayView<'a, A, D> { }
@@ -78,7 +78,6 @@ impl<'a, A, D: Clone> Clone for BlasArrayView<'a, A, D> {
         BlasArrayView(self.0.clone())
     }
 }
-*/
 
 /// ***Requires `features = "rblas"`***
 pub struct BlasArrayViewMut<'a, A: 'a, D>(ArrayViewMut<'a, A, D>);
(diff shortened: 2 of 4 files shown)


snd:
looking forward to the immutable blas views on arrays which are already continuous ! 

was about to open an issue asking for this as it will solve some issues for me.

bluss:
There are some challenges, and that may mean breaking changes down the line.

cblas API takes a rowmajor/colmajor option, but it is not per matrix but must be the same for them all. We can't enforce that without re-wrapping the blas operations.

blas does not support overlapping inputs! This is not easy to enforce in the type system. I.e. A times A  would be overlapping if A is the same array view.

snd:
i see !

> blas does not support overlapping inputs!
too bad. i wasn't aware.

> This is not easy to enforce in the type system. I.e. A times A would be overlapping if A is the same array view.

while not necessarily elegant the current mutable-only blas views enforce this as mutable borrows are exclusive.
maybe a reason to keep it that way in a documented way ?

bluss:
I am not sure where I read about overlapping read-only inputs. It does seem a bit strict when you think over it. That said, we can easily create an input that next to overlaps with the output (if the input = even rows of a matrix, output = odd rows of a matrix). Those are disjoint in current ndarray and have disjoint elements, but some blas implementation may consider them overlapping.

Deserves experimentation!

@snd What do you think about allowing this interface in the PR? It would allow both row major and column major arrays to be used without copying. However, the user must be careful not to try to multiply (for example) matrices with mismatching order.

snd:
i apologize for the delay. i've got a lot going on at the moment.

> I am not sure where I read about overlapping read-only inputs. 

i'd be interested to know.

> It does seem a bit strict when you think over it.

i agree.

i did some research. my current understanding is that it's ok for readonly/input arguments to alias one another. it is not allowed (undefined behaviour) for a readonly/input argument to alias a modifyable/output argument.

that would be ideal !

the exclusivity of mutable borrows enforces that an input argument can't also be an output argument.

> What do you think about allowing this interface in the PR?

i'd say: **go for it !** but add some tests to verify this.

> However, the user must be careful not to try to multiply (for example) matrices with mismatching order.

is there a reason why you can't add assertions for that ?

if not i'd add some assertions with helpful error messages.
instead of having to be careful i'd prefer it blowing asap with a helpful error message rather than getting some confusing blas error or (even worse) no error at all.

what do you think of modeling row-major and col-major via the type system ?
then it should be possible to verify matching order at compile time via the type system.
just an idea. 

> cblas API takes a rowmajor/colmajor option, but it is not per matrix but must be the same for them all.

am i correct in assuming that it must be the same for all matrixes involved in a single call to BLAS or must it be the same for all matrixes globally ?

here are my sources:

GSL which builds on BLAS:
> The library assumes that arrays, vectors and matrices passed as modifiable arguments are not aliased and do not overlap with each other
> If the arguments will not be modified (for example, if a function prototype declares them as const arguments) then overlapping or aliased memory regions can be safely used.

http://www.csse.uwa.edu.au/programming/gsl-1.0/gsl-ref_2.html

LAPACK which builds on BLAS:
> Unless specified otherwise, only input arguments (that is, scalars passed by values and arrays specified with the const qualifier) may be legally aliased on a call to C interface to LAPACK.

http://www.netlib.org/lapack/lapacke.html#_aliasing_of_arguments

gonum which builds on BLAS:
> It is forbidden for the modified data region of the receiver to overlap the used data area of the input arguments.

https://godoc.org/github.com/gonum/matrix/cmat128#hdr-Element_Aliasing

bluss:
Thank you for looking into this.

cblas functions take one layout (ordering) parameter per operation. For example for `cblas_dgemm` (matrix multiplication) it regards the two input matrices and the output matrix of that operation.

This interface was to interoperate with rblas specifically, and that's why I can't enforce that the inputs have the same ordering: rblas' operations don't enforce it.

In the long run we may want to add our own high level blas-based operations.

bluss:
If you're interested, a tricky question remains, imagine a 4 x 4 matrix in row major order. The columns of the matrix are technically both F-order and C-order since they are one dimensional (but with a leading dimension of 4).

Anyway, adjacent columns do not overlap, but their memory interleaves very closely. Is this compatible with blas? :smile:

bluss:
View support merged, without f-order support yet (since that remains to be clarified).

bluss:
Follow up issues:

- #78 
- #79

PR #51, opened 2016-01-14
Add row_mut and column_mut methods for matrices

vbarrielle:
They are simply mutable versions of `fn row` and `fn column`.


bluss:
Thank you!

Issue #22, opened 2015-12-21
Add PartialOrd impls


bluss:

spencerlyon2:
What struct(s) should this be implemented for? I would like to take a stab a this

bluss:
To start with, like `PartialEq` does it. Then possibly expand to mismatching dimension parameters and broadcasting later. I'd keep it simple first.

bluss:
Oh `PartialEq` simply says `false` for unequal shapes. We can't really translate that to PartialOrd (except using partiality, but that doesn't affect the comparison operators..). Maybe returning None in `partial_cmp` method for unequal sizes, and panicking in the lt, etc methods will be ok?

bluss:
See `src/arraytraits.rs` line 60.

spencerlyon2:
How could I return None in `partial_cmp`? 

Isn't the signature:

```rust
fn partial_cmp(&self, other: &Rhs) -> Option<Ordering>;
```

bluss:
Yes, `Option` admits `None`. Maybe you see better what makes sense while implementing it.

spencerlyon2:
Ahh, I get it. Sorry. First real experience with rust. Working on it now. Should have a PR ready soon.

spencerlyon2:
Ok I think I have something that makes some sense, but I think I could use some feedback.

```rust
impl<S, S2, D> PartialOrd<ArrayBase<S2, D>> for ArrayBase<S, D>
    where D: Dimension,
          S: Data,
          S2: Data<Elem = S::Elem>,
          S::Elem: PartialOrd,
{
    /// Return `None` if the array shapes `self` and `rhs` are not equal.
    /// Return appropriate `Option<Ordering>` otherwise.
    fn partial_cmp(&self, rhs: &ArrayBase<S2, D>) -> Option<Ordering>
    {
        // Balk on arguments with different shapes. Should probably
        // broadcast at some point
        if self.shape() != rhs.shape() {
            return None;
        }
        // Use partial_cmp from the slice version if contiguous
        if let Some(self_s) = self.as_slice() {
            if let Some(rhs_s) = rhs.as_slice() {
                return self_s.partial_cmp(rhs_s);
            }
        }

        let mut cmps = self.iter().zip(rhs.iter()).map(|(a, b)| a.partial_cmp(b));
        let first_cmp_opt = cmps.next();
        if let Some(first_cmp) = first_cmp_opt {
[...]

PR #52, opened 2016-01-14
WIP implementation of PartialOrd

spencerlyon2:
Before merging, I have a few things that I think should be resolved:

- [ ] Is the implementation of `partial_cmp` ok? I didn't really love having to make `cmps` by itself and then iterate over it, but it seemed to do the trick.
- [ ] I don't like all the repetition in the `lt`, `gt`, `le`, `ge` methods. They are identical except for which method I call on `self_s` and `a`. I think there must be some way (higher order function, macro) to reduce this bloat, but I don't know enough rust or have quite enough time to figure it out on my own.
- [ ] Docstrings and tests

### DIFF
--- a/src/arraytraits.rs
+++ b/src/arraytraits.rs
@@ -9,6 +9,8 @@ use std::ops::{
     IndexMut,
 };
 
+use std::cmp::Ordering;
+
 use super::{
     Dimension, Ix,
     Elements,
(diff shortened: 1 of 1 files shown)


bluss:
I have many questions around semantics, so I'll probably invite more people to discuss this. In the end, plain comparison operators are maybe not that important though.

For arithmetic ops we have the method zip_mut_with to collect the zipping logic in one place. We need something similar here with read only access and shortcircuiting.

Partialord semantics are clumsy but we should strive for both semantics and implementation that has good perf for float elements.

src/arraytraits.rs
```diff
@@ -79,6 +81,88 @@ impl<S, S2, D> PartialEq<ArrayBase<S2, D>> for ArrayBase<S, D>
     }
 }
 
+impl<S, S2, D> PartialOrd<ArrayBase<S2, D>> for ArrayBase<S, D>
+    where D: Dimension,
+          S: Data,
+          S2: Data<Elem = S::Elem>,
+          S::Elem: PartialOrd,
+{
+    /// Return `None` if the array shapes `self` and `rhs` are not equal.
```
bluss: Docs on trait impl methods unfortunately don't show. Rustdoc gives you one doc block per impl.. Put this on the line above `impl`

bluss: It could be the doc on partialeq is so old it's from the time when ruatdoc did show...

src/arraytraits.rs
```diff
@@ -79,6 +81,88 @@ impl<S, S2, D> PartialEq<ArrayBase<S2, D>> for ArrayBase<S, D>
...
+            if let Some(rhs_s) = rhs.as_slice() {
+                return self_s.partial_cmp(rhs_s);
+            }
+        }
+
+        let mut cmps = self.iter().zip(rhs.iter()).map(|(a, b)| a.partial_cmp(b));
+        if let Some(first_cmp) = cmps.next() {
+            if cmps.all(|x| x == first_cmp) {return first_cmp}
+        }
+        None
```
bluss: Sorry I didn't review this before. To make a lexicographical ordering, continue while each pair compares equal. If an element pair is non-equal, that's the result of the whole comparison. If you reach the end, result is Equal (since we already know the length is equal).

bluss:
I'm mostly worried over how the mismatching shape case should be handled. Can't find a good solution.

bluss:
I didn't expect to get a different perspective on what `a < b` should mean, but you're right, it could mean that every element of a is smaller than the corresponding element in `b`. (However, in that case, we can't use slice comparison methods at all).

aldanor:
In my experience, operations like "check if all elements are greater than other element" aren't really that useful for n-d arrays.

I pretty much concur with what `numpy` does:
```python
>>> bool(np.array([0]))
False

>>> bool(np.array([1]))
True

>>>bool(np.array([1, 1]))
---------------------------------------------------------------------------
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

bluss:
@aldanor What I've been looking at is an extension of lexical ordering for arrays, which is just like a &[T] compares under <, <=, >=, >. Do you see a way for that to make sense for ndarrays? I know numpy doesn't impl this either.

aldanor:
@bluss No standard method that I know off, really. If you're doing something like this, you're probably (arg) sorting the list of arrays, in which case the chances you want to sort some functor mapped over the list and not the raw elements, e.g. `argsort(map(trace, list_of_arrays)))`.

bluss:
Ok, it seems we don't know exactly what semantics for `PartialOrd` would make sense. That's not so strange, these issues show up once you try to put the idea to life. I'll put some thoughs back in the issue I think. #22

spencerlyon2:
Sorry I've been silent on this, but that sounds good to me.

bluss:
@spencerlyon2 Thank you for working with us! Suggestions on rules for comparisons are very welcome of course, but if it's not a good fit for ndarray then that's just how it is.

spencerlyon2:
No problem, I totally understand. There are some design/semantic behavior questions that need to be answered first. 

I was still able to learn how rust's trait and struct system works, so I got something out of this!

PR #57, opened 2016-01-15
Add an iter_axis function to get subviews along an axis

vbarrielle:
This enables for instance iterating along the columns of a matrix.

Right now the test coverage is basic, so I'll probably add more tests before it is mergeable.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1247,6 +1247,22 @@ impl<A, S, D> ArrayBase<S, D> where S: Data<Elem=A>, D: Dimension
         iterators::new_outer_iter_mut(self.view_mut())
     }
 
+    /// Return an iterator that traverses over the `axis` dimension
+    /// and yields each mutable subview.
+    ///
+    /// Iterator element is `ArrayViewMut<A, D::Smaller>`
+    /// (read-write array view).
+    ///
+    /// # Panics
+    ///
+    /// If axis is out of bounds.
+    pub fn axis_iter_mut(&mut self, axis: usize) -> OuterIterMut<A, D::Smaller>
+        where S: DataMut,
+              D: RemoveAxis,
+    {
+        iterators::new_axis_iter_mut(self.view_mut(), axis)
+    }
+
     // Return (length, stride) for diagonal
     fn diag_params(&self) -> (Ix, Ixs)
     {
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -1218,6 +1218,20 @@ impl<A, S, D> ArrayBase<S, D> where S: Data<Elem=A>, D: Dimension
...
     }
 
+    /// Return an iterator that traverses over the `axis` dimension
+    /// and yields each subview.
+    ///
+    /// For example, in a 2 × 2 × 3 array, with `axis` equal to 1,
+    /// the iterator element
+    /// is a 2 × 2 subview (and there are 3 in total).
+    ///
+    /// Iterator element is `ArrayView<A, D::Smaller>` (read-only array view).
```
bluss: Should say that it panics if `axis` is out of bounds

bluss:
Thanks. Tests look good enough, it's the same iterator struct. I think we can reuse `OuterIter` without penalty, seems natural in fact.

Can we rename this? `.subviews()` seems like a natural name? We also want the mutable version.

vbarrielle:
Sure subviews is a nice name. I'll change that when I'm back on my computer.

bluss:
`axis_iter` is fine too (same convention as the rest of the iterator methods I think). Methods shouldn't really have `iter`-names I think, pluralized names are better? I'm not sure..

vbarrielle:
So I renamed to `axis_iter`, looked like the best of the 3 possibilities. Ad I implemented the mutable iteration as well.

tests/iterators.rs
```diff
@@ -235,6 +252,31 @@ fn outer_iter_mut() {
...
+    let a = a.reshape((2, 3, 2));
+    // [[[0, 1],
+    //   [2, 3],
+    //   [4, 5]],
+    //  [[6, 7],
+    //   [8, 9],
+    //    ...
+    let mut a = a.to_owned();
+
+    for mut subview in a.axis_iter_mut(1) {
```
bluss: `mut subview`; it's sad quirks like this because we have no user-defined DSTs (views should be unsized like slices I think).

bluss:
Thank you! I'll fixup the panic info; I don't like using headings in module docs (it doesn't work in many ways).

PR #60, opened 2016-01-18
Implement owning constructors taking dimensions and strides

### LINKED ISSUE #59: Constructor with specific stride
In some cases it can be nice to be able to specify a specific stride when constructing a new array (for instance to wrap over foreign data).

vbarrielle:
Fixes #59.

This is a first draft of constructors taking strides into account. Tests are required, and I need to think it through a bit more to see if the checks are ok.

### DIFF
--- a/src/dimension.rs
+++ b/src/dimension.rs
@@ -80,16 +80,16 @@ pub fn can_index_slice<A, D: Dimension>(data: &[A],
             // offset is guaranteed to be positive so no issue converting
             // to usize here
             if (offset as usize) >= data.len() {
-                return Err(StrideError::OutOfBoundsStride);
+                return Err(StrideError::OutOfBounds);
             }
             if dim_stride_overlap(dim, strides) {
-                return Err(StrideError::OutOfBoundsStride);
+                return Err(StrideError::Aliasing);
             }
         }
         Ok(())
     }
     else {
-        Err(StrideError::NegativeStride)
+        Err(StrideError::Aliasing)
     }
 }
 
@@ -635,12 +635,36 @@ unsafe impl<'a> NdIndex for &'a [Ix] {
 
 #[cfg(test)]
 mod test {
-    use super::Dimension;
+    use super::{Dimension};
+    use stride_error::StrideError;
 
     #[test]
     fn fastest_varying_order() {
         let strides = (2, 8, 4, 1);
         let order = super::fastest_varying_order(&strides);
         assert_eq!(order.slice(), &[3, 0, 2, 1]);
+    }
+
+    #[test]
+    fn slice_indexing_uncommon_strides()
+    {
+        let v: Vec<_> = (0..12).collect();
+        let dim = (2, 3, 2);
+        let strides = (1, 2, 6);
+        assert!(super::can_index_slice(&v, &dim, &strides).is_ok());
+
+        let strides = (2, 4, 12);
+        assert_eq!(super::can_index_slice(&v, &dim, &strides),
+                   Err(StrideError::OutOfBounds));
+    }
+
+    #[test]
+    fn overlapping_strides_dim()
+    {
+        let dim = (2, 3, 2);
+        let strides = (5, 2, 1);
+        assert!(super::dim_stride_overlap(&dim, &strides));
+        let strides = (6, 2, 1);
+        assert!(!super::dim_stride_overlap(&dim, &strides));
     }
 }
--- a/src/dimension.rs
+++ b/src/dimension.rs
@@ -10,6 +10,38 @@ pub fn stride_offset(n: Ix, stride: Ix) -> isize
     (n as isize) * ((stride as Ixs) as isize)
 }
 
+/// Check whether `stride` is strictly positive
+#[inline]
+pub fn stride_is_positive(stride: Ix) -> bool
+{
+    (stride as Ixs) > 0
+}
+
+/// Check whether the given dimension and strides are memory safe
+/// to index the provided slice.
+///
+/// To be safe, no stride may be negative, and the 
+pub fn can_index_slice<A, D: Dimension>(data: &[A],
+                                        dim: &D,
+                                        strides: &D
+                                       ) -> bool
+{
+    if strides.slice().iter().cloned().all(stride_is_positive) {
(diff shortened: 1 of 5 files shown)


bluss:
Seems nice. We need a simpler way than this to create f-order arrays too, later. I was thinking you maybe know how to handle that best.

Can we use a `Result` based API here?

vbarrielle:
> We need a simpler way than this to create f-order arrays too, later. I was 
thinking you maybe know how to handle that best.

I think a nice API for this could be to have `fn arr1_f` and similar to create 
f-order arrays easily.

> Can we use a Result based API here?

Yes it's probably better.

I think my current tests may not be enough. Some strides may enable a same 
element to be referenced by different indices, and I don't think this should be 
allowed (probably could break some unsafe code). I'll try and add checking for 
that as well.

bluss:
Right, accessing the same element under multiple indices is the way broadcasting is implemented, and it can only be allowed for read only array views, so not for these constructors.

bluss:
Cool stuff happening here. Say something here if you want me to look at it.  I don't think we want to expose the helper functions in src/dimension.rs.

vbarrielle:
> I don't think we want to expose the helper functions in src/dimension.rs.

Yeah I wasn't sure about exposing those either. Once thing that made me think they had to be exposed was that it allowed documenting clearly what kind of strides are allowed.

Otherwise I think this is in a pretty good shape now, so you can review it. Depending on what you prefer I can un-expose the helper functions before a merge.

src/stride_error.rs
```diff
@@ -0,0 +1,32 @@
+use std::fmt;
+use std::error::Error;
+
+/// An error to describe invalid stride states
+#[derive(Clone, Debug, PartialEq)]
+pub enum StrideError {
+    /// stride leads to out of bounds indexing
+    OutOfBoundsStride,
```
bluss: Error looks good. I'd like to remove `Stride` from all the variant names.

bluss: Can we combine Aliasing and Negative? I don't know if we need the details.

bluss:
One thing for info: In the long run I hope to switch all dimension impls to be on fixed size arrays, so `[Ix; 2]` instead of `(Ix, Ix)` for example.

It would mean that in the future, arguments like dimension and stride will be provided as arrays. But I'm hopeful that we can bridge the transition with associated types and a trait, just like how `NdIndex` works today to accept `(i, j)` and `[i, j]` the same way.

One thing I'm thinking about is if we ever want to accept signed strides. Maybe not. Maybe that's mostly relevant "for raw parts"

bluss:
Let's skip exporting the helper functions. I want the least API liability :D

bluss:
Note to anyone following, we optimize for c-order arrays in arithmetic right now, but we will fix that down the line (and add fast cases for f-order too).

vbarrielle:
The latest changes rewrites the StrideError variants and no longer exposes the 
stride check methods at the API level.

vbarrielle:
> One thing I'm thinking about is if we ever want to accept signed strides. Maybe not. Maybe that's mostly relevant "for raw parts"

Negative strides look useful if you want to reverse iteration on some axis (at least that's what I use them for in numpy).
At the type level I would see some benefits having the strides as signed integers, it would make it impossible to use a `strides` data in place of a `dim`.

bluss:
The `Dimension` design that uses the signed strides in unsigned integers predates associated types! So it's def something that will be fixed at some point.

src/dimension.rs
```diff
@@ -10,6 +11,88 @@ pub fn stride_offset(n: Ix, stride: Ix) -> isize
...
+///
+/// Assumes that no stride value appears twice. This cannot yield the correct
+/// result the strides are not positive.
+fn fastest_varying_order<D: Dimension>(strides: &D) -> D
+{
+    let mut sorted = strides.clone();
+    sorted.slice_mut().sort();
+    let mut res = strides.clone();
+    for (ind, val) in strides.slice().iter().enumerate() {
+        // cannot binary search in unsorted...
```
bluss: It should probably just be a linear search. Linear search is faster than binary search anway, for slices surprisingly long, up to 32 or more elements.

bluss: I'd use `.unwrap()` since it's expected to never be None. No reason to insert a panic with formatting here, that's just awful in the generated code unfortunately.

Compare asm in release mode here: http://is.gd/UFzf2y

vbarrielle: Right, I originally thought `.expect()` would document the fact that it should never be `None` but it's better to not have dead formatting code.

bluss: It's an inspiration to finally fix this issue at least :smile: rust-lang/rust/pull/31116

src/dimension.rs
```diff
@@ -10,6 +11,89 @@ pub fn stride_offset(n: Ix, stride: Ix) -> isize
...
+{
+    let mut sorted = strides.clone();
+    sorted.slice_mut().sort();
+    let mut res = strides.clone();
+    for (ind, &val) in strides.slice().iter().enumerate() {
+        let sorted_ind = sorted.slice()
+                               .iter()
+                               .enumerate()
+                               .find(|&(_, &x)| x == val)
+                               .unwrap().0; // cannot panic by construction
```
bluss: This is actually shorter with the iterator method position.

bluss:
Ok everything looks good. I think we want to scrap the abbreviation "uchk", I want to remove that from ndarray (the other use is deprecated)!

No worries, at this point I can do any adjustments after merge as well.

I would probably rename StrideError::Aliased to something more generic too.

src/dimension.rs
vbarrielle: Oh nice I didn't know about that method.

vbarrielle:
Ok I'll leave the renames to you after the merge since you probably know better which names are more consistent with the rest of the API :)

vbarrielle:
Wow weird looks like the CI check failed because it could not download the rust toolchain... I don't know how to trigger another CI check.

bluss:
Thank you

Rust 1.7 is adding .sort_by_key() to slices, which can be used in the stride sort scenario, in the future.

PR #64, opened 2016-01-26
Adjustments for the stride constructors & overflow error handling

bluss:
Follow up on PR #59 

@vbarrielle I discovered another thing that must be checked generally, and that's wraparound when computing the number of elements in a shape.

It's a bit sad, requires adding error cases to all multidimensional constructors (eye, from_elem, zeros).

Old code would for example allow this, amusingly (in release mode, where we don't have debug assertions for this wrap around in the multiplication)

```rust
let five = OwnedArray::<f32, _>::zeros(5);
let _five_large = five.into_shape((3, 7, 29, 36760123, 823996703)).unwrap();
```

### DIFF
--- a/src/dimension.rs
+++ b/src/dimension.rs
@@ -13,7 +13,7 @@ pub fn stride_offset(n: Ix, stride: Ix) -> isize
 
 /// Check whether `stride` is strictly positive
 #[inline]
-pub fn stride_is_positive(stride: Ix) -> bool
+fn stride_is_positive(stride: Ix) -> bool
 {
     (stride as Ixs) > 0
 }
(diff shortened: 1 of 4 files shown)


vbarrielle:
Nice catch. It's a bit sad indeed to have error cases in basic constructors such as `zeros`, but I think having them panic on wraparound is a good behaviour, since there's not much the user could do in such cases.

src/dimension.rs
```diff
@@ -66,12 +66,13 @@ pub fn dim_stride_overlap<D: Dimension>(dim: &D, strides: &D) -> bool
...
 /// referenced by two different index.
-pub fn can_index_slice<A, D: Dimension>(data: &[A],
-                                        dim: &D,
-                                        strides: &D
-                                       ) -> Result<(), StrideError>
+pub fn can_index_slice<A, D: Dimension>(data: &[A], dim: &D, strides: &D)
+    -> Result<(), StrideError>
 {
     if strides.slice().iter().cloned().all(stride_is_positive) {
+        if dim.size_checked().is_none() || strides.size_checked().is_none() {
```
vbarrielle: I'm wondering if `strides.size_checked().is_none()` is a good check here, since strides have to be interpreted as isize.

Maybe a good check here would be to check if the product of a stride with its corresponding dimension does not overflow.

bluss: This is inside the all strides positive check.

Computing the stride product seems to be very wrong; I should probably check the (dimension - 1) * stride computation for overflow instead.

bluss:
Oh, I need to check all dimensions fit in `isize` actually.

PR #65, opened 2016-01-26
Iterate by chunks over specified axis

vbarrielle:
This relies on iterating using `OuterIterCore` while monitoring the location of the current index. When the iterator detects the last chunk, a different shape is used.

I'm not sure this is the proper design yet, it works for forward iteration but probably not for reversed iteration.

### DIFF
--- a/src/iterators.rs
+++ b/src/iterators.rs
@@ -659,8 +659,10 @@ pub struct ChunkIter<'a, A: 'a, D> {
 fn chunk_iter_parts<A, D: Dimension>(v: ArrayView<A, D>, axis: usize, size: usize)
     -> (OuterIterCore<A, D>, *mut A, D)
 {
-    let last_index = v.shape()[axis] / size;
-    let rem = v.shape()[axis] % size;
+    let axis_len = v.shape()[axis];
+    let size = if size > axis_len { axis_len } else { size };
+    let last_index = axis_len / size;
+    let rem = axis_len % size;
     let shape = if rem == 0 { last_index } else { last_index + 1 };
     let stride = v.strides()[axis] * size as isize;
 
@@ -670,8 +672,13 @@ fn chunk_iter_parts<A, D: Dimension>(v: ArrayView<A, D>, axis: usize, size: usiz
     let mut last_dim = v.dim.clone();
     last_dim.slice_mut()[axis] = if rem == 0 { size } else { rem };
 
-    let last_ptr = unsafe {
-        v.ptr.offset(stride * last_index as isize)
+    let last_ptr = if last_index < axis_len {
+        unsafe {
+            v.ptr.offset(stride * last_index as isize)
+        }
+    }
+    else {
+        v.ptr
     };
     let iter = OuterIterCore {
         index: 0,
--- a/src/iterators.rs
+++ b/src/iterators.rs
@@ -672,7 +672,7 @@ fn chunk_iter_parts<A, D: Dimension>(v: ArrayView<A, D>, axis: usize, size: usiz
     let mut last_dim = v.dim.clone();
     last_dim.slice_mut()[axis] = if rem == 0 { size } else { rem };
 
-    let last_ptr = if last_index < axis_len {
+    let last_ptr = if rem != 0 {
         unsafe {
             v.ptr.offset(stride * last_index as isize)
         }
--- a/src/iterators.rs
+++ b/src/iterators.rs
@@ -657,11 +657,10 @@ pub struct ChunkIter<'a, A: 'a, D> {
     life: PhantomData<&'a A>,
 }
 
-pub fn new_chunk_iter<A, D>(v: ArrayView<A, D>,
-                        axis: usize,
-                        size: usize,
-                        ) -> ChunkIter<A, D>
-    where D: Dimension
+fn chunk_iter_parts<A, D: Dimension>(v: ArrayView<A, D>,
+                                     axis: usize,
+                                     size: usize
+                                    ) -> (OuterIterCore<A, D>, usize, D)
 {
     let last_index = v.shape()[axis] / size;
     let rem = v.shape()[axis] % size;
@@ -683,6 +682,17 @@ pub fn new_chunk_iter<A, D>(v: ArrayView<A, D>,
         ptr: v.ptr,
     };
 
+    (iter, last_index, last_dim)
+}
+
+pub fn new_chunk_iter<A, D>(v: ArrayView<A, D>,
+                            axis: usize,
+                            size: usize,
+                            ) -> ChunkIter<A, D>
+    where D: Dimension
+{
+    let (iter, last_index, last_dim) = chunk_iter_parts(v.view(), axis, size);
+
     ChunkIter {
         iter: iter,
         last_index: last_index,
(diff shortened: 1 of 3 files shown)


vbarrielle:
Reversed iteration works with the last changes. I'll proof read that again later but it should be ok.

tests/iterators.rs
```diff
@@ -295,6 +295,55 @@ fn axis_iter_mut() {
 }
 
 #[test]
+fn axis_chunks_iter() {
+    let a = Array::from_iter(0..24);
+    let a = a.reshape((2, 6, 2));
+
+    let mut it = a.axis_chunks_iter(1, 2);
+    assert_eq!(it.next().unwrap(),
```
bluss: Using `assert_equal` would make these asserts nicer. See tests above.

src/iterators.rs
```diff
@@ -641,3 +641,157 @@ pub fn new_axis_iter_mut<A, D>(v: ArrayViewMut<A, D>,
...
+    iter: OuterIterCore<A, D>,
+    last_ptr: *mut A,
+    last_dim: D,
+    life: PhantomData<&'a A>,
+}
+
+fn chunk_iter_parts<A, D: Dimension>(v: ArrayView<A, D>,
+                                     axis: usize,
+                                     size: usize
+                                    ) -> (OuterIterCore<A, D>, *mut A, D)
```
bluss: I can't really handle these hanging `)` and returns. Sure, the style in the crate is not super consistent but it does not use these.

bluss:
Reverse iteration doesn't need to be a goal, only when it's cost free to do so.

src/iterators.rs
```diff
@@ -641,3 +641,157 @@ pub fn new_axis_iter_mut<A, D>(v: ArrayViewMut<A, D>,
...
+        { }
+    )
+}
+
+/// An iterator that traverses over the specified axis
+/// and yields mutable subviews of the specified size on this axis.
+///
+/// For example, in a 2 × 8 × 3 array, if the axis of iteration
+/// is 1 and the chunk size is 2, the yielded elements
+/// are 2 × 2 × 3 subviews (and there are 4 in total).
```
bluss: we can call them views instead of "subviews" since they have the same dimensionality as the original array.

src/iterators.rs
vbarrielle: Ok, I'll try not to use that style. Btw maybe there should be a rustfmt file for the crate at some point?

src/iterators.rs
bluss: Good example text otherwise! :smile:

src/iterators.rs
bluss: I've never been happy with rustfmt, maybe I should try with the newest version!

bluss:
Oh, my review question that I haven't managed to answer myself is, does the raw pointer of the iteration stay in bounds? We cannot have it point outside the array's allocation, even when we just use the last value to stop on.  I think it should be ok, by the construction of OuterIterCore.

vbarrielle:
Yes the raw pointer should always be in bounds because the stride and length given to OuterIterCore ultimately point to the last block and not any further. Thus the only additional risk is if the last shape is not correct.

src/lib.rs
```diff
@@ -1432,6 +1434,47 @@ impl<A, S, D> ArrayBase<S, D> where S: Data<Elem=A>, D: Dimension
...
+    ///
+    /// ```
+    /// use ndarray::Array;
+    ///
+    /// let a = Array::from_iter(0..28).reshape((2, 7, 2));
+    /// let mut iter = a.axis_chunks_iter(1, 2);
+    /// assert_eq!(2, iter.next().unwrap().shape()[1]);
+    /// assert_eq!(2, iter.next().unwrap().shape()[1]);
+    /// assert_eq!(2, iter.next().unwrap().shape()[1]);
+    /// assert_eq!(1, iter.next().unwrap().shape()[1]);
```
bluss: This example is unfortunately too cryptic. We can show just the first view in the iterator instead. Or some pretty ascii pictures... (I can add that if nothing else comes up)

vbarrielle:
The latest changes should address your remarks. Though for the style change I'm not sure whether the style I've used suits you.

bluss:
Beautiful!

I'm back at my review question. I think `last_ptr` can point outside of the array, which we can't do unfortunately.

Consider the case where `size` equals the length of the axis. last_ptr will point outside of the array.

For a regular slice we can point "one past the end" but no pointing further outside is legal (when using `ptr.offset`)
```
[a, b, c, d] x
 ^  ^  ^  ^  ^    valid pointer targets in and next to a slice
```

With a non-unitary stride, we easily land way past the end (or before it, if it's negative). We may not point to one row past the end.

bluss:
To clarify, it is UB to use `ptr.offset` to point at anything but the “valid pointer targets”, even if we don't deref the pointer.

bluss:
Doing that safely is a digression for another day (and when this is stable) http://doc.rust-lang.org/nightly/std/intrinsics/fn.arith_offset.html

vbarrielle:
> To clarify, it is UB to use ptr.offset to point at anything but the “valid pointer targets”, even if we don't deref the pointer.

I see, I didn't know it was UB, I thought only derefing was UB. Looks like I need to add some special case for when `size == self.dim[axis]`

bluss:
Additionally maybe clamp size to the length of the dimension if needed. Silently accepting a larger size is consistent with the uneven last view anyway.

vbarrielle:
Right that looks like a good behaviour.

vbarrielle:
I've thought about that offseting a bit more. With `size` clamper to the dimension length, there is an invariant `last_index * iterator_stride <= dim_len * dim_stride`, and the inequality is an equality when `size == dim_len`. So at worst the offset will point to right after the last allocated element if I'm correct.

This is however subject to the assumption that, for a dimension of length 1, the stride does not go out of bounds. It should be enforced by the strides checks but I'm unsure about this special case.

However we're better sure than sorry, so I've implemented the special handling for `size == dim_len`.

bluss:
That assumption does not hold for views of length 1. Consider `[0, 1, 2, 3]` sliced like `1..; 2`; it has elements `[1, 3], len = 2, stride = 2` the "one past the end" in stride * dim_len = 4 terms is outside of the valid range.

We don't know where the last allocated element is (stride may be negative). I don't see any other way but making sure the pointer always points at something we know is a valid element.

src/iterators.rs
```diff
@@ -641,3 +640,157 @@ pub fn new_axis_iter_mut<A, D>(v: ArrayViewMut<A, D>,
...
+
+    let mut inner_dim = v.dim.clone();
+    inner_dim.slice_mut()[axis] = size;
+
+    let mut last_dim = v.dim.clone();
+    last_dim.slice_mut()[axis] = if rem == 0 { size } else { rem };
+
+    let last_ptr = if last_index < axis_len {
+        unsafe {
+            v.ptr.offset(stride * last_index as isize)
```
bluss: I don't want to tire you out, but if `size == axis_len` here, we still compute an offset outside of the array, because `stride = axis_stride * size`.

I guess the condition `size < axis_len` would work well?

bluss:
Here's ~~a~~two demos, where last_ptr ends up pointing outside the range. Just to make it a bit more concrete.

```rust
    let a = Array::<f32, _>::zeros((8, 1));
    let sl = a.as_slice().unwrap();
    println!("pointers: {:p}, end={:p}", sl.as_ptr(),
        unsafe { sl.as_ptr().offset(sl.len() as isize) });
    let a = a.slice(s![..;-1,..]);
    let it = a.axis_chunks_iter(0, 8);


    let a = Array::<f32, _>::zeros((8, 2));
    let sl = a.as_slice().unwrap();
    println!("pointers: {:p}, end={:p}", sl.as_ptr(),
        unsafe { sl.as_ptr().offset(sl.len() as isize) });
    let a = a.slice(s![1..;2,..]);
    let it = a.axis_chunks_iter(0, 8);
```

src/iterators.rs
vbarrielle: Don't worry you won't tire me out, I want this to be safe too :)
It's probable `size < axis_len` works, I'll test that against the examples you provided later.

vbarrielle: Actually this won't work either, I think there's a possibility to index out of bounds if `rem == 0`. That's because any time the size divides the dimension length, `last_index` is not a "valid" index (ie `last_index * stride >= axis_len`).

So it seems the proper test is `rem != 0`.

vbarrielle:
I've added a better check to avoid offseting out of bounds. Checking `rem != 0` is the proper check imho. I've added more diverse test cases and can no longer witness an offseting going out of bounds.

bluss:
Yep I agree. Now it's good :smile:

bluss:
:fish_cake:

vbarrielle:
Cool, thanks for your help with this, I'm not yet used to the requirements of unsafe rust, so it was nice having you pointing out the flaws.

bluss:
Yep, no problem, it's great that we can work on it like this. Rust didn't give us the largest challenges in the world this time..

PR #70, opened 2016-01-30
Support for transposition

vbarrielle:
This PR adds support for having a transposed view (mutable or not) of an array. We probably also want to have a transposition with copy that doesn't change the storage order.

### DIFF
--- a/src/dimension.rs
+++ b/src/dimension.rs
@@ -210,18 +210,6 @@ pub unsafe trait Dimension : Clone + Eq {
         strides
     }
 
-    /// Return a Dimension with inverted values, useful for transposition
-    fn reverse(&self) -> Self {
-        let mut reversed = self.clone();
-        {
-            let iter = reversed.slice_mut().iter_mut().rev().zip(self.slice());
-            for (rev, &cur) in iter {
-                *rev = cur;
-            }
-        }
-        reversed
-    }
-
     #[inline]
     fn first_index(&self) -> Option<Self>
     {
--- a/src/dimension.rs
+++ b/src/dimension.rs
@@ -210,6 +210,18 @@ pub unsafe trait Dimension : Clone + Eq {
         strides
     }
 
+    /// Return a Dimension with inverted values, useful for transposition
+    fn reverse(&self) -> Self {
+        let mut reversed = self.clone();
+        {
+            let iter = reversed.slice_mut().iter_mut().rev().zip(self.slice());
+            for (rev, &cur) in iter {
+                *rev = cur;
+            }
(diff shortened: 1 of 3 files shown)


vbarrielle:
I've also added a version of transpose that consumes the array, returning an `OwnedArray`.
What do you think?

bluss:
Hi, a single method could serve all use cases, i.e. transpose views or ownedarray in place.

bluss:
.swap_axes already exists, but I guess it's only useful for 2d arrays. Not sure when you ever need transposing anything else, though.

vbarrielle:
> Hi, a single method could serve all use cases, i.e. transpose views or ownedarray in place.

I hadn't thought of that, so if someone wants a transposed view it would be `a.view().transpose()`?

In that case though the `transpose_into` method I made would need an additional `into_owned` method able to get out of Rc with copying if necessary.

> .swap_axes already exists, but I guess it's only useful for 2d arrays. Not sure when you ever need transposing anything else, though.

I had missed `swap_axes`, maybe it should have transposition in its doc then. But in my opinion transposition is useful in more cases than 2d arrays, a pattern I likeis to implement some method on C-order, then rely on some double transposition trick to have it work on F-order arrays too. I've used that pattern quite some times when interfacing python with C++ code.

aldanor:
Just a note: transposition for ndim > 2 could mean any permutation, not necessarily in reverse order: e.g. http://docs.scipy.org/doc/numpy-1.10.1/reference/generated/numpy.transpose.html

For 2d the only other permutation is the inverse, so there's that

vbarrielle:
I see, that would need a dedicated function, something like `fn permute_axes(self, perm: D) where D: Dimension`.

vbarrielle:
@bluss the new version has a single `transpose()` implementation.

bluss:
We can make this super simple, just reverse dimensions in place (`.slice_mut().reverse()`). Not sure we want the into owned method? Why is it needed?

vbarrielle:
> We can make this super simple, just reverse dimensions in place (.slice_mut().reverse()).

Neat. I'll do that.

> Not sure we want the into owned method? Why is it needed?

I thought the functionality might be useful. However I have no use case for it yet, so I'll remove it from this PR and it might come back if someday I find I need it :)

vbarrielle:
It's done. It makes a much simpler PR :)

bluss:
I'll change the name of the method - .reversed_axes, given the discussion about transposition being an even more general operation.

bluss:
Also, thanks, simple is always better..

PR #77, opened 2016-02-08
Allow reshape() to work on OwnedArray

SuperFluffy:
Relaxing the type constraints from `DataShared` to `DataClone` allows `reshape()` to be used on `OwnedArray` as well. Looking at the definitions of `DataClone for Rc<Vec<A>>`, this won't change anything for `Array`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 
 name = "ndarray"
-version = "0.4.0-alpha.1"
+version = "0.4.0-alpha.2"
 authors = ["bluss"]
 license = "MIT/Apache-2.0"
 


bluss:
Note that `into_shape()` already exists and works with owned arrays.

SuperFluffy:
I noticed that there is also `into_shape()`. Is there a reason why both are around? When I tried using `reshape()` this morning it was unexpected that it didn't work on my `OwnedArray`.

bluss:
`into_shape` is less flexible, but works with all arrays (views too). Since it does no copying, it can't rearrange everything.

This addition makes sense, it's just the panicking and error rules that strain.

- Reshape: Succeeds if new shape has the same number of elements as old shape. Clones when needed (always needed for OwnedArray, sometimes for Array).
- Into_shape: Requires same number of elements, as well as a contiguous layout (or reshapeable layout)

The simpler contract for reshape (must be the same number of elements) sort of allows it to use `panic` for error handling. It would fall under contract violation. It makes sense `into_shape` has more detailed error handling, since it has more complicated requirements to be able to succeed.

What do you think about the error handling aspects?

reshape's docs need to be updated to reflect that it clones all elements. It should note that in particular for `Array`, when the layout is compatible, it can avoid cloning.

In the end it's a bit unfortunate to have two methods. It's a result of the evolution (Array and reshape used to exist, into_shape was added to support views).

vbarrielle:
As far as I can tell, `reshape()` was there originally but was designed to work with Rc-based storage. It can clone the data if necessary. `into_shape()` is lighter as it will never clone the data, but that comes at the cost of possible errors. So I think there is a need for both.

Maybe there should be a "See also" section in the documentations of these methods to point at each other.

SuperFluffy:
It seems like there is indeed a lot of redundancy in here. In the case of `OwnedArray`, `reshape()` copying `data` does not make any sense. Also, I am not sure if it is obvious to the user that `reshape()` will yield another strong pointer to the data in the case of `Array`.

I think it would be good for the user if he needed to ensure contiguity of the data explicitly, leading to the following points:

* provide a `make_contiguous()` method;
* rename `into_shape()` to `reshape();
* never clone/copy any data;
* only change the fields `strides` and `dim`;
* maybe change `ShapeError` into `ContiguityError`, but keep the `Result<>` error handling of `into_shape()`.

bluss:
Copying the data makes sense, because that's what enables `reshape` to handle more input configurations successfully than what `into_shape` can do.

Still, a factorization into independent features is always attractive, for example `fn make_contiguous(self, layout: Layout) -> Self` which may clone.

I wonder if there's an algorithm for relayouting the elements in place, regardless of their input layout.

bluss:
I'm not sure how we should resolve .reshape(). Should we remove it?

vbarrielle:
How about having reshape return a view and panic on all situations that would require cloning, and adding a reshape_clone method that always clones? Thus a user would have to be sure about the layout to use reshape, but would have no cost. Somehow this proposal is a shortcut for `a.view().into_shape(shape).unwrap()`.

SuperFluffy:
I like @vbarrielle suggestion. I think we should keep `.reshape()` simply for the fact that I believe a consumer would expect a method of this name to be around.

PR #85, opened 2016-02-23
Rename Array to RcArray (old name is deprecated)

bluss:
This is clearer and makes it sound less like `Array` is the default type
to use.


bluss:
`Array` is a too nice name for the rc array type, that we don't want to seem like it is the default. Maybe we can have a trait named `Array` or something like that later.

Issue #81, opened 2016-02-10
Implement split_at for outer iter


bluss:

vbarrielle:
I'd like to try and implement this functionality.
Just to be clear, the signature would be somehting like:
``` rust
fn outer_split_at(&self, index: usize) -> (ArrayView<A, D>, ArrayView<A, D>)
```

I'm also wondering if the split is possible along other dimensions, which would call for a `fn axis_split_at(&self, axis: usize, index: usize)` method.

bluss:
That sounds interesting too, but for this issue I was thinking of split at for the outer iter itself. `OuterIter::split_at(usize) -> (OuterIter, OuterIter)`.

vbarrielle:
I see, that's interesting too. Well I'll have a look at that then :)

vbarrielle:
Should be closed as #86 got merged.

bluss:
oh right, thanks.

PR #86, opened 2016-02-23
split_at for OuterIter

vbarrielle:
This PR implements `split_at` for `OuterIter` (see #81). It's still a work in progress as there is one API point I'm not sure about. The current implementation panics if the split index is out of bounds, but there is an alternative: the returned left iterator could be the same as the original iterator, and the right iterator would immediately return `None`.

Also the method needs documentation but it'll wait until the precedent point is decided.

### DIFF
--- a/src/iterators.rs
+++ b/src/iterators.rs
@@ -487,10 +487,10 @@ pub struct OuterIter<'a, A: 'a, D> {
 impl<'a, A, D> OuterIter<'a, A, D>
     where D: Dimension
 {
-    /// Split the iterator at index
+    /// Split the iterator at index, yielding two disjoint iterators.
     ///
-    /// *panics* if `index > len`.
-    pub fn split_at(&self, index: Ix)
+    /// *panics* if `index` is strictly greater than the iterator's length
+    pub fn split_at(self, index: Ix)
         -> (OuterIter<'a, A, D>, OuterIter<'a, A, D>)
     {
         assert!(index <= self.iter.len);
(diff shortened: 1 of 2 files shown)


bluss:
OuterIterMut also needs split_at (and it's the primary one since it can not be implemented by the user, just like split_at_mut for a slice).

- Panic seems fine. Our role model is slice's split_at, split_at_mut
- API wise we can use `self` receiver and name `split_at` for both I think. Consistency is the best? Then we don't need to copy the dimensions twice (not sure if the compiler can improve the situation and optimize it out).

vbarrielle:
Indeed the API feels nicer when taking self by value. So it's implemented for `OuterIterMut`too now.

bluss:
Nice. I guess OuterIter already has a Clone impl. If not, we'll add it later.

vbarrielle:
There was no Clone impl on OuterIter, so I've just added one.

src/iterators.rs
```diff
@@ -482,6 +484,61 @@ pub struct OuterIter<'a, A: 'a, D> {
...
+    where D: Dimension
+{
+    /// Split the iterator at index, yielding two disjoint iterators.
+    ///
+    /// *panics* if `index` is strictly greater than the iterator's length
+    pub fn split_at(self, index: Ix)
+        -> (OuterIter<'a, A, D>, OuterIter<'a, A, D>)
+    {
+        assert!(index <= self.iter.len);
+        let right_ptr = unsafe { self.iter.offset(index) };
```
bluss: Ah how annoying, we end up with the same pointer past the end trouble here. Needs to handle the == len case differently.

vbarrielle: Right. I need to be more careful about that issue...

vbarrielle:
I've added special handling to prevent UB when `index == len`. I'll check it again tomorrow, and I'll probably use a macro to factor out the implementation (nearly forgot to fix the UB in the mutable case...).

bluss:
It looks all good to me.

vbarrielle:
I've just factored both implementations using a macro. If it's ok for you I think I'm done :)

bluss:
Thanks!

I'll change a final thing, just style. I think I prefer `life = self.life` for some reason, just to copy the old "value" along.

vbarrielle:
I had never thought of the `life = self.life` style, I like that too, it makes it more explicit what the resulting lifetime will be.

PR #94, opened 2016-02-28
split_at for Array

vbarrielle:
This PR introduces a `split_at`like family of functions for arrays. This is currently work in progress.

Status:
- [x] `axis_split_at`
- [x] `axis_split_at_mut`
- [x] tests

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -935,8 +935,7 @@ impl<'a, A, D> ArrayView<'a, A, D>
         let left_ptr = self.ptr;
         let right_ptr = if index == self.shape()[axis] {
             self.ptr
-        }
-        else {
+        } else {
             let mut indices = self.dim.clone();
             for (ax, ind) in indices.slice_mut().iter_mut().enumerate() {
                 if ax != axis {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1439,6 +1439,57 @@ impl<A, S, D> ArrayBase<S, D> where S: Data<Elem=A>, D: Dimension
         iterators::new_axis_iter_mut(self.view_mut(), axis)
     }
 
+    /// Split the array along `axis` and return one view strictly before the
+    /// split and one view after the split.
+    ///
+    /// **Panics** if `axis` is out of bounds.
+    pub fn axis_split_at(&self, axis: usize, index: Ix)
+        -> (ArrayView<A, D>, ArrayView<A, D>)
+    {
+        assert!(index <= self.shape()[axis]);
+        let left_ptr = self.ptr;
+        let right_ptr = if index == self.shape()[axis] {
+            self.ptr
+        }
+        else {
+            let mut indices = self.dim.clone();
+            for (ax, ind) in indices.slice_mut().iter_mut().enumerate() {
+                if ax != axis {
+                    *ind = 0;
+                }
+                else {
+                    *ind = index;
+                }
+            }
+            let offset = self.dim.stride_offset_checked(&self.strides,
+                                                        &indices).unwrap();
+            unsafe {
+                self.ptr.offset(offset)
+            }
+        };
+
+        let mut dim_left = self.dim.clone();
+        dim_left.slice_mut()[axis] = index;
+        let left = ArrayView {
+            data: ViewRepr::new(),
+            ptr: left_ptr,
+            dim: dim_left,
+            strides: self.strides.clone()
+        };
+
+        let mut dim_right = self.dim.clone();
+        dim_right.slice_mut()[axis] = self.dim.slice()[axis] - index;
+        let right = ArrayView {
+            data: ViewRepr::new(),
+            ptr: right_ptr,
+            dim: dim_right,
+            strides: self.strides.clone()
+        };
+
+        (left, right)
+    }
+
+
     /// Return an iterator that traverses over `axis` by chunks of `size`,
     /// yielding non-overlapping views along that axis.
     ///
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -94,6 +94,8 @@ pub use dimension::{
     RemoveAxis,
 };
 
+use dimension::stride_offset;
+
 pub use dimension::NdIndex;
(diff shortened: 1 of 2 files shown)


bluss:
I think I prefer if we only provide axis split (not outer split, since it's a simple special case, isn't it?)

src/lib.rs
```diff
@@ -1395,6 +1395,57 @@ impl<A, S, D> ArrayBase<S, D> where S: Data<Elem=A>, D: Dimension
...
+            let offset = self.dim.stride_offset_checked(&self.strides,
+                                                        &indices).unwrap();
+            unsafe {
+                self.ptr.offset(offset)
+            }
+        };
+
+        let mut dim_left = self.dim.clone();
+        dim_left.slice_mut()[axis] = index;
+        let left = ArrayView {
```
bluss: You can use `ArrayView::new_` here.

vbarrielle:
Indeed it's just a special case, but I was wondering if we didn't want to have some symmetry with the `outer_iter`, `axis_iter` methods?

src/lib.rs
vbarrielle: right, that's less boilerplate

bluss:
The most flexible API is not the easiest to use. Again a view based `self -> (Self, Self)` function may be the easiest.

Consider a function like this:

`fn divide_and_conquer<'a, A>(array: ArrayView<'a, A, Ix>) { }`

The only way to pass on a view with the same lifetime `'a` is the `self` based methods I think. We want to be able to split an array view without tying the split result to the stack frame where the split happens.

This is the same way `.into_iter()` for an array view works, it is lifetime preserving unlike `.iter()`. So if you examine this problem, you find that maybe you want `into_inner_iter()` etc methods for array views too. (Lifetime-preserving iterator constructors). It's all quite complicated.

bluss:
Consistency is good, but we need to stem the rising tide of a million methods too. Maybe I shouldn't have added `OuterIter` either.

vbarrielle:
Yes I see your point with the lifetimes issue. I wanted to mimick the slice's API but I guess there is some special treatment to slices here.

One (probably ugly) way I found to circumvent that problem in other crates (sprs) was to have additional methods defined only on views which would return views based on the lifetime of the data, not the lifetime of self in the method call. I wish there were some sort of "lifetime specialization" to overcome this issue. Not sure I'm really clear here, my thoughts are a bit blurry and I think I lack the proper vocabulary to express them.

bluss:
Yes! Slices are dsts, so they implement methods on the type `[T]`. If you look closely, you'll find lots of traces of this issue where it would be cool if an `ArrayView` could be a custom DST like that.

I think that `ArrayView::axis_split_at(self)` and `ArrayViewMut::axis_split_at(self)` is what we want here. We have to solve view methods not showing up in the docs, though.

vbarrielle:
> We have to solve view methods not showing up in the docs, though.

If I remember correctly, I've had a similar issue in sprs, and the solution was to define the impl block on the full type, not the alias type. That was quite some time ago but it might be worth a try.

bluss:
Yes, that will work

bluss:
The for loop doesn't seem right.

Is this right: consider a 3 x 4 x 5 array (3 pages, 4 rows, 5 columns)

Split on axis = 2, index = 2 should produce two views of shapes
3 x 4 x 2 and 3 x 4 x 3

vbarrielle:
I'm not sure I got your point here, that for loop is only supposed to help find the correct pointer offset, but the resulting shapes should be what you said.
However it's getting late here so it's quite probable I'm too sleepy to see the issue.

bluss:
I'll review it more carefully. Some pen and paper doodling (that's when you know you're actually thinking..).

vbarrielle:
I've switched to the view-by-value pattern and added some tests. I'll probably add a macro to factor out the code once again, and I need to test more complicated cases.

bluss:
Awesome

src/lib.rs
```diff
@@ -879,6 +879,51 @@ impl<'a, A, D> ArrayView<'a, A, D>
...
+            let mut indices = self.dim.clone();
+            for (ax, ind) in indices.slice_mut().iter_mut().enumerate() {
+                if ax != axis {
+                    *ind = 0;
+                }
+                else {
+                    *ind = index;
+                }
+            }
+            let offset = self.dim.stride_offset_checked(&self.strides,
```
bluss: There's the `stride_offset` method which does this without bounds checking.

src/lib.rs
```diff
@@ -879,6 +879,51 @@ impl<'a, A, D> ArrayView<'a, A, D>
...
+        let right_ptr = if index == self.shape()[axis] {
+            self.ptr
+        }
+        else {
+            let mut indices = self.dim.clone();
+            for (ax, ind) in indices.slice_mut().iter_mut().enumerate() {
+                if ax != axis {
+                    *ind = 0;
+                }
+                else {
```
bluss: Can you put the `else` on the same line as the `}`? Thanks.

src/lib.rs
bluss: On a closer look, we can just skip the whole for loop, can't we? Using stride * index directly (function `stride_offset` in dimension for example).

vbarrielle:
I've fixed the formatting issue, replaced the for loop by an appropriate call to `dimension::stride_offset`, and added more tests.

I'm no longer sure I should try and factor both impls in a macro, they're quite short now. As you wish :)

bluss:
Awesome, this is good. I'll merge the Axis PR too.

bluss:
Before I release this, what do you think of renaming it?

`pub fn split_at(self, axis: Axis, index: Ix)`

versus

`pub fn split_at(self, axis: Axis, index: Ix)`

I bring this up since `split_at` seems nice an unambiguous too.

vbarrielle:
Yes `split_at` seems nice and unambiguous, I don't think we could ever have a split_at taht does not require an axis argument.

bluss:
Done.

PR #96, opened 2016-02-28
Not going to be merged: Statically checked axis numbers

bluss:
This approach is posted here to show it, but this will not be merged.

It uses `Axis(usize)` for dynamic axis indexing, and statically checked `Axis0, Axis1` etc that can be checked against the array dimension using the type system.

This allows some degree of static checking. However, I think this system does not carry its own costs -- it introduces more generics for very little benefit. Instead I will post a simpler non generic approach.

### DIFF
--- a/examples/axis.rs
+++ b/examples/axis.rs
@@ -2,22 +2,15 @@ extern crate ndarray;
 
 use ndarray::{
     OwnedArray,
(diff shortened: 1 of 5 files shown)


bluss:
Superseded by #97

PR #97, opened 2016-02-28
Strongly typed Axis argument (newtype called Axis)

bluss:
Use a newtype `Axis(usize)` for all method arguments that need an axis.

### DIFF
--- a/benches/bench1.rs
+++ b/benches/bench1.rs
@@ -12,6 +12,7 @@ use rblas::matrix::Matrix;
 
 use ndarray::{
     OwnedArray,
+    Axis,
 };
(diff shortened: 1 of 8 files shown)


bluss:
See #96 for an alternative approach, and explanation why the generic version will not be used.

cc @vbarrielle Do you think this is ok? It's clearly more to type, but it's not possible to confuse the axis and index arguments anymore.

vbarrielle:
Yes I think this is nice. It really removes some possible bugs and it's not that much to type. I also feel it makes the code more readable.

I agree with you that the approach in #96 is a bit too much generics.

Issue #21, opened 2015-12-21
Add `From` impls


bluss:

bluss:
`From<&[A]> for ArrayView` isn't actually easy to use at all, so slice → ArrayView impls have not been done.

```rust
<anon>:7:9: 7:24 error: the trait `core::convert::From<&[_; 4]>` is not implemented for the type `ndarray::ArrayBase<ndarray::ViewRepr<&_>, _>` [E0277]
<anon>:7         ArrayView::from(&[1., 2., 3., 4.]).into_shape((2, 2)).unwrap()
                 ^~~~~~~~~~~~~~~
<anon>:7:9: 7:24 help: the following implementations were found:
<anon>:7:9: 7:24 help:   <ndarray::ArrayBase<ndarray::ViewRepr<&'a A>, usize> as core::convert::From<&'a [A]>>
<anon>:7:9: 7:24 help:   <ndarray::ArrayBase<ndarray::ViewRepr<&'a mut A>, usize> as core::convert::From<&'a mut [A]>>
<anon>:7:9: 7:24 note: required by `core::convert::From::from`
error: aborting due to previous error(s)
```

PR #98, opened 2016-02-29
Add ArrayView::from_slice(&[T]) and the same for ArrayViewMut

bluss:


bluss:
See #21 for a comment about From (trait) impls.

PR #113, opened 2016-03-02
Enable indexing Dimension by Axis

### LINKED ISSUE #111: Implementing Index<Axis> for Dimension
This would allow using `a.dim()[Axis(0)]`, which is not that much of a win. However, when implementing a method that takes an Axis as an argument, this could be a real win:
`a.dim().slice()[axis.axis()]` vs `a.dim()[axis]`.

What do you think?

vbarrielle:
Fixes #111 

Due to coherence rules I've had to use a macro. Looks like it works fine.

### DIFF
--- a/src/dimension.rs
+++ b/src/dimension.rs
@@ -327,6 +327,18 @@ pub unsafe trait Dimension : Clone + Eq + Debug {
         }
         offset
     }
+
+    #[doc(hidden)]
+    #[inline]
+    fn index(&self, axis: Axis) -> &Ix {
+        &self.slice()[axis.axis()]
+    }
+
+    #[doc(hidden)]
+    #[inline]
+    fn index_mut(&mut self, axis: Axis) -> &mut Ix {
+        &mut self.slice_mut()[axis.axis()]
+    }
 }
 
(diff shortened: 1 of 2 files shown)


bluss:
It seems this is not enough for it to be useful with just `D: Dimension`? Maybe there is another way, so that we don't have to make this interface public. My idea is that it should not be a part of ndarrays external API.

vbarrielle:
Yes avoiding to expose this too much would be nice, but I don't know if it's possible. Right now I have no idea how to restrict this.

bluss:
We can add a new method to `Dimension` (or two).

vbarrielle:
Right, simply being able to use `a.dim().index(axis)` would be nice.

vbarrielle:
The methods are added to the trait now. I haven't added doc for them yet, but I will.

bluss:
Looks good. This gives me food for thought regarding the issue of hiding these implementation details. `Dimension` is good because we can use it to specialize each method while also providing a general implementation. But that's a larger question outside of this PR.

vbarrielle:
I added some documentation on these new members.

bluss:
Thank you!

vbarrielle:
You're welcome.

PR #114, opened 2016-03-03
Add a trait `NdFloat`

### LINKED ISSUE #112: Add a float trait based on `Float`
The idea is to add a trait for just f32, f64 that allows access to all relevant ndarray API, as long as the element has that trait.

We now have Any and ScalarOperand involved in various bounds for technical reasons, so the trait would for example include those.

bluss:
The intention is to make it easy to work with f32 or f64, and have access to all ndarray operations.

However — the scalar operations on the left hand side are not cooperating with this :disappointed:

Closes #112 

### Is `NdFloat` needed?

It's not needed to use ndarray, but it makes it easier for the case where you only care about f32, f64.

### Why do we need a trait more than `num::Float`?

It would be the easiest if no new trait bounds were needed.

- `Any` is needed for type-based specialization (needed for future transparent blas support)
- `ScalarOperand` is needed as a marker trait so that arithmetic operators `Array + k` where `k` is a scalar can be implemented. I suspect this is not strictly needed, but limitations in Rust's current trait matching seem to require it.

With these two already being needed, we can take the opportunity to add-on traits that we want to have accessible in float-generic code, like the += operators and formatting traits.

### Why is `Array + scalar` included but `scalar + Array` is not?

*Left hand side* scalar operands seem entirely unreachable from generics. It's unfortunate, but not much we can do in rust as it is now. f32 + Array is implemented, it's just that we can't really capture that in `NdFloat`.

### DIFF
--- a/benches/bench1.rs
+++ b/benches/bench1.rs
@@ -1,6 +1,5 @@
 #![feature(test)]
 #![allow(unused_imports)]
-#![cfg_attr(feature = "assign_ops", feature(augmented_assignments))]
 
 extern crate test;
 #[macro_use(s)]
@@ -246,7 +245,6 @@ fn add_2d_regular(bench: &mut test::Bencher)
     });
 }
 
-#[cfg(feature = "assign_ops")]
 #[bench]
 fn add_2d_assign_ops(bench: &mut test::Bencher)
 {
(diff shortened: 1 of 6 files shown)


bluss:
cc @palango @daniel-vainsencher Maybe you want to comment on the trait or just like the information.

Trait code is here: https://github.com/bluss/rust-ndarray/pull/114/files#diff-a351598b40b2c7f2d849232cc86f0784R68

bluss:
Feedback is still welcome!

PR #115, opened 2016-03-03
Updated serde serialization to `0.7`

SuperFluffy:
No deserialization yet. But the output looks like the one produced by `rustc-serialize`. Check it with:

````
% cargo test serial --features "serialize_serde rustc-serialize" -- --nocapture
````

Also, I have a spurious commit due to updating my fork using github's pull requests. No idea how to fix that. :(

### DIFF
--- a/src/arrayserialize.rs
+++ b/src/arrayserialize.rs
@@ -6,62 +6,73 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 use serde::ser::impls::SeqIteratorVisitor;
-use serde::{self, Serialize, Deserialize};
+use serde::{self, Serialize};
 
-use super::{
-    Array,
-    Dimension,
-    Ix,
-    Elements,
-};
+use imp_prelude::*;
 
-struct AVisitor<'a, A: 'a, D: 'a> {
-    arr: &'a RcArray<A, D>,
+use super::arraytraits::ARRAY_FORMAT_VERSION;
+use super::Elements;
+
+struct AVisitor<'a, D: 'a, S: 'a>
+    where S: Data
+{
+    arr: &'a ArrayBase<S, D>,
     state: u32,
 }
 
-impl<A: Serialize, D: Serialize> Serialize for RcArray<A, D>
-    where D: Dimension
-{
-    fn serialize<S>(&self, serializer: &mut S) -> Result<(), S::Error>
-        where S: serde::Serializer
-    {
-        serializer.visit_named_map("Array",
-                                   AVisitor {
-                                       arr: self,
-                                       state: 0,
-                                   })
-    }
-}
-
-impl<'a, A: Serialize, D: Serialize> Serialize for Elements<'a, A, D>
-    where D: Dimension
-{
-    fn serialize<S>(&self, serializer: &mut S) -> Result<(), S::Error>
-        where S: serde::Serializer
-    {
-        serializer.visit_seq(SeqIteratorVisitor::new(self.clone(), None))
-    }
-}
-
-impl<'a, A, D> serde::ser::MapVisitor for AVisitor<'a, A, D>
+impl<A, D, S> Serialize for ArrayBase<S, D>
     where A: Serialize,
-          D: Serialize + Dimension
+          D: Dimension + Serialize,
+          S: DataOwned<Elem = A>
+
 {
-    fn visit<S>(&mut self, serializer: &mut S) -> Result<Option<()>, S::Error>
+    fn serialize<Se>(&self, serializer: &mut Se) -> Result<(), Se::Error>
+        where Se: serde::Serializer
+    {
+        serializer.serialize_struct("Array",
+            AVisitor {
+                arr: self,
+                state: 0,
+        })
+    }
+}
+
+impl<'a, A, D> Serialize for Elements<'a, A, D>
+    where A: Serialize,
+          D: Dimension + Serialize
+{
+    fn serialize<S>(&self, serializer: &mut S) -> Result<(), S::Error>
         where S: serde::Serializer
+    {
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by SuperFluffy** [comment]
I should note that the `"Array"` key is not written (at least in `serde_json`) for this definition:

````
serializer.serialize_struct("Array",
    AVisitor {
        arr: self,
        state: 0,
})
````

I am not really sure why the key is dropped.

**Comment by bluss** [comment]
You have a stray merge in the commit history. It needs to be removed. You can for example rebase your branch on top of ndarray's master.

File: src/arrayserialize.rs [hunk 1]
```diff
@@ -46,13 +52,17 @@ impl<'a, A, D> serde::ser::MapVisitor for AVisitor<'a, A, D>
...
+            },
             1 => {
                 self.state += 1;
-                Ok(Some(try!(serializer.visit_map_elt("data", self.arr.iter()))))
-            }
+                Ok(Some(try!(serializer.serialize_map_elt("dim", self.arr.dim()))))
+            },
+            2 => {
+                self.state += 1;
+                Ok(Some(try!(serializer.serialize_map_elt("data", self.arr.iter()))))
```
bluss: I think there's a specific method to serialize a struct field, can we use that?

bluss:
It looks like `serde_json` does not need to be a dependency.

src/arrayserialize.rs
SuperFluffy: Are you thinking of `serialize_struct_elt`? Replacing `serializer.serialize_map_elt("data", self.arr.iter()` with `serializer.serialize_struct_elt("data", self.arr)` leads to:

````
thread 'serial_serde' has overflowed its stack
fatal runtime error: stack overflow
Process didn't exit successfully: `/home/janis/github/rust-ndarray/target/debug/serialize-0fbcfd722bad2949 serial --nocapture` (signal: 4, SIGILL: illegal instruction)
````

Comment (bluss) [inline src/arrayserialize.rs]: Ah, you don't want `self.arr` there but `self.arr.iter()`.

Comment (SuperFluffy) [inline src/arrayserialize.rs]: Ok, it works as a drop-in replacement, the results are the same. Truth be told, I don't know what the semantically version would be. Pushed a commit for using the `serialize_struct_elt` version.

**Comment by bluss** [comment]
Thanks for working on this. We need to support serialize for all array types, and deserialize for `DataOwned`-kinds of arrays I guess. I'd like to remove the dependency on serde_json; we shouldn't pull this in when it is not needed.

File: src/arraytraits.rs [hunk 2]
```diff
@@ -207,10 +207,9 @@ unsafe impl<S, D> Send for ArrayBase<S, D>
     where S: Send + Data, D: Send
 { }
 
-
-#[cfg(feature = "rustc-serialize")]
+#[cfg(any(feature = "rustc-serialize", feature = "serde"))]
 // Use version number so we can add a packed format later.
-static ARRAY_FORMAT_VERSION: u8 = 1u8;
+pub static ARRAY_FORMAT_VERSION: u8 = 1u8;
```
bluss: This should be a `const` now. I don't think const existed when the old code was first written.

bluss:
Ok this is good. We need to implement Serialize for all `ArrayBase` kinds (views too). Then it needs a documentation string like rustc-serialize has that mentions the crate feature.

It's ok to merge with just Serialization, but I'd prefer if Deserialization existed too.

The question is how we should handle serde instability. User that depend on us will break if we upgrade serde, and the reverse..

To solve the tests, we can make a separate internal crate for that. That's easier to configure to pull in crates only needed for testing.

SuperFluffy:
Can we hide the `serde` serialization/deserialization behind a feature gate and make it work with nightly only? Next to keeping it as a feature one has to exlicitly opt into with `--features "serde"` I think this will be sufficient. Add a warning to the docs too.

It also just occured to me that we might want to consider actually integrating a `c` or `fortran` order flag now. Can this be done with the `ARRAY_FORMAT_VERSION`?

I will work on this towards the end of next week. Got to wrap up a few other projects.

bluss:
np. Yes maybe we should support a different memory order, just to serialize faster. Right now it's using the logical order of the elements, so it's agnostic.

SuperFluffy:
Are we rather interested in a simple but straightforward serialization, writing the underlying `Vec` directly into a list and providing the shape information with an extra field? Or do we prefer to encode the shape by storing the `Vec` as nested lists? Compare to how one could do it in `numpy`:

````python
In [0]: import json
In [1]: import numpy as np
In [2]: a = np.arange(12).reshape(2,3,2)

In [3]: a
Out[3]: 
array([[[ 0,  1],
        [ 2,  3],
        [ 4,  5]],

       [[ 6,  7],
        [ 8,  9],
        [10, 11]]])

In [4]: b = a.tolist()

In [5]: json.dumps(b)
Out[5]: '[[[0, 1], [2, 3], [4, 5]], [[6, 7], [8, 9], [10, 11]]]'
````

bluss:
Shape up front is useful for several reasons, so we want that. I'd stick with the existing format. I don't see too much value in the nested format really? I guess it's valuable for pretty small data structures.

SuperFluffy:
Yeah, I was thinking the same. It's great for eyeballing some data, but you don't want to do that once it gets too large.

bluss:
Json isn't even the most sensible target, ideally the user uses something like bincode I think.

SuperFluffy:
Ok, looks like serialization and deserialization on `ArrayBase` is working now. All round-trip tests, where I have adapted those from `rustc-serialize`, are working now. The error handling is also similar to the one you use for `rustc-serialize`.

Please have a look at it and tell me what you think. I would be quite excited if this got merged.

PR #116, opened 2016-03-03
Add `From` impls and convenience trait `AsArray` that makes it easy to receive array views

### LINKED ISSUE #21: Add `From` impls

bluss:
We solve the array input issue for `ArrayView::from` (see #21) by using `AsRef<[A]>`, which works well. The only real issue with AsRef is that `ArrayView::from` will accept strings and create a view from `&[u8]` based on that. Let's laugh and forget about that..

This replaces `ArrayView::from_slice` with just `ArrayView::from` instead.

Also add `AsArray` which is a simple restriction of the `Into<ArrayView<..>>` bound.

Fixes #21

### DIFF
--- a/src/arraytraits.rs
+++ b/src/arraytraits.rs
@@ -275,3 +275,38 @@ impl<A, S, D> Decodable for ArrayBase<S, D>
         })
     }
 }
+
+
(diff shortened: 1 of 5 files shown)


bluss:
@daniel-vainsencher This is since you asked about From/Into conversion. I think the `AsArray` example is as nice as it can be made.

PR #124, opened 2016-03-05
Implement an `arange` and `arange_step`constructors

palango:
I added `arange` and and `arange_step`, where `arange` is a specialisation with unit step length.
If you like it I can refactor a lot of internal uses of `linspace`.

There are some numeric problems in `arange_step`, but they exist equally in numpy:
```
In [1]: import numpy as np

In [2]: np.arange(1.0, 2.2, 0.1)
Out[2]:
array([ 1. ,  1.1,  1.2,  1.3,  1.4,  1.5,  1.6,  1.7,  1.8,  1.9,  2. ,
        2.1,  2.2])
```

Update: don't know why it shows all these commits, only the last one matters.

### DIFF
--- a/src/impl_constructors.rs
+++ b/src/impl_constructors.rs
@@ -33,13 +33,22 @@ impl<S> ArrayBase<S, Ix>
         Self::from_vec(iterable.into_iter().collect())
     }
 
-    /// Create a one-dimensional array from inclusive interval
+    /// Create a one-dimensional array from the inclusive interval
     /// `[start, end]` with `n` elements. `F` must be a floating point type.
     pub fn linspace<F>(start: F, end: F, n: usize) -> ArrayBase<S, Ix>
         where S: Data<Elem=F>,
               F: libnum::Float,
     {
         Self::from_iter(linspace::linspace(start, end, n))
+    }
+
+    /// Create a one-dimensional array from the half-open interval
+    /// `[start, end)` with elements spaced by `step`. `F` must be a floating point type.
+    pub fn range<F>(start: F, end: F, step: F) -> ArrayBase<S, Ix>
+        where S: Data<Elem=F>,
+              F: libnum::Float,
+    {
+        Self::from_iter(linspace::range(start, end, step))
     }
 }
 
@@ -215,4 +224,3 @@ impl<S, A, D> ArrayBase<S, D>
     }
 
 }
-
--- a/src/linspace.rs
+++ b/src/linspace.rs
@@ -78,3 +78,26 @@ pub fn linspace<F>(a: F, b: F, n: usize) -> Linspace<F>
         len: n,
     }
 }
+
+/// Return an iterator of floats spaced by `step`.
+///
+/// The `Linspace` has `n` elements, where the first
+/// element is `a` and `b` is not included.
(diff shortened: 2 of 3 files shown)


bluss:
Your branch is based on `gh-pages`. Can you rebase the thing to be on top of master? Since it's just one commit, `git cherry-pick <thecommithash>` is a simple way to transfer it to a new branch you check out.

bluss:
This is where I battle my frugality in providing methods. Do we provide both arange and arange_step, or only the more general longer to type one. Either one we add, I'll surely use it a lot too.

I think that in our naming scheme, this is better called just `range`. No need for the a for array disambiguation, and there's no importing static methods. That's for contributing to us.

SuperFluffy:
I am against having either. It adds very little on top of `linspace`, and it believe it would fit in better with the rest of Rust to rather `collect()` a `Linspace` into an `Array`.

bluss:
We need the convenience. I could have stopped at just from_vec and from_iter, but they are not simple enough.

Explicit is a lot better than using .collect() too, I think `OwnedArray::range(a, b, 0.1)` is nice code.

bluss:
By the way, the `Array` typedef being renamed to RcArray, the `Array` name is now free to take.. Not sure what you use it. Maybe `OwnedArray`,  `ArrayBase` or a new trait for all arrays?

SuperFluffy:
What's to bad about `collect::<OwnedArray<_,Ix>>()` (not checked if this works). I am not sure if more convenience than that is needed? It's really quite explicit. I might be wrong, of course.

palango:
I agree that `range_step` is maybe to much, I can remove it. But I think that it's a nicer interface than having to collect().

bluss:
Explicit Type::method is my preference after working with type inference and .collect() for a long time.

@palango Can you force-reset the branch in this PR? It would clean up git (all PRs are branches) for me, so it would do me a favour. Thanks

SuperFluffy:
An exclusive or inclusive range/linspace would also be something nice, btw. In numpy:

````python
numpy.linspace(start, stop, num=50, endpoint=True, retstep=False, dtype=None)[
````

That `endpoint=True/False` can be quite handy.

palango:
@bluss: done. Should I remove `range` and rename `range_step`?

bluss:
Yes, that's what I prefer. Thanks for fixing the branch!

We don't really have a chance at python's convenient interface here short of builder patterns or macros. I'm open to discussing that of course.

bluss:
@SuperFluffy Two bad parts with `collect::<OwnedArray<_,Ix>>()`. The type is sometimes needed and sometimes not. I think it's a good *generic* interface, but not so good when you want which type you want. Second bad part is that it's a heap of punctuation in a language already full of it. I don't like typing punctuation.

palango:
OK, renamed everything. Should I squash at well?

bluss:
Please do. It looks good.

bluss:
The range_step -> range rename disappeared.

palango:
@bluss: What the hell!? I have to improve my git skills.

palango:
@bluss: Seems like I figured it out.

bluss:
Thanks!

PR #134, opened 2016-03-09
Switch to blas-sys and use BLAS in .mat_mul

bluss:
Use transparent BLAS support in .mat_mul().

- Switch to using `blas-sys` and raw cblas function calls. The savings on overhead are noticeable compared to `rblas`'s trait object interface.
- `blas-sys` allows selecting which library provides blas.
- You must now use feature flag `blas` to use BLAS in .dot() and .mat_mul()
- This half-deprecates the feature flag `rblas`. If the rblas API is attractive, it can move to a separate crate.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,16 +1,24 @@
 language: rust
-sudo: false
+# use trusty for newer openblas
+sudo: required
+dist: trusty
 matrix:
   include:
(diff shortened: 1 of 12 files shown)


bluss:
needs to be consistent in memory order of the returned array (blas vs no blas).

PR #146, opened 2016-03-12
Add randn for arrays of standard gaussian distributed random numbers.

daniel-vainsencher:
This time trying to rebase, rather than merge, let me know if I've done it correctly.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,16 +1,24 @@
 language: rust
-sudo: false
+# use trusty for newer openblas
+sudo: required
+dist: trusty
 matrix:
   include:
     - rust: stable
+      env:
+       - FEATURES='test'
     - rust: beta
     - rust: nightly
       env:
-       - FEATURES='assign_ops rustc-serialize'
+       - FEATURES='test'
        - BENCH=1
 branches:
   only:
     - master
+addons:
+  apt:
+    packages:
+      - libopenblas-dev
 script:
   - |
       cargo build --verbose &&
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 
 name = "ndarray"
-version = "0.4.0-alpha.8"
+version = "0.4.0"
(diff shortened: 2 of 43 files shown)


bluss:
Thanks. I'm not promising that we will integrate rand directly like this. It could for example belong to a crate of its own "ndarray-rand" or so.
Something is definitely amiss when there are so many commits listed in the PR. Would be great if you could reset your branch to just contain one patch on top of master and force push again.

daniel-vainsencher:
This is just a first cut. I agree we need to get the dependencies and design right. And I'll have no choice but to read up on git rebasing... out of time for now, will get back to this, as I need it. BTW, is there a good alternative to depending on rand?

bluss:
It's just one commit, so you can use cherry-pick to move it to the correct branch. It's simple, like this https://github.com/bluss/rust-ndarray/pull/124#issuecomment-192680940

I don't know a better crate than rand.

`rand` could be an optional dependency, but even better than that (also what crates.io advises) is to make a separate crate if the feature is easy to split out.

A new crate ndarray-rand could be hosted by me, it would depend on ndarray and rand, and it provides these constructors.

bluss:
Maybe using rand for the gaussian samples is the wrong thing? I wonder if we can do this much more efficiently if we're generating a big set.

daniel-vainsencher:
I am not sure what is the significant advantage of avoiding rand as a dependency of ndarray.

- If we create our own code that takes advantage of making a whole array of them at a time, that's fine, but currently a hypothetical.
- Otherwise random numbers are not esoteric in numerical computing, they are a fundamental tool used all the time. randn is one of my most common numpy imports in python.
- While cargo is good enough that having users add another crate is not difficult, this only works if they are aware it exists; discovery is still made harder. I haven't thought through any ergonomic differences for users of randn.

daniel-vainsencher:
Created a new pull request with clean history and addressing the points above.

PR #159, opened 2016-03-16
Methods .mapv(), .mapv_into(), .apply(), .applyv(), .visit()

### LINKED ISSUE #58: Add unordered element iterator or visitor (using closure)
Like `.iter()`, but visits the array elements in "arbitrary" order. Will use memory order or whatever order is most efficient.

### LINKED ISSUE #153: Add method to modify elements in place
A value-to-value mapper would be convenient, something like `array.apply(f32::abs)` it could even be called `lift`.

bluss:
Fixes #58 
Fixes #153

I did experiment with a trait to unify the map and mapv case into a single method, and while it worked, it also introduced type inference ambiguities everywhere, so it was so much less useful.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -53,16 +53,20 @@ Status and Lookout
 
 - Performance status:
 
-  + Arithmetic involving arrays of contiguous inner dimension optimizes very well.
-  + ``.fold()`` and ``.zip_mut_with()`` are the most efficient ways to
+  + Performance of an operation depends on the memory layout of the array
+    or array view. Especially if it's a binary operation, which
+    needs matching memory layout to be efficient (with some exceptions).
+  + Arithmetic optimizes very well if the arrays are have contiguous inner dimension.
+  + The callback based methods like ``.mapv()``, ``.applyv()`` and
+    ``.zip_mut_with()`` are the most efficient ways to
     perform single traversal and lock step traversal respectively.
-  + ``.iter()`` and ``.iter_mut()`` are efficient for contiguous arrays.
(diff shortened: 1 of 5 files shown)


bluss:
I am taking suggestions for a good naming scheme for the methods .map(), .mapv(), .apply(), .applyv(), .visit().

We do want them all as this kind of "internal iteration" or closure visiting is the only way we can optimize the traversal across nested loops if needed.

We do want both .map and .mapv and the same with .apply, .applyv, because the by reference methods are the general solution and the by value methods are convenient.

daniel-vainsencher:
Map should definitely stay map. `Apply` doesn't explain that mutation is going to happen. Maybe `transform`? I doubt I'd ever find out about the `v` versions, I would be using & because I've gotten used to this inconvenience :( I don't have a better name, though. Visit is fine, IMO.

bluss:
Thanks for the input! They will be next to each other in the docs, so I hope you'd see them there.

This is tricky because mathematically, all four methods do the same thing. So we're trying to find words for that..

bluss:
Oh no, there's yet another one that comes up, `mapv_into`. (Map by value, return the input).

Again, all of these are the same mathematically, just with different rust type system effects.

```rust
 .map(&self, |&A| -> B) -> OwnedArray<B, D>
 .mapv(&self, |A| -> B) -> OwnedArray<B, D>
 .mapv_into(self, |A| -> A) -> Self
 .apply(&mut self, |&mut A|)
 .applyv(&mut self, |A| -> A)
 .visit(&self, |&A|)
```

PR #166, opened 2016-03-19
Add crate ndarray-rand

bluss:
This is a stab at just ndarray-rand with four simple constructor methods.


bluss:
cc @daniel-vainsencher 

The f64 vs f32 issue is making me unhappy here. It would be the best if rand just implemented Normal for f32.

bluss:
Pushed a new version that even removed the ::normal() constructor again. We'll add it later. Ideally, Normal from rand should just implement generation of any F: Float or something like that..

daniel-vainsencher:
Wasted some time looking at this, but unfortunately don't have much smart to say. I see your frustration with rand's Normal being f64 only. Maybe there aren't any speed savings to be had (without too much numerical precision tradeoff) so they didn't bother?

One way to depend on the design of rand less (hence make it easier to change implementations later on) is to export only particular important convenience constructors like normal, binomial, rademacher, hence not expose IndependentSample etc in the interfaces. Ideally generic in NdFloat, but this isn't critical, IMO.

bluss:
I thought that was a smart contribution about a possible rand-less API surface.

daniel-vainsencher:
Glad that helps, I meant I'd hoped to have a fix rather than a workaround, but I don't see a simple one and don't have the bandwidth to try to improve rand.

bluss:
Haha the more I fret over this, the harder it seems.

It's related to traits and common interfaces. It's really uncommon that a trait really fits your use case like a glove.

rand and num are immature, just like ndarray is.

`Float` is another trait that does not fit so well. Yet I want to use it because it's what I **assume** numerical code likes to use, and we want to be compatible with that as far as possible. As soon as we try to create the best possible library, we are drawn towards creating and controlling traits ourselves though.

I'll merge this, it has the right directory structure, so that I can fix up documentation on top of this.

But, no release of ndarray-rand yet.

PR #172, opened 2016-03-25
Update blas-sys and openblas-provider

IvanUkhov:
I have updated `openblas-provider` to compile OpenBLAS v0.2.17, which was released five days ago.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,8 +36,8 @@ optional = true
 
 [dependencies]
 # Use via the `blas` crate feature!
-blas-sys = { version = "0.5", optional = true, default-features = false }
-openblas-provider = { version = "0.3", optional = true, default-features = false }
+blas-sys = { version = "0.6", optional = true, default-features = false }
+openblas-provider = { version = "0.4", optional = true, default-features = false }
 
 # deprecated! use ndarray-rblas instead
 rblas = { version = "0.0.13", optional = true }
@@ -50,7 +50,7 @@ blas = ["blas-sys"]
 blas-openblas-sys = [
     "blas",
     "blas-sys/openblas",
-    "openblas-provider/system-openblas",
+    "openblas-provider/system",
 ]
 
 # This feature is used for testing


bluss:
Thank you!

bluss:
I've been looking at BLIS since it in theory fits ndarray better with its support for arbitrary strides. However, integrating them with rust is not so easy, like most scientific projects it requires a lot of manual work to configure.

daniel-vainsencher:
Just my vote as a particular user: for me an assert preventing stride
mismatch on calls to another blas is just as good.

OTOH, the primitives I use are mostly simple matrix vector products, so I
don't care about using the one "best" BLAS library (scare quotes because
its a multi criteria problem). Encountering configuration problems would be
cost me more than any potential benefit.

On Fri, Mar 25, 2016 at 10:30 AM, bluss <notifications@github.com> wrote:

> I've been looking at BLIS since it in theory fits ndarray better with its
> support for arbitrary strides. However, integrating them with rust is not
> so easy, like most scientific projects it requires a lot of manual work to
> configure.
>
> —
> You are receiving this because you are subscribed to this thread.
> <https://github.com/bluss/rust-ndarray/pull/172#issuecomment-201312250>
>



-- 
Daniel Vainsencher

bluss:
.mat_mul() never fails to compute as long as the matrix dimensions match as the should, it just falls back to the generic matrix multiplication implementation. If the matrix is large enough, the performance difference is huge (more than 10 times slower for a 128x128 matrix multiplication).

But I have said I will not reimplement linear algebra in rust. So for ndarray to support more linear algebra I would prefer to use an external library.

IvanUkhov:
@bluss, where can one read your argumentation agains reimplementing linear algebra in Rust?

bluss:
ndarray wants to provide good solutions, why not use the decades long expertise in linear algebra that others have put into for example BLAS/LAPACK?

IvanUkhov:
Why implementing Servo if we have Gecko?

bluss:
@IvanUkhov ?? I've said I will not implement linear algebra. That's my decision. You can write Servo or linear algebra in rust if you want to.

IvanUkhov:
Sorry, no offense was intended. I just thought that you meant that, in general, it was not worth it rewriting linear algebra in Rust (like it would be drastically slower or something like that).

daniel-vainsencher:
A bit off topic, but I think that linear algebra and web browsers have
completely different tradeoffs. In the first complexity is all about
optimizations, the interfaces are small and very well known. In particular
the ratio potential for unsafety bugs / code size is probably much much
lower than in web browsers that need to e.g. parse and run arbitrary html
etc.

On Fri, Mar 25, 2016 at 12:23 PM, bluss <notifications@github.com> wrote:

> @IvanUkhov <https://github.com/IvanUkhov> ?? I've said I will not
> implement linear algebra. That's my decision. You can write Servo or linear
> algebra in rust if you want to.
>
> —
> You are receiving this because you commented.
> <https://github.com/bluss/rust-ndarray/pull/172#issuecomment-201349444>
>



-- 
Daniel Vainsencher

bluss:
Aha. The more I looked into BLIS the more I thought they could benefit from Rust. They don't use generics (instead void *), and use C macros a lot. Quite the mess! I think rust can do it just as well, but it will be a lot of raw pointers, unsafe code and asm kernels. Maybe an open question is, how easy is it to use asm from Rust?

bluss:
@daniel-vainsencher Do you use blas + ndarray at the moment? If you don't enjoy hassle with dependencies, I suspect you don't do that. It's an optional ("and experimental") crate feature and I haven't documented it properly yet (it requires picking a BLAS provider deliberately.)

daniel-vainsencher:
I'm not; for matrix.dot(vec) don't see a need to. If your point is that a
more "hassle-y" blas would cause me the exact same amount of hassle (zero,
because not using the feature) you are right at least for now :)

bluss:
@IvanUkhov I apparently had too much time this weekend: http://bluss.github.io/rust/2016/03/28/a-gemmed-rabbit-hole/  (it's an implementation of dgemm in Rust).

PR #175, opened 2016-03-29
Depend on crate matrixmultiply for f32, f64 matrix multiplication

bluss:
The crate matrixmultiply (written by me) uses a solid approach to matrix
multiplication for floats. It is so far in plain rust and outperforms
naive multiplication by a ton.

See result of using matrixmultiply instead of the naive algorithm,
here for a 64 by 64 matrix multiplication, f32.

```
name               before ns/iter  after ns/iter    diff ns/iter   diff %
mat_mul_f32::m064  258,969         31,093               -227,876  -87.99%
```

Another benefit of matrixmultiply is that it supports our general striding scheme. That means it's a good fallback even if we use OpenBLAS for example (which does not handle general strides, i.e where both strides are != 1).

(Also added more benchmarks and tests.)


daniel-vainsencher:
Just as a curiosity, have you tried using OpenBLAS as the microkernel? if
it closes most of the gap to OpenBLAS, this says good things about the BLIS
approach.

On Tue, Mar 29, 2016 at 4:58 PM, bluss <notifications@github.com> wrote:

> Merged #175 <https://github.com/bluss/rust-ndarray/pull/175>.
>
> —
> You are receiving this because you are subscribed to this thread.
> <https://github.com/bluss/rust-ndarray/pull/175#event-607324906>
>



-- 
Daniel Vainsencher

bluss:
I don't think OpenBLAS has any function that we can call directly to use as the microkernel.

What I did try was to take the sgemm (f32) inline asm kernel from BLIS and plug it into this crate, and that [went well, ](https://github.com/bluss/matrixmultiply/pull/1#issuecomment-203089659) clocked  at 150,000 ns/iter for the size 127 matrix problem.

However, to use their microkernels requires

1. matrixmultiply learns to align the packed a, packed b and mask buffers to 32 bytes (needed for AVX aligned load)
2. We need to compile them using a C compiler in a cargo build script. Which isn't that hard actually.
3. We need to feature detect when it's appropriate to use AVX, which seems basically impossible to do at build time.

I guess (3) can be solved by opting in to feature detection at build time, which removes the possibility of cross compilation.

bluss:
[This is the BLIS microkernel I tried.](https://github.com/flame/blis/blob/64b41fa554dff44b2f9ad48901b67c63836407a8/kernels/x86_64/avx/3/bli_gemm_asm_d8x4.c#L40) It was remarkably easy to just put that file in a .c by itself and compile it using crates.io crate `gcc`.

Issue #141, opened 2016-03-12
Breaking changes for next version


bluss:
strikeout: it was decided to skip this

- [x] Remove deprecated `ndarray::blas` in favour of ndarray_rblas crate
- [x] ~~Add Send + Sync to LinalgScalar~~
- [x] ~~Consider AsPrim for NdFloat~~
- [x] Remove deprecations: free function zeros, type alias Array, .raw_data(), ~~.fold()~~
- [x] ~~Remove iadd, imul etc as deprecated too, in favour of += etc~~
- [x] ~~.swap_axes() should use Axes arguments!~~
- [x] ~~Restrict .map() to `Fn()` closures for now.~~
- [x] consuming neg and not should not be defined for arrayviewmut nor consuming lhs scalar operations
- [x] Require Zero or LinalgScalar in .sum()

PR #179, opened 2016-04-08
Initial changes for the 0.5 series

bluss:
Do removals of deprecated items. Enable assign_ops by default (Rust 1.8 feature).

Addresses #141 partly

- Remove Array, zeros, module blas, raw_data, raw_data_mut, OuterIter, OuterIterMut
- Remove crate features assign_ops, rblas.
- Mark deprecated all the iadd and similar methods (will be removed in a later alpha release before 0.5).


daniel-vainsencher:
I proposed for fold to change the documentation to not promise any order, instead of removing/changing the method, not sure if you saw that.

bluss:
I did.. and I decided that was great advice, so I'll do that.

PR #189, opened 2016-04-16
support for non-contiguous index selection along axis

pegasos1:
Wrote some tests for the implementation we discussed. I thought about handling errors around `stack`, but index errors will crash the program in `isubview` beforehand anyway.

### DIFF
--- a/src/impl_methods.rs
+++ b/src/impl_methods.rs
@@ -197,25 +197,7 @@ impl<A, S, D> ArrayBase<S, D> where S: Data<Elem=A>, D: Dimension
         debug_assert!(self.pointer_is_inbounds());
     }
 
-    pub fn select_along(&self, axis: Axis, indices: &[Ix]) -> OwnedArray<A, D>
-        where A: Copy,
-              D: RemoveAxis,
-    {
-        let v = self.view();
-        let mut subs = vec![v; indices.len()];
-        for (&i, sub) in zipsl(indices, &mut subs[..]) {
-            sub.isubview(axis, i);
-        }
-        if subs.is_empty() {
-            let mut dim = self.dim();
-            dim.set_axis(axis, 0);
-            unsafe {
-                OwnedArray::from_vec_dim_unchecked(dim, vec![])
-            }
-        } else {
-            stack(axis, &subs).unwrap()
-        }
-    }
+
 
     /// Return a reference to the element at `index`, or return `None`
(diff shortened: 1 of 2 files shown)


src/impl_methods.rs
```diff
@@ -196,6 +197,26 @@ impl<A, S, D> ArrayBase<S, D> where S: Data<Elem=A>, D: Dimension
         debug_assert!(self.pointer_is_inbounds());
     }
 
+    pub fn select_along(&self, axis: Axis, indices: &[Ix]) -> OwnedArray<A, D>
```
bluss: I think select is a good keyword (probably got it from daniel's issue). Do you have a suggestion for an even better method name here? I was thinking that this is maybe only one kind of such a method, this is the along an axis selection. `select_along` would suggest "along an axis" maybe `select_axis` is better? Only `select`? Something else.

This method is actually in the wrong place. It's not so important that it should be before .get() in the docs. I guess it's not a perfectly sorted order. Below the subview methods seems ok.

pegasos1:
I think `select` makes sense. This type of indexing is generic enough. Also, any indexing along 2 or more axes is probably more suited to methods like get().

whoops, forgot to move that! Also, i'll add some basic docs.

bluss:
Btw, the PR says it has a merge conflict. Select sounds good. @daniel-vainsencher is very welcome with feedback also after the PR is merged, should be no trouble to change things around to better names.

docs are great. Notice that one can select the same view multiple times and that's just fine I think.

pegasos1:
oh, we could do `select_subview` to keep consistent with other `subview` methods. 
I'll look into those conflicts.

bluss:
hm that sounds pretty good. with an s in subviews?

pegasos1:
nah I was thinking singular subview. why make it plural?

bluss:
Since it picks several subviews and concats them it seems appropriate. Even if the terminology is ad hoc.

bluss:
It's a bit long though. Naming things..

pegasos1:
haha yeah, probably the hardest part of programming.

pegasos1:
if you want, we can leave it at `select`, and then change it if we want later.

bluss:
seems best for now

pegasos1:
good to go

bluss:
thanks! this is a good feature. I'll edit the docs slightly.

PR #191, opened 2016-04-17
Unify array with shape construction using Shape & ShapeBuilder

### LINKED ISSUE #163: Papercut: mismatch between name and signature
The name of `from_slice_dim_stride(dim: D, strides: D, xs: &'a mut [A])` suggests a parameter ordering inconsistent with what it actually expects.

Same goes for `pub fn from_vec_dim(dim: D, v: Vec<A>)` and related constructors.

I suggest to change the API now despite the breakage, rather than wait to get more users. I don't have good ideas on which ordering to choose.

bluss:
Now please use these: from_elem, zeros, from_shape_vec, from_shape_vec_unchecked

instead of the following which are deprecated:

+ from_elem_f
+ zeros_f
+ from_vec_dim
+ from_vec_dim_f
+ from_vec_dim_unchecked
+ from_vec_dim_unchecked_f
+ from_vec_dim_stride
+ from_vec_dim_stride_unchecked

Fixes #163

### DIFF
--- a/src/data_traits.rs
+++ b/src/data_traits.rs
@@ -83,8 +83,8 @@ unsafe impl<A> DataMut for Rc<Vec<A>>
             // Create a new vec if the current view is less than half of
             // backing data.
             unsafe {
-                *self_ = ArrayBase::from_vec_dim_unchecked(self_.dim.clone(),
-                                                           self_.iter()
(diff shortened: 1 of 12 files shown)


bluss:
cc @daniel-vainsencher

daniel-vainsencher:
Pretty... :)

bluss:
That's all we need to hear.. go!

Issue #209, opened 2016-08-02
QUESTION: ArrayView into C allocated memory


fkjogu:
I'm trying to do a 2D Fourier transform calling to [FFTW3](www.fftw.org) over FFI. For internal performance reasons, FFTW3 offers a custom malloc function, that returns a pointer to a RowMajor C-array.

Can I somehow bind this to an ndarray ArrayView?

Please feel free to close the issue, since it is more a question than an issue.

vbarrielle:
Unfortunately, it looks like `ArrayView::new_` is private, even though you could use it for that purpose.
@bluss, if you want I can put it public (under a more explicit name maybe) and PR.

bluss:
You can use the existing aview1, ArrayView::from or ArrayView::from_shape(shape, slice), if you combine them with std::slice::from_raw_parts.

We can add separate direct constructors though.

vbarrielle:
Oh right, I didn't think of creating the slice first...

> We can add separate direct constructors though.

I do think it would be nice, since interfacing with C allocated arrays might be frequent if people want to e.g. interface with numpy. I'll PR a direct constructor so you can tell me what you think.

PR #210, opened 2016-08-02
expose ArrayView(Mut)::from_ptr, useful for FFI

vbarrielle:
See #209 for the intent: fast-path for interfacing with FFI-allocated arrays.

### DIFF
--- a/src/impl_views.rs
+++ b/src/impl_views.rs
@@ -69,6 +69,19 @@ impl<'a, A, D> ArrayBase<ViewRepr<&'a A>, D>
         ArrayView::new_(ptr, dim, strides)
     }
 
+    /// Create an ArrayView from a raw pointer and shape information
+    ///
+    /// The caller is responsible for ensuring that the pointer is valid
+    /// and coherent with the dim and stride information.
+    pub unsafe fn from_shape_ptr<Sh>(shape: Sh, ptr: *const A) -> Self
+        where Sh: Into<StrideShape<D>>
+    {
+        let shape = shape.into();
+        let dim = shape.dim;
+        let strides = shape.strides;
+        ArrayView::new_(ptr, dim, strides)
+    }
+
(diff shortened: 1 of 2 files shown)


bluss:
Thank you.

Looking with consistency eyes, here are some thoughts:

`std::slice::from_raw_parts` suggests that maybe it can be named from_raw_parts.

The other array constructors now use a size/shape object as one parameter for convenience, we should use that. Example link: https://bluss.github.io/rust-ndarray/master/ndarray/struct.ArrayBase.html#method.from_shape_vec

Going all the way would be to have both `from_shape_ptr(shape, ptr) -> Result<ArrayView<..>>` and `from_shape_ptr_unchecked(shape, ptr) -> ArrayView<..>` and this would undeniably be consistent, and promote a checked version as the preferred approach.

However there is only so much checking that we can do. We have to trust the caller that the pointer is valid for the whole needed range. Maybe the checked version won't see any use anyway.

`from_shape_ptr(shape, ptr) -> ArrayView<..>` would be fine for that reason, and it's simple to call if you just use default strides, or you can just use for example `ArrayView::from_shape_ptr((m, n).f(), ptr)` to have f-order.

vbarrielle:
As far as I can tell, the only check that can be performed would be verifying that the strides do not allow indexing the same element with different indices. So I'd say having only ``from_shape_ptr(shape, ptr) -> ArrayView<..>`` should be enough.

bluss:
yes

vbarrielle:
Here's a new version with the rename, and the unchecked ``from_shape_ptr`` version.

fkjogu:
Doing a quick test, I can confirm this works quiet nicely. Thank you so much.

bluss:
Thank you! Merging and will do changes on top of that. The from_raw_parts variants are more or less redundant and will be removed.

PR #211, opened 2016-08-04
0.6.1 with from_shape_ptr

bluss:


bluss:
Thanks @vbarrielle for making it happen.

Ok, @SuperFluffy, I will look at serde next

PR #212, opened 2016-08-12
Use serde 0.8

bluss:
Builds upon #115

### DIFF
--- a/src/arrayserialize.rs
+++ b/src/arrayserialize.rs
@@ -12,6 +12,7 @@ use imp_prelude::*;
 use super::arraytraits::ARRAY_FORMAT_VERSION;
 use super::Elements;
 
+/// **Requires crate feature `"serde"`**
 impl<A, D, S> Serialize for ArrayBase<S, D>
     where A: Serialize,
           D: Dimension + Serialize,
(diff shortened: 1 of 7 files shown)


bluss:
@SuperFluffy What do you think the versioning strategy should be? We can't upgrade to a future serde 0.9 without a major version bump.

SuperFluffy:
I think moving serialization into its own (sub-) crate might make sense,
look e.g. at how num does it with complex, etc.

This way we'd only update the serialization crate and can leave the main
crate untouched, sans a minor version bump maybe.

bluss:
That would be ideal, but it's not possible since we're implementing traits for the type ArrayBase, so it has to be in ndarray.

SuperFluffy:
Ah, shoot, I didn't think of that. Well, I guess bumping the major number
is the only way then. I expect serde to be a very important dependency, so
I suspect this is fine.

bluss:
I don't like it at all, but it's a well known problem in the ecosystem.

src/arrayserialize.rs
```diff
@@ -24,20 +24,24 @@ impl<A, D, S> Serialize for ArrayBase<S, D>
         let mut struct_state = try!(serializer.serialize_struct("Array", 3));
         try!(serializer.serialize_struct_elt(&mut struct_state, "v", ARRAY_FORMAT_VERSION));
         try!(serializer.serialize_struct_elt(&mut struct_state, "dim", self.dim()));
-        try!(serializer.serialize_struct_elt(&mut struct_state, "data", self.iter()));
+        try!(serializer.serialize_struct_elt(&mut struct_state, "data", Sequence(self.iter())));
```
SuperFluffy: Just curious, what's the thinking behind this?

bluss: Export nothing more publically than needed. A serialize impl for the iterator would be public.

PR #213, opened 2016-08-14
Implement Default for owned arrays

### LINKED ISSUE #204: `Default` trait for `ArrayBase`?
What is the opinion on implementing `std::default::Default` for `ArrayBase`? When skipping deserialization of a field with `serde`, it wants to insert the `Default::default()` value in there, so this would be one case where this would be very useful.

We already have a default function for `ArrayBase`, which however expects some kind of shape/dimension:

```rust
    pub fn default<Sh>(shape: Sh) -> ArrayBase<S, D>
        where A: Default,
              Sh: Into<Shape<D>>,
    {
        let shape = shape.into();
        let v = (0..shape.dim.size()).map(|_| A::default()).collect();
        unsafe { Self::from_shape_vec_unchecked(shape, v) }
    }
```

A default in the spirit of `std::default::Default` would look like:

```rust
    pub fn default() -> ArrayBase<S, D>
        where A: Default,
              D: Default + Default,
              Sh: Into<Shape<D>>,
    {
        let shape = <D as Default>::default().into();
        let v = (0..shape.dim.size()).map(|_| A::default()).collect(); // == Vec::new()
        unsafe { Self::from_shape_vec_unchecked(shape, v) }
    }
```

And `Default` impls for the various dimension-like types would be akin to:

```rust
impl Default for Ix {
    fn default() -> Self {
        0 as Ix
    }
}

impl Default for (Ix,Ix) {
    fn default() -> Self {
        (0,0) as (Ix,Ix)
    }
}
```

Is this desirable?

**EDIT:** I just realize that this might not be as straightforward as I thought, since e.g. `Ix` acts as just a type alias, preventing an easy implementation of `Default` for it. :-/

bluss:
zero-dimensional arrays (using () or Vec::new() as the dimension/D type
parameter) will have 1 element in the default state.
    
nonzero-dimensional arrays (using Ix, (Ix, Ix) and so on) will have
0 elements in their default state.

Fixes #204


bluss:
ok @SuperFluffy, the reality of zero-dimensional arrays say that these semantics are needed for at least some arrays.

Is it too insane to start with either 1 or 0 elements depending on the `D` parameter? We could make all arrays have exactly one element too, but then we need to allocate a vector in the common case.

SuperFluffy:
I think your solution is a good one. Even though I'd prefer to have the 0-dimensional case to have 0 elements as well, though that's obviously not going to work. I am strongly opposed to the other arrays having a default element... I can't articulate why that is, but I'd rather have the 0-d case be a special one.

PR #214, opened 2016-08-18
num-traits and -complex to 0.1.35 to support serde 0.8

SuperFluffy:
Upgraded `num-traits` and `num-complex` to `0.1.35` so that all of (?) `ndarray` supports `serde 0.8`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,11 +20,11 @@ bench = false
 test = false
 
 [dependencies.num-traits]
-version = "0.1.32"
+version = "0.1.35"
 default-features = false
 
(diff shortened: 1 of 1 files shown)


bluss:
This bump isn't strictly needed, you can just use cargo upgrade and it will work.

SuperFluffy:
Yeah, you are right. Sorry for the panic. I just rechecked how to [specify crate dependencies](http://doc.crates.io/specifying-dependencies.html), and there it's noted that `0.1.32` is equivalent to `^0.1.32 := >=0.1.32 <1.0.0`.

PR #218, opened 2016-09-19
Add method .map_axis()

bluss:


mokasin:
After a quick test, it seems to be doing, what it is supposed to do. Although you could consider ´.reduce_axis()´ as a name, since it suggests that it takes something (an axis) away and *reduces* it to a single scalar.

bluss:
The improved formulation is beautiful (just an application of existing .map()) https://github.com/bluss/rust-ndarray/commit/0c10593eea3a9e804b3c801cb702946dd5e05123?diff=split

I think we will stick with map since both fold and reduce have an association to a function that is applied successively on each element of the access IMO.

bluss:
Released in 0.6.4 by the way, name map_axis.

PR #220, opened 2016-09-26
Add method into_raw_vec for Array

### LINKED ISSUE #207: (ArrayBase<S, Ix>) -> Vec<S::Elem>
There is a `from_vec` function which consumes a `Vec<_>` to create a 1D array.

Is it possible to construct an inverse operation as well?

    (ArrayBase<S, Ix>) -> Vec<S::Elem>

bluss:
Fixes #207

### DIFF
--- a/src/impl_owned_array.rs
+++ b/src/impl_owned_array.rs
@@ -1,0 +1,15 @@
+
+use imp_prelude::*;
+
+impl<A, D> ArrayBase<Vec<A>, D>
+    where D: Dimension
+{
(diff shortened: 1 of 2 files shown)


bluss:
This is just the simple one, the one with just the memory layout

Rufflewind:
Thanks!

bluss:
Sorry for the delay. It's not perfect, but the docs make it clear that the result vector is not so straightforward for arbitrary arrays.

PR #222, opened 2016-10-20
Better code for Array::default(). Also add type aliases for common array sizes.

bluss:


bluss:
@SuperFluffy What do you think about having all the type alises directly under the top level?

SuperFluffy:
I can't think of a reason to put them under `alias`. This way they are more visible and might actually be picked up. :-)

bluss:
Yes, that sounds good.

PR #225, opened 2016-11-07
The great dimension migration to `Dim<[Ix; N]>`

### LINKED ISSUE #6: Transition to using fixed size arrays for dimensions
Main win is syntax, `array[[1, 2]]` is pretty sweet.

bluss:
If you thought I was slacking, finally the rework of ndarray's dimension handling is ready (almost).

To have your code be compatible with ndarray 0.6.8 and with this and the future: Use the dimension type aliases. Use `Array<X, Ix1>` or `Array1` for 1d arrays, `Array<X, Ix2>` or `Array2` for 2d arrays and so on. Then transition will be very smooth.

- Tuples are still used to describe shapes when creating arrays (`Array::zeros((3, 4))`).
- The dimension values are represented using the new `Dim` struct. `Dim<[usize; 2]>` is a 2D dimension.
- Dimensions/Indices now **implement indexing and arithmetic**

Requires Rust 1.12 (macro expansion of tt).

Fixes #6


bluss:
@SuperFluffy Serialization changes, now serializes as arrays instead of tuples. I think the defacto format for json is the same in both serialization libs, but it needs checking. Handling that properly feels like a project of its own.

PR #232, opened 2016-11-12
Update the zip-many example

bluss:


bluss:
cc @SuperFluffy first step, update the example. Next, make it a real feature in ndarray.

PR #236, opened 2016-11-14
Implement ArrayBase::fill

### LINKED ISSUE #235: Replace ArrayBase::assign_scalar with ArrayBase::fill
Is the shorter name worth it? The longer name has some consistency (but not much).

cc @SuperFluffy  I heard you wanted to change .assign_scalar. My suggestion is to just deprecate it (without changing it), and adding .fill() instead, that has the signature that you wanted.

SuperFluffy:
Also changed the tests and depracated `ArrayBase::assign_scalar`.

Fixes #235


bluss:
Thanks, looks good! I'll just wait for travis to finish.

PR #241, opened 2016-11-19
Rename raw_dimension → raw_dim, better docs, expand prelude

bluss:
- The prelude gains all the dimensionality-specific type aliases.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -105,7 +105,7 @@ Recent Changes (ndarray)
   - Constructors and methods that take tuples for array sizes, like ``Array::zeros,``
     ``Array::from_shape_vec``, ``.into_shape()`` and so on will continue to work
     with tuples.
-  - The array method ``.raw_dimension()`` returns the shape description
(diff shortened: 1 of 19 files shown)


bluss:
cc @SuperFluffy do you want to review this? The prelude becomes very large.

SuperFluffy:
Would you like a review on the entire PR or only prelude?

I like the prelude. One can always choose to not import it. But it looks like it contains everything that I tend to use with `ndarray`. I might actually switch over to prelude. :D

bluss:
mainly the prelude changes. Yeah it's optional to use it, and it's hard to pick a smaller subset. It could for example only include dimensionalities 0, 1, 2, but that doesn't reduce the logical complexity of the prelude at all, only the number of items.

SuperFluffy:
Yes, that was my thought, too. Furthermore, it would be inconsistent. Why would import a certain subset of indices or dimensionalities, but not all of them? I think that would be quite confusing and not very useful.

I think either import all of the sets, or none of them.

PR #242, opened 2016-11-19
Serde generic deserialization

### LINKED ISSUE #240: serde chokes on binary deserialization
# Problem description

Serde chokes on deserialization of `ndarray::Array` when deserializing from binary. This might extend to non-json formats as well: I have been trying to serialize and deserialize an `Array2<f64>` with both `bincode` and `rmp` (`msgpack`), and in both cases errors are thrown.

The two code blocks at the bottom of the page are the versions for `rmp` and `bincode`, respectively. The error for `rmp` is as follows:

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: TypeMismatch(Array32)', ../src/libcore/result.rs:837
```

While the error for `bincode` is:

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Serde(Custom("Invalid type. Expected `Seq`"))', ../src/libcore/result.rs:837
```

Deserializing the buffer written to a file in the `rmp` example Python yields the following:

```python
In [1]: import msgpack

In [2]: f = open('./test.txt', 'rb')

In [3]: msgpack.unpackb(f.read())
Out[3]: 
[1,
 [4, 4],
 [1.0,
  2.0,
  3.0,
  4.0,
  5.0,
  6.0,
  7.0,
  8.0,
  9.0,
  10.0,
  11.0,
  12.0,
  13.0,
  14.0,
  15.0,
  16.0]]
```

# Error hypothesis

Given the result of the Python deserialization, I believe the serialization itself to be working ok.

I suspect, but need to verify, that maybe the order of serialization and deserialization is not the same.

# Code examples to reproduce the error

## `rmp` version

+ `serde 0.8`
+ `ndarray 0.6.9`
+ `rmp_serde 0.10`

```rust
#![feature(proc_macro)]

extern crate ndarray;
extern crate rmp_serde;
extern crate serde;

#[macro_use]
extern crate serde_derive;

use serde::{Deserialize,Serialize};
use rmp_serde::Serializer;
use std::env;
use std::fs::File;
use std::io::{Cursor,SeekFrom};
use std::io::prelude::*;

fn main() {
    let a: ndarray::Array2<_> = ndarray::Array::linspace(1.0, 16.0, 16).into_shape((4,4)).unwrap();

    println!("{:?}", a);

    let mut bufcursor = Cursor::new(Vec::new());

    Serialize::serialize(&a, &mut Serializer::new(&mut bufcursor)).ok().unwrap();

    bufcursor.seek(SeekFrom::Start(0));

    let mut f = File::create("test.txt").unwrap();
    f.write(bufcursor.get_ref());

    bufcursor.seek(SeekFrom::Start(0));

    let mut deserializer = rmp_serde::Deserializer::new(bufcursor);
    let a_de: ndarray::Array2<f64> = Deserialize::deserialize(&mut deserializer).unwrap();

    println!("{:?}", a_de);
}
```

## `bincode` version

+ `ndarray: 0.6.9`
+ `serde: 0.8`
+ `bincode: 0.6`

```rust
#![feature(proc_macro)]

extern crate bincode;
extern crate ndarray;
extern crate serde;

#[macro_use]
extern crate serde_derive;

use bincode::SizeLimit::Infinite;
use serde::{Deserialize,Serialize};
use std::fs::File;
use std::io::BufWriter;
use std::io::prelude::*;

fn main() {
    let a: ndarray::Array2<_> = ndarray::Array::linspace(1.0, 16.0, 16).into_shape((4,4)).unwrap();

    println!("{:?}", a);

    let mut buf = bincode::serde::serialize(&a, Infinite).unwrap();

    let a_de: ndarray::Array2<f64> = bincode::serde::deserialize(&buf).unwrap();

    println!("{:?}", a_de);
}
```

SuperFluffy:
Fixes #240

### DIFF
--- a/src/array_serde.rs
+++ b/src/array_serde.rs
@@ -17,6 +17,19 @@ use super::Iter;
 use Dim;
 use dimension::DimPrivate;
 
+/// Verifies that the version of the deserialized array matches the current
+/// `ARRAY_FORMAT_VERSION`.
+pub fn verify_version<E>(v: u8) -> Result<(), E>
+        where E: de::Error
+{
+    if v != ARRAY_FORMAT_VERSION {
+        let err_msg = format!("unknown array version: {}", v);
+        try!(Err(de::Error::custom(err_msg)));
+    }
+
+    Ok(())
+}
+
 /// **Requires crate feature `"serde"`**
 impl<I> Serialize for Dim<I>
     where I: Serialize,
@@ -151,9 +164,7 @@ impl<A, Di, S> Visitor for ArrayVisitor<S,Di>
             }
         };
 
-        if v != ARRAY_FORMAT_VERSION {
-            try!(Err(de::Error::custom(format!("unknown array version: {}", v))));
-        }
+        try!(verify_version(v));
 
         let dim: Di = match try!(visitor.visit()) {
             Some(value) => value,
@@ -191,10 +202,7 @@ impl<A, Di, S> Visitor for ArrayVisitor<S,Di>
             match key {
                 ArrayField::Version => {
                     let val = try!(visitor.visit_value());
-                    if val != ARRAY_FORMAT_VERSION {
-                        let err_msg = format!("unknown array version: {}", val);
-                        try!(Err(de::Error::custom(err_msg)));
-                    }
+                    try!(verify_version(val));
                     v = Some(val);
                 },
                 ArrayField::Data => {
--- a/src/array_serde.rs
+++ b/src/array_serde.rs
@@ -139,6 +139,46 @@ impl<A, Di, S> serde::de::Visitor for ArrayVisitor<S,Di>
 {
     type Value = ArrayBase<S, Di>;
 
+    fn visit_seq<V>(&mut self, mut visitor: V) -> Result<ArrayBase<S, Di>, V::Error>
+        where V: serde::de::SeqVisitor
+    {
+        let v: u8 = match try!(visitor.visit()) {
+            Some(value) => value,
+            None => {
+                try!(visitor.end());
+                return Err(serde::de::Error::invalid_length(0));
+            }
+        };
+
+        let dim: Di = match try!(visitor.visit()) {
+            Some(value) => value,
+            None => {
+                try!(visitor.end());
+                return Err(serde::de::Error::invalid_length(1));
+            }
+        };
+
+        let data: Vec<A> = match try!(visitor.visit()) {
+            Some(value) => value,
+            None => {
+                try!(visitor.end());
+                return Err(serde::de::Error::invalid_length(2));
+            }
+        };
+
+        try!(visitor.end());
+
+        if v != ARRAY_FORMAT_VERSION {
+            try!(Err(serde::de::Error::custom(format!("unknown array version: {}", v))));
+        }
+
+        if let Ok(array) = ArrayBase::from_shape_vec(dim, data) {
+            Ok(array)
+        } else {
+            Err(serde::de::Error::custom("data and dimension must match in size"))
+        }
+    }
+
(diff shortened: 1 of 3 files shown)


src/array_serde.rs
```diff
@@ -175,7 +175,6 @@ impl<A, Di, S> serde::de::Visitor for ArrayVisitor<S,Di>
         };
 
         if let Ok(array) = ArrayBase::from_shape_vec(dim, data) {
-            try!(visitor.end());
```
bluss: Since the docs say "This signals to the MapVisitor that the Visitor does not expect any more items.", it seems strange to call end before making the missing field errors?

bluss:
I want to move the serialization tests to their own crate in the same repo, so that the tests can run in travis but not as part of the regular testsuite. That moves serde_json and rmp-serde out of dev-dependencies for ndarray itself.

src/array_serde.rs
bluss: ok, then we can do it like they do.

bluss:
And not ndarray-tests, but yet another different crate.

SuperFluffy:
Do you want this in this PR or in a future one?

src/array_serde.rs
```diff
@@ -139,21 +139,61 @@ impl<A, Di, S> serde::de::Visitor for ArrayVisitor<S,Di>
...
         let mut data: Option<Vec<A>> = None;
         let mut dim: Option<Di> = None;
 
-        loop {
-            match try!(visitor.visit_key()) {
-                Some(ArrayField::Version) => { v = Some(try!(visitor.visit_value())); },
-                Some(ArrayField::Data) => { data = Some(try!(visitor.visit_value())); },
-                Some(ArrayField::Dim) => { dim = Some(try!(visitor.visit_value())); },
-                None => { break; },
+        while let Some(key) = try!(visitor.visit_key::<ArrayField>()) {
```
dtolnay: This turbofish isn't required.

src/array_serde.rs
```diff
@@ -139,21 +139,61 @@ impl<A, Di, S> serde::de::Visitor for ArrayVisitor<S,Di>
...
+            Some(value) => value,
+            None => {
+                try!(visitor.end());
+                return Err(serde::de::Error::invalid_length(2));
+            }
+        };
+
+        try!(visitor.end());
+
+        if v != ARRAY_FORMAT_VERSION {
```
dtolnay: I would check this immediately after deserializing `v`, in case other formats use different fields. Same thing in `visit_map` if possible.

src/array_serde.rs
```diff
@@ -139,31 +139,78 @@ impl<A, Di, S> serde::de::Visitor for ArrayVisitor<S,Di>
...
             None => try!(visitor.missing_field("v")),
         };
 
-        if v != ARRAY_FORMAT_VERSION {
-            try!(Err(serde::de::Error::custom(format!("unknown array version: {}", v))));
-        }
-
         let data = match data {
             Some(data) => data,
             None => try!(visitor.missing_field("data")),
```
bluss: These missing field calls should be replaced by calls to `Error::missing_field`

src/array_serde.rs
```diff
@@ -91,6 +91,17 @@ impl<S, Di> ArrayVisitor<S, Di> {
     pub fn new() -> Self {
         ArrayVisitor { _marker_a: PhantomData, _marker_b: PhantomData, }
     }
+
+    pub fn verify_version<E>(v: u8) -> Result<(), E>
```
bluss: Thanks. This is best as a free function actually, since it doesn't use the S, Di type parameters.

bluss:
Thank you!

And thanks @dtolnay for all the help

Issue #234, opened 2016-11-14
ZST soundness challenge


bluss:
Find an ndarray test that fails if it is rewritten to use ZST (Zero sized type) array elements. For example the value `()`. If you find one, report a bug.

floatn:
Not sure if it's a ZST problem,
```Rust
let v1 = Vec::<i32>::new();
let A = Array::<i32, _>::from_elem((), 1i32);
let B = Array::<i32, _>::from_vec(v1);
```
println gives different value and size for those two.

bluss:
@floatn Thanks for looking into this. That's expected, `()` is the dimension description for a 0-dimensional array (always has 1 one element ~ isomorphic with a scalar). So it's expected that `A` and `B` both have `.len()` one, but `A` has ndim 0 and `B` has ndim 1.

This issue is for ZST array elements (the element type is `i32` which is not a ZST in your snippet).

floatn:
Guess I get it a bit better now. So, the right way to construct ZST Array is like this?
```Rust
let A = Array::<(), _>::from_elem((), ());
```
Also, shouldn't `size_of_val()` return 0 for a ZST?

bluss:
yes. `size_of_val(&())` should be zero yes.

`Array3::<()>::default((2, 3, 4));` is an array of zst elements, too.

bluss:
And one can make new such types easily of course:

```rust
struct Empty;
```

floatn:
thread 'test_swap' panicked at 'attempt to divide by zero', src/data_traits.rs:127
```Rust
#[test]
fn test_swap() {
    let mut a = arr2(&[[();3];3]);

    let b = a.clone();

    for i in 0..a.rows() {
        for j in i + 1..a.cols() {
            a.swap((i, j), (j, i));
        }
    }
    assert_eq!(a, b.t());
}
```

bluss:
You found something, thanks, that's a bug!

floatn:
one more 'division by size' at 'src/data_traits.rs:93'
```Rust
let C = RcArray::<(), _>::default((3, 4));
let mut D = C.clone();
for _ in D.iter_mut() {}
```

bluss:
Thank you. Fixed in master

bluss:
Ok, we consider this fixed now, new bugs can be reported if any are found.

PR #247, opened 2016-11-25
Fix two bugs in Array::clone

bluss:
You'd think that a simple function would not have bugs.

1. Did not handle the ZST case at all. See issue #234, this bug was found by @floatn
2. Did not move the pointer's relative offset correctly (applies when the array had a view of its data storage that puts the first element away from the first element of the storage). This is a more serious bug, IMO.

### DIFF
--- a/src/data_traits.rs
+++ b/src/data_traits.rs
@@ -8,7 +8,7 @@
 
 //! The data (inner representation) traits for ndarray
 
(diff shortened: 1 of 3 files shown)


bluss:
Will be pushed manually. Thank you @floatn for finding this bug and leading me to the next one.

PR #250, opened 2016-12-14
Implement rayon's ParallelIterator on a wrapper around AxisIter, AxisIterMut

### LINKED ISSUE #248: Implement Rayon parallel iterator traits for Axis Iter

bluss:
Fixes #248


bluss:
I've reduced scope again to just axis iter (and mut). We don't want to impl this for `.iter()/.iter_mut()` until the parallel iterator can customize their inner loop in rayon (something along `fold` or even `fold_ok` as the inner loop).

bluss:
Replaced by  #252

PR #252, opened 2016-12-21
Implement parallelization using rayon

### LINKED ISSUE #248: Implement Rayon parallel iterator traits for Axis Iter

bluss:
- IntoParallelIterator for `ArrayView`, `ArrayViewMut`, `&Array` and `&mut Array`
- IntoParallelIterator for `AxisIter` and `AxisIterMut`

The axis iterators are straightforward. They are indexed and exact length, and in order.

The element iterators have no guaranteed order (so that we can split the array where it's best for memory locality). The inner loop can not be customized yet, but will be when upstream rayon releases `fold_with` (this is important when the operation is cheap and wants to be autovectorized).

Fixes #248 and much more


bluss:
@pegasos1 even though implementing these general interfaces first seem like the best starting point, the whole thing already feels very complex. User needs to pick between using the built in higher order methods like `map_inplace` (that exist because iterators are not as efficient as a custom loop inside map inplace), and using these parallel iterators.

bluss:
To go further, the higher order methods themselves can be parallelized.

- `map: array f -> array` Needs custom collect
- `map_inplace` can use this PR internally
  - Possible sketch API: `parallel(&mut a).map_inplace(|x| *x = x.exp())`
  - Possible sketch API: `a.map_inplace(parallel(|x| *x = x.exp()))`
  - I favour the second one. It can be extended with options later (`parallel(f).max_threads(2)`).

Much later: Parallelize array zip.

bluss:
Superseded for now rust-ndarray/pull/264

PR #255, opened 2016-12-23
Implement .par_map_inplace() and .par_map(f) -> Array

bluss:
Successor to #252


bluss:
rust-ndarray/pull/264 adopts most of this, in a separate crate. `par_map` next, I guess.

PR #266, opened 2017-01-24
Add RcArray::into_owned() -> Array

bluss:


bluss:
A more generic version of this must wait.

PR #271, opened 2017-02-21
Update to serde 0.9

### LINKED ISSUE #270: Update to serde 0.9
https://github.com/serde-rs/serde/releases/tag/v0.9.0

dtolnay:
Fixes #270.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,7 +41,7 @@ blas-sys = { version = "0.6.5", optional = true, default-features = false }
 matrixmultiply = { version = "0.1.13" }
 
 [dependencies.serde]
-version = "0.8.20"
+version = "0.9"
 optional = true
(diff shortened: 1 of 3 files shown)


bluss:
Thank you!

@SuperFluffy this should be good to go as it is right?

SuperFluffy:
Yes, I think so! :-)

bluss:
Finally released in ndarray 0.8.0. I was a bit wary of the bump, but it should be ok. It's how it is. We have serde and num as public dependencies and when they bump, we bump. (We're in triple pace that way). Thanks to serde we are dropping rustc-serialize soon enough, so it is not a factor.

dtolnay:
I filed #274 to consider an alternative approach that mitigates the triple pace situation.

PR #280, opened 2017-03-23
BUG: Fix blas + axpy by only using it for certain cases

bluss:
Only use blas axpy for 1D arrays or contiguous arrays.

### DIFF
--- a/benches/bench1.rs
+++ b/benches/bench1.rs
@@ -457,26 +457,11 @@ fn iadd_2d_strided(bench: &mut test::Bencher)
     });
 }
 
-const SCALE_ADD_SZ: usize = 64;
-
 #[bench]
 fn scaled_add_2d_f32_regular(bench: &mut test::Bencher)
 {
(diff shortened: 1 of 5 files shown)


bluss:
cc @SuperFluffy

SuperFluffy:
Other than the negative and zero stride case, did you find anything that would prevent using blas for non-contiguous but equispaced/compatible arrays?

bluss:
It should be fine. Equispaced stride doesn't cover one case though, which is F1 + F2 (both arrays are 2+ dimensions, same shape, but in fortran memory order).

bluss:
Changed to a revert since it broke existing uses of is_contiguous (for negative stride arrays). It's my responsibility to decide on merges, so it's my responsibility there is a bug.

bluss:
A clearer thought on the situation is that we have two input arrays, and we need to figure out if they are separately blas compatible, but also that they are memory order compatible with each other. Same problem that zip needs to solve.

PR #287, opened 2017-03-26
Add n-ary Zip object

### LINKED ISSUE #231: Add “n-ary” zip
For given arities (2, 3, 4, ..) implement zip2, zip3, etc functions.

https://github.com/bluss/rust-ndarray/blob/master/examples/zip_many.rs

bluss:
The Zip object represents a lock step function application over *n* input arrays (where n is fixed by trait impls, so 1-6 input arrays in the current state of this PR).

### From its Documentation

> Lock step function application across several arrays or other producers.
>
> Zip allows matching several arrays to each other elementwise and applying a function over all tuples of elements (one element from each input at a time).
>
> In general, the zip uses a tuple of producers (Producer trait) that all have to be of the same shape. The Producer implementation defines what its element type is (for example if it's a shared reference, mutable reference or an array view etc).
>
> If all the input arrays are of the same memory order the zip performs much better and the compiler can usually vectorize the loop.
>
> The order elements are visited is not specified. The producers don’t have to have the same element type.
>
> See also the azip!() macro which offers a convenient shorthand to common ways to use Zip.

```rust
use ndarray::Zip;
use ndarray::Array2;

type M = Array2<f64>;

let mut a = M::zeros((64, 64));
let b = M::zeros((64, 64));
let c = M::zeros((64, 64));
let d = M::zeros((64, 64));

Zip::from(&mut a).and(&b).and(&c).and(&d).apply(|w, &x, &y, &z| {
    *w += x + y * z;
});
```

### Splitting / Parallelization

The zip object is designed to be splittable so it should be easy to augment it with a rayon-enhanced `par_apply`.

#### Broadcasting

+ We could broadcast by default, or have it as a separate method (`.and_broadcast()`).

#### Syntax

This is full of pointless repetition!

```rust
Zip::from(&mut a).and(&b).and(&c).and(&d).apply(|w, &x, &y, &z| {
    *w += x + y * z;
});
```

Of course, we already have shorthands for the common operations like  `a += &b` and so on, so we can afford to be a bit verbose in exchange for flexibility here.

We should try to use a macro to fix this anyway.

### Performance

The contiguous cases look good, no problem there. The compiler can autovectorize for us across many inputs. There's a 20% overhead in the add_2d_zip_strided benchmark that I looked at. It's related to how .zip_mut_with has the inner loop fourfold (something like that) unrolled by the compiler, and that doesn't happen in `Zip` for some reason.

### Acknowledgement

We introduce @cuviper's PrivateMarker that he uses in rayon.

Fixes #231


bluss:
New macro.

The equivalent of


```rust
Zip::from(&mut a).and(&b).and(&c).and(&d).apply(|w, &x, &y, &z| {
    *w += x + y * z;
});
```

is 

```rust
array_zip!(mut a, b, c, d in { *a += b + c * z });
```

Ampersands missing here is a feature. We want less clutter.


or with explicit expressions:

```rust
array_zip!(mut w @ [&mut a], x @ [&b], y @ [&c], z @ [&d] in { *w += x + y * z });
```

cuviper:
> Acknowledgement
> 
> We introduce @cuviper's PrivateMarker that he uses in rayon.

It was originally [your suggestion](https://github.com/nikomatsakis/rayon/pull/188#issuecomment-269636395), implemented and acknowledged [here](https://github.com/nikomatsakis/rayon/pull/214) -- thanks! 😄

bluss:
Lol! I had in the back of my mind that I discussed this in the rayon issues and that I saw a proposal there that was better than my idea. So then I went over there now and didn't find the discussion but I could easily just take the implementation from the tree..

Robbepop:
I really like this n-ary Zip. Could make my ndarray based code much cleaner and maybe even improve performance!

bluss:
Yes! The binary case we can't improve much upon, the existing `array.zip_mut_with(&b, f)` is already good for the common cases. However, for 3, 4, 5, etc operands we can improve.

bluss:
Ok. parallelization changes are factored out of this PR (need rayon git). Time to go!

Robbepop:
I have successfully used this new zip facility in order to improve the performance of [Prophet](https://github.com/Robbepop/prophet) that can be seen [here](https://github.com/Robbepop/prophet/blob/master/src/neural_net.rs#L158).

However at [another part](https://github.com/Robbepop/prophet/blob/master/src/neural_net.rs#L177) of the code I was not able to replace the current `multizip` based functionality with your new `ndarray::Zip`.
Tests are failing for the commented out part here.

bluss:
Thanks for the report, I'll have to look at the failing test. What happens? It computes the wrong result?

One difference to multizip is that multizip stops at the shortest input, while azip/Zip needs all inputs to be the same size.

There is no OuterIter in ndarray anymore (it's the same as AxisIter). There are missing implementations otherwise (for example missing for AxisChunksIter{,Mut}).

bluss:
ndarray should output a more clear panic for this case.

Robbepop:
> One difference to multizip is that multizip stops at the shortest input, while azip/Zip needs all inputs to be the same size.

In the failing case [this](https://github.com/Robbepop/prophet/blob/master/src/neural_net.rs#L173) assertion holds true and so it zips arrays with unequal length together. So I think it is time to fix this design failure on my side, ... finally. :D

bluss:
In this case one can just slice gradients to be one shorter for the loop, that preserves the same behavior. If you find any bugs that result in the wrong value being calculated, those are very important. There might be bugs like that since the code is new.

Robbepop:
I have implemented slicing for the arrays that were longer than the others and now everything works as intended according to my tests! :)

Robbepop:
However, no performance gains for the 2nd rewrite [here](https://github.com/Robbepop/prophet/blob/master/src/neural_net.rs#L173).

bluss:
Ok, are they always very short vectors? The activation function might be dominating everything else anyway.

Robbepop:
My [bench](https://github.com/Robbepop/prophet/blob/master/src/benches/neural_net.rs#L3) neural network has 10 hidden layers with 1000 neurons each.
So vectors are huge. I will benchmark the thing without activation function but the used activation functions are pretty cheap as far as I know.

bluss:
But with only 1 output, the length of that particular loop is always 1, right?

This one https://github.com/Robbepop/prophet/blob/master/src/neural_net.rs#L178

bluss:
Looks like the predict bench spends 90% of its time in matrix-vector multiplication.

The problem with the activation functions is that they don't vectorize (like exp and tanh turn into function calls per element). However that's not what is using the time here.

Issue #276, opened 2017-03-14
Implement non-axis windows and chunks iterators


Robbepop:
Already implemented features and todos:
----

- [x] `ExactChunks` and `ExactChunksMut` aswell as their `NdProducer` impl
- [ ] `RaggedChunks` or simply `Chunks` aswell as their `NdProducer` impl
- [x] `Windows`
- [x] `NdProducer` impl for `Windows`

----

Implement n-dimensional `windows` and `chunks` iterators.

Windows-iterators are especially useful for **convolutional neural networks** while chunk iterations may prove useful for paralell algorithms.

N-Dimensionality for both iterators use N-dimensionally shaped items.

For example for a 1-dimensional Array the `windows` and `chunks` items are also 1-dimensional.
For a 2-dimensional Array the `windows` and `chunks` items are 2-dimensional, and so on.

For `windows` iterator it could be useful to support two different kinds of `windows` iterators.
One for "valid" windows only that iterate only over windows within a valid region.
For example in a valid-windows iterator over a `10x10` shaped 2d-array and window sizes of `4x4` the iterator would iterate totally over `(10-4+1)x(10-4+1)` items of size `4x4`.
A full-windows iterator would also iterate over the windows with invalid elements (on the edges of the array) and would simulate missing elements with zero elements.

Semi-nice visualization of n-dimensional [chunks](http://ufldl.stanford.edu/tutorial/images/Pooling_schematic.gif) and [windows](http://ufldl.stanford.edu/tutorial/images/Convolution_schematic.gif) iterators.

A similar approach could be used for n-dimensional chunks that iterate beyond the array sizes and replace missing/invalid array elements with zero elements.

bluss:
That sounds good, and I love a good illustration to go together with expressing an idea.

One concern I have here is that it's probably still not an efficient way to implement convolution, to go through this.

Robbepop:
> That sounds good, and I love a good illustration to go together with expressing an idea.
Thanks! :)

> One concern I have here is that it's probably still not an efficient way to implement convolution, to go through this.

True! For performance the most promising implementation for convolution is to use a GPGPU approach. But what if one doesn't want to go that far? The main question for this issue for me is, if it is possible to provide such iterators with equal performance as extern definitions. But as far as I can imagine such iterators would perform much better if implemented internally in this crate with hidden data of `ArrayBase`.

[...]

PR #291, opened 2017-03-30
Add .whole_chunks_mut()

bluss:
- Fix bug in .whole_chunks() for IxDyn (did not check chunk dimensionality)
- Add macro for implementing NdProducer based on an array view
- Add `.whole_chunks_mut()` producer and intoiterator
- Show memory layout description in an array's debug fmt


bluss:
cc @Robbepop

This is related to #276, but just the mutable corresponding thing to whole chunks.

PR #301, opened 2017-04-03
Reimplement IxDyn, with no allocation for the small dimension case

### LINKED ISSUE #285: Reimplement IxDyn
It's using a Vec right now; it should be simplified.

- Use “small vec” optimization; no allocation for small dimensions
- Cap maximum number of axes to some limit (u16?)

bluss:
- Current parameter for inline case is CAP = 4, above which the dimension is stored as a `Box<[T]>`.
- Breaking change where `Dim<Vec<Ix>>` as a dimension is replaced by `Dim<IxDynImpl>`. Vec and slices still implement `IntoDimension → IxDyn` the same way, though.
- Performance improvements for usage of `IxDyn` across the board, maybe not as much as you would think. (For example: Binary operations on arrays don't care about dimension type if both operands are contiguous, so then we can't improve that.)

Fixes #285

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,6 +46,9 @@ matrixmultiply = { version = "0.1.13" }
 version = "0.9"
 optional = true
 
+[dev-dependencies]
+defmac = "0.1"
+
 [features]
 blas = ["blas-sys"]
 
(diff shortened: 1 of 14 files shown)


bluss:
Ok, beta ICE we see is fixed upstream <3 rust-lang/rust/pull/40813

PR #304, opened 2017-04-05
Implement breaking changes for next version

bluss:
- Remove Ix, Ixs from prelude (use usize, isize or import directly)
- Encapsulate Vec/RcVec
- Replace `.inner_iter()` with `.inners()`, `.genrows()`, `.gencolumns()`
  + Awkward names!! Because `.rows()` is already the row count, and `.row_iter` is not applicable -- `.genrows()` is an NdProducer + IntoIterator
- Private marker on Data and Dimension
- Stop using # flag in formatting
- Require Rust 1.15
- Add Item associated type to IntoNdProducer

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ sudo: required
 dist: trusty
 matrix:
   include:
-    - rust: 1.13.0
+    - rust: 1.15.0
       env:
        - FEATURES='test'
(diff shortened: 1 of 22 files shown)


bluss:
Feel free to find a better name for generalized rows (`.genrows()`!) before the release.

PR #306, opened 2017-04-06
Added basic implementation of windows iterators

Robbepop:
**Features:**
- `Windows` struct based on `BaseIter`
- free function `fn windows` used internally, similar to `fn exact_chunks`
- added to ndarray trait `fn windows(&self, ...)` for `a.windows(Dim(...))` usage
- also include some tests that also test for edge cases
- maybe some useful docs are still missing (no time, yet)

Addresses issue #276.


bluss:
Thanks, it looks good.

1. I guess the macro_use on si can be removed since `s![]` is not used anymore.
2. Same problem I found in my chunk code, needs to check that the window's ndim is the same as the view's ndim.  https://github.com/bluss/rust-ndarray/blob/master/src/iterators/chunks.rs#L127-L130  One can use ndassert, it's an assertion that is always enabled but has a long output with debug assertions.
3. We want this to be an NdProducer too, I think it fits into the macro over in the new and better chunk code. I can fix that up after merge.
4. Tests are failing due to the prerelease version of ndarray in master, don't worry.

Good tests so it looks good.

The whole ndproducer thing has really taken off since we started discussing this, so there hasn't been a clear goal to work towards, shifting all the time :D  Now that we have an Indices ndproducer we know that we can in theory produce anything we want as an ndproducer (including the ragged edges chunks, if we need that).

Robbepop:
Thanks for the fast feedback.

Should be fixed now. :)

bluss:
Thanks! the ndarray tests themselves pass, so we can merge.

src/iterators/windows.rs
```diff
@@ -0,0 +1,55 @@
+
+use imp_prelude::*;
+use IntoDimension;
+
+pub struct Windows<'a, A: 'a, D> {
+    iter  : ::iter::Iter<'a, A, D>,
```
bluss: alignment like this, it's a minor issue. I see what it's trying to do, but I prefer to keep this consistent and not do this. Such small things can always be fixed up afterwards, so I try to not focus on them. But it's good to mention, for the future.

bluss:
Quick plug: #rust-sci is a new irc channel we are trying to start and get going. One more rust channel!

src/iterators/windows.rs
Robbepop: Okay good to know. I just made it out of habit as I try to simply stick to the project's conventions. :)
I wish one day we can automate this formatting.^^

Robbepop:
What is #rust-sci about? gnu-scientific-library bindings?

bluss:
We wanted it to be everything about numeric and scientific stuff in Rust

PR #308, opened 2017-04-09
Add general matrix-vector multiplication

### LINKED ISSUE #289: Implement general_matvec_mul
As far as I can imagine ndarray's [general matrix multiplication](http://bluss.github.io/rust-ndarray/master/ndarray/linalg/fn.general_mat_mul.html) is based on [this](https://crates.io/crates/matrixmultiply) crate which has very generalized parameters.

The ndarray `general_mat_mul` function only works for 2-dimensional arrays (matrices) but it should be pretty easy to provide a `general_matvec_mul` function for matrix-vector-multiplication of a **n**x**k** Matrix and a **k**-sized Vector that is reinterpreted under the hood as a **k**x**1** Matrix. The returned **n**x**1** Matrix should be returned as **n**-sized Vector.

```rust
pub fn general_matvec_mul<A, S1, S2, S3>(
        alpha: A,
        a: &Array2<S1, Ix2>,
        b: &Array1<S2, Ix1>,
        beta: A,
        c: &mut Array1<S3, Ix1>)
    where
        S1: Data<Elem=A>,
        S2: Data<Elem=A>,
        S3: DataMut<Elem=A>,
        A: LinalgScalar
```

bluss:
- Add general matrix-vector multiplication “gemv” including BLAS bindings.
- The existing gemv `A.dot(&x)` uses the general function, which means that it
  too gains the unrolled floating point loop, improving performance for f32, f64.

Also fix a bug in BLAS bindings for gemm with column matrices (dimension n × 1).

Benchmarks show (not all included) that OpenBLAS doesn't redirect to gemv for column matrices in gemm either, so we don't. We let the user know their data and pick the corect out of gemm and gemv up front. (Both work with columns, but gemm will be slower).

Fixes #289

### DIFF
--- a/benches/gemv.rs
+++ b/benches/gemv.rs
@@ -1,0 +1,42 @@
+#![feature(test)]
+
+extern crate test;
+use test::Bencher;
+
+extern crate ndarray;
+use ndarray::prelude::*;
+
+use ndarray::linalg::general_mat_vec_mul;
+
+#[bench]
+fn gemv_64_64c(bench: &mut Bencher) {
+    let a = Array::zeros((64, 64));
+    let (m, n) = a.dim();
+    let x = Array::zeros(n);
+    let mut y = Array::zeros(m);
+    bench.iter(|| {
+        general_mat_vec_mul(1.0, &a, &x, 1.0, &mut y);
+    });
+}
+
+#[bench]
+fn gemv_64_64f(bench: &mut Bencher) {
+    let a = Array::zeros((64, 64).f());
+    let (m, n) = a.dim();
+    let x = Array::zeros(n);
+    let mut y = Array::zeros(m);
+    bench.iter(|| {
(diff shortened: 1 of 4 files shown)


bluss:
cc @Robbepop. The fallback implementation is simple, but it has good performance for smallish arrays, uses manually unrolled dot product. Benchmarks show it can be improved 5-10% by lifting the check for contiguous data out of the loop, but we start simple.

Robbepop:
Sounds nice to me! :D 

I will try this out in my prophet crate as soon as possible! :)

bluss:
New `Zip` really is very versatile, I'm glad we have it. I should run the benchmarks to see if outer_iter() vs genrows() makes any difference at all..

bluss:
Ok, outer_iter and genrows are exactly the same. That's good, both abstractions have the same cost..

bluss:
The external crate tests fail due to prerelease version stuff, but the important tests/oper.rs tests pass.

PR #309, opened 2017-04-09
parallel: Fix .collect() for the element par iters

bluss:
- They must not define opt_len to return Some (reserved for indexed iterators).
- Added test that shows problem and verifies the fix


bluss:
tests fail due to prerelease version stuff..

PR #334, opened 2017-07-21
Add support for scalar operation on bignums

### LINKED ISSUE #333: impl ScalarOperand for {BigInt, BigUint, Ratio}
It would be very useful to implement ScalarOperand for BigInt, BigUint and Ratio so that ndarrays can be multiplied by bigints, using a "num-bigint" feature flag.

winsut:
Adds support for scalar operations on bignums, and also introduces more macros to avoid code duplication.

It would be better to have more generic impls for Ratio (and Complex), but that would take significantly more code, so it's left as a future extension.

Fixes #333.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,6 +30,14 @@ default-features = false
 version = "0.1.32"
 default-features = false
 
+[dependencies.num-bigint]
+version = "0.1.32"
+default-features = false
+
+[dependencies.num-rational]
+version = "0.1.32"
+default-features = false
+
 [dependencies.itertools]
 version = "0.6.0"
 
--- a/src/impl_ops.rs
+++ b/src/impl_ops.rs
@@ -6,7 +6,13 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
+use std::mem;
 use num_complex::Complex;
+use num_bigint::{BigInt, BigUint};
+use num_rational::Ratio;
+use libnum::Zero;
+
+type BigRational = Ratio<BigInt>;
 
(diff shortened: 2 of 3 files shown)


bluss:
Hi, thanks.

Loose thoughts:

- This should be an optional feature
- The existing arithmetic operation impls are based on elements that are cheap to clone. They are really not very good implementations for bignums, lots of unnecessary cloning of the elements.

We can use this for example:

```rust
 /// Perform elementwise
 #[doc=$doc]
 /// between `self` and reference `rhs`,
 /// and return the result (based on `self`).
 ///
 /// If their shapes disagree, `rhs` is broadcast to the shape of `self`.
 ///
 /// **Panics** if broadcasting isn’t possible.
 impl<'a, A, S, S2, D, E> $trt<&'a ArrayBase<S2, E>> for ArrayBase<S, D>
     where A: Clone + $trt<A, Output=A>,
           S: DataMut<Elem=A>,
           S2: Data<Elem=A>,
           D: Dimension,
           E: Dimension,
 {
     type Output = ArrayBase<S, D>;
     fn $mth(mut self, rhs: &ArrayBase<S2, E>) -> ArrayBase<S, D>
     {
         self.zip_mut_with(rhs, |x, y| {
             *x = x.clone() $operator y.clone();
         });
         self
     }
 }
```

For bignums we should not clone both operands. It requires some work. It's a bit unsatisfactory indeed, we want to base the array's Add impl for scalars on by value operations, but more general ones on by reference implementations.

bluss:
closing since it's inactive.

I'd like to handle this with by-reference (not cloning) implementations and an opt-in bigint crate feature in ndarray.

PR #336, opened 2017-07-28
Impl `From<ArrayN>` for ArrayD, for N in 0..6

bobogei81123:
* Implement the From<ArrayN> trait for ArrayD to allow conversion
  into dynamic dimensional array.
* Add some tests.

### DIFF
--- a/src/dimension/dimension_trait.rs
+++ b/src/dimension/dimension_trait.rs
@@ -371,6 +371,11 @@ pub trait Dimension : Clone + Eq + Debug + Send + Sync + Default +
             .map_or(Axis(0), |ax| ax.axis())
     }
 
+    /// Convert the dimensional into a dynamic dimensional (IxDyn).
+    fn into_dyn(self) -> IxDyn {
+        IxDyn(self.slice())
+    }
+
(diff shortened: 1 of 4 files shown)


bluss:
I'd prefer to simply have a method for this. We can implement it once for ArrayBase, so that all array kinds (including views) support this conversion.

bobogei81123:
Ok, but since `ArrayBase<S, D>` only required `D: Dimension`, this means that we have to be able
to cast any `Dimension` into `IxDyn`.

bluss:
Converting any dimension to IxDyn is good. Thanks for the PR — it's a design I agree a lot with. I'll merge and might do adjustments on top of it.

As an aside, individual methods can require what they need. See for example [`.select()`](https://docs.rs/ndarray/0.10.0/ndarray/struct.ArrayBase.html#method.select) which has a where clause, adding extra requirements.

bluss:
Remaining travis failure is the upstream bug ICE that affects matrixmultiply. It has been fixed in Rust, just needs a new nightly.

bluss:
Now in version 0.10.1

PR #343, opened 2017-08-18
Method .into_dimensionality::<Dim>()

bluss:


bluss:
I wonder what you think of the design of this. IxDyn isn't succinct or nice to use, but we can make it nicer step by step. Also vote for names, `into_dimension`, `into_dimensionality`, `into_` something else?

PR #344, opened 2017-08-20
Use cargo cache for travis

bluss:


bluss:
Doesn't make a good difference.

mgeisler:
I wanted to submit the same change, but noticed that there is already a `travis-cache` branch :-D

In my projects, most of the CI time is taken up by compiling dependencies, so there the cache has reduced the time significantly.

Stupid question, did you try a second Travis build after letting the cache become populated? I only see [build #944](https://travis-ci.org/bluss/rust-ndarray/builds/266504997) for this PR.

PR #345, opened 2017-08-22
serde: Deserialize from bytes

Libbum:
I was implementing a [RON](https://github.com/ron-rs/ron) channel for some data and was unable to deserialize a ron object to an `Array3` type. Considering that there is no issue doing this through the `serde-1` option to json, messagepack, yaml etc. I figured it was an issue with the ron implementation and opened ron-rs/ron#52 on its tracker.

Turns out it was not an issue there, but an omission here.

This PR extends the `Deserialize` for `ArrayField` method to include the `visit_bytes` function, and adds tests to confirm that ron now deserializes correctly. All previous tests also pass, inferring that this addition doesn't alter any current pathways.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,10 @@ script:
       cargo build --verbose --features "$FEATURES" &&
       cargo test --verbose --features "$FEATURES" &&
       cargo test --manifest-path=parallel/Cargo.toml --verbose &&
-      cargo test --manifest-path=serialization-tests/Cargo.toml --verbose &&
+      if [[ "$TRAVIS_RUST_VERSION" == "1.18.0" ]]; then
+        cargo test --manifest-path=serialization-tests/Cargo.toml --verbose --no-default-features
+      else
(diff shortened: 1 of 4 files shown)


Libbum:
Ron fails to build on Rust v1.18, so perhaps it's best to drop the tests before merging.

bluss:
Looks good, let's merge when travis passes. Using ron for tests seems fine, as long as they don't break a working Rust on patch versions. Using it needs to exclude older Rust from using the ron tests. Serialization tests are their own crate so they can do what they want, for example cargo features.

Libbum:
Ron is now an default feature & optional dependency for the serialization-tests crate (dev-dependencies can't be optional). There's no way for cargo to test for rust versions AFAIK, so travis now checks the version and adds `--no-default-features` to the v1.18.0 case.

bluss:
Thanks. Feel free to ping me in a while if I haven't had time to publish a new version

Issue #349, opened 2017-09-01
Improve bounds check elision in indexing


bluss:
Note: We encourage higher level functionality including map, Zip, azip!() and so on before resorting indexing. Even `azip!` with an index a.k.a. `Zip::indexed` is a big improvement.

Still, it is a good benchmark to consider bounds check elision in the most trivial cases — assuming it will improve it in some less trivial cases too.

Set up a goal that the following examples compile without extra bounds check, because their loop counters are already in bounds.

```rust
pub fn test1d_1(a: &Array1<f64>) -> f64 {
    let mut sum = 0.;
    for i in 0..a.len() {
        sum += a[i];
    }
    sum
}

pub fn test2d(a: &Array2<f64>) -> f64 {
    let mut sum = 0.;
    for i in 0..a.rows() {
        for j in 0..a.cols() {
            sum += a[[i, j]];
        }
    }
    sum
}
```

bobogei81123:
Is this possible? I thought it is rust compiler's job to do this kind of optimization.

bluss:
Sure, it is possible to a degree. It would be to write Rust code that the compiler understands like we want it to understand it.

Sure, there are some road blocks. The following doesn't even remove the bounds checks in a test where `a` is a `&Vec<f64>`, but it does with `&[f64]`:

```rust
    for i in 0..a.len() {
        sum += a[i];
    }
```

so a component of it could be to fix that as a bug in rustc.

bluss:
To clarify, we will look for a way to implement the same thing but that the compiler optimises like we want. It requires some analysis of the current code, and experiments with actual results if we changet it. Current checked indexing is relatively high level and computes a cumulative poster offset ad an optional isize, no surprise the compiler does not see through it.

To be honest I'd probably work on the vector case first.. if it can't do it we don't have much chance.

bluss:
To clarify, we will look for a way to implement the same thing but that the compiler optimises like we want. It requires some analysis of the current code, and experiments with actual results. Current checked indexing is relatively high level and computes a cumulative pointer offset as an optional isize, no surprise the compiler does not see through it.

To be honest I'd probably work on the vector case first.. if it can't do it we don't have much chance.

bluss:
We can put up a lower goal to start.

Rewrite indexing code so that the following does not have a panic for bounds check:

```rust
pub fn test1d_single(a: &Array1<f64>, i: usize) -> f64 {
    if i < a.len() { a[i] } else { 0. }
}
```

PR #350, opened 2017-09-07
 Tweak bounds checked indexing code (Index, IndexMut)

bluss:
Tweak bounds checked indexing code (Index, IndexMut)

Tweak bounds checked indexing code so that the compiler understands to
remove the bounds check in some simple cases. See examples in the new
example file bounds_check_elim.rs

This helps in *some* cases where a conditional or loop conditional
already do the required bounds check. Simple examples are if i < len and
while i < len loops; the range iterator loop is a worse example, it
doesn't do this as successfully! Further works is needed with that.

* Improve: test1d_single, test1d_while, test1d_range, test2d_while, test2d_range
* Need more improvement: test1d_range, test2d_range

Example input code:

```rust
pub fn test1d_single(a: &Array1<f64>, i: usize) -> f64 {
    if i < a.len() { a[i] } else { 0. }
}
```

Codegen diff:

```diff
 test1d_single:
 	.cfi_startproc
-	push	rax
-.Lcfi0:
-	.cfi_def_cfa_offset 16
 	xorps	xmm0, xmm0
 	cmp	qword ptr [rdi + 32], rsi
-	jbe	.LBB0_3
+	jbe	.LBB0_2
+	mov	rax, qword ptr [rdi + 24]
 	imul	rsi, qword ptr [rdi + 40]
-	shl	rsi, 3
-	add	rsi, qword ptr [rdi + 24]
-	je	.LBB0_4
-	movsd	xmm0, qword ptr [rsi]
-.LBB0_3:
-	pop	rax
+	movsd	xmm0, qword ptr [rax + 8*rsi]
+.LBB0_2:
 	ret
-.LBB0_4:
-	call	_ZN7ndarray11arraytraits19array_out_of_bounds17h009d4a482e88d6aaE@PLT
 .Lfunc_end0:
 	.size	test1d_single, .Lfunc_end0-test1d_single
 	.cfi_endproc
```

Full codegen diff for the bounds check elim example: https://gist.github.com/bluss/e5d9b33a1eea36894534503b0e06341a/revisions?diff=split

Note that not all examples do so well.

### DIFF
--- a/src/arraytraits.rs
+++ b/src/arraytraits.rs
@@ -69,7 +69,10 @@ impl<S, D, I> Index<I> for ArrayBase<S, D>
     #[inline]
     fn index(&self, index: I) -> &S::Elem {
         debug_bounds_check!(self, index);
-        self.get(index).unwrap_or_else(|| array_out_of_bounds())
+        unsafe {
+            &*self.ptr.offset(index.index_checked(&self.dim, &self.strides)
+                                   .unwrap_or_else(|| array_out_of_bounds()))
+        }
     }
 }
 
@@ -84,7 +87,10 @@ impl<S, D, I> IndexMut<I> for ArrayBase<S, D>
     #[inline]
     fn index_mut(&mut self, index: I) -> &mut S::Elem {
         debug_bounds_check!(self, index);
-        self.get_mut(index).unwrap_or_else(|| array_out_of_bounds())
+        unsafe {
+            &mut *self.as_mut_ptr().offset(index.index_checked(&self.dim, &self.strides)
+                                                .unwrap_or_else(|| array_out_of_bounds()))
+        }
     }
 }
 
--- a/examples/bounds_check_elim.rs
+++ b/examples/bounds_check_elim.rs
@@ -1,0 +1,103 @@
+#![crate_type="lib"]
+
+// Test cases for bounds check elimination
+
+extern crate ndarray;
+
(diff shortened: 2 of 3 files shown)


bluss:
Related to #349

bluss:
It actually looks like this is another place where **avoiding** `Option<&T>` is good for performance. The compiler doesn't reason as well about that "optimized" option representation; it adds extra null checks.

* Old code: `Option<isize>` mapped to an `Option<&T>`, then unwrapping.
* New code: `Option<isize>.unwrap()` before making a `&T`

src/arraytraits.rs
```diff
@@ -84,7 +87,10 @@ impl<S, D, I> IndexMut<I> for ArrayBase<S, D>
     #[inline]
     fn index_mut(&mut self, index: I) -> &mut S::Elem {
         debug_bounds_check!(self, index);
-        self.get_mut(index).unwrap_or_else(|| array_out_of_bounds())
+        unsafe {
+            &mut *self.as_mut_ptr().offset(index.index_checked(&self.dim, &self.strides)
+                                                .unwrap_or_else(|| array_out_of_bounds()))
```
bluss: For all the @bluss and other hot shots out there, you need to remember that the mutable case is really tricky with RcArray, which is clone-on-write. This code is wrong, because it uses the stride before calling as_mut_ptr.

Issue #352, opened 2017-09-14
Add method to insert an axis


jturner314:
Currently, converting an `Array1` into a one-row or one-column `Array2` is quite awkward. This comes up often when doing linear algebra. The best way I've found is to use `into_shape`:

```rust
let long_array_name = array![1, 2, 3];
let col = {
    let len = long_array_name.len();
    long_array_name.into_shape((len, 1)).unwrap()
};
assert_eq!(col, array![[1], [2], [3]]);
```

I suggest adding something like `fn insert_axis(self, axis: Axis) -> ArrayBase<S, D::Larger>`, which would be equivalent to NumPy's `numpy.expand_dims(a, axis)`. This would also generalize to higher-dimensional arrays. The example would become:

```rust
let long_array_name = array![1, 2, 3];
let col = long_array_name.insert_axis(Axis(1));
assert_eq!(col, array![[1], [2], [3]]);
```

In my own code, I've implemented a custom trait on `Array1` to add `into_row` and `into_column` methods, but a more general `insert_axis` would be very useful. By the way, thanks for writing `ndarray`! I use it for a lot of my scientific code.

bluss:
Sounds good.

1. I think into_shape is already pretty good for this but it's far from perfect.
2. This has come up recently. [into_dimensionality()](https://github.com/bluss/rust-ndarray/pull/343) is strict, but we floated the idea of a weaker version. Something that creates the extra axes so that `.into_ndim::<Ix2>()` on an 1d array makes a 2d array just like that. I guess we need two version for choosing where to insert the axes
3. We don't need a new associated type, because we already have the inverse relation:

```rust
fn insert_axis<A, D>(self_: Array<A, D::Smaller>, axis: Axis) -> Array<A, D>
```

bluss:
That said, no reason to not add a new associated type if that's better. Then the Ix6 -> IxDyn thing is smooth, but I generally don't consider Ix6 to be an important use case.

jturner314:
2. `.into_ndim::<Ix2>()` is an interesting idea, and it would be sufficient for the `Ix1 -> Ix2` case in combination with the `.reversed_axes()` method. (Use `arr.into_ndim::<Ix2>()` to get a row, and `arr.into_ndim::<Ix2>().reversed_axes()` to get a column.)

[...]

PR #354, opened 2017-09-15
Add ArrayBase.insert_axis() and Dimension::Larger

jturner314:
This adds a method to ArrayBase to insert a new (length 1) axis. See #352 for the reasoning behind this feature. I modeled the implementation on the `.remove_axis()` method.

I originally tried implementing `.insert_axis()` using `Dimension::Smaller` for the input type (as suggested by @bluss). The signature was
```rust
impl<A, S, D> ArrayBase<S, D> where S: Data<Elem=A>, D: Dimension {
    pub fn insert_axis<D2: Dimension<Smaller=D>>(self, axis: Axis) -> ArrayBase<S, D2> where D2: InsertAxis {...}
}
```
The disadvantage of that approach was that the caller of `.insert_axis()` had to disambiguate the return type, since multiple types could implement `Dimension<Smaller=D>`. For example,
```rust
let a = arr1(&[1, 2, 3]);
let col = a.insert_axis(Axis(1)); // didn't compile; had to specify return type Array2<_>
println!("{:?}", col);
```
As a result, the cleanest method was to add the `Dimension::Larger` associated type, which makes the above example work without specifying the return type. The signature is then
```rust
impl<A, S, D> ArrayBase<S, D> where S: Data<Elem=A>, D: Dimension {
    pub fn insert_axis(self, axis: Axis) -> ArrayBase<S, D::Larger> where D: InsertAxis {...}
}
```

Please let me know what you think.


bluss:
Looks good. I'd shoot for not having any new trait in the public API here, and I'll try to resolve that too if I have time. feel free to ping me about this later.

jturner314:
That's a good point about avoiding adding an `InsertAxis` trait. I just pushed a new version. I expect that it's slightly less efficient since it creates a `Vec` for all cases, not just when `D::Larger` is `IxDyn`, but it's simpler and doesn't add a new trait to the public API.

jturner314:
I think if you wanted to keep the more efficient implementation, you could make `InsertAxis` private and provide an implementation for all types that implement `Dimension` (using the body of the `dim_insert_axis` function in dbe871edea073823f53ccd8dc0e3a90519dc0a53). Then, `insert_axis` for `Ix0`, `Ix1`, ... `Ix5` could still be separately implemented (as in 0edc15310ecce8c4f4950f45c7ca48c7cfb0add6). The disadvantage is that this would require more code than dbe871edea073823f53ccd8dc0e3a90519dc0a53.

bluss:
Sorry to send you on the goose chase here. I'd add a new method to the Dimension trait. That trait is locked down so that we can add stuff to it as we want. You can have a default impl and some specific items. In the end we want this to be so good so that we only go into using a Vec if we are in the non-short case for IxDynIml. It uses a short vector optimization nowadays.

jturner314:
Okay, I pushed a new version that adds `.insert_axis()` to the `Dimension` trait and provides specific implementations for the various dimension types. What do you think?

bluss:
Thanks a lot for this

PR #355, opened 2017-09-15
Add is_empty() to ArrayBase

### LINKED ISSUE #353: Add is_empty to ArrayBase
I found myself expecting this function while writing array code. Would this make sense to have? I'm totally willing to make a PR if you can point me to the best place to implement. I was thinking the best way would be to add a method to `Dimension` checking if any of the axes are 0, but maybe there is a better way?

iamed2:
Closes #353

### DIFF
--- a/src/impl_methods.rs
+++ b/src/impl_methods.rs
@@ -69,6 +69,11 @@ impl<A, S, D> ArrayBase<S, D> where S: Data<Elem=A>, D: Dimension
         self.dim[axis.index()]
     }
 
+    /// Return whether the array has any elements
+    pub fn is_empty(&self) -> bool {
+        self.len() == 0
+    }
+
     /// Return the number of dimensions (axes) in the array
     pub fn ndim(&self) -> usize {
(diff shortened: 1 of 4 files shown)


src/impl_methods.rs
```diff
@@ -59,6 +59,11 @@ impl<A, S, D> ArrayBase<S, D> where S: Data<Elem=A>, D: Dimension
         self.dim.size()
     }
 
+    /// Return whether the array has any elements
+    pub fn is_empty(&self) -> bool {
```
bluss: Thanks for adding this and tests.

Sorry to nitpick on such a small PR -- but `len` & `len_of` must be next to each other in the docs. So this method should move down a step.

iamed2: Oh no problem! That makes sense.

bluss:
Thanks!

iamed2:
Thank you!

bluss:
In ndarray 0.10.7

PR #356, opened 2017-09-16
Update ndarray-rand random function, format with rustfmt

v-shmyhlo:
Hello community, my very first contribution to open source, please feel free to criticize what i am doing so i could learn and become better contributor.
* formatted code with rustfmt
* random function updated to use `from_shape_fn` constructor, which i think is more suitable (please correct me if i am wrong)
* added explicit types to benches to better show the intent of using `F32`

### DIFF
--- a/ndarray-rand/src/lib.rs
+++ b/ndarray-rand/src/lib.rs
@@ -86,9 +86,7 @@ impl<S, D> RandomExt<S, D> for ArrayBase<S, D>
               R: Rng,
               Sh: ShapeBuilder<Dim=D>,
     {
-        let shape = shape.into_shape();
-        let elements = Vec::from_iter((0..shape.size()).map(move |_| dist.ind_sample(rng)));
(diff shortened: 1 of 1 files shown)


bluss:
Hi, thanks for your interest in ndarray!

Please hold off formatting for now -- the whole project wants to move to that relatively orderly, including using a rustfmt formatting configuration file in the repo. Also keep in mind to not mix feature commits with formatting commits. With that said, the feature change (using from_shape_fn) looks good to me, and it's good to submit that one alone. It might even be a performance improvement this way.

v-shmyhlo:
@bluss, I have updated PR according to your comment, thank you very much for descriptive answer!

bluss:
Thanks!

bluss:
This is now in ndarray-rand 0.6.1. Just thought I'd notice you. Normally we make very little fuss about releases though, just make new ones all the time.

PR #362, opened 2017-09-28
Fix indexing of zero-size nonzero-shape Fortran arrays

### LINKED ISSUE #361: Segfault in ArrayBase::from_shape_fn()
The following program causes a segfault with `ndarray` 0.10.8:

```rust
extern crate ndarray;

use ndarray::prelude::*;

fn main() {
    let a: Array2<f64> = ArrayBase::from_shape_fn((1, 0).f(), |_| 0.);
}
```

jturner314:
Fixes #361.

This is fix (and tests) for indexing a Fortran array that has zero size but has a shape with nonzero elements (e.g. shape `(1, 0)`). The bug was in `indices_iter_f()`, but this caused segfaults in `ArrayBase::from_shape_fn()`.

### DIFF
--- a/src/indexes.rs
+++ b/src/indexes.rs
@@ -212,7 +212,7 @@ pub fn indices_iter_f<E>(shape: E) -> IndicesIterF<E::Dim>
     let dim = shape.into_dimension();
     let zero = dim.zero_index();
     IndicesIterF {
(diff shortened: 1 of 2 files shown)


bluss:
Thanks for the fix and the test!

PR #363, opened 2017-09-28
Add tests for from_shape_fn() with Ix0

jturner314:
I just noticed that #362 had the unintentional side-effect of fixing `test_from_fn_f0` in this PR, so I thought it would be worth adding these tests too.

### DIFF
--- a/tests/array-construct.rs
+++ b/tests/array-construct.rs
@@ -45,6 +45,14 @@ fn test_uninit() {
 }
 
 #[test]
+fn test_from_fn_c0() {
+    let a = Array::from_shape_fn((), |i| i);
+    assert_eq!(a[()], ());
+    assert_eq!(a.len(), 1);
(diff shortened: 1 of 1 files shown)


bluss:
Awesome! Thanks

bluss:
With out the fix, the f test trips a debug assertion inside `from_vec_dim_stride_unchecked`: this assertion is not exactly easy to read. It also has false positives, only can catch some mistakes (it's an "unchecked" method so that's ok.)

```rust
---- test_from_fn_f0 stdout ----
        thread 'test_from_fn_f0' panicked at 'assertion failed: match dimension::can_index_slice(&v, &dim, &strides) {
    Ok(_) => true,
    Err(ref e) =>
    match e.kind() {
        ErrorKind::OutOfBounds => false,
        ErrorKind::RangeLimited => false,
        _ => true,
    },
}', src/impl_constructors.rs:305:8
```

PR #364, opened 2017-09-29
Add unchecked swap method (uswap)

jturner314:
This adds an unchecked swap method, which is necessary for performing in-place transposition of square matrices (i.e. actually moving the data around) without bounds checking. (Using `uget_mut` doesn't work for this case `::std::mem::swap(unsafe { a.uget_mut((i, j)) }, unsafe{ a.uget_mut((j, i)) })` because it would require the array to be mutably borrowed twice.)

### DIFF
--- a/src/impl_methods.rs
+++ b/src/impl_methods.rs
@@ -341,6 +341,24 @@ impl<A, S, D> ArrayBase<S, D> where S: Data<Elem=A>, D: Dimension
         }
     }
 
+    /// Swap elements *unchecked* at indices `index1` and `index2`.
+    ///
+    /// Indices may be equal.
+    ///
+    /// **Note:** only unchecked for non-debug builds of ndarray.<br>
+    /// **Note:** (For `RcArray`) The array must be uniquely held.
+    pub unsafe fn uswap<I>(&mut self, index1: I, index2: I)
+        where S: DataMut,
+              I: NdIndex<D>,
+    {
+        debug_assert!(self.data.is_unique());
+        arraytraits::debug_bounds_check(self, &index1);
+        arraytraits::debug_bounds_check(self, &index2);
+        let off1 = index1.index_unchecked(&self.strides);
+        let off2 = index2.index_unchecked(&self.strides);
+        std_ptr::swap(self.ptr.offset(off1), self.ptr.offset(off2));
+    }
+
     // `get` for zero-dimensional arrays
     // panics if dimension is not zero. otherwise an element is always present.
     fn get_0d(&self) -> &A {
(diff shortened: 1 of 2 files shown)


bluss:
Looks good. I've implemented divide and conquer square transposition at some point, it might be useful to you? It didn't need this method at least, maybe using more deviant tricks (we can get the raw pointer directly, too).

looking now, yes, one can use `uget_mut` if we cast the result to raw pointer (the borrow is gone that way).

```rust
#[inline(never)]
fn trans_base_case<A, O>(m: Ix, mut v: ArrayViewMut<A, Ix2>)
    where O: Order,
{
    for i in 0..m - 1 {
        if O::inner_contig() {
            unsafe {
                let mut ptr1: *mut _ = v.uget_mut((i, i + 1));
                for j in i + 1..m {
                    let ptr2: *mut _ = v.uget_mut((j, i));
                    mem::swap(&mut *ptr1, &mut *ptr2);
                    ptr1 = ptr1.offset(1);
                }
            }
        } else {
            unsafe {
                for j in i + 1..m {
                    let ptr1: *mut _ = v.uget_mut((i, j));
                    let ptr2: *mut _ = v.uget_mut((j, i));
                    mem::swap(&mut *ptr1, &mut *ptr2);
                }
            }
        }
    }
}
```

jturner314:
> I've implemented divide and conquer square transposition at some point, it might be useful to you?

Yeah, if you still have the code lying around somewhere, it would be helpful. I'm working on adding some functionality to `ndarray-linalg`, and I need to be able to convert from arbitrary `ndarray` layouts to column-major layout for passing to the Fortran LAPACK routines. For most matrices, the simplest thing to do is to allocate a new array and copy all the elements, but when possible for row-major square matrices, it would be nice to transpose the data in-place.

> one can use `uget_mut` if we cast the result to raw pointer (the borrow is gone that way)

Good point; I didn't think of bypassing the borrow checker that way. I need to reread the Rust book at some point. :) Since using pointers is an option, I don't think it's necessarily worth merging this PR.

bluss:
I think this is a worthwhile feature

bluss:
Hey, sorry for forgetting about this. This is the transpose code which is MIT/Apache-2.0 (c) bluss as usual and use as you wish. What's blocking putting it in ndarray itself is that it's only implementing square transpose (the easy part!) and of course the usual doubts if it's efficient enough.. but I think it is. https://gist.github.com/bluss/8221102b3a3941bb68deb07cbd96d9cc

bluss:
Thank you for a well written PR!

jturner314:
You're welcome! Thank for providing the gist for the cache-friendly in-place transpose.

PR #365, opened 2017-10-01
Fix .dot() with zero-length non-float arrays

jturner314:
Before, matrix multiplication would panic (on debug builds) or segfault (on release builds) for non-`f32`/`f64` arrays that had zero rows or zero columns. The issue was that `linalg::impl_linalg::mat_mul_general()` always tried to access the `(0,0)` element in the output array `c`, even if `c` had zero length. This commit fixes the issue and adds a test for this case.

### DIFF
--- a/tests/oper.rs
+++ b/tests/oper.rs
@@ -1,3 +1,4 @@
+#[macro_use] extern crate defmac;
 #[macro_use(s)] extern crate ndarray;
 extern crate num_traits;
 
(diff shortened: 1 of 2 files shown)


bluss:
Thank you! It's the fallback code that's the problem isn't it, and not matrixmultiply? We could move the check down to the fallback maybe.

jturner314:
Yes, the fallback code is the issue, particularly [this line](https://github.com/bluss/rust-ndarray/blob/0.10.9/src/linalg/impl_linalg.rs#L451). I just pushed a commit that moves the check into the fallback case. I can squash the two commits in this PR if you'd like.

tests/oper.rs
```diff
@@ -267,15 +268,15 @@ fn fold_and_sum() {
 }
 
 fn range_mat(m: Ix, n: Ix) -> Array2<f32> {
-    Array::linspace(0., (m * n - 1) as f32, m * n).into_shape((m, n)).unwrap()
+    Array::linspace(0., (m * n) as f32 - 1., m * n).into_shape((m, n)).unwrap()
 }
 
 fn range_mat64(m: Ix, n: Ix) -> Array2<f64> {
-    Array::linspace(0., (m * n - 1) as f64, m * n).into_shape((m, n)).unwrap()
+    Array::linspace(0., (m * n) as f64 - 1., m * n).into_shape((m, n)).unwrap()
```
bluss: I was confused about why the - 1 moved here -- it's actually a the same result either way, right?

bluss:
Thanks a lot for your bugfixes in ndarray

bluss:
!

tests/oper.rs
jturner314: `m` and `n` are of type `usize`, so if `m` or `n` is zero, then `m * n` is `0usize`. `0usize - 1` panics due to underflow. By casting the product (`m * n`) to `f64` before subtracting `1`, there is no underflow. This change makes `range_mat64` work for cases where `m` or `n` is zero.

PR #366, opened 2017-10-09
Test thinlto in travis (Travis build only)

bluss:
- Will not be merged, only for testing


bluss:
Tests passed (except for cancelled builds)

PR #370, opened 2017-10-19
Implement product of row vector and matrix

jturner314:
This adds an implementation for a row vector times a matrix. It exploits the identity *AB* = (*B*<sup>T</sup>*A*<sup>T</sup>)<sup>T</sup> to reuse the existing implementation for a matrix times a column vector.

Edit: Note that the 1-D times 2-D case being interpreted as a row vector times a matrix is the same behavior as NumPy.


bluss:
Thanks a lot for this

Issue #371, opened 2017-10-19
Lifetime parameter in the implementation of `Index` trait for `ArrayView` is wrong


termoshtt:
The following code does not work:

```rust
extern crate ndarray;
use ndarray::*;

fn get_ref<'a>(a: &'a Array<i32, Ix1>, i: usize) -> &'a i32 {
    let v: ArrayView<'a, i32, Ix1> = a.view();
    &v[i]  // `v` does not live long enough
}

fn main() {
    let a = array![1, 2, 3];
    get_ref(&a, 2);
}
```

and fixed version works:

```rust
extern crate ndarray;
use ndarray::*;

fn get_ref<'a>(a: &'a Array<i32, Ix1>, i: usize) -> &'a i32 {
    let v: ArrayView<'a, i32, Ix1> = a.view();
    index2(&v, i)
}

fn index2<'a, 'b>(a: &'b ArrayView<'a, i32, Ix1>, i: usize) -> &'a i32 {
    let sl = a.into_slice().unwrap();  // lifetime of `into_slice` is correct and use it.
    &sl[i]
}

fn main() {
    let a = array![1, 2, 3];
    get_ref(&a, 2);
}
```

This is due to the `Index::index` for `ArrayView<'a, A, D>` has the following signature:

```rust
fn index<'a, 'b>(a: &'b ArrayView<'a, i32, Ix1>, i: usize) -> &'b i32 { ... }
```

The implementation of `index2` seems to be correct, and I think it is a bug.

bluss:
hi @termoshtt, consider that the Index trait prescribes the signature for us. We can't do `index2` in the `Index` trait, but I promise we would if it were possible.

PR #372, opened 2017-10-20
Get a correct reference to an element from ArrayView

termoshtt:
Revival of #371 
Since `Index` trait does not allow us to change the lifetime, I implement it as an associated function `ArrayView::elem_ref` and `ArrayViewMut::elem_ref_mut`

### DIFF
--- a/src/arraytraits.rs
+++ b/src/arraytraits.rs
@@ -94,6 +94,46 @@ impl<S, D, I> IndexMut<I> for ArrayBase<S, D>
     }
 }
 
+impl<'a, A, D> ArrayView<'a, A, D>
+where
+    D: Dimension,
+{
+    /// Get a reference of a element through the view.
+    ///
+    /// This is a replacement of `Index::index` since it forces us a wrong lifetime
+    /// https://github.com/bluss/rust-ndarray/issues/371
+    pub fn elem_ref<I: NdIndex<D>>(&self, index: I) -> &'a A {
+        debug_bounds_check!(self, index);
+        unsafe {
+            &*self.as_ptr().offset(
+                index
+                    .index_checked(&self.dim, &self.strides)
+                    .unwrap_or_else(|| array_out_of_bounds()),
+            )
+        }
+    }
+}
+
+impl<'a, A, D> ArrayViewMut<'a, A, D>
+where
+    D: Dimension,
+{
+    /// Get a mutable reference of a element through the view.
+    ///
+    /// This is a replacement of `IndexMut::index_mut` since it forces us a wrong lifetime
+    /// https://github.com/bluss/rust-ndarray/issues/371
+    pub fn elem_ref_mut<I: NdIndex<D>>(&mut self, index: I) -> &'a mut A {
+        debug_bounds_check!(self, index);
+        unsafe {
+            &mut *self.as_mut_ptr().offset(
+                index
+                    .index_checked(&self.dim, &self.strides)
+                    .unwrap_or_else(|| array_out_of_bounds()),
+            )
+        }
+    }
+}
+
 /// Return `true` if the array shapes and all elements of `self` and
 /// `rhs` are equal. Return `false` otherwise.
 impl<S, S2, D> PartialEq<ArrayBase<S2, D>> for ArrayBase<S, D>
--- a/src/arraytraits.rs
+++ b/src/arraytraits.rs
@@ -102,7 +102,7 @@ where
     ///
     /// This is a replacement of `Index::index` since it forces us a wrong lifetime
     /// https://github.com/bluss/rust-ndarray/issues/371
-    pub fn elem_ref<I: NdIndex<D>>(&self, index: I) -> &'a A {
+    pub fn elem<I: NdIndex<D>>(&self, index: I) -> &'a A {
         debug_bounds_check!(self, index);
         unsafe {
             &*self.as_ptr().offset(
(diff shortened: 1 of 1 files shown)


bluss:
Good idea, will look closer at it soon

bluss:
`elem_ref_mut` is not sound with Rust borrowing rules. This is a tricky case, where the shared and mutable reference rules don't really correspond to another.

The summary is:

* from `&'a &'b T` we can get a `&'b T` even if `'b` is a greater scope than `'a`
* from `&'a mut &'b mut T` we **cannot** get a `&'b mut T` because we can only get the shorter of `'a` and `'b`.

Instead of trying to convince anyone of that rule, here is a program that shows that `elem_ref_mut` allows producing two simultaneous and equal mut references:

[Code link](https://gist.github.com/5a1d6bed59b50bb0fd18214eba3541b8)

When I run it, it has this output:

```
$ cargo test --test aliasing_mut -- --nocapture
running 1 test
In two references with reference 0x7f9fd79fe5cc and 0x7f9fd79fe5cc
test test ... ok
```

But, since this is undefined behaviour, this output is not required to be there, it could change with compiler optimizations and compiler version.

src/arraytraits.rs
```diff
@@ -94,6 +94,46 @@ impl<S, D, I> IndexMut<I> for ArrayBase<S, D>
...
 
+impl<'a, A, D> ArrayView<'a, A, D>
+where
+    D: Dimension,
+{
+    /// Get a reference of a element through the view.
+    ///
+    /// This is a replacement of `Index::index` since it forces us a wrong lifetime
+    /// https://github.com/bluss/rust-ndarray/issues/371
+    pub fn elem_ref<I: NdIndex<D>>(&self, index: I) -> &'a A {
```
bluss: I think this is a good idea. We should make something that closely aligns with the design of the methods for element access that we have, I'm thinking of the Index trait and of the methods `.get()` and `.uget()`.

I'd like to find a good name for this -- something that's similar to `get` or `index`. Also, we probably want an unchecked version of this method too (like `uget`).

src/arraytraits.rs
```diff
@@ -94,6 +94,46 @@ impl<S, D, I> IndexMut<I> for ArrayBase<S, D>
...
+
+impl<'a, A, D> ArrayViewMut<'a, A, D>
+where
+    D: Dimension,
+{
+    /// Get a mutable reference of a element through the view.
+    ///
+    /// This is a replacement of `IndexMut::index_mut` since it forces us a wrong lifetime
+    /// https://github.com/bluss/rust-ndarray/issues/371
+    pub fn elem_ref_mut<I: NdIndex<D>>(&mut self, index: I) -> &'a mut A {
```
bluss: See the comment in the PR thread first. We'll have to be creative here, I don't know what we can do that is also correct with Rust's borrow checker rules.

(As a digression, it's a shame we have this problem! Slices are special in Rust, they are built in dynamically sized types. We'd like to have a custom version of that instead of what we have with the ArrayView and ArrayViewMut. We wouldn't have this problem.)

termoshtt:
> `elem_ref_mut` is not sound with Rust borrowing rules.

Yes, and thanks for your careful check. This is an invalid pointer dereferencing.

I cannot find a way to keep this interface, and my answer for this problem is to consume `ArrayVeiwMut` like `ArrayViewMut::into_slice` done:

```rust
pub fn elem_ref_mut<I: NdIndex<D>>(self, index: I) -> &'a mut A { ... }
```

Then the two mutable reference cannot exist. (*Note*: We should rename this `into_*`.)

> I'd like to find a good name for this -- something that's similar to get or index.

Do you find any name? My current candidate is `elem` and `into_elem`.

termoshtt:
I change the interface of `&mut` to use move. I also renamed into `elem`, `uelem`, `into_elem`, and `into_elem_unchecked`

src/arraytraits.rs
```diff
@@ -94,6 +94,62 @@ impl<S, D, I> IndexMut<I> for ArrayBase<S, D>
...
 }
 
+impl<'a, A, D> ArrayView<'a, A, D>
+where
+    D: Dimension,
+{
+    /// Get a reference of a element through the view.
+    ///
+    /// This is a replacement of `Index::index` since it forces us a wrong lifetime
+    /// https://github.com/bluss/rust-ndarray/issues/371
```
bluss: I can edit this doc myself. We're going to use a matter-of-fact approach, but still be positive. This is a version of `index` that gives out references of a longer lifetime. :smile:

bluss:
Thanks. If you don't mind, I'll merge this in and brood over the name myself. It's easier to get a feel for it when editing.

src/arraytraits.rs
```diff
@@ -94,6 +94,62 @@ impl<S, D, I> IndexMut<I> for ArrayBase<S, D>
...
+                    .unwrap_or_else(|| array_out_of_bounds()),
+            )
+        }
+    }
+
+    /// Get a reference of a element through the view without boundary check
+    ///
+    /// This is a replacement of `Index::index` since it forces us a wrong lifetime
+    /// https://github.com/bluss/rust-ndarray/issues/371
+    pub fn uelem<I: NdIndex<D>>(&self, index: I) -> &'a A {
```
bluss: This method must be `unsafe`. Sorry, I should not have merged without this fixed. I'll fix it now, though.

termoshtt:
Thanks for your revise and merging! and sorry for my later reply...

PR #375, opened 2017-10-28
MAINT: use len_of(...) instead of shape().axis(...)

mgeisler:
The `len_of` method is an equivalent and slightly more direct way of
getting the length of an axis.


bluss:
That's nice, thank you!

PR #376, opened 2017-10-28
MAINT: let TravisCI cache Cargo artifacts

mgeisler:
This can speed up the CI run significantly and in my experience, Cargo
is very good at invalidating the cache when needed.

Take two of #344.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: rust
+cache: cargo
 # use trusty for newer openblas
 sudo: required
 dist: trusty
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -19,7 +19,7 @@
 //! - [**`ArrayBase`**](struct.ArrayBase.html):
 //!   The *n*-dimensional array type itself.<br>
 //!   It is used to implement both the owned arrays and the views; see its docs
-//!   for an overview of all array features.  
+//!   for an overview of all array features.
 //! - The main specific array type is [**`Array`**](type.Array.html), which owns
 //! its elements.
 //!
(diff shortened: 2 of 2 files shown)


bluss:
I don't know what I was thinking, not sure if there was more than 1 build.

Anyway, in ndarray, building all the tests, examples and benchmarks seems to take the longest time (apart from downloading and installing rust).

mgeisler:
Yeah, it looks like the compilation is not the biggest part of the CI run. Also, the cache itself needs to transferred to and from S3, so it's not necessarily a win.

When https://travis-ci.org/bluss/rust-ndarray/builds/294077086 is done, I'll try pushing a whitespace only change to the branch so we get timings for the ideal case where the cache is already up to date.

bluss:
Sure. I think you can just refresh the metadata of the commit and don't have to change anything. Like git commit --amend  and save the message.

mgeisler:
Ah, that might have been more clever :-) The build is now running and recompiling ndarray (which seems fair since most PRs will touch something in the crate itself): https://travis-ci.org/bluss/rust-ndarray/builds/294080752

bluss:
There's not much to save with those deps. Maybe serde and serde-derive that are further down the build.

mgeisler:
Yeah, I agree with you! Two two builds:

* https://travis-ci.org/bluss/rust-ndarray/builds/294077086: Without cache, ran for 12 min 35 sec.
* https://travis-ci.org/bluss/rust-ndarray/builds/294080752: With cache, ran for 11 min 39 sec.

Now it's at least documented for others in the future. I'll just close the PR now.

mgeisler:
For completeness, I played a bit with using `export CARGO_INCREMENTAL=1` locally, but that doesn't seem to save much time either.

bluss:
Thanks. Yes, it seems hard to say if it's faster either way.

PR #381, opened 2017-11-06
Add .into_owned() for all arrays with S: DataOwned

jturner314:
This is useful for generic code that needs to get a uniquely owned array but only knows that `S: DataOwned`. Before, `.into_owned()` only existed for `RcArray`.


bluss:
Thanks, this is very nice!

PR #382, opened 2017-11-09
parallel: Update to rayon 0.9

cuviper:


bluss:
Thanks a lot. Great job with rayon 0.9!

bluss:
(This is breaking only to the ndarray-parallel crate.)

PR #385, opened 2017-11-20
Add support for more index types in s![] macro

jturner314:
This adds support for more index types in the `s![]` macro. (This was [discussed briefly](https://github.com/bluss/rust-ndarray/pull/369#issuecomment-339104061) in #369.)

The only thing I don't like about this PR is that it uses `as` to convert the `step` argument (of arbitrary type) to `isize` in the `s![]` macro. It would be possible to add a conversion trait to handle this, but the `as` approach is much simpler.


bluss:
Thanks a lot!

Issue #312, opened 2017-04-20
Slice with negative step inconsistent with python result


minstrel271:
First of all, I understand that this crate is not mean to reproduct all feature in python, but since in [the docs about Si](https://docs.rs/ndarray/0.9.1/ndarray/struct.Si.html) python slice is used as comparison, I believe this is worth to bring up.

Here is some test case with ndarray:

    #[macro_use(s)]
    extern crate ndarray;
    use ndarray::prelude::*;

    fn main() {
        let x: Array1<usize> = Array::zeros(16);

        println!("{}", x.slice(s![..4;1]).len());
        println!("{}", x.slice(s![4..;1]).len());
        println!("{}", x.slice(s![..-4;1]).len());
        println!("{}", x.slice(s![-4..;1]).len());
        println!("{}", x.slice(s![..4;-1]).len());
        println!("{}", x.slice(s![4..;-1]).len());
        println!("{}", x.slice(s![..-4;-1]).len());
        println!("{}", x.slice(s![-4..;-1]).len());
    }

which return:

    4
    12
    12
    4
    4
    12
    12
    4

and in the cooresponding python test case:

    def main():
        x = [0] * 16
        print(len(x[:4:1]))
        print(len(x[4::1]))
        print(len(x[:-4:1]))
        print(len(x[-4::1]))
        print(len(x[:4:-1]))
        print(len(x[4::-1]))
        print(len(x[:-4:-1]))
        print(len(x[-4::-1]))

    if __name__ == '__main__':
        main()

which return:

    4
    12
    12
    4
    11
    5
    3
    13

bluss:
Interesting. This was definitely a surprise to me:

```python
In [25]: x = np.arange(16.)

In [26]: x[:4:-1]
Out[26]: array([ 15.,  14.,  13.,  12.,  11.,  10.,   9.,   8.,   7.,   6.,   5.])
```

jturner314:
I just noticed this. In #377, I copied the existing behavior.

Other than the Python and `ndarray 0.10` approaches, there's a third approach I think is the best fit for `ndarray`. These are examples of the three options (indexing an axis of length `4` (indices `0`, `1`, `2`, `3`)):

1. Python-style
   1. `2..0;-1` → `2`, `1`
   2. `..0;-2` → `3`, `1`
   3. `3..0;-2` → `3`, `1`
   4. `3..;-2` → `3`, `1`
   5. `..;-2` → `3`, `1`
[...]

PR #386, opened 2017-11-21
Add .nth() method to Slice type

jturner314:
This PR adds an `nth()` method for the `Slice` type.

Instead of adding an `nth()` method, it would be possible to implement `IntoIterator` for `Slice` and specialize `nth()`, but I think it's useful to provide `nth()` on `Slice` itself. It would be nice to implement `Index` for `Slice` instead of `nth()`, but I don't think that's possible because `Index::index()` must return a reference.

I was [planning to include a `len()` method](https://github.com/jturner314/rust-ndarray/commit/4437063697a7670987941b2f278f6ae900d953b6) as part of this PR, but after thinking about it some more, I don't see a good use-case for `len()`.

If you don't think the `nth()` method is sufficiently useful to include in `ndarray`, that's fine too. I just found it useful in my own code and thought I'd propose it.

### DIFF
--- a/src/slice.rs
+++ b/src/slice.rs
@@ -48,6 +48,48 @@ impl Slice {
     pub fn step_by(self, step: Ixs) -> Self {
         Slice { step, ..self }
     }
+
+    /// Returns the `n`th index in this slice.
+    ///
+    /// The `n` argument starts from zero, so `nth(0)` is the first index. If
+    /// `n` is greater than or equal to the length of the slice, then the
+    /// return value is `None`.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use ndarray::Slice;
+    ///
+    /// // indices 0, 1, 2, 3, 4
+    /// assert_eq!(Slice::new(0, Some(5), 1).nth(2), Some(2));
+    ///
+    /// // indices 1, 3
+    /// assert_eq!(Slice::new(1, Some(5), 2).nth(1), Some(3));
+    ///
+    /// // indices 1, -1, -3
+    /// assert_eq!(Slice::new(1, Some(-4), -2).nth(2), Some(-3));
+    ///
+    /// // indices 1, 3, 5, ...
+    /// assert_eq!(Slice::new(1, None, 2).nth(2), Some(5));
+    ///
+    /// // indices 3, 3, 3, ...
+    /// assert_eq!(Slice::new(3, Some(5), 0).nth(1000), Some(3));
+    ///
+    /// // indices 1, 3
+    /// assert_eq!(Slice::new(1, Some(5), 2).nth(2), None);
+    /// ```
+    pub fn nth(&self, n: usize) -> Option<isize> {
+        let nth = self.start + self.step * n as isize;
+        if let Some(end) = self.end {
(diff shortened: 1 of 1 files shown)


src/slice.rs
```diff
@@ -48,6 +48,48 @@ impl Slice {
...
+    /// assert_eq!(Slice::new(1, Some(5), 2).nth(1), Some(3));
+    ///
+    /// // indices 1, -1, -3
+    /// assert_eq!(Slice::new(1, Some(-4), -2).nth(2), Some(-3));
+    ///
+    /// // indices 1, 3, 5, ...
+    /// assert_eq!(Slice::new(1, None, 2).nth(2), Some(5));
+    ///
+    /// // indices 3, 3, 3, ...
+    /// assert_eq!(Slice::new(3, Some(5), 0).nth(1000), Some(3));
```
bluss: What meaning do we give to a step of 0? It used to be an error when using it for slicing, I'll have to check again.

jturner314: If you actually call `slice()` with a step of 0, then you'll get a panic. (See [`dimension::do_slice`](https://github.com/bluss/rust-ndarray/blob/3d9134bdf5b35c0a4f70e04590f5e0a4cc6d07c6/src/dimension/mod.rs#L264).) You'll also get a panic when calling `slice()` if the slice begin/end is out-of-range.

In contrast, `nth()` doesn't panic in these cases. It's possible to change it to panic if `step == 0`, without changing the function signature. To make it panic for out-of-range indices, it would need an argument for the length of the axis you're going to slice.

If you'd prefer it panic in those cases, I'd probably choose a different name (e.g. `index()`) instead of `nth()`.

bluss: Great, thanks for clarifying!

That sounds you have no particular behaviour in mind for 0, so we can choose one freely. I'm not too fond of mixing panics in explicitly checked or fallible methods either, so I'm not sure.  I do lean towards that it should be a panic -- maybe a debug assertion. Maybe indeed a debug assertion already in Slice::new ? Or is that redundant?

bluss: Ah with public fields a debug assertion in new doesn't do much good anyway.

jturner314: It's certainly possible to make the fields non-public (or maybe just `step`) and add an assertion to `new()` and `step_by()`. In some ways, I think that's a better design anyway; what do you think? It's still not possible to handle out-of-range indices this way, but it would catch `step == 0` early.

If we make some of the fields in `Slice` non-public, I would also change `enum SliceOrIndex { Slice { start, end, step }, ... }` to `enum SliceOrIndex { Slice(Slice), ... }` to hide the fields there too.

jturner314:
I just realized that the implementation of `nth()` in this PR doesn't match the current slicing behavior for negative steps, so please don't merge it yet. (I incorrectly assumed that the current behavior matched Python.) I'll fix this PR based on the outcome of the discussion in #312.

jturner314:
As I mentioned on IRC, I realized that it's not possible to correctly handle negative `start`/`end` indices without taking the length of the axis as an argument, which makes `nth()` not very useful. So, I'm closing this PR.

PR #388, opened 2017-11-29
Add multislice! macro

jturner314:
This PR adds a `multislice!()` macro which makes it possible to take multiple simultaneous mutable (and read-only) slices of an array. It panics at runtime if slices violate Rust's aliasing rules.

Regarding the implementation:

* It's a lot of new lines of code, but most of the new lines are comments, docs, and tests.

* Most of the new functions are in `dimension/mod.rs`. They could also be put in `slice.rs`, whatever makes more sense.

* Additions to the public API are hidden from the docs (except for `multislice!()`).

* Basically, the way `multislice!()` works is to borrow the array a single time by calling `view_mut()`, and then copy the borrow as many times as needed to create the resulting `ArrayView` and `ArrayViewMut` instances. This ensures that the underlying array is mutably borrowed and all the lifetimes are correct.

  I didn't see a good way to do that with the existing public API, so I added two (hidden) methods to `ArrayViewMut`: `aliasing_view()` and `aliasing_view_mut()`.

  The primary disadvantage of this approach is that it always mutably borrows the array, even if there aren't any `mut` slices, but that's not a big issue because if you aren't taking any `mut` slices, you can just call `.slice()` multiple times anyway.

* This is the first time I've used `quickcheck`, and I love it! I caught a few bugs in my implementation on edge cases with `quickcheck` tests.

Edit: Would it be better to return an `Err` instead of panicking if the slices violate the aliasing rules?

### DIFF
--- a/src/dimension/mod.rs
+++ b/src/dimension/mod.rs
@@ -6,9 +6,10 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
-use {Ix, Ixs, Slice};
+use {Ix, Ixs, Slice, SliceOrIndex};
 use error::{from_kind, ErrorKind, ShapeError};
 use itertools::izip;
+use num_integer::div_floor;
 
 pub use self::dim::*;
 pub use self::axis::Axis;
@@ -392,6 +393,199 @@ pub fn do_slice(dim: &mut usize, stride: &mut usize, slice: Slice) -> isize {
     offset
 }
 
+/// Solves `a * x + b * y = gcd(a, b)` for `x`, `y`, and `gcd(a, b)`.
+///
+/// Returns `(g, (x, y))`, where `g` is `gcd(a, b)`, and `g` is always
+/// nonnegative.
+///
+/// See https://en.wikipedia.org/wiki/Extended_Euclidean_algorithm
+fn extended_gcd(a: isize, b: isize) -> (isize, (isize, isize)) {
+    if a == 0 {
+        (b.abs(), (0, b.signum()))
+    } else if b == 0 {
+        (a.abs(), (a.signum(), 0))
+    } else {
+        let mut r = (a, b);
+        let mut s = (1, 0);
+        let mut t = (0, 1);
+        while r.1 != 0 {
+            let q = r.0 / r.1;
+            r = (r.1, r.0 - q * r.1);
+            s = (s.1, s.0 - q * s.1);
+            t = (t.1, t.0 - q * t.1);
+        }
+        if r.0 > 0 {
+            (r.0, (s.0, t.0))
+        } else {
+            (-r.0, (-s.0, -t.0))
+        }
+    }
+}
+
+/// Solves `a * x + b * y = c` for `x` where `a`, `b`, `c`, `x`, and `y` are
+/// integers.
+///
+/// If the return value is `Some((x0, xd))`, there is a solution. `xd` is
+/// always positive. Solutions `x` are given by `x0 + xd * t` where `t` is any
+/// integer. The value of `y` for any `x` is then `y = (c - a * x) / b`.
+///
+/// If the return value is `None`, no solutions exist.
+///
+/// **Note** `a` and `b` must be nonzero.
+///
+/// See https://en.wikipedia.org/wiki/Diophantine_equation#One_equation
+/// and https://math.stackexchange.com/questions/1656120#1656138
+fn solve_linear_diophantine_eq(a: isize, b: isize, c: isize) -> Option<(isize, isize)> {
+    debug_assert_ne!(a, 0);
+    debug_assert_ne!(b, 0);
+    let (g, (u, _)) = extended_gcd(a, b);
+    if c % g == 0 {
+        Some((c / g * u, (b / g).abs()))
+    } else {
+        None
+    }
+}
+
+/// Returns `true` if two (finite length) arithmetic sequences intersect.
+///
+/// `min*` and `max*` are the (inclusive) bounds of the sequences, and they
+/// must be elements in the sequences. `step*` are the steps between
+/// consecutive elements (the sign is irrelevant).
+///
+/// **Note** `step1` and `step2` must be nonzero.
+fn arith_seq_intersect(
+    (min1, max1, step1): (isize, isize, isize),
+    (min2, max2, step2): (isize, isize, isize),
+) -> bool {
+    debug_assert!(max1 >= min1);
+    debug_assert!(max2 >= min2);
+    debug_assert_eq!((max1 - min1) % step1, 0);
+    debug_assert_eq!((max2 - min2) % step2, 0);
+
+    // Handle the easy case where we don't have to solve anything.
+    if min1 > max2 || min2 > max1 {
+        false
+    } else {
+        // The sign doesn't matter semantically, and it's mathematically convenient
+        // for `step1` and `step2` to be positive.
+        let step1 = step1.abs();
+        let step2 = step2.abs();
+        // Ignoring the min/max bounds, the sequences are
+        //   a(x) = min1 + step1 * x
+        //   b(y) = min2 + step2 * y
+        //
+        // For intersections a(x) = b(y), we have:
+        //   min1 + step1 * x = min2 + step2 * y
+        //   ⇒ -step1 * x + step2 * y = min1 - min2
+        // which is a linear Diophantine equation.
+        if let Some((x0, xd)) = solve_linear_diophantine_eq(-step1, step2, min1 - min2) {
+            // Minimum of [min1, max1] ∩ [min2, max2]
+            let min = ::std::cmp::max(min1, min2);
+            // Maximum of [min1, max1] ∩ [min2, max2]
+            let max = ::std::cmp::min(max1, max2);
+            // The potential intersections are
+            //   a(x) = min1 + step1 * (x0 + xd * t)
+            // where `t` is any integer.
+            //
+            // There is an intersection in `[min, max]` if there exists an
+            // integer `t` such that
+            //   min ≤ a(x) ≤ max
+            //   ⇒ min ≤ min1 + step1 * (x0 + xd * t) ≤ max
+            //   ⇒ min ≤ min1 + step1 * x0 + step1 * xd * t ≤ max
+            //   ⇒ min - min1 - step1 * x0 ≤ (step1 * xd) * t ≤ max - min1 - step1 * x0
+            //
+            // Therefore, the least possible intersection `a(x)` that is ≥ `min` has
+            //   t = ⌈(min - min1 - step1 * x0) / (step1 * xd)⌉
+            // If this `a(x) is also ≤ `max`, then there is an intersection in `[min, max]`.
+            //
+            // The greatest possible intersection `a(x)` that is ≤ `max` has
+            //   t = ⌊(max - min1 - step1 * x0) / (step1 * xd)⌋
+            // If this `a(x) is also ≥ `min`, then there is an intersection in `[min, max]`.
+            min1 + step1 * (x0 - xd * div_floor(min - min1 - step1 * x0, -step1 * xd)) <= max
+                || min1 + step1 * (x0 + xd * div_floor(max - min1 - step1 * x0, step1 * xd)) >= min
+        } else {
+            false
+        }
+    }
+}
+
+/// Returns the minimum and maximum values of the indices (inclusive).
+///
+/// If the slice is empty, then returns `None`, otherwise returns `Some((min, max))`.
+fn slice_min_max(axis_len: usize, slice: Slice) -> Option<(usize, usize)> {
+    let (start, end, step) = to_abs_slice(axis_len, slice);
+    if start == end {
+        None
+    } else {
+        if step > 0 {
+            Some((start, end - 1 - (end - start - 1) % (step as usize)))
+        } else {
+            Some((start + (end - start - 1) % (-step as usize), end - 1))
+        }
+    }
+}
+
+/// Returns `true` iff the slices intersect.
+#[doc(hidden)]
+pub fn slices_intersect<D: Dimension>(
+    dim: &D,
+    indices1: &D::SliceArg,
+    indices2: &D::SliceArg,
+) -> bool {
+    debug_assert_eq!(indices1.as_ref().len(), indices2.as_ref().len());
+    for (&axis_len, &si1, &si2) in izip!(dim.slice(), indices1.as_ref(), indices2.as_ref()) {
+        // The slices do not intersect iff any pair of `SliceOrIndex` does not intersect.
+        match (si1, si2) {
+            (
+                SliceOrIndex::Slice {
+                    start: start1,
+                    end: end1,
+                    step: step1,
+                },
+                SliceOrIndex::Slice {
+                    start: start2,
+                    end: end2,
+                    step: step2,
+                },
+            ) => {
+                let (min1, max1) = match slice_min_max(axis_len, Slice::new(start1, end1, step1)) {
+                    Some(m) => m,
+                    None => return false,
+                };
+                let (min2, max2) = match slice_min_max(axis_len, Slice::new(start2, end2, step2)) {
+                    Some(m) => m,
+                    None => return false,
+                };
+                if !arith_seq_intersect(
+                    (min1 as isize, max1 as isize, step1),
+                    (min2 as isize, max2 as isize, step2),
+                ) {
+                    return false;
+                }
+            }
+            (SliceOrIndex::Slice { start, end, step }, SliceOrIndex::Index(ind)) |
+            (SliceOrIndex::Index(ind), SliceOrIndex::Slice { start, end, step }) => {
+                let ind = abs_index(axis_len, ind);
+                let (min, max) = match slice_min_max(axis_len, Slice::new(start, end, step)) {
+                    Some(m) => m,
+                    None => return false,
+                };
+                if ind < min || ind > max || (ind - min) % step.abs() as usize != 0 {
+                    return false;
+                }
+            }
+            (SliceOrIndex::Index(ind1), SliceOrIndex::Index(ind2)) => {
+                let ind1 = abs_index(axis_len, ind1);
+                let ind2 = abs_index(axis_len, ind2);
+                if ind1 != ind2 {
+                    return false;
+                }
+            }
+        }
+    }
+    true
+}
+
 pub fn merge_axes<D>(dim: &mut D, strides: &mut D, take: Axis, into: Axis) -> bool
     where D: Dimension,
 {
@@ -423,11 +617,15 @@ pub fn merge_axes<D>(dim: &mut D, strides: &mut D, take: Axis, into: Axis) -> bo
 #[cfg(test)]
 mod test {
     use super::{
-        can_index_slice, can_index_slice_not_custom, max_abs_offset_check_overflow, IntoDimension
+        arith_seq_intersect, can_index_slice, can_index_slice_not_custom, extended_gcd,
+        max_abs_offset_check_overflow, slice_min_max, slices_intersect,
+        solve_linear_diophantine_eq, IntoDimension
     };
     use error::{from_kind, ErrorKind};
-    use quickcheck::quickcheck;
-    use {Dimension, Ix0, Ix1, Ix2, Ix3, IxDyn};
+    use num_integer::gcd;
+    use quickcheck::{quickcheck, TestResult};
+    use slice::{Slice, SliceOrIndex};
+    use {Dim, Dimension, Ix0, Ix1, Ix2, Ix3, IxDyn};
 
     #[test]
     fn slice_indexing_uncommon_strides() {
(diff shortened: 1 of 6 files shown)


src/dimension/mod.rs
```diff
@@ -329,5 +529,183 @@ mod test {
...
+    fn slices_intersect_true() {
+        assert!(slices_intersect(
+            &Dim([4, 5]),
+            &[(..).into(), (..).into()],
+            &[(..).into(), (..).into()],
+        ));
+        assert!(slices_intersect(
+            &Dim([4, 5]),
+            &[0.into(), (..).into()],
+            &[0.into(), (..).into()],
```
bluss: This is not a very interesting comment, but I will note my passing disapproval of Rust code that ends up looking like x.into(), y.into() where it is not about argument conversion (inside the function receiving the parameters).

This is testing code, and style doesn't matter so much here, so it's not about that. We need to design an interface that doesn't encourage the .into() style. multislice doesn't, so that's good.

The disapproval is founded on a) easier to read the X::from(..) variant and b) Seeing breaking changes due to once-unambiguous `.into()` calls fail with minor changes in Rust or other crates.

bluss:
Very well written PR!

src/impl_views.rs
```diff
@@ -533,5 +533,108 @@ impl<'a, A, D> ArrayViewMut<'a, A, D>
...
+    /// let v = {
+    ///     let mut arr = arr1(&[0, 1, 2]);
+    ///     unsafe { arr.view_mut().aliasing_view_mut() }
+    ///     //       --- borrow occurs here
+    /// };
+    /// // `arr` dropped here while still borrowed
+    /// ```
+    #[doc(hidden)]
+    pub unsafe fn aliasing_view_mut(&self) -> ArrayViewMut<'a, A, D> {
+        ArrayViewMut::new_(self.ptr, self.dim.clone(), self.strides.clone())
```
bluss: Here I think we must choose a different design. It's tricky, and connected with the guidelines for unsafe code and the exact safety rules we have to follow, which is not entirely set in stone yet.

The gist of it is that I think `ArrayViewMut<'a, T, D>` has the same semantic as `&'a mut T`, a mutable reference.

Even in unsafe code we are not allowed to create multiple mutable pointers to the same value and use them as the type `&mut T`, so I think this method crosses that line. What we do in usual unsafe code is to keep raw pointers.

The reasoning isn't exactly bullet proof since it relies on the incompleteness of the unsafe code guidelines; so all the more reason to be careful.

I haven't understood the whole code here yet, but the ndarray equivalent would be something like `type RawView<T, D> = ArrayBase<PhantomData<*mut T>, D>`

src/dimension/mod.rs
jturner314: This is something I should keep in mind more. The biggest thing I dislike about `.into()` is that you have to look at the docs to figure out what type it's converting into, but that's also a good point about issues with type inference due to minor changes.

In this case, it's actually possible to use the `s![]` macro, which is something I didn't consider when writing the PR. I think the `s![]` macro will work for most cases in similar client code. I just pushed another commit to this PR switching to `s![]`.

src/impl_views.rs
jturner314: Yeah, getting this right is tricky. We need to ensure that (1) the array being sliced is mutably borrowed and (2) the views have the correct lifetime. Applying Rust's pointer aliasing rules to (temporary) `ArrayView`/`ArrayViewMut` instances complicates things further.

If it matters, `.aliasing_view_mut(&self)` is hidden from the docs, and it's not possible to create aliasing mutable views with `multislice!()`. (It panics if you try to do so.) However, it sounds like you're saying that the implementation should never create (even temporary, internal) aliasing `ArrayViewMut` instances.

At first glance, I see a couple of options. Both options include the following:

* Add a method `.into_repr(self) -> ViewRepr<&'a mut A>` to `ArrayViewMut<'a, A, D>`.
* Add a method `.raw_strides(&self)` to `ArrayBase` similar to `.raw_dim(&self)` but for strides.
* Add an associated function `from_raw_parts(data: ViewRepr<&'a mut A>, ptr: *mut A, dim: D, strides: D)` to `ArrayViewMut<'a, A, D>`.
* Remove the `A: Copy` constraint on the `Copy` implementation for `ViewRepr<A>`. This should be okay, right? (We should also remove the `A: Clone` constraint on the `Clone` implementation for consistency.)

#### Option 1

* Refactor the body of `.slice()` (and as a result, `.slice_move()`, `.slice_inplace()`, `.slice_axis_inplace()`, and `.subview_inplace()`) into a function `slice()` that operates on `ptr: *mut S::Elem, dim: &D, strides: &D, info: &SliceInfo<D::SliceArg, Do>` instead of operating on `ArrayBase` and `info`.
* Then, `multislice!(arr, (mut info1, mut info2))` would expand to something like

  ```rust
  let view = arr.view_mut();
  let ptr = view.as_mut_ptr();
  let dim = view.raw_dim();
  let strides = view.raw_strides();
  let repr = view.into_repr();
  (
      {
          let (new_ptr, new_dim, new_strides) = slice(ptr, &dim, &strides, info1);
          unsafe { ArrayViewMut::from_raw_parts(repr, new_ptr, new_dim, new_strides) }
      },
      {
          let (new_ptr, new_dim, new_strides) = slice(ptr, &dim, &strides, info2);
          unsafe { ArrayViewMut::from_raw_parts(repr, new_ptr, new_dim, new_strides) }
      },
  )
  ```

  plus checks that the views are non-aliasing.

#### Option 2

* Remove the `S: Data` constraint on `ArrayBase`, `.slice_move()`, `.slice_inplace()`, `.slice_axis_inplace()`, `.subview_inplace()`, `.ndim()`, and `.len_of()`. (We can't have `PhantomData<*mut A>` implement `Data` because that would allow dereferencing essentially a raw pointer in safe code (e.g. with `.index()`).)
* Add a method `.as_mut_ptr_view(&self) -> ArrayBase<PhantomData<*mut A>, D>` to `ArrayViewMut<'a, A, D>`.
* Then, `multislice!(arr, (mut info1, mut info2))` would expand to something like

  ```rust
  let view = arr.view_mut();
  let ptr_view = view.as_mut_ptr_view();
  let repr = view.into_repr();
  (
      {
          let sliced = ptr_view.clone().slice_move(info1);
          unsafe {
              ArrayViewMut::from_raw_parts(
                  repr,
                  sliced.as_mut_ptr(),
                  sliced.raw_dim(),
                  sliced.raw_strides(),
              )
          }
      },
      {
          let sliced = ptr_view.clone().slice_move(info2);
          unsafe {
              ArrayViewMut::from_raw_parts(
                  repr,
                  sliced.as_mut_ptr(),
                  sliced.raw_dim(),
                  sliced.raw_strides(),
              )
          }
      },
  )
  ```

  plus checks that the views are non-aliasing.

What do you think of these approaches?

bluss: - IMO it doesn't matter that `aliasing_view_mut(&self)` is hidden from docs. The `&X -> &mut Y` signature is just inadmissible.
- The  the A: Copy constraint for a ViewRepr is important, without it, `ArrayViewMut` can be `Copy`.
- I think we will find that “Remove the S: Data constraint on ArrayBase” is not possible - we need something to give us the type of `*const S::Data`.
- FYI `ArrayViewMut`

I think we can just deconstruct the view into raw parts and construct each `ArrayViewMut` in turn. The main concern seems to be about how to reach the `slice_move` method?

Basically Option 2, but I wouldn't want to change any of the methods in the way you suggest. I know this answer is not complete.

Feel free to ping me when I forget to answer.

jturner314: > the A: Copy constraint for a ViewRepr is important, without it, `ArrayViewMut` can be `Copy`.

I'm confused about this, because the constraints on the `Copy` implementation of `ArrayBase` are `S: DataClone + Copy, D: Copy`, and `DataClone` is not implemented for `ViewRepr<&'a mut A>`. As a result, it seems like the `A: Copy` constraint shouldn't influence whether `ArrayViewMut` is `Copy`. Am I missing something?

Would it be okay to remove the `A: Clone` constraint for the `Clone` implementation for `ViewRepr<A>`?

Regardless, neither of these constraint changes are critical; there are alternatives.

> “Remove the S: Data constraint on ArrayBase” is not possible - we need something to give us the type of `*const S::Data`.

Good point. (I assume you meant `S::Elem`.) I wouldn't want to have to add a third type parameter.

> The main concern seems to be about how to reach the `slice_move` method?

Yeah, that's the issue. We have to either refactor the slicing methods so that we can slice the deconstructed pieces (`ptr`, `dim`, `strides`) or create (temporarily) aliasing `ArrayBase` instances in order to call `slice_move`. We can't slice `ArrayViewMut` views of the full array because that would require multiple (temporary) aliasing mutable views.

Here are a couple more options to avoid refactoring the slicing methods:

#### Option 3

Add a (hidden, unsafe) method to convert `ArrayView` to `ArrayViewMut`. Then, `multislice!()` could slice the array into `ArrayView` instances, and then "upgrade" them to `ArrayViewMut` instances. I don't really like this idea, though, because it seems weird to have a method that effectively turns `&T` into `&mut T`. It would also take a little work to make sure the underlying array is mutably borrowed.

#### Option 4

Add a new marker trait `pub unsafe trait DataDeref: Data {}` that would be required to read/write data (i.e. deref `self.ptr`) in an `ArrayBase` with safe methods. `DataDeref` would be implemented for all of the existing data types, but would not be implemented for `PhantomData<*mut A>`. This would allow us to safely implement `Data` for `PhantomData<*mut A>` so that you could create an `ArrayBase<PhantomData<*mut A>, D>`.

With this approach, `multislice!()` could create "pointer views" (`ArrayBase<PhantomData<*mut A>, D>`) of the entire array, slice them, and then "dereference" them (or use `from_raw_parts()`) to get the final `ArrayViewMut` instances.

This seems like a fairly nice, clean approach, but it is a breaking change because the constraint on many `ArrayBase` methods (e.g. `.index()` and `.scalar_sum()`) would change from `S: Data` to `S: DataDeref`.

What do you think?

bluss: Option 3 is not an alternative for the same reason that we want to fix this method to start with.

I'd go with Option 4, except making Data require a new trait instead, `DataRaw` or something like that. It shouldn't be breaking then. It would be very interesting to go in this direction anyway -- a raw pointer ArrayBase with no specific ownership semantic

bluss: You are right about the Copy thing, I didn't think about that.

jturner314: Yeah, I like the idea to create a "raw pointer array" too. I'm not sure I understand what you're proposing with `DataRaw`, though. In order to be able to create an `ArrayBase<PhantomData<*mut A>, D>`, `PhantomData<*mut A>` must implement `Data`. However, it's not safe to read/write data from this type because that's equivalent to dereferencing the pointer. So, we need another trait to mean "you can safely read/write the data" that we can use as a bound for methods that read/write data. We can't make `Data` require this new trait, though, because then `PhantomData<*mut A>` can't implement `Data`.

bluss: Yes you are down the same thread of what I mean.

bluss:
Do you want me to look into the Raw pointer array view?

jturner314:
> Do you want me to look into the Raw pointer array view?

That would be great, thanks. I have a start on the implementation in [jturner314/rust-ndarray/dataraw-trait](https://github.com/jturner314/rust-ndarray/tree/dataraw-trait) but I don't have much time to work on it at the moment.

jturner314:
I've updated the PR to use the `RawArrayViewMut` type so that it's no longer necessary to temporarily create mutable aliasing views.

Now (using `s_a`, `s_b`, `s_c`, and `s_d` variables to simplify the expansion),

```rust
let (a, b, c, d) = multislice!(arr, (mut s_a, s_b, mut s_c, s_d));
```

expands to the following (generated with `cargo +nightly rustc --example foo -- -Z unstable-options --pretty=expanded,hygiene` and manually un-expanding the `assert!` calls):

```rust
let (a /* 107#0 */, b /* 108#0 */, c /* 109#0 */, d /* 110#0 */) =
    {
        let (life /* 11149#22 */, raw_view /* 11021#22 */) =
            {
                let mut view /* 10978#22 */ =
                    ::ArrayBase /* 10959#22 */::view_mut /* 10979#22
                        */(&mut arr /* 93#0 */);
                (::life_of_view_mut /* 10783#22 */(&view /* 10978#22 */),
                 view /* 10978#22 */.raw_view_mut /* 11022#22 */())
            };
        match s_a /* 100#0 */ {
            info /* 11259#23 */ => {
                match s_b /* 103#0 */ {
                    info /* 11259#24 */ => {
                        assert!(!::slices_intersect /* 10954#25
                                 */(&raw_view /* 11021#22 */.raw_dim /*
                                         10799#25 */(),
                                    info /* 11259#24 */,
                                    info /* 11259#23 */));
                        match s_c /* 104#0 */ {
                            info /* 11259#36 */ => {
                                assert!(!::slices_intersect /* 10954#37
                                         */(&raw_view /* 11021#22
                                                 */.raw_dim /* 10799#37
                                                 */(),
                                            info /* 11259#36 */,
                                            info /* 11259#23 */));
                                assert!(!::slices_intersect /* 10954#48
                                         */(&raw_view /* 11021#22
                                                 */.raw_dim /* 10799#48
                                                 */(),
                                            info /* 11259#36 */,
                                            info /* 11259#24 */));
                                match s_d /* 106#0 */ {
                                    info /* 11259#60 */ => {
                                        {
                                            assert!(!::slices_intersect /*
                                                     10954#62
                                                     */(&raw_view /*
                                                             11021#22
                                                             */.raw_dim /*
                                                             10799#62
                                                             */(),
                                                        info /* 11259#60
                                                            */,
                                                        info /* 11259#23
                                                            */));
                                            assert!(!::slices_intersect /*
                                                     10954#73
                                                     */(&raw_view /*
                                                             11021#22
                                                             */.raw_dim /*
                                                             10799#73
                                                             */(),
                                                        info /* 11259#60
                                                            */,
                                                        info /* 11259#36
                                                            */));
                                        };
                                        (#[allow(unsafe_code)] unsafe {
                                                                   ::deref_raw_view_mut_into_view_mut_with_life
                                                                       /*
                                                                       10785#84
                                                                       */(raw_view
                                                                              /*
                                                                              11021#22
                                                                              */.clone
                                                                              /*
                                                                              474#84
                                                                              */().slice_move
                                                                              /*
                                                                              10983#84
                                                                              */(info
                                                                                     /*
                                                                                     11259#23
                                                                                     */),
                                                                          life
                                                                              /*
                                                                              11149#22
                                                                              */)
                                                               },
                                         #[allow(unsafe_code)] unsafe {
                                                                   ::deref_raw_view_mut_into_view_with_life
                                                                       /*
                                                                       10784#85
                                                                       */(raw_view
                                                                              /*
                                                                              11021#22
                                                                              */.clone
                                                                              /*
                                                                              474#85
                                                                              */().slice_move
                                                                              /*
                                                                              10983#85
                                                                              */(info
                                                                                     /*
                                                                                     11259#24
                                                                                     */),
                                                                          life
                                                                              /*
                                                                              11149#22
                                                                              */)
                                                               },
                                         #[allow(unsafe_code)] unsafe {
                                                                   ::deref_raw_view_mut_into_view_mut_with_life
                                                                       /*
                                                                       10785#86
                                                                       */(raw_view
                                                                              /*
                                                                              11021#22
                                                                              */.clone
                                                                              /*
                                                                              474#86
                                                                              */().slice_move
                                                                              /*
                                                                              10983#86
                                                                              */(info
                                                                                     /*
                                                                                     11259#36
                                                                                     */),
                                                                          life
                                                                              /*
                                                                              11149#22
                                                                              */)
                                                               },
                                         #[allow(unsafe_code)] unsafe {
                                                                   ::deref_raw_view_mut_into_view_with_life
                                                                       /*
                                                                       10784#87
                                                                       */(raw_view
                                                                              /*
                                                                              11021#22
                                                                              */.clone
                                                                              /*
                                                                              474#87
                                                                              */().slice_move
                                                                              /*
                                                                              10983#87
                                                                              */(info
                                                                                     /*
                                                                                     11259#60
                                                                                     */),
                                                                          life
                                                                              /*
                                                                              11149#22
                                                                              */)
                                                               })
                                    }
                                }
                            }
                        }
                    }
                }
            }
        }
    };
```

I'd like to add a few more tests, and I'm strongly considering returning an error instead of panicking when the slices violate the aliasing rules, but otherwise, this PR is mostly done.

Any thoughts on the error vs. panic question?

src/slice.rs
```diff
@@ -623,3 +623,273 @@ macro_rules! s(
...
+/// long as Rust's aliasing rules are followed for *elements* in the slices.
+/// For example, it's possible to take two disjoint, mutable slices of an
+/// array, with one referencing the even-index elements and the other
+/// referencing the odd-index elements. If you tried to achieve this by calling
+/// `.slice_mut()` twice, the borrow checker would complain about mutably
+/// borrowing the array twice (even though it's safe as long as the slices are
+/// disjoint).
+///
+/// The syntax is `multislice!(` *expression, (pattern [, pattern [, …]])* `)`,
+/// where *expression* evaluates to an `ArrayBase<S, D>` where `S: DataMut`,
```
bluss: I guess you disagree (since you wrote it this way), but I'd prefer to say "mutable array" rather than "`ArrayBase<S, D>` where `S: DataMut`". (Yes I'm constantly unhappy with exposing ArrayBase, DataMut like this :slightly_smiling_face: .)

jturner314: I like "mutable array" better too. I think I originally wrote it that way because macros don't have type signatures like methods do, which tends to make me want to be fairly explicit about the types they expect and return. In this case, though, "mutable array" is clear and less verbose.

src/slice.rs
```diff
@@ -623,3 +623,273 @@ macro_rules! s(
...
+///
+/// ```
+/// extern crate ndarray;
+///
+/// use ndarray::multislice;
+/// use ndarray::prelude::*;
+///
+/// # fn main() {
+/// let mut arr = Array1::from_iter(0..12);
+/// let (a, b, c, d) = multislice!(arr, (s![0..5], mut s![6..;2], s![1..6], mut s![7..;2]));
```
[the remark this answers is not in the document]
jturner314: That's a good point. I wrote it that way because it was the most obvious way to allow the user to supply a `&SliceInfo` instance instead of always requiring them to use `s!`-style notation. However, it would be possible to use pattern-matching to differentiate between `s!`-style notation and an expression that evaluates to `&SliceInfo`, so the `s!` isn't strictly necessary. That said, I think that we should keep it as-is for the following reasons:

* It's easier to read `multislice!(arr, (s![0..5], mut s![6..;2]))` at a glance than `multislice!(arr, ([0..5], mut [6..;2]))` if you have syntax highlighting because the `s!` is shown in a different color, which makes it really easy to see the grouping of the slicing info.
* When the user writes `multislice!(arr, (s![0..5], mut s![6..;2]))`, it's clear what the `s!` means because the `s!` macro is used fairly often. Without the `s!`, it's not quite as obvious.
* The `multislice!` implementation is a little simpler without having to provide special parsing for `s!`-style expressions.

jturner314:
> I would first think that indexing errors should, “as usual”, result in a panic.

Out-of-bounds indices should panic, but there isn't precedent for how to handle overlapping slices, so we could go either way. The primary issue with panicking is that determining whether slices will overlap for arbitrary indices is somewhat tricky, and I like to keep panic conditions simple and easy-to-avoid. Additionally, if we return an error, the user can easily convert it into a panic with `.unwrap()` or `.expect()`.

> If it's an error -- what would the user do with it?

That's a really good question. If there is an overlap, is there a feasible way to recover? I have trouble coming up with a realistic case where you'd want to do anything other than panic.

I've been using `multislice!` (the old, questionable `aliasing_view`/`aliasing_view_mut` implementation) for a while in my own code. For my use cases, the way I'm using it always makes it obvious that the slices won't overlap. I don't think tricky cases where it's difficult to determine whether the slices might overlap would actually come up in practice.

So, on further thought, I think that `multislice!` should panic instead of returning an error. If necessary, we can expose the `slices_intersect` functionality to allow the user to easily determine if the macro will panic or not before calling it.

src/slice.rs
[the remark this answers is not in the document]
bluss: I'm curious, is it technically possible to use this syntax: `multislice!(arr, [0..5], mut [6..;2])`?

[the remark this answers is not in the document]
jturner314: Yes; we can just add rules that apply the `s![]` macro, matching on `[$($info:tt)*]`, above the rules matching on `$info:expr`:

```diff
diff --git a/src/slice.rs b/src/slice.rs
index 0d358f1..a06a8fd 100644
--- a/src/slice.rs
+++ b/src/slice.rs
@@ -770,6 +770,108 @@ macro_rules! multislice(
             )
         }
     };
+    // Parse last slice (mutable), no trailing comma, applying `s![]` macro.
+    (
+        @parse $view:expr, $life:expr,
+        ($($sliced:tt)*),
+        ($($mut_info:tt)*),
+        ($($immut_info:tt)*),
+        (mut [$($info:tt)*])
+    ) => {
+        // Apply `s![]` macro to info.
+        $crate::multislice!(
+            @parse $view, $life,
+            ($($sliced)*),
+            ($($mut_info)*),
+            ($($immut_info)*),
+            (mut s![$($info)*],)
+        )
+    };
+    // Parse last slice (read-only), no trailing comma, applying `s![]` macro.
+    (
+        @parse $view:expr, $life:expr,
+        ($($sliced:tt)*),
+        ($($mut_info:tt)*),
+        ($($immut_info:tt)*),
+        ([$($info:tt)*])
+    ) => {
+        // Apply `s![]` macro to info.
+        $crate::multislice!(
+            @parse $view, $life,
+            ($($sliced)*),
+            ($($mut_info)*),
+            ($($immut_info)*),
+            (s![$($info)*],)
+        )
+    };
+    // Parse last slice (mutable), with trailing comma, applying `s![]` macro.
+    (
+        @parse $view:expr, $life:expr,
+        ($($sliced:tt)*),
+        ($($mut_info:tt)*),
+        ($($immut_info:tt)*),
+        (mut [$($info:tt)*],)
+    ) => {
+        // Apply `s![]` macro to info.
+        $crate::multislice!(
+            @parse $view, $life,
+            ($($sliced)*),
+            ($($mut_info)*),
+            ($($immut_info)*),
+            (mut s![$($info)*],)
+        )
+    };
+    // Parse last slice (read-only), with trailing comma, applying `s![]` macro.
+    (
+        @parse $view:expr, $life:expr,
+        ($($sliced:tt)*),
+        ($($mut_info:tt)*),
+        ($($immut_info:tt)*),
+        ([$($info:tt)*],)
+    ) => {
+        // Apply `s![]` macro to info.
+        $crate::multislice!(
+            @parse $view, $life,
+            ($($sliced)*),
+            ($($mut_info)*),
+            ($($immut_info)*),
+            (s![$($info)*],)
+        )
+    };
+    // Parse a mutable slice, applying `s![]` macro.
+    (
+        @parse $view:expr, $life:expr,
+        ($($sliced:tt)*),
+        ($($mut_info:tt)*),
+        ($($immut_info:tt)*),
+        (mut [$($info:tt)*], $($t:tt)*)
+    ) => {
+        // Apply `s![]` macro to info.
+        $crate::multislice!(
+            @parse $view, $life,
+            ($($sliced)*),
+            ($($mut_info)*),
+            ($($immut_info)*),
+            (mut s![$($info)*], $($t)*)
+        )
+    };
+    // Parse a read-only slice, applying `s![]` macro.
+    (
+        @parse $view:expr, $life:expr,
+        ($($sliced:tt)*),
+        ($($mut_info:tt)*),
+        ($($immut_info:tt)*),
+        ([$($info:tt)*], $($t:tt)*)
+    ) => {
+        // Apply `s![]` macro to info.
+        $crate::multislice!(
+            @parse $view, $life,
+            ($($sliced)*),
+            ($($mut_info)*),
+            ($($immut_info)*),
+            (s![$($info)*], $($t)*)
+        )
+    };
     // Parse last slice (mutable), no trailing comma.
     (
         @parse $view:expr, $life:expr,
```

This change would be backwards compatible if we want to add it in the future.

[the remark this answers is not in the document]
bluss: Just counting in syntactic noise, it seems like a big win to drop both the extra () around the slices and each `s!`, I would go for that right away.  The `s!` would keep working, because you've designed it so we can insert arbitrary expressions for each axis, right?  You could do (?)

```rust
let ax1 = s![1..];
let ax2 = s![2..];
multislice!(arr, ax1, ax2);
```

I guess it would count as us breaking array expressions in those locations, but there is no reason to use those there at this point.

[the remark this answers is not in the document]
jturner314: Oh, I didn't notice the dropped () in your earlier comment. Yeah, we can drop the () too. With the extra () dropped, I think it's fine to drop the `s!` too. (With the (), it looks like a tuple, which makes me want to require the extra `s!` so that the elements are expressions. Without the (), the parameters are clearly parameters to the macro and not part of a tuple, so IMO it's fine to do things like interpreting them as `s!`-style syntax without requiring the `s!`.)

I'll update the PR to drop the extra () and `s!`.

> The `s!` would keep working, because you've designed it so we can insert arbitrary expressions for each output, right?

Yes, normal expressions that evaluate to `&SliceInfo` will still work. (They just have a lower precedence than `[$($info:tt)*]`.)

> I guess it would count as us breaking array expressions (like `[1,2,3]`) in those locations, but there is no reason to use those there at this point.

Yeah, these don't matter right now, because the slicing expressions must have type `&SliceInfo`, and an expression of the form `[$($info:tt)*]` is always a fixed-length array, not a `&SliceInfo`. If we want to add support for slicing using fixed-length array expressions in the future (e.g. `[1, 2, 3]` and `[0..5, 3..7, 2..8]` for `Ix3`) that don't use the `s!` macro machinery, we can do so by adding more rules with even higher precedence. (They'd match on e.g. `[$($info:expr),*]`.)

One interesting thing to note is that we could implement `multislice!` as a method on `ArrayBase`, instead of a macro, that could be called like this:

```rust
arr.multislice((s![0..5], Mut(s![6..;2])))
// or
arr.multislice(ss!([0..5], mut [6..;2]))
```

but that would work up to only a limited tuple length (however long we'd decide to support), like `Zip` works up to only 6 producers.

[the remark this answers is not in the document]
bluss: As a method sounds cool but the syntax doesn't work quite so well there, so it doesn't seem as good. Hopefully multislice as a macro has its own attraction by virtue of its functionality so that users find it.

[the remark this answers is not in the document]
jturner314: I actually think `arr.multislice(ss!([0..5], mut [6..;2]))` is quite nice, but it would be more complex to implement than the `multislice!` macro.

I've updated the PR to drop the extra () and `s!`.

bluss:
This is ready to go, right? Everything looks good to me! Then we can soon release 0.13 with this and updated matrixmultiply

jturner314:
I looked over the PR again and merged it. I'm glad to get this merged; I'll finally be able to use mainline `ndarray` again!

> Then we can soon release 0.13 with this and updated matrixmultiply

I have a couple more breaking changes I'd like to make in order to support more slicing operations. I'll submit the PR today.

PR #391, opened 2017-12-05
Add 0.11.0 changes to README

jturner314:
Will you please also review the [release announcement](https://jim.turner.link/pages/ndarray-0.11/)? You may also be interested in [this article talking about the implementation of combined slicing and subviews](https://jim.turner.link/pages/impl-slicing-subviews-ndarray/).

Edit: Please also see #392 before releasing 0.11.0.

PR #392, opened 2017-12-06
Change SliceInfo::new() to return a Result

jturner314:
This is a last-minute change for 0.11.0. I think it would be better for `SliceInfo::new()` to return `Err` instead of panicking if `D` is not consistent with `indices`. It's more flexible for the caller and is more similar to `.into_dimensionality()`.

This change doesn't affect #391 and should merge cleanly.

Edit: I initially chose to make `SliceInfo::new()` panic becuase I thought it would be weird for a method named `new()` to return a `Result` instead of the object itself. However, there is precedent for returning a `Result` from `new()`, e.g. [`regex::Regex::new()`](https://docs.rs/regex/0.2.3/regex/struct.Regex.html#method.new).

### DIFF
--- a/src/slice.rs
+++ b/src/slice.rs
@@ -5,6 +5,7 @@
 // <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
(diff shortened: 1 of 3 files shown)


bluss:
Thanks!

PR #393, opened 2017-12-06
Fix evaluation of arguments in s![]

jturner314:
This PR contains fixes and tests for two issues related to evaluation of arguments in the `s![]` macro. Before, the `$r` expression was used in two places: `SliceNextDim::next_dim` and `s!(@convert ...)`, which caused it to be evaluated twice. In some cases, this also failed the borrow checker.

The fix is to evaluate `$r` only once, binding it to `r`, then determine `out_dim` using `&r`, and then convert `r` to `SliceOrIndex`. Rust's macro hygiene keeps the various `r` bindings from conflicting. The `$s` expression doesn't need to be bound to a variable because it's used only once.

### DIFF
--- a/src/slice.rs
+++ b/src/slice.rs
@@ -533,67 +533,83 @@ impl<D1: Dimension> SliceNextDim<D1, D1::Larger> for RangeFull {
 macro_rules! s(
     // convert a..b;c into @convert(a..b, c), final item
     (@parse $dim:expr, [$($stack:tt)*] $r:expr;$s:expr) => {
-        {
-            let out_dim = $crate::SliceNextDim::next_dim(&$r, $dim);
-            unsafe {
-                &$crate::SliceInfo::new_unchecked(
-                    [$($stack)* s!(@convert $r, $s)],
-                    out_dim,
-                )
+        match $r {
+            r => {
+                let out_dim = $crate::SliceNextDim::next_dim(&r, $dim);
+                unsafe {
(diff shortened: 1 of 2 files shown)


bluss:
Can we use match to bind and lift it out of the unsafe? I was actually thinking of the unsafe clean issue today. Thanks.

bluss:
The diff is not pretty. I'll have to be back at my computer to look closer at this.

jturner314:
> Can we use match to bind and lift it out of the unsafe?

Yeah, it would be good to move the variable bindings outside the `unsafe` block. I'm not exactly sure what you mean in this case by "use `match`", though.

I pushed a new version that moves the variable bindings out of the `unsafe` block. I just wrapped the expansion in `{ }` to keep it as an expression. Is that what you're looking for?

bluss:
`match $r { r => ` *expression* ` }` keeps any temporaries in $r alive for the whole rest of the expression

jturner314:
That's true. `match $r { r =>` *expression* `}` should be equivalent to `{ let r = $r;` *expression* `}`, right? The second option what I pushed an hour ago. I can change it to `match` if you'd prefer.

bluss:
It's not equivalent

jturner314:
Okay, I changed it to the `match` version. Out of curiosity, what is the difference between `match $r { r =>` *expression* `}` and `{ let r = $r;` *expression* `}` in this context?

bluss:
Thanks!

PR #394, opened 2017-12-19
Add .permuted_axes() method

jturner314:
This is equivalent to [`numpy.transpose()`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.transpose.html) with the `axes` argument and [Matlab's `permute()` function](https://www.mathworks.com/help/matlab/ref/permute.html).

Note that this adds another associated type to the `Dimension` trait. In some ways, I'd prefer to reuse the `NdIndex` trait instead, but I don't see a clean way to do that, so adding another associated type looks like the best option to me.

It's also worth considering if it'd be better to return a `Result` instead of panicking if an axis is missing or repeated in the `axes` argument.

### DIFF
--- a/src/impl_methods.rs
+++ b/src/impl_methods.rs
@@ -1329,6 +1329,59 @@ impl<A, S, D> ArrayBase<S, D> where S: Data<Elem=A>, D: Dimension
         self.strides.slice_mut().swap(ax, bx);
     }
 
+    /// Permute the axes.
+    ///
+    /// This does not move any data, it just adjusts the array’s dimensions
+    /// and strides.
+    ///
+    /// *i* in the *j*-th place in the axes sequence means `self`'s *i*-th axis
+    /// becomes `self.permuted_axes()`'s *j*-th axis
+    ///
+    /// **Panics** if any of the axes are out of bounds, if an axis is missing,
+    /// or if an axis is repeated more than once.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use ndarray::{arr2, Array3};
+    ///
+    /// let a = arr2(&[[0, 1], [2, 3]]);
+    /// assert_eq!(a.view().permuted_axes([1, 0]), a.t());
+    ///
+    /// let b = Array3::<u8>::zeros((1, 2, 3));
+    /// assert_eq!(b.permuted_axes([1, 0, 2]).shape(), &[2, 1, 3]);
+    /// ```
+    pub fn permuted_axes<T>(self, axes: T) -> ArrayBase<S, D>
+    where
+        T: IntoDimension<Dim = D>,
+    {
+        let axes = axes.into_dimension();
+        // Ensure that each axis is used exactly once.
+        let mut usage_counts = D::zero_index_with_ndim(self.ndim());
+        for axis in axes.slice() {
+            usage_counts[*axis] += 1;
+        }
+        for count in usage_counts.slice() {
+            assert_eq!(*count, 1, "each axis must be listed exactly once");
+        }
+        // Determine the new shape and strides.
+        let mut new_dim = usage_counts; // reuse to avoid an allocation
+        let mut new_strides = D::zero_index_with_ndim(self.ndim());
+        {
+            let dim = self.dim.slice();
+            let strides = self.strides.slice();
+            for (new_axis, &axis) in axes.slice().iter().enumerate() {
+                new_dim[new_axis] = dim[axis];
+                new_strides[new_axis] = strides[axis];
+            }
+        }
+        ArrayBase {
+            dim: new_dim,
+            strides: new_strides,
+            ..self
+        }
+    }
+
     /// Transpose the array by reversing axes.
     ///
     /// Transposition reverses the order of the axes (dimensions and strides)
--- a/tests/array.rs
+++ b/tests/array.rs
@@ -542,6 +542,62 @@ fn swapaxes()
 }
 
 #[test]
+fn permuted_axes()
+{
+    let a = array![1].into_subview(Axis(0), 0);
+    let permuted = a.view().permuted_axes([]);
+    assert_eq!(a, permuted);
+
+    let a = array![1];
+    let permuted = a.view().permuted_axes([0]);
+    assert_eq!(a, permuted);
+
+    let a = Array::from_iter(0..24).into_shape((2, 3, 4)).unwrap();
+    let permuted = a.view().permuted_axes([2, 1, 0]);
+    for ((i0, i1, i2), elem) in a.indexed_iter() {
+        assert_eq!(*elem, permuted[(i2, i1, i0)]);
+    }
+    let permuted = a.view().into_dyn().permuted_axes(&[0, 2, 1][..]);
+    for ((i0, i1, i2), elem) in a.indexed_iter() {
+        assert_eq!(*elem, permuted[&[i0, i2, i1][..]]);
+    }
+
+    let a = Array::from_iter(0..120).into_shape((2, 3, 4, 5)).unwrap();
+    let permuted = a.view().permuted_axes([1, 0, 3, 2]);
+    for ((i0, i1, i2, i3), elem) in a.indexed_iter() {
+        assert_eq!(*elem, permuted[(i1, i0, i3, i2)]);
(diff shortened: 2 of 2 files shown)


bluss:
The IntoDimension trait is actually appropriate. At least looking at its signature and implementations.

Eventually; I think ndarray must separate index values and types from dimension values and types, but that's not implemented yet.

jturner314:
> The IntoDimension trait is actually appropriate.

Good point. I missed that.

I just pushed a new version of this PR that uses `IntoDimension` and is rebased off the latest master.

jturner314:
What do you think of this PR? How do you feel about panicking on invalid input (should it return a `Result` instead)? Is there something else you'd like me to change?

bluss:
I think it's fine. I'm sorry that it has been waiting. I'm mulling over the argument type, since it's not perfect but that's probably something we can fix another time.

Since you ask, we can actually permute the dimension and strides in place, can't we? That said I'm not sure that even matters? The algorithm I know is conveniently placed @ the example https://docs.rs/unchecked-index/0.2.2/unchecked_index/

bluss:
This method can use `&mut self` by the way. I'd prefer that I think, for utility's sake, but the by value signature is maybe easier to understand for users. When using this method on a struct field that owns an Array it would be great to avoid the swaparoo.

jturner314:
> we can actually permute the dimension and strides in place, can't we?

Yeah, that might be a little more efficient. Would you like me to change the PR to that algorithm?

> This method can use `&mut self` by the way. I'd prefer that I think, for utility's sake, but the by value signature is maybe easier to understand for users.

That would be okay. I chose to move the array instead of mutating in-place to match `.reversed_axes()` and because it makes my client code cleaner. (I'm typically using `.permuted_axes()` in the middle of an expression. If it mutated in-place, I'd need to do something like `some_expr_or_fn(&{ let v = arr.view(); v.permute_axes([1, 0, 2]); v })` instead of `some_expr_or_fn(&arr.view().permuted_axes([1, 0, 2]))`.) However, operating on `&mut self` is fine too. In that case, I'd probably call it `.permute_axes()`.

bluss:
We don't need to fix the permutation, only if it's an improvement. Good comparison with reversed_axes. It is fine to let the signature stay as it is, but it's regrettable that sometimes you need the one and sometimes the other (and we already have so many methods).

jturner314:
I did some benchmarking in 28626f3575ea45054e7780ffb6e040390c6b6bca. Basically, the benchmarks generate random permutations of the axis indices and call `.permute_axes()` with those random permutations. (The RNG seed is the same for all benchmarks.) It's worth noting that in real-world use, some permutations (swap and reverse) would probably be done with other methods. I'm also not sure if the memory allocator would behave differently in real-world use than in these benchmarks. The results are in the tables below.

The **Dimension** column is the dimension of the array, where `IxDyn, x` means an `IxDyn` dimension of length `x`. The **Baseline** column is the time to generate the input for `.permute_axes()` (create the array and generate the permutation indices). The **Original** column is the original implementation I submitted in this PR. The **In-place** column is the [in-place algorithm you linked above](https://docs.rs/unchecked-index/0.2.2/unchecked_index/#example), but with checked indexing. The **In-place, unchecked** column is the same as **In-place**, but with unchecked indexing of the permutation indices.

The benchmarks were run with `cargo +nightly bench --bench permuted-axes` with nightly Rust from 2017-01-05 on a laptop with an Intel Core i5-2520M CPU and 8 GB DDR3 memory.

**Total times, ns/iter**

Dimension | Baseline | Original | In-place | In-place, unchecked
--------- | -------- | -------- | -------- | -------------------
`Ix3`       |  31±0    |  56±1    |  56±1    |  56±0
`Ix4`       |  38±1    |  62±4    |  69±1    |  70±5
`Ix5`       |  48±0    |  76±41   |  84±2    |  84±2
`Ix6`       |  58±1    |  87±2    |  99±2    |  99±6
`IxDyn, 3`  | 136±2    | 251±4    | 203±4    | 201±3
`IxDyn, 4`  | 165±5    | 281±5    | 237±5    | 236±4
`IxDyn, 5`  | 226±6    | 411±4    | 337±7    | 338±3
`IxDyn, 6`  | 251±1    | 448±25   | 368±4    | 371±2
`IxDyn, 7`  | 285±2    | 475±15   | 404±14   | 400±7

**Differences from baseline (means only), ns/iter**

Dimension | Original | In-place | In-place, unchecked
--------- | -------- | -------- | -------------------
`Ix3`       | 25       | 25       | 25 
`Ix4`       | 24       | 31       | 32 
`Ix5`       | 28       | 36       | 36 
`Ix6`       | 29       | 41       | 41 
`IxDyn, 3`  | 115      | 67       | 65 
`IxDyn, 4`  | 116      | 72       | 71 
`IxDyn, 5`  | 185      | 111      | 112
`IxDyn, 6`  | 197      | 117      | 120
`IxDyn, 7`  | 190      | 119      | 115

The original method is somewhat faster for fixed-size dimensions, while the in-place methods are noticeably faster for `IxDyn` dimensions. There is no noticeable performance difference between the checked and unchecked in-place methods.

I don't have a strong preference between the original and in-place methods, although I do find the original implementation easier to understand. I'm not really sure how representative these benchmarks are anyway, and we're looking at a difference of less than 80 ns.

Which method would you prefer?

bluss:
Wow cool! If you prefer the original it is fine to use it.

src/impl_methods.rs
```diff
@@ -1329,6 +1329,58 @@ impl<A, S, D> ArrayBase<S, D> where S: Data<Elem=A>, D: Dimension
...
+        // Ensure that each axis is used exactly once.
+        let mut usage_counts = D::zero_index_with_ndim(self.ndim());
+        for axis in axes.slice() {
+            usage_counts[*axis] += 1;
+        }
+        for count in usage_counts.slice() {
+            assert_eq!(*count, 1, "each axis must be listed exactly once");
+        }
+        // Determine the new shape and strides.
+        let mut new_dim = D::zero_index_with_ndim(self.ndim());
```
bluss: Instead of creating a new `D` here for `new_dim`, we can re-purpose `usage_counts`.

jturner314: Good idea! This brings the speed for `IxDyn` closer to that of the in-place implementation without noticeably changing the fixed-dimension cases. Now, the largest difference from the in-place implementation is ~40 ns/iter.

bluss: Nice

bluss:
I feel like there is a chance that the dataset project will have use for a `&mut self` permute_axes, but it can be added at that time then, it is not need right now.

bluss:
Thanks for working on this!

jturner314:
I noticed a slight speed increase with changing

```rust
self.dim = new_dim;
self.strides = new_strides;
self
```

to

```rust
ArrayBase {
    data: self.data,
    ptr: self.ptr,
    dim: new_dim,
    strides: new_strides,
}
```

so I made that change. I also rebased off of the latest master and reused `usage_counts`, as you suggested above.

This PR should be ready to merge.

> I feel like there is a chance that the dataset project will have use for a `&mut self` permute_axes, but it can be added at that time then, it is not need right now.

Yeah, one of the things I dislike about working in Rust is that you have to write variations on the same thing that only differ in ownership. I like the explicitness of using Rust libraries, but writing the extra methods seems like unnecessary work. Anyway, adding a `&mut self` method would be simple—move the existing implementation from `permuted_axes` to `permute_axes`, and then have `permuted_axes` call `permute_axes` on `self` and return `self`.

bluss:
ok ok that's fine, just waiting for travis to complete.

Minor note: we should strive for non-proliferation of `ArrayBase {` constructors, just to keep things more refactorable (does this make sense?)

jturner314:
> Minor note: we should strive for non-proliferation of ArrayBase { constructors, just to keep things more refactorable (does this make sense?)

Yeah, that makes sense. I edited my above post and the PR to change the `ArrayBase { }` to use `..self`, which should help a little (now we can add/remove fields other than `dim` and `strides` in `ArrayBase` without changing `permuted_axes`).

bluss:
ok FRU works well when we are not changing any type params thanks.

PR #397, opened 2017-12-28
Update to newest (c)blas-sys, blas-src

bluss:
Update dependencies, code and documentation for new changes to blas-sys,
cblas-sys and blas-src.

ndarray now links to blas-src. Use it to configure the blas backend for
ndarray.

Also add a subcrate that tests this setup, so that it works. It's using
system openblas, which we have set up so that it should work in travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,6 +25,7 @@ addons:
   apt:
     packages:
       - libopenblas-dev
+      - gfortran
 script:
   - |
       cargo build --verbose --no-default-features &&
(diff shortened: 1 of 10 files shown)


bluss:
cc @IvanUkhov just for interest, I think the new crates work ok, it's at least easy to get at the system openblas (and avoid compiling them). :smile:

cc @termoshtt as ndarray-linalg author

bluss:
Travis passed, except for the jobs that were skipped.

PR #405, opened 2018-01-11
Add ArcArray and turn RcArray into ArcArray

bluss:
- ArcArray is like RcArray but using Arc instead, so with thread safe reference
counting. The use case is threading compatible shared ownership of
arrays.
- Remove RcArray; The names remain but they now point to ArcArray.


bluss:
Waiting with this to see if RcArray and ArcArray can exist at the same time. I would be very tempted to say you aren't going to need it to that duplication, though.

PR #409, opened 2018-01-16
Implement Hash for Dim<I> where Dim<I>: Dimension

jturner314:
I noticed #408 and thought I'd write a quick implementation. It uses `Dimension::slice()` to easily ensure that equivalent `Ix2` and `IxDyn` hash to the same value, for example. Fwiw, `[T; N]` and `std::vec::Vec` in the standard library use the same hashing strategy (hash a slice of the contents).

cc @Forty-Bot

### DIFF
--- a/src/dimension/dim.rs
+++ b/src/dimension/dim.rs
@@ -7,6 +7,7 @@
 // except according to those terms.
 
 use std::fmt;
+use std::hash;
 use itertools::zip;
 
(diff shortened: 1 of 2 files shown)


bluss:
Thanks!

@Forty-Bot your suggestion was implemented quickly, feel free to ask on any of the help wanted (or other!) issues if you want to help with them

PR #412, opened 2018-01-25
Add .reborrow() methods to array views

bluss:
`ArrayView<'a, T, D>` is invariant in parameters `T` and `'a`, because
`ArrayBase<S, D>` is invariant in the parameter `S` because the array
base is using the projection `<S as Data>::Elem` in the struct
definition.

This means that the equivalent of the `&'a T -> &'b T` conversion (valid
whenever `'a: 'b` is not naturally available for array views; we can
provide it manually, with a method.

### DIFF
--- a/src/impl_views.rs
+++ b/src/impl_views.rs
@@ -82,6 +82,16 @@ impl<'a, A, D> ArrayView<'a, A, D>
         ArrayView::new_(ptr, dim, strides)
     }
 
+    /// Convert the view into an `ArrayView<'b, A, D>` where `'b` is a lifetime
+    /// outlived by `'a'`.
+    pub fn reborrow<'b>(self) -> ArrayView<'b, A, D>
+        where 'a: 'b
+    {
+        unsafe {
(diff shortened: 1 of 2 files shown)


bluss:
cc @jturner314; I'm unsure here if "reborrow" is the right name.

jturner314:
"reborrow" seems like a good name to me. You are effectively borrowing the data again; the `'a: 'b` constraint is just necessary to ensure that the lifetime of the new borrow is valid. You could call it "shrink_lifetime", "reduce_lifetime", or "shorten_life", but I like "reborrow" better.

For those not on IRC, [this was the case that caused me to encounter the issue.](https://gist.github.com/anonymous/929976231020dbf82c5fb82a510d67ad) I did find that calling `.view()`  on `state` (`Foo { mean: state.view(), covariance: view }`) was a workaround for this particular case, but I'm not sure that workaround would work for all cases.

Robbepop:
I was really in need of this API some time ago. Thanks for introducing it! :)

bluss:
Good. It is a temporary solution.

PR #417, opened 2018-02-19
create array of ones similar to numpy for convenience

ehsanmok:
@bluss this is my very first commit to a Rust project. Looking for more learning opportunities :)

As mentioned [here](https://github.com/bluss/rust-ndarray/issues/416) I'd like to add those ops after this, if you think they're ok!


bluss:
Hi, this looks good, thanks a lot. I don't always have so much time, so I don't always respond very quickly.

Issue #419, opened 2018-02-25
How to perform in-place operations on parts of a vector provided an index array/vector?


rcarson3:
First off, I just want to say that I'm still very much getting the hang of ndarray. So far, I understand the basics like how to perform matrix multiplication, slicing, and in place mapping functions. However, I've found the documentation to be at times a little hard to understand. 

Now on to my actual question, so in Numpy or Fortran I can easily do something like the following:

```
vec = np.zeros((6,1))

index = [1, 3, 5]

mat = np.ones((3,3))
x = np.ones((3,1))

vec[index] = mat.dot(vec)
```

It's the being able to assign values to a vector or an array at these indices that I'm trying to get a better hold of using ndarray. 

Currently, the only way I can really think of how to do this is the following:

```
let mut vec = Array::<f32,_>::zeros((3,3));

let mut mat = Array::<f32, _>::zeros((3,3));
mat.fill(1.0);

let mut x = Array::<f32, _>::zeros((3,1));
x.fill(1.0);

ind = vec!(1, 3, 5);

let temp = mat.dot(&x);

for (i, &index) = ind.iter().enumerate(){
      vec[[index, 0]] = temp[[i, 0]];
}
```

I could make this into function, but I'd like to believe there's a built in way to do this in ndarray that I'm just missing or have over looked that avoids the need for the temporary arrays/vectors creation.

jturner314:
I'm not quite sure what your exact objectives are, so here are a few solutions depending on what you're trying to achieve. (Note that all these examples use `f64` instead of `f32`; if you want `f32`, you can specify it when creating the arrays, e.g. `Array::<f32>::zeros(6)`.)

This is the simplest example:

```rust
#[macro_use]
extern crate ndarray;

use ndarray::prelude::*;

fn main() {
    let mut vec = Array1::zeros(6);
    let mat = Array2::from_elem((3, 3), 1.);
    let x = Array1::from_elem(3, 1.);
    vec.slice_mut(s![1..;2]).assign(&mat.dot(&x));
}
```

Some notes:

* This treats `vec` and `x` as one-dimensional arrays (which is what I prefer when working with column vectors since with `Array1` you get compile-time checking that they're one-dimensional). Dot products with one-dimensional arrays are the same in `ndarray` as they are in NumPy: 1D dot 1D is a vector dot product, 1D dot 2D is a row vector times a matrix, 2D dot 1D is a matrix times a column vector, and 2D dot 2D is a matrix product.
[...]

PR #421, opened 2018-03-01
Add ndarray for NumPy users doc

### LINKED ISSUE #399: Add quick reference page (Docs)
Inspiration: https://eigen.tuxfamily.org/dox/group__QuickRefPage.html

jturner314:
Motivated by @rcarson3 in #419, I put together an initial "`ndarray` for NumPy users" document ([rendered here](https://github.com/jturner314/rust-ndarray/blob/ndarray-for-numpy-users/ndarray-for-numpy-users.md)). It needs more stuff added (maybe some examples solving the same problems with NumPy and `ndarray`?), and I'm not convinced that a big table is the best way to organize things, but it's a start.

@rcarson3 What do you think so far?

Where's the best place to put this? Just host it on GitHub at `bluss/rust-ndarray` and link to it from the README and docs, or is there a good way to put it on its own page in the docs themselves?

Fixes #399

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,18 +6,18 @@ matrix:
   include:
     - rust: 1.22.1
       env:
-       - FEATURES='test'
+       - FEATURES='test docs'
     - rust: stable
       env:
-       - FEATURES='test'
+       - FEATURES='test docs'
        - CARGO_INCREMENTAL=0
     - rust: beta
       env:
-       - FEATURES='test'
+       - FEATURES='test docs'
        - CARGO_INCREMENTAL=0
     - rust: nightly
       env:
-       - FEATURES='test'
+       - FEATURES='test docs'
        - IS_NIGHTLY=1
        - CARGO_INCREMENTAL=0
 branches:
--- a/src/doc/mod.rs
+++ b/src/doc/mod.rs
@@ -1,0 +1,1 @@
+//! Standalone documentation pages.
--- a/src/doc/mod.rs
+++ b/src/doc/mod.rs
@@ -1,1 +1,3 @@
 //! Standalone documentation pages.
+
+pub mod ndarray_for_numpy_users;
--- a/src/doc/ndarray_for_numpy_users/coord_transform.rs
+++ b/src/doc/ndarray_for_numpy_users/coord_transform.rs
@@ -1,0 +1,138 @@
+//! Example of rotation with Euler angles.
+//!
+//! This is an example of some coordinate transformations (using Euler angles)
+//! for illustrative purposes. Note that other crates such as
+//! [`cgmath`](https://crates.io/crates/cgmath) or
+//! [`nalgebra`](https://crates.io/crates/nalgebra) may be better-suited if
+//! most of your work is coordinate transformations, since they have built-in
+//! geometry primitives.
+//!
+//! This is the original Python program:
+//!
+//! ```python
+//! # Euler angles (rows) for four coordinate systems (columns).
+//! nelems = 4
+//! bunge = np.ones((3, nelems))
+//!
+//! # Precompute sines and cosines
+//! s1 = np.sin(bunge[0, :])
+//! c1 = np.cos(bunge[0, :])
+//! s2 = np.sin(bunge[1, :])
+//! c2 = np.cos(bunge[1, :])
+//! s3 = np.sin(bunge[2, :])
+//! c3 = np.cos(bunge[2, :])
+//!
+//! # Rotation matrices.
+//! rmat = np.zeros((3, 3, nelems), order='F')
+//! for i in range(nelems):
+//!     rmat[0, 0, i] = c1[i] * c3[i] - s1[i] * s3[i] * c2[i]
+//!     rmat[0, 1, i] = -c1[i] * s3[i] - s1[i] * c2[i] * c3[i]
+//!     rmat[0, 2, i] = s1[i] * s2[i]
+//!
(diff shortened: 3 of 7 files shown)


rcarson3:
@jturner314 I really like this.

I do agree that a giant table might not be a best way to convey the information at once.
I could see potentially something along the lines of breaking the table down into sections like the following: 

Array Indexing;
Array Bounds;
Array Operations
Array Creation;
Array Manipulation so like flattening, reshape, transpose;
... maybe Miscellaneous topic.

I feel like examples could be nice to have, but I would definitely say what you have right now is a great start for those of us coming over from other languages. I'll see if I can't come up with a few examples this weekend when I have some more time that could potentially go in here.

Also, one thing that I might like to see added is just a simple mention for how to create a Fortran memory layout array. While it didn't take me too long to find in the docs, I think it'd be nice to have it just listed here as well.

jturner314:
I added some more stuff and reorganized the big table into sections. ([rendered here](https://github.com/jturner314/rust-ndarray/blob/ndarray-for-numpy-users/ndarray-for-numpy-users.md)) Any thoughts?

@rcarson3 That would be awesome if you have any examples to include.

Would some discussion of `Array` vs. `ArrayView` vs. `ArrayViewMut` be helpful?

rcarson3:
@jturner314  I think the current setup is fine. I don't know if discussions on ```Array```, ```ArrayView```, and ```ArrayViewMut``` is really necessary, since the docs do a decent enough job on it. However, if other people have thoughts on the manner then maybe?

I've written some ```ndarray``` examples [here](https://github.com/rcarson3/Rust_Language_Trials/blob/master/ndarray_trials/src/main.rs). It includes some common operations that I thought people might want to do. It could probably use an index array example, but I wanted to get some quick examples together first.

The equivalent ```numpy``` version of the code is [here](https://github.com/rcarson3/Rust_Language_Trials/blob/master/ndarray_trials/src/numpy_rust_eqv.py).

If you can think of any other common features that might be useful to include let me know, and I'll iterate on these a bit more. I'm not sure if the best place to put them would be in the document just due to length, but maybe they could go in the [examples folder](https://github.com/bluss/rust-ndarray/tree/master/examples). I currently just have them hosted on my repo I made to learn Rust, but I can just as easily make a pull request on your ndarray-for-numpy-users branch for the examples as well.

jturner314:
Yeah, I think it makes the most sense to put the example(s) in the `examples` directory because then they get built when running `cargo test`. This makes it easier to ensure that they don't get out-of-date with changes to the library. I'd prefer to put the Python and Rust programs in their own directory within `examples`, like this:

```
examples/
\- ndarray-for-numpy-users/
   |- main.py
   \- main.rs
```

Will you please submit a PR to my `ndarray-for-numpy-users` branch?

I added another section to the `ndarray` for NumPy users document: "Other Rust array/matrix crates". `ndarray` is the most similar to NumPy, but for some use-cases, other crates may be a better choice (e.g. `cgmath` for coordinate transformations for computer graphics).

bluss:
That's really great work.

For C & C on the document, I'd like us to be clear in the style / algorithms we promote.

From the top.

`a[a.len()-1]` should be `a[a.len() - 1]` to use spaces around the operators.

Instead of current `a[(1,4)]` it should be `a[[1, 4]]` to use the preferred style in both whitespace and brackets IMO.

Slices are fine, the range dots are usually cuddled even when using spaces. So the style would prefer this: `x..x + 1` and that's fine.

There's an unqualified `.into()` call in there, that's not something I'd approve for inclusion.  `a.slice_axis(Axis(0), (-5..).into())` should be `a.slice_axis(Axis(0), Slice::from(-5..))`

Where we mention `a.dim()` and `a.shape()`, should we mention their differing return types? Probably something about that, to explain why there are two.

For `a.t()` we can mention it's a view.

Some things we can add:

- `Array::ones` (it's new, not yet in a released version)

- `np.zeros_like(a)` with equivalent in `Array::zeros(a.raw_dim())`  (but this doesn't match the memory layout, just the dimensions)

- Maybe we can mention `a.outer_iter()` as the equivalent of numpy `iter(a)`, but there is an unfortunate explosion of equivalents for the specific case of a 2D array. For example `.genrows()` and `.gencolumns()` are neat producers / iterables.

I don't know if we should avoid mentioning `uninitialized`. They do approximately the same thing in Python and Rust, but in Rust it's easier to get a broken program by using it incorrectly.

bluss:
In the spirit of style, I think I'd skip `a.shape()[axis]` in favour of the `len_of` method

jturner314:
Okay, I changed those things. I also did the following:

* Move the page into a module in the API docs (which is only included with the `docs` feature).
* Mention the `ndarray` for NumPy users page on the landing page of the API docs.
* Add a table of contents.
* Add a separate "Iteration" section.
* Add some more examples (and move @rcarson3's examples into the `ndarray_for_numpy_users` module).

The examples maybe could use a little more description or reorganization, but I think they're good enough for the time being. Everything else seems pretty good.

You can view the docs by running `cargo +nightly doc --open --features=docs`. (The docs also build on stable, but the Markdown parser on stable is buggy and produces incorrect output in some of the tables.)

What do you think? Is this ready to merge? (I'll clean up the commit history first.)

bluss:
Yes, it's ready for merge. It looks great!

bluss:
(Feel free to merge when you have rebased it)

bluss:
:1st_place_medal: fantastic!

Issue #441, opened 2018-04-27
What is the standard way to bring arrays into standard order?


ExpHP:
Suppose I have an arbitrary array; any model of ownership, any number of dimensions, possibly non-contiguous, with strides in whatever order.  How do I get an owned array in standard layout?  And could it be better documented?

Based on the documentation of ArrayBase alone, it seems to me that the following is the only safe bet: 

```rust
Array::from_shape_vec(arr.raw_dim(), arr.iter().cloned().collect())
```

jturner314:
Yeah, that's the simplest way, I think. The other obvious things to try (`.to_owned()` and `.map()`/`.mapv()`) don't alter the layout of contiguous arrays, so they don't help in this case.

Something like this might be faster for handling contiguous arrays in standard layout, but you'd have to benchmark to know for sure:

```rust
if let Some(slc) = self.as_slice() {
    Array::from_shape_vec(arr.raw_dim(), slc.to_vec())
} else {
    Array::from_shape_vec(arr.raw_dim(), arr.iter().cloned().collect())
}
```

If your element type is `Copy`, an alternative approach is to allocate a new array with `::uninitialized()` and then `.assign()` the data.

> And could it be better documented?

I've created #442.

jturner314:
Resolved by #442.

It would be nice in the future to provide a more concise way to convert between memory layouts, possibly using something like the `Cow`-like types described in #390.

PR #442, opened 2018-04-27
Add note about memory layout in .to_owned() docs

jturner314:
Questions:

* Is this the right place to put this?
* Can we improve upon these example implementations? They're pretty verbose.

### DIFF
--- a/src/impl_methods.rs
+++ b/src/impl_methods.rs
@@ -124,7 +124,36 @@ impl<A, S, D> ArrayBase<S, D> where S: Data<Elem=A>, D: Dimension
         }
     }
 
-    /// Return an uniquely owned copy of the array
+    /// Return an uniquely owned copy of the array.
+    ///
+    /// If the input array is contiguous and its strides are positive, then the
+    /// output array will have the same memory layout. Otherwise, the layout of
+    /// the output array is unspecified. If you need a particular layout, you
+    /// can allocate a new array with the desired memory layout and
+    /// [`.assign()`](#method.assign) the data. Alternatively, you can collect
+    /// an iterator, like this for a result in standard layout:
+    ///
+    /// ```
+    /// # use ndarray::prelude::*;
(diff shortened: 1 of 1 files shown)


ExpHP:
My feelings:

* `to_owned` is one of the places I looked; I think it is a good location
    * This location has the advantage of also being to describe Fortran layout, which wouldn't work so well on `is_standard_layout`
* I'm also glad that it includes the search phrase "standard layout"

I would consider #441 to be solved by this.

src/impl_methods.rs
```diff
@@ -124,7 +124,35 @@ impl<A, S, D> ArrayBase<S, D> where S: Data<Elem=A>, D: Dimension
...
     }
 
-    /// Return an uniquely owned copy of the array
+    /// Return an uniquely owned copy of the array.
+    ///
+    /// For performance, the memory layout of the resulting array is arbitrary.
+    /// If you need a particular layout, you can allocate a new array with the
+    /// desired memory layout and [`.assign()`](#method.assign) the data.
+    /// Alternatively, you can collect an iterator, like this for a result in
+    /// standard layout:
```
bluss: This can be fine as a contract, but I don't really agree with it. The layout of the output should be the same as that of the input, if possible. Then we restrict and say it's only possible if the input is contiguous (which is more or less true). Then yes, that's not entirely correct with the current implementation, since not all contiguous arrays are covered right now (but most..).

jturner314: Okay, how about

> If the input array is contiguous and its strides are positive, then the output array will have the same memory layout. Otherwise, the layout of the output array is unspecified. If you need a particular layout…

As far as I can tell, this description matches the current implementation. Once we add support for negative strides in `.is_contiguous()`, we can remove the "strides are positive" constraint.

bluss: Yes that's good

PR #444, opened 2018-04-28
add tests with discontiguous owned data

ExpHP:
Sprinkle about a couple of tests that produce discontiguous owned arrays, including one RcArray.  I figured there's no need for the tests to be that intensive, because ndarray handles ownership entirely orthogonally to slicing/strides; but anything is better than nothing.

This also adds a smoke test for `Drop` data.  Given that ndarray never manually impls `Drop` and that all usages of `forget` and unsafe `std::ptr` methods easily fit in a screenful of grep output[^1], there's little reason to believe this test should fail; but it does give bragging rights.

---

[^1] I'll note there was *one* call to `ptr::write` in `from_shape_fn` that made me give a double take before finally concluding that it was panic safe.  Maybe that is what deserves to be tested instead...

### DIFF
--- a/tests/array.rs
+++ b/tests/array.rs
@@ -274,28 +274,37 @@ fn test_slice_inplace_with_subview_inplace() {
         *elt = i;
     }
 
-    let mut vi = arr.view();
-    vi.slice_inplace(s![1.., 2, ..;2]);
-    assert_eq!(vi.shape(), &[2, 1, 2]);
-    assert!(
-        vi.iter()
-            .zip(arr.slice(s![1.., 2..3, ..;2]).iter())
-            .all(|(a, b)| a == b)
-    );
+    {
+        let mut vi = arr.view();
+        vi.slice_inplace(s![1.., 2, ..;2]);
+        assert_eq!(vi.shape(), &[2, 1, 2]);
+        assert!(
+            vi.iter()
+                .zip(arr.slice(s![1.., 2..3, ..;2]).iter())
+                .all(|(a, b)| a == b)
+        );
 
-    let mut vi = arr.view();
-    vi.slice_inplace(s![1, 2, ..;2]);
-    assert_eq!(vi.shape(), &[1, 1, 2]);
-    assert!(
-        vi.iter()
-            .zip(arr.slice(s![1..2, 2..3, ..;2]).iter())
-            .all(|(a, b)| a == b)
-    );
+        let mut vi = arr.view();
+        vi.slice_inplace(s![1, 2, ..;2]);
+        assert_eq!(vi.shape(), &[1, 1, 2]);
+        assert!(
+            vi.iter()
+                .zip(arr.slice(s![1..2, 2..3, ..;2]).iter())
+                .all(|(a, b)| a == b)
+        );
 
-    let mut vi = arr.view();
+        let mut vi = arr.view();
+        vi.slice_inplace(s![1, 2, 3]);
+        assert_eq!(vi.shape(), &[1, 1, 1]);
+        assert_eq!(vi, Array3::from_elem((1, 1, 1), arr[(1, 2, 3)]));
+    }
+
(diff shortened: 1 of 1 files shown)


tests/array.rs
```diff
@@ -807,6 +816,49 @@ fn owned_array_with_stride() {
...
+
+#[test]
+fn owned_array_discontiguous_drop() {
+    use ::std::rc::Rc;
+    use ::std::cell::RefCell;
+    use ::std::collections::BTreeSet;
+
+    struct InsertOnDrop<T: Ord>(Rc<RefCell<BTreeSet<T>>>, Option<T>);
+    impl<T: Ord> InsertOnDrop<T> {
+        fn new(set: Rc<RefCell<BTreeSet<T>>>, value: T) -> Self {
```
jturner314: Will you please remove this `new` method and just construct `InsertOnDrop` manually with `InsertOnDrop(set.clone(), Some(x))`? (`new` is used in only one place, and it would be nice to make the test more compact. I see the advantage of hiding the implementation detail that `value` should start as `Some`, but since the use is so close to the implementation, I prefer conciseness over encapsulation in this case.)

tests/array.rs
```diff
@@ -807,6 +816,49 @@ fn owned_array_with_stride() {
...
+    }
+
+    let set = Rc::new(RefCell::new(BTreeSet::new()));
+    {
+        let v: Vec<_> = (0..12).map(|x| InsertOnDrop::new(set.clone(), x)).collect();
+        let mut a = Array::from_shape_vec((2, 6), v).unwrap();
+        // discontiguous and non-zero offset
+        a.slice_inplace(s![.., 1..]);
+    }
+    // each item was dropped
```
jturner314: Will you please change this comment to say "each item was dropped exactly once"?

tests/array.rs
```diff
@@ -1372,6 +1424,19 @@ fn to_owned_neg_stride() {
...
+fn discontiguous_owned_to_owned() {
+    let mut c = arr2(&[[1, 2, 3],
+                       [4, 5, 6]]);
+    c.slice_inplace(s![.., ..;2]);
+
+    let co = c.to_owned();
+    assert_eq!(c.strides(), &[3, 2]);
+    assert_eq!(co.strides(), &[2, 1]);
+    assert_eq!(c.shape(), co.shape());
+    itertools::assert_equal(c.iter(), co.iter());
```
jturner314: I'd prefer just `assert_eq!(c, co)` instead of `itertools::assert_equal()`, unless there's a specific reason for using `itertools::assert_equal()` here.

tests/array.rs
ExpHP: No problem, that was an artefact of copypasta from another project where it was used more heavily

tests/array.rs
ExpHP: No reason, I think I was just confused and thought PartialEq would check strides for some reason

jturner314:
Everything looks good to me. If you squash the PR into a single commit and force-push it to your branch (`ExpHP:owned-strided-tests`), I'll merge it.

(I wouldn't mind squashing it myself, but if I did that, then GitHub wouldn't let me mark the PR as merged, and you wouldn't be listed as the committer.)

ExpHP:
> (I wouldn't mind squashing it myself, but if I did that, then GitHub wouldn't let me mark the PR as merged, and you wouldn't be listed as the committer.)

I think Github has an option you can pick from a dropdown near the "Merge" button that makes it do a squash?  (unless that's what you're referring to, though I would have hoped/assumed that GitHub would have it work in such a way as to preserve this metadata)

jturner314:
On further investigation, it looks like if "Allow edits from maintainers" is checked on the PR, then it might be possible for me to squash and force-push directly to the contributor's PR branch. I'll have to try that the next time I encounter this situation so that I don't have to bother the contributor with squashing.

PR #448, opened 2018-05-10
Remove unnecessary lifetime parameter

jturner314:
Rust 1.23 fixed the original issue (#103) that required `'b` to be added. (See rust-lang/rust#32008, rust-lang/rust#45425, and rust-lang/rust#45435 for the relevant Rust issues/PRs.)


bluss:
Thanks!

PR #451, opened 2018-05-12
Added detailed error messages to improve ease of use.

gilescope:
Rust compiler puts a lot of effort into great error messages and I really appreciate them. I'm only going to be a ndarray noob once, so I'm taking this oppotunity to make it easier for other noobs.

This is to close: https://github.com/bluss/ndarray/issues/449

Please let me know your thoughts? If you're happy I'll add in a few more tests to try and hit most of the error messages.


bluss:
Better error messages are great, but this solution increases the size and overhead of the ShapeError struct **dramatically**, so it can't be merged.

PR #456, opened 2018-05-23
Update ndarray-rand to rand 0.5.0

### LINKED ISSUE #455: ndarray-rand 0.5
An update to rand 0.5 breaks ndarray-rand. For example

    let m = Array3::random((6, 6, 6), Uniform::new(0.0, 1.0));

the trait `rand::distributions::IndependentSample<_>` is not implemented for `rand::distributions::Uniform<{float}>`

jturner314:
This PR tries to replicate the old behavior as closely as possible.

This PR removes support for `Sample` and `IndependentSample`, which are deprecated in `rand 0.5.0`, and replaces them with the new `Distribution` trait.

Fixes #455.

### DIFF
--- a/ndarray-rand/src/lib.rs
+++ b/ndarray-rand/src/lib.rs
@@ -12,9 +12,9 @@
 extern crate rand;
 extern crate ndarray;
 
-use rand::Rng;
-use rand::distributions::Sample;
-use rand::distributions::IndependentSample;
+use rand::{thread_rng, Rng, SeedableRng};
+use rand::distributions::Distribution;
+use rand::rngs::SmallRng;
 
 use ndarray::{
     ArrayBase,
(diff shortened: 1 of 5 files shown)


ndarray-rand/src/lib.rs
```diff
@@ -73,18 +81,20 @@ impl<S, D> RandomExt<S, D> for ArrayBase<S, D>
           D: Dimension,
 {
     fn random<Sh, IdS>(shape: Sh, dist: IdS) -> ArrayBase<S, D>
-        where IdS: IndependentSample<S::Elem>,
+        where IdS: Distribution<S::Elem>,
               Sh: ShapeBuilder<Dim=D>,
     {
-        Self::random_using(shape, dist, &mut rand::weak_rng())
+        let mut rng = SmallRng::from_rng(thread_rng())
+            .unwrap_or_else(|err| panic!("create SmallRng from thread_rng failed: {:?}", err));
```
nilgoyette: Isn't this the same as `expect("create SmallRng from thread_rng failed")`? The way I understand the [documentation](https://doc.rust-lang.org/std/result/enum.Result.html#method.expect), I believe it is equivalent. If so, I think we should prefer `expect`.

jturner314: Yes, you're right. I just made that change. Thanks.

jturner314:
No one has objected to this PR, so I've merged it.

jturner314:
By the way, @nilgoyette thanks for reviewing this PR.

PR #458, opened 2018-05-28
Bump num-traits, num-complex

alexbool:
This is a breaking change since `num-complex` is a part of `ndarray`'s a public API


alexbool:
@bluss friendly ping

jturner314:
I haven't heard from @bluss in quite a while, so I'm going ahead and merging this.

Thanks @alexbool.

Issue #452, opened 2018-05-17
map_axis with FnMut(ArrayViewMut1<'a, A>)


LukeMathWalker:
Currently `map_axis` accepts only a mapping taking as argument a 1-dimensional `ArrayView` (https://docs.rs/ndarray/0.11.2/ndarray/struct.ArrayBase.html#method.map_axis).
What if I wanted to use a function that takes an `ArrayViewMut1` as argument?

Is there another way to achieve the effect of `map_axis` using other iterators/Zip/etc.?
I have tried to reason around that but I have not yet been able to come up with a solution...

The use-case I have in mind is a reducer that takes a lane in the array as argument and applies a function that modifies the one-dimensional array in place to return a single value for each lane.

rcarson3:
Normally, I would use the [`azip!` macro](https://docs.rs/ndarray/0.11.2/ndarray/macro.azip.html) to do something like what you want. I actually use this quite often to zip together different ArrayView types and change one of them or even just a single ArrayViewMut type with nothing else zipped with it. Typically I use the `.axis_iter()` or `.axis_iter_mut()` methods in the `azip!` macro.

However, I can't seem to get it to work with the `.lanes()` method. I really haven't had a need to use that method, so I'm more unfamiliar with it. However looking at the documents, it looks like it should work, since it implements the `NdProducer` trait. If anyone more familiar with this method has gotten it to work I'd be curious to see how to use this with `azip!` or even the `Zip` iterator.

LukeMathWalker:
You gave me an idea! I'll work on it tomorrow and then I'll report on the results.
Even though, having had a look at `map_axis` implementation, my question becomes: what could go wrong replacing `ArrayView::new_` with `ArrayViewMut::new_` in
```
    pub fn map_axis<'a, B, F>(&'a self, axis: Axis, mut mapping: F)
        -> Array<B, D::Smaller>
        where D: RemoveAxis,
              F: FnMut(ArrayView1<'a, A>) -> B,
              A: 'a,
    {
        let view_len = self.len_of(axis);
        let view_stride = self.strides.axis(axis);
        // use the 0th subview as a map to each 1d array view extended from
        // the 0th element.
        self.subview(axis, 0).map(|first_elt| {
            unsafe {
                mapping(ArrayView::new_(first_elt, Ix1(view_len), Ix1(view_stride)))
            }
        })
    }
```
with all the necessary adjustments in the signature to produce a `map_axis_mut`?

[...]

PR #461, opened 2018-06-02
Order statistics

LukeMathWalker:
### Motivation
Percentiles are a nice-to-have feature, but they turn out to be extremely useful when dealing with Machine Learning projects (in particular, to support efficient and distribution-agnostic splitting criteria for Decision Trees operating on continuous values,  [paper - Sec 3.3](https://www.microsoft.com/en-us/research/publication/efficient-determination-dynamic-split-points-decision-tree/?from=http%3A%2F%2Fresearch.microsoft.com%2Fen-us%2Fum%2Fpeople%2Fdmax%2Fpublications%2Fsplits.pdf))

### Overview
I have added an array method called `percentile_axis_mut` that shuffles in-place each 1-dimensional lane along the specified axis and returns the desired percentile.
The function signature is pretty similar to [np.percentile](https://docs.scipy.org/doc/numpy/reference/generated/numpy.percentile.html). Main external differences/limitations:
- `q` is required to be in range `[0, 1]` instead of range `[0, 100]`;
- no `out` parameter;
- `overwrite_input` is `True`, being a `_mut` method. The `False` behaviour can be recovered by adding a `percentile_axis` method taking its argument as immutable reference;
- `interpolation=lower` is the only supported (and thus default) behaviour. The function can be easily extended to support `highest` and `nearest` strategy, because they both require the recovery of a single element by index;
- `keepdims=False`.

To keep functions small and separate concerns I have also implemented two new methods for `ArrayViewMut1`: `ith_mut` and `partition_mut`.
They can be used as stepping stones to provide `ith_axis_mut` and `partition_axis_mut` methods for n-dimensional arrays. Ideally I would have liked to implement both `ith_mut` and `partition_mut` for n-dimensional arrays, but I ran into some doubts:
- should we flatten the array implicitly, as numpy does in [np.ndarray.partition](https://docs.scipy.org/doc/numpy/reference/generated/numpy.ndarray.partition.html) if `axis=-1`, or is it better to make it explicit that the method runs on 1-dimensional objects?
- how do I flatten an array? In [ndarray for Numpy users](https://docs.rs/ndarray/0.11/ndarray/doc/ndarray_for_numpy_users/) @jturner314 suggests to use
```
Array::from_iter(a.iter())
``` 
to emulate `np.flatten` but I'd actually like to emulate `np.ravel` behaviour in this case to avoid extra memory allocation.

### Algorithm
`np.percentile` uses [introselect](https://en.wikipedia.org/wiki/Introselect) while my `partition_axis_mut`/`ith_mut` relies on [randomized quickselect](https://en.wikipedia.org/wiki/Quickselect): same expected complexity in the average case, O(n), while quickselect has O(n^2) complexity in the worst case compared to O(n) complexity in the worst case for `introselect`.
Nothing against introselect - I have started with quickselect because it was easier to implement, but I structured the code to make it simple to switch to introselect with a minimum amount of effort (i.e. we just need a working implementation of [median of medians](https://en.wikipedia.org/wiki/Median_of_medians)).

To use randomized quickselect I need to sample random integers => I need to use the `rand` crate.
As far as I understand, `ndarray` does not want to add `rand` as a dependency, thus justifying the existence of [`ndarray-rand`](https://github.com/bluss/ndarray/tree/master/ndarray-rand).
Choosing the pivot index randomly improves running time increasing the number of inputs falling in the "average case" bucket. But if `rand` is an issue, it is sufficient to modify `random_pivot`, a private helper function used by `ith_mut`, into a function returning a pivot index using a deterministic algorithm (first value, last value, middle value, whatever) to get rid of the dependency.

### Trait bounds
All new methods work for `A: [...] + Ord`, thus excluding `f32` and `f64` as valid element types. Because of the peculiarities of`NaN` handling, I'd imagine that specific float equivalents of these methods will have to be added (as in [`np.nanpercentile`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.nanpercentile.html)).
This will be an on-going problem for all order-induced statistics (maximum, minimum, etc.): I am happy to work on it, but I'd like to first understand what kind of design choices would be preferred @bluss @jturner314.

### P.S.
- If #460 gets merged I can use it to simplify `percentile_axis_mut` function body, which now relies on @jturner314's clever workaround (see #452);
- I am not very convinced by `ith_mut`/`ith` as method names. I was considering `pick_mut`/`pick` or `find_mut`/`find` as alternatives.


src/impl_1d.rs
```diff
@@ -23,5 +26,77 @@ impl<A, S> ArrayBase<S, Ix1>
...
+    /// where n is the number of elements in the array.
+    ///
+    /// **Panics** if `i` is greater than or equal to `n`.
+    pub fn ith_mut(&mut self, i: usize) -> A
+        where A: Ord + Clone,
+              S: DataMut,
+    {
+        let n = self.len();
+        if n == 1 {
+            (&self[0]).clone()
```
nilgoyette: I don't have a compiler in front of me so I may be wrong, but I don't understand why you can't use `self[0].clone()`? Why do you need to take the reference before?

Idem for 2 other clones below.

src/impl_1d.rs
```diff
@@ -23,5 +26,77 @@ impl<A, S> ArrayBase<S, Ix1>
...
+        } else {
+            let pivot_index = random_pivot(n);
+            let partition_index = self.view_mut().partition_mut(pivot_index);
```
nilgoyette: I don't understand your usage of `view_mut()` here. Why can't you write

    let partition_index = self.partition_mut(pivot_index);

? We already know that `self` is mutable. Is it important if `self` is a `OwnedRepr` or a `ViewRepr`?

src/numeric/impl_numeric.rs
```diff
@@ -115,6 +115,49 @@ impl<A, S, D> ArrayBase<S, D>
...
+    ///
+    /// Some examples:
+    /// - `q=0.` returns the minimum along each 1-dimensional lane;
+    /// - `q=0.5` returns the median along each 1-dimensional lane;
+    /// - `q=1.` returns the maximum along each 1-dimensional lane.
+    /// (`q=0` and `q=1` are considered improper percentiles)
+    ///
+    /// The array is shuffled **in place** along each 1-dimensional lane in
+    /// order to produce the required percentile without allocating a copy
+    /// of the original array. Each 1-dimensional lane is shuffled indipendently
```
nilgoyette: independently

src/impl_1d.rs
LukeMathWalker: No, it is not necessary. Fixed it in every occurrence.

src/numeric/impl_numeric.rs
LukeMathWalker: Fixed.

src/impl_1d.rs
LukeMathWalker: It is actually unnecessary, given that we have the `S: DataMut` bound. I removed it.

LukeMathWalker:
Thanks for taking the time of reviewing it @nilgoyette - I've still got a lot of room to improve.

LukeMathWalker:
Out of curiosity, I did a small benchmark exercise against `np.percentile`.
I compare the time necessary to compute the median of a 1-dimensional array over 256 randomly-drawn arrays of different dimensions (@jturner314's `ndarray-npy` has been pretty handy to get this done quickly and reliably).
These are the results:
[benchmark_w_numpy.tar.gz](https://github.com/bluss/ndarray/files/2088061/benchmark_w_numpy.tar.gz)
Some samples: [`n` is the length of the 1-d array]
256_runs_128_elements_1048576_high
256_runs_1024_elements_1048576_high
256_runs_16384_elements_1048576_high
256_runs_131072_elements_1048576_high


NumPy has a smaller constant with a slightly higher overhead [`high` is the upper limit of the range I randomly drew integers from]:
`x-axis: number of elements in the array, y-axis average running time`
means_1024_high
means_1048576_high

LukeMathWalker:
I managed to close the gap modifying the partition routine called by `ith_mut`:
image

Tonight I'll commit the required changes.

LukeMathWalker:
Added `hoare_partition_mut` as a non-public method for 1-dimensional arrays in `impl_1d.rs`.
I have also added a small test for it.

`partition_mut` still uses Lomuto's partition algorithm because it provides stronger guarantees on the positioning of the elements after the partition, as well as returning the exact index of the pivot, which Hoare's algorithm cannot provide.

LukeMathWalker:
Merged changes from `master`. I simplified `percentile_axis_mut`'s body using `map_axis_mut` (#460).

src/impl_1d.rs
```diff
@@ -23,5 +26,146 @@ impl<A, S> ArrayBase<S, Ix1>
...
+    /// `self` is shuffled **in place** to operate the desired partition:
+    /// no copy of the array is allocated.
+    ///
+    /// The method uses Lomuto's partition algorithm.
+    /// Complexity: O(`n`), where `n` is the number of elements in the array.
+    /// Average number of element swaps: (n - 1)/2 (see
+    /// (link)[https://cs.stackexchange.com/questions/11458/quicksort-partitioning-hoare-vs-lomuto/11550])
+    ///
+    /// **Panics** if `partition_index` is greater than or equal to `n`.
+    pub fn partition_mut(&mut self, partition_index: usize) -> usize
```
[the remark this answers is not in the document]
LukeMathWalker: The current implementation of Lomuto's algorithm guarantees that all elements equal to the pivot value are placed to the right of the partition index. Furthermore, `self[partition_index]` is equal to the pivot value.
Both these facts were not granted with my implementation of Hoare's partition algorithm.
I have reworked it to provide both guarantees - thus I have removed the auxiliary function and used Hoare's algorithm directly into `partition_mut`.

LukeMathWalker:
Ok, I guess I implemented/fixed everything you pointed to @jturner314.
I have also renamed `ith_mut` to `sorted_get_mut`, which sounded much clearer!

Re: floats - For usecases where no NaN is expected people could leverage things like [noisy_float](https://docs.rs/noisy_float/0.1.5/noisy_float/struct.NoisyFloat.html). In general I agree we need an implementation that ignores NaN values such as `np.nanpercentile`.
I guess we can discuss this further in an issue and include it into another pull request.

Thanks for the clarification on `ravel` - I'll look into it!

LukeMathWalker:
I had to add a couple of `pub` to make `InterpolationStrategy` importable with `use ndarray::numeric::InterpolationStrategy`. I hope I didn't expose anything that was meant to be private @jturner314

jturner314:
I did some brainstorming about percentile with floats. These are some observations:

1. We should be able to provide lower/upper/nearest percentile for `A: PartialOrd` as long as we return an `Option<A>` instead of `A` to deal with cases where the array contains a NaN (or other value for which the ordering is undefined). To do this, we need to use [`.partial_cmp`](https://doc.rust-lang.org/stable/std/cmp/trait.PartialOrd.html#tymethod.partial_cmp) for all comparisons to detect cases where the ordering is undefined. Note that this is how `np.percentile` handles arrays that contain one or more `np.nan` values: it returns `np.nan` (and issues a warning).

2. We need a separate parameter or method to allow the user to filter out NaN values instead of just returning `None`. Using a separate method would be equivalent to NumPy's `np.nanpercentile`.

3. Finding a lower/upper/nearest percentile is fairly straightforward in terms of `sorted_get_mut`, so it doesn't seem very important to provide a percentile method for non-numeric types since the user can use `sorted_get_mut` instead to accomplish the same thing fairly easily.

4. Other types similar to `f32` and `f64` are `Option<i32>`, `Option<u64>`, etc. These types have a "not a number" value and support arithmetic operations. They also implement `PartialOrd`.

This is one approach I like pretty well (showing only the methods for 1-D arrays for simplicity):

```rust
enum InterpolationStrategy {
    Upper,
    Lower,
    Nearest,
    Linear,
    Midpoint,
}

impl<A, S> ArrayBase<S, Ix1>
where
    S: Data<Elem=A>,
{
    fn percentile_mut(&mut self, q: f64, interp: InterpolationStrategy) -> Option<A>
    where
        A: PartialOrd + Add + Div + Clone + ToPrimitive + FromPrimitive,
        S: DataMut,
    {}

    fn percentile_filtered_mut(&mut self, q: f64, interp: InterpolationStrategy, predicate: impl FnMut(&A) -> bool) -> Option<A>
    where
        A: PartialOrd + Add + Div + Clone + ToPrimitive + FromPrimitive,
        S: DataMut,
    {}
}
```

The ordering constraint is now `PartialOrd` instead of `Ord` to provide support for `f32`, `f64`, `Option<i32>`, etc., and the return type is now `Option<A>` to handle the "undefined ordering" case. The `percentile_filtered_mut` method is intended to provide "ignore NaNs" functionality with a predicate `|x| !x.is_nan()`, but could also work for things like `|x| x.is_some()`.

Note that this requires `Add`, `Div`, `ToPrimitive`, and `FromPrimitive`, which prevents use with non-numeric types, but as I mentioned above, I don't think that's a big deal because the user can use `.sorted_get_mut()` instead.

Instead of (or in addition to) `percentile_filtered_mut`, we could provide predefined NaN-specific functionality:

```rust
// Trait providing generic NaN handling for numeric types.
pub trait MaybeNaN: PartialOrd {
    // Returns `true` for any value of this type which is not a numeric value.
    //
    // For example, `::std::f64::NAN.is_nan() == `true` and `None.is_nan() == true`. Values for which `.is_nan() == true` are ignored in some statistical calculations.
    //
    // **Note** The set of all values for which `value.is_nan() == false` must form a totally ordered set.
    fn is_nan(&self) -> bool;
}

impl MaybeNaN for i8 {}
impl MaybeNaN for i16 {}
impl MaybeNaN for i32 {}
impl MaybeNaN for i64 {}
impl MaybeNaN for i128 {}
impl MaybeNaN for u8 {}
impl MaybeNaN for u16 {}
impl MaybeNaN for u32 {}
impl MaybeNaN for u64 {}
impl MaybeNaN for u128 {}
impl MaybeNaN for f32 {}
impl MaybeNaN for f64 {}
impl<T: MaybeNaN> for Option<T> {}

impl<A, S> ArrayBase<S, Ix1>
where
    S: Data<Elem=A>,
{
    fn percentile_skipnan_mut(&mut self, q: f64, interpolation_strategy: InterpolationStrategy) -> A
    where
        A: MaybeNaN + Add + Div + Clone + ToPrimitive + FromPrimitive,
        S: DataMut,
    {}
}
```

This has the nice advantage that we can return `A` instead of `Option<A>` since the filtered set of values is totally ordered, but it has the disadvantage of an additional trait and a little less flexibility.

We could also use `Float` instead of `MaybeNaN`. However, this would restrict the method to floating-point types, and I don't think we could rely on the set of non-NaN values being totally ordered for all types that implement `Float`, so we'd have to return `Option<A>`.

Instead of requiring so many constraints on `A` just to be able to provide the `Midpoint` and `Linear` interpolation strategies, we could require different constraints for different interpolation strategies with something like this:

```rust
pub trait Interpolate<T> {
    fn needs_lower(q: f64, len: usize) -> bool;
    fn needs_upper(q: f64, len: usize) -> bool;
    fn interpolate(lower: Option<T>, upper: Option<T>, q: f64, len: usize) -> T;
}

struct Upper;
struct Lower;
struct Nearest;
struct Linear;
struct Midpoint;

impl<T> Interpolate for Upper<T> {
    fn needs_lower(q: f64, len: usize) -> bool {
        false
    }
    fn needs_upper(q: f64, len: usize) -> bool {
        true
    }
    fn interpolate(lower: Option<T>, upper: Option<T>, q: f64, len: usize) -> T {
       upper.unwrap()
    }
}
impl<T> Interpolate for Lower<T> {
    // ...
}
impl<T> Interpolate for Nearest<T> {
    // ...
}
impl<T: Add + Div + ToPrimitive + FromPrimitive> Interpolate for Linear<T> {
    // ...
}
impl<T: Add + Div + FromPrimitive> Interpolate for Midpoint<T> {
    // ...

    /// Performs the interpolation as `(upper + lower) / 2`.
    ///
    /// **Panics** if `lower` is `None`, `upper` is `None`, or `T` cannot represent the number 2.
    fn interpolate(lower: Option<T>, upper: Option<T>, q: f64, len: usize) -> T {
       (upper.unwrap() + lower.unwrap()) / T::from_u8(2).unwrap()
    }
}

impl<A, S> ArrayBase<S, Ix1>
where
    S: Data<Elem=A>,
{
    fn percentile_mut<I>(&mut self, q: f64) -> Option<A>
    where
        A: PartialOrd + Clone,
        S: DataMut,
        I: Interpolate<A>,
    {}

    fn percentile_filter_mut<I>(&mut self, q: f64, predicate: impl FnMut(&A) -> bool) -> Option<A>
    where
        A: PartialOrd + Clone,
        S: DataMut,
        I: Interpolate<A>,
    {}
}
```

This approach actually doesn't look all that bad to me, but I'm not sure that it's worth the additional complexity.

What do @LukeMathWalker and others think?

LukeMathWalker:
I think that your idea to make `percentile_mut` generic over the interpolation strategy using an `Interpolate` trait is great!
That would allow us to easily tune the level of requirements on `A` with a single function. It is particularly important to keep supporting numeric types that do not implement `Div` and `Mul` without having to limit our feature set for those that do.

Going back to your main point, I would be ok to implement this signature
```rust
fn percentile_mut(&mut self, q: f64, interp: InterpolationStrategy) -> Option<A>
    where
        A: PartialOrd + Add + Div + Clone + ToPrimitive + FromPrimitive,
        S: DataMut,
    {}
```
if Rust allowed us to provide a specialized implementation for `Ord` - dealing with `None` values does introduce an overhead cost to every comparison operation in the method, which would be better not to force on those who don't need it. 

Right now I'd lean towards something like:
```rust
fn percentile_mut<I>(&mut self, q: f64) -> A
    where
        A: Ord + Clone,
        S: DataMut,
        I: Interpolate<A>,
    {...}

fn percentile_[partially_ordered?]_mut<I, N>(&mut self, q: f64, null_strategy: NullStrategy) -> Option(A)
    where
        A: PartialOrd + Clone,
        S: DataMut,
        I: Interpolate<A>,
        N: NullChecker<A>
    {...}
```

where

```
pub trait NullChecker<T> {
    fn is_null(x: T) -> bool;
}

pub enum NullStrategy {
    Ignore
    Raise
}
```

In this way we would be able to retain speed for `Ord` types while providing a sufficiently general implementation for `partialOrd` types capable of handling floats as well as `Option<T>`.

LukeMathWalker:
I'd like to wrap up this PR in the next couple of weeks - can I get your view on how we want to move forward with it with respect to the last comments? @bluss @jturner314

LukeMathWalker:
`percentile_axis_mut` now supports `Lower`, `Upper`, `Midpoint`, `Nearest` and `Linear` as strategies using this signature:
```rust
    pub fn percentile_axis_mut<I>(&mut self, axis: Axis, q: f64) -> Array<A, D::Smaller>
        where D: RemoveAxis,
              A: Ord + Clone,
              S: DataMut,
              I: Interpolate<A>,
```
with
```rust
pub trait Interpolate<T> {
    [private utility methods]
}

pub struct Upper;
pub struct Lower;
pub struct Nearest;
pub struct Midpoint;
pub struct Linear;

[implementations]
```

`percentile_axis_mut` can be made much faster by using `min` on a subset of the array when `Midpoint` or `Linear` strategies are being used - I'll work on implementing `min` and `max` in order to leverage them in the implementation.

`PartialOrd` types are still not supported in this implementation.

jturner314:
I'm sorry for taking so long to get back to this.

I think that for the time being, it makes the most sense to put this functionality in a separate crate. It doesn't need access to `ndarray` internals, and it would be good to refine the API before we commit to it in `ndarray`. It's easier to make breaking changes in a crate separate from `ndarray`. I've created a repository [`ndarray-stats`](https://github.com/jturner314/ndarray-stats) that I'd like to release on crates.io.  I've copied the percentile functions you provided in this PR, added a few more functions, and worked on NaN handling. (View the docs with `cargo doc --open`.)

See the `MaybeNan` trait and the method variants in the `QuantileExt` trait for the NaN handling. It provides support for `f32`, `f64`, and `Option<integer>` types. I'd like to add a `quantile_axis_partialord_mut` method to `QuantileExt`, similar to `min_partialord` and `max_partialord`.

Other functionality that would be good to add at some point are NaN-supporting sum, mean, variance, standard deviation, and histogram methods (basically everything in https://docs.scipy.org/doc/numpy/reference/routines.statistics.html).

What do you think?

jturner314:
Okay, sounds good. By the way, if you'd like to be the maintainer of the `ndarray-stats` crate, that would be fine with me. I haven't claimed the name on crates.io yet. I just created a repository myself because that was the most expedient thing to do.

LukeMathWalker:
I'd be happy to do it - I definitely plan to engage for the long term in this niche of the Rust ecosystem.
On 4 Sep 2018, at 18:43, Jim Turner <notifications@github.com<mailto:notifications@github.com>> wrote:

Okay, sounds good. By the way, if you'd like to be the maintainer of the ndarray-stats crate, that would be fine with me. I haven't claimed the name on crates.io yet. I just created a repository myself because that was the most expedient thing to do.

PR #464, opened 2018-06-13
Remove trailing whitespace

jturner314:
I realize that this makes `git blame` harder to use on those lines where I'm removing whitespace, but there aren't many whitespace removals outside of comments and blank lines, so that shouldn't be a big issue.

I'd like to do a mass-reformat once `rustfmt` is stable, so whitespace removal could wait until then. However, the negative impact of removing the trailing whitespace now is small, and it would be really nice for people like @LukeMathWalker and me who have our editors configured to automatically remove trailing whitespace.


bluss:
Thanks, that's nice. Using rustfmt sounds good.

PR #476, opened 2018-08-07
Relax constraint on closure for map* methods

jturner314:
I'm not sure why these aren't already `FnMut` bounds instead of `Fn` bounds. Maybe it would be confusing to have a mutable closure since the elements are visited in arbitrary order, but `map` and `map_mut` already have a `FnMut` bound, and there are cases with `FnMut` where the order isn't important (e.g. when generating independent random numbers from a PRNG).


bluss:
There was some kind of thought about transparent parallelization (using specialization when possible), but I'm not sure that kind of specialiation will be allowed anymore anyway. This looks good to me, we can do the original idea when that time arrives.

PR #477, opened 2018-08-18
Attempt to document usage of ref in azip

paulkernfeld:
Hi, I was recently confused when trying to use the `azip` macro in a function that borrowed its input. I've tried to document how to correctly do this, but I have not been able to figure out how to get it to compile. Could someone help me figure out what line 61 of this file needs to look like?

### DIFF
--- a/src/zip/zipmacro.rs
+++ b/src/zip/zipmacro.rs
@@ -56,9 +56,9 @@
 ///
 /// type M = Array2<f32>;
 ///
-/// // Since this function borrows its inputs, captures must use ref
+/// // Since this function borrows its inputs, captures must use the x (x) pattern
 /// fn borrow_multiply(a: &mut M, b: &M, c: &M) {
-///     azip!(mut a (a), ref b, ref c in { *a = b * c });
+///     azip!(mut a (a), b (b), c (c) in { *a = b * c });
 /// }
 ///
 /// fn main() {
@@ -86,6 +86,8 @@
 ///     });
 ///
 ///     assert_eq!(a, &b - &c);
+///
+///     // Example of azip!() on borrowed values
 ///
 ///     borrow_multiply(&mut a, &b, &c);
 ///
--- a/src/zip/zipmacro.rs
+++ b/src/zip/zipmacro.rs
@@ -56,6 +56,11 @@
 ///
 /// type M = Array2<f32>;
 ///
+/// // Since this function borrows its inputs, captures must use ref
+/// fn borrow_multiply(a: &mut M, b: &M, c: &M) {
+///     azip!(mut a (a), ref b, ref c in { *a = b * c });
+/// }
+///
 /// fn main() {
 ///     let mut a = M::zeros((16, 16));
 ///     let mut b = M::zeros(a.dim());
@@ -81,6 +86,10 @@
 ///     });
 ///
 ///     assert_eq!(a, &b - &c);
+///
+///     borrow_multiply(&mut a, &b, &c);
+///
+///     assert_eq!(a, &b * &c);
 /// }
 /// ```
 macro_rules! azip {


jturner314:
I find the documentation of the `azip!` macro somewhat confusing too. I'd be interested in any suggestions you might have.

When I try to compile your example, I get the following:

```
---- src/zip/zipmacro.rs - azip (line 51) stdout ----
error[E0277]: the trait bound `&&ndarray::ArrayBase<ndarray::OwnedRepr<f32>, ndarray::Dim<[usize; 2]>>: ndarray::NdProducer` is not satisfied
  --> src/zip/zipmacro.rs:61:5
   |
11 |     azip!(mut a (a), ref b, ref c in { *a = b * c });
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `ndarray::NdProducer` is not implemented for `&&ndarray::ArrayBase<ndarray::OwnedRepr<f32>, ndarray::Dim<[usize; 2]>>`
   |
             <ndarray::ArrayBase<ndarray::ViewRepr<&'a mut A>, D> as ndarray::NdProducer>
             <ndarray::ArrayBase<ndarray::ViewRepr<&'a A>, D> as ndarray::NdProducer>
   = note: required because of the requirements on the impl of `ndarray::IntoNdProducer` for `&&ndarray::ArrayBase<ndarray::OwnedRepr<f32>, ndarray::Dim<[usize; 2]>>`

error[E0599]: no method named `and` found for type `ndarray::Zip<(ndarray::ArrayBase<ndarray::ViewRepr<&mut f32>, ndarray::Dim<[usize; 2]>>, &&ndarray::ArrayBase<ndarray::OwnedRepr<f32>, ndarray::Dim<[usize; 2]>>), ndarray::Dim<[usize; 2]>>` in the current scope
  --> src/zip/zipmacro.rs:61:5
   |
11 |     azip!(mut a (a), ref b, ref c in { *a = b * c });
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: the method `and` exists but the following trait bounds were not satisfied:
           `&&ndarray::ArrayBase<ndarray::OwnedRepr<f32>, ndarray::Dim<[usize; 2]>> : ndarray::NdProducer`

thread 'src/zip/zipmacro.rs - azip (line 51)' panicked at 'couldn't compile the test', librustdoc/test.rs:330:13
```

The first error message is the clue. It tells us that "the trait bound `&&ndarray::ArrayBase<ndarray::OwnedRepr<f32>, ndarray::Dim<[usize; 2]>>: ndarray::NdProducer` is not satisfied" and that this trait bound is "required because of the requirements on the impl of `ndarray::IntoNdProducer` for `&&ndarray::ArrayBase<ndarray::OwnedRepr<f32>, ndarray::Dim<[usize; 2]>>`". This is slightly misleading because it jumps straight to the `NdProducer` trait, but the important part here is that `IntoNdProducer` is not implemented for our type. Why is this happening?

Well, the `b` parameter to `borrow_multiply` has type `&Array<f32>`, and the pattern in the `azip` call for `b` is `ref b`. The [docs for `azip`](https://docs.rs/ndarray/0.11.2/ndarray/macro.azip.html) say that for a pattern of the form `ref b`, the producer is `&b` and the variable pattern is `b`. This implies that `azip` will try to use `&b` as the producer, and this has type `&&Array<f32>`, which is equivalent to the `&&ndarray::ArrayBase<ndarray::OwnedRepr<f32>, ndarray::Dim<[usize; 2]>>` type from the error message. However, `IntoNdProducer` is not implemented for `&&Array<f32>`.

How do we avoid this? Well, [`IntoNdProducer` is implemented for `&Array<f32>`](https://docs.rs/ndarray/0.11.2/ndarray/trait.IntoNdProducer.html) (note the single ampersand), which is the type of `b`, so we don't want `azip` to add another `&` by using a pattern of the form `b` or `ref b`. Instead, we want to use `b (expr)` or `ref b (expr)`, with `b` as the `expr`.

So, we should change the `azip` call to

```rust
azip!(mut a (a), b (b), c (c) in { *a = b * c });
```

We can also add `ref` to the patterns because of the `impl<'a, 'b> Mul<&'a f32> for &'b f32` in the standard library:

```rust
azip!(mut a (a), ref b (b), ref c (c) in { *a = b * c });
```

but I prefer the version without the extra `ref`s.

Does this explanation make sense?

Perhaps an easier way to look at this is to determine the expansion. The example you provided expands from

```rust
azip!(mut a (a), ref b, ref c in { *a = b * c });
```

to

```rust
Zip::from(a).and(&b).and(&c).apply(|mut a, b, c| { *a = b * c });
```

while the fixed version expands from

```rust
azip!(mut a (a), b (b), c (c) in { *a = b * c });
```

to

```rust
Zip::from(a).and(b).and(c).apply(|mut a, &b, &c| { *a = b * c });
```

paulkernfeld:
Your explanation makes sense, but I'm not sure if I would have had the patience to work through the entire thing on my own. I think I would probably be able to write `azip` macros better if there were at least one example using each feature (normal, `ref`, `mut`, `x (x)`, and `index`).

paulkernfeld:
Is it possible to formulate an example where `ref` is necessary? I have not been able to figure out how to do so.

I have updated the `x (x)` example as you suggested; is this worth merging now?

bluss:
@paulkernfeld when you want a reference to the element instead of a copy. Copies are the default since we often work with copyable scalars like floats. You might use ref if you have a bigger element in the array, like a matrix or a String.

paulkernfeld:
Ah, got it. Thanks for the explanation!

I would be happy to open an issue for this, but what do you think of removing the `ref` keyword and functionality from `azip`? I'm guessing that putting non-copy types into `ndarray` is uncommon enough that users could just fall back to `Zip::from` in those cases.

src/zip/zipmacro.rs
```diff
@@ -56,6 +56,11 @@
 ///
 /// type M = Array2<f32>;
 ///
+/// // Since this function borrows its inputs, captures must use the x (x) pattern
+/// fn borrow_multiply(a: &mut M, b: &M, c: &M) {
+///     azip!(mut a (a), b (b), c (c) in { *a = b * c });
+/// }
```
bluss: If you have time, move this function to below the main function. I think it can be read last.

bluss:
@paulkernfeld Just another thought. Can you use Zip? How is that in comparison, which interface is better?

bluss:
Merging to apply the final edits

bluss:
Thanks for this!

PR #480, opened 2018-08-28
Change type of end field in slices to Bound

jturner314:
I just noticed the [`Bound`](https://doc.rust-lang.org/std/ops/enum.Bound.html) enum in the standard library while I was reading about the new [`RangeBounds`](https://doc.rust-lang.org/std/ops/trait.RangeBounds.html) trait. I thought I'd propose switching to the `Bound` type, now that we have support for inclusive ranges (#467). There is no memory cost to doing this (since `Bound<isize>` and `Option<isize>` are the same size). This is a breaking change.

I'm pretty ambivalent about this PR, since `Option` has the benefit of being more common than `Bound` and having lots of convenient methods. It's also nice to only have to deal with exclusive end bounds after the `Slice`/`SliceOrIndex` is created. So, I'm curious to see what everyone thinks about this change.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -94,6 +94,7 @@ Recent Changes (ndarray)
   - Add ``map_mut`` and ``map_axis_mut`` methods (mutable variants of ``map`` and ``map_axis``) by @LukeMathWalker.
   - Add support for 128-bit integer scalars (``i128`` and ``u128``).
   - Add support for slicing with inclusive ranges (``start..=end`` and ``..=end``).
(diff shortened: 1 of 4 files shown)


bluss:
Pretty ambivalent / I'd decline the complication since we should do this to get the upsides, not just because we can. So with a plan, it could be good.

jturner314:
Okay, I'll close this. I don't see any significant upsides, so we might as well avoid unnecessary breakage. If we find a specific use-case in the future, we can always make this change then.

PR #484, opened 2018-09-09
Documentation for simple examples for sorting and shuffling

tsoernes:


jturner314:
Thanks for working on this! I think it would be good to document ways to sort and shuffle arrays/views.

It looks like CI failed, so there are a few things to fix to make the tests pass. It looks like the primary issue is missing `extern crate` and `use` declarations.

Regarding the documentation itself -- we should mention that approaches using `.as_slice_mut()` only work if the array/view is contiguous. ([See the docs.](https://docs.rs/ndarray/0.12.0/ndarray/struct.ArrayBase.html#method.as_slice_mut)) This means that, for example, trying to sort a column in a 2-D row-major array doesn't work with this approach.

```rust
extern crate ndarray;

use ndarray::prelude::*;

fn main() {
    let mut a = arr2(&[[1, 2], [3, 4]]);
    a.column_mut(0).as_slice_mut().unwrap().sort_unstable();
}
```

This panics with ``thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', libcore/option.rs:345:21`` since the column is non-contiguous.

We don't yet have a good way to sort/shuffle non-contiguous arrays/views. (Maybe there's a crate that would work for this? Otherwise, we could to add the necessary methods to `ndarray`.) So, if the array/view is non-contiguous, there are basically three options at the moment:

1. Copy the data into a continuous `Vec`/array/view, and then sort/shuffle the copy. This doesn't modify the order of the original data.
2. Sort/shuffle a slice of indices with `.sort_by_key()`/`.shuffle()`, and then rearrange the data according to the sorted/shuffled indices. [`examples/sort-axis.rs`](https://github.com/bluss/ndarray/blob/master/examples/sort-axis.rs) provides an example of this.
3. Add support for sorting strided data to a sorting crate or contribute the necessary methods to `ndarray`. :)

tsoernes:
@jturner314 I added your comment nearly verbatim and fixed the import errors,

src/doc/various_utils/mod.rs
```diff
@@ -0,0 +1,84 @@
...
+//! Two arrays can be shuffled in unison, at the cost of copying both arrays:
+//! ```
+//! extern crate rand;
+//! extern crate ndarray;
+//! use ndarray::*;
+//! use rand::{thread_rng, Rng};
+//! let arr1: Array1<f64> = Array::from_vec(vec![3.0, -1.0, 8.0, 2.0]);
+//! let arr2: Array1<f64> = Array::from_vec(vec![3.0, 2.0, 5.0, 1.0]);
+//!
+//! let mut indecies: Vec<usize> = (0..arr1.len_of(Axis(0))).collect();
```
nilgoyette: We can find the word `indecies` five times in this PR. I can't find any information on that word. Do you mean the plural of 'index'? In that case, it's 'indexes' or 'indices', but you already use 'indices' so you might as well continue.

bluss:
Hi @tsoernes, thanks for this. We unfortunately can't include examples that only work for some arrays, so a solution based on `as_slice_mut` is not enough.

There is an example implementation of sorting an array along an axis inside examples/sort-axis.rs in the repository, and it should work for all arrays.

When we have time, we should work in that example as a feature in the library. It should be solid, but of course it has not received the same level of review the ndarray crate itself and I'm not sure I'm happy with its performance.

PR #496, opened 2018-09-28
Add raw array pointer types

jturner314:
The motivation for this is #388 and another project I'm working on where I'd like to be able to properly deal with non-`Copy` uninitialized data.

This PR introduces two new traits, `DataRaw` and `DataRawMut`, which are less restrictive than `Data` and `DataMut`. It also introduces `ArrayPtr` and `ArrayPtrMut` which are similar to `ArrayView` and `ArrayViewMut` but don't carry lifetime information. I believe all the changes are backwards compatible.

`ArrayPtr` and `ArrayPtrMut` are more restrictive than normal raw pointers in Rust: they require pointers to all elements in the array to be within bounds or one byte past the end of a single allocated object. The reason for this restriction is to ensure that `.offset()` operations on the array's pointer can be performed safely. If we lifted this restriction, then then following methods would have to be unsafe for `ArrayPtr` and `ArrayPtrMut`:

* `.slice_move()`
* `.slice_inplace()`
* `.slice_axis_inplace()`
* `.subview_inplace()`
* `.into_subview()`
* `.invert_axis()`
* `.split_at()`

While we could make these methods safe for normal arrays and `unsafe` for array pointers by implementing them separately for `ArrayBase<S: Data, D: Dimension>`, `ArrayPtr<A, D: Dimension>`, and `ArrayPtrMut<A, D: Dimension>`, doing so requires some duplication and hurts code organization. Additionally, as a user, I'd prefer for most methods on `ArrayPtr`/`ArrayPtrMut` to be safe for convenience, and in all of my realistic use-cases, my `ArrayPtr`/`ArrayPtrMut` instances will point to valid allocations.

@bluss Will you please review this? This PR deals with more unsafe code and potential for undefined behavior than I'm used to.

Edit: I've now added a `DataRawClone` trait to allow cloning `ArrayPtr` and `ArrayPtrMut`. (I couldn't just implement `DataClone` for `PtrRepr<*const A>` and `PtrRepr<*mut A>` because `DataClone: Data`.)


src/impl_methods.rs
```diff
@@ -1048,18 +1119,36 @@ impl<A, S, D> ArrayBase<S, D> where S: Data<Elem=A>, D: Dimension
...
+    pub fn as_array_ptr(&self) -> ArrayPtr<A, D> {
+        unsafe { ArrayPtr::new_(self.ptr, self.dim.clone(), self.strides.clone()) }
+    }
+
+    /// Return a mutable pointer array.
+    #[inline(always)]
+    pub fn as_array_ptr_mut(&mut self) -> ArrayPtrMut<A, D>
+        where S: DataRawMut
+    {
+        self.try_ensure_unique(); // for RcArray
```
bluss: Nice PR.

I've been reading it and wondering if we need both a "const" and "mut" ArrayPtr type, we don't really have a real difference between raw pointer types. I guess this here is the real difference: We need this split if to "unshare" reference counted data before mutating.

The alternative would be to not have automatic unsharing with array ptrs, and it would be a footgun to us and consumers of the library?

src/impl_methods.rs
```diff
@@ -1048,18 +1119,36 @@ impl<A, S, D> ArrayBase<S, D> where S: Data<Elem=A>, D: Dimension
...
-        where S: DataMut
+        where S: DataRawMut
     {
-        self.ensure_unique(); // for RcArray
+        self.try_ensure_unique(); // for RcArray
         self.ptr
     }
 
+    /// Return a pointer array.
+    #[inline(always)]
```
bluss: Can be inline or no attribute, same below

src/impl_methods.rs
jturner314: > I guess this here is the real difference: We need this split if to "unshare" reference counted data before mutating.

Yeah, I think that was the only real difference implementation-wise. There is a semantic advantage, too, to having separate types, and they help reduce mistakes. Rust has separate `*const T` and `*mut T` types, after all.

One other potential difference is [covariance](https://doc.rust-lang.org/nightly/nomicon/subtyping.html) over `A`. (This will come up once we switch `ArrayBase.ptr` to `NonNull`. `ArrayPtr` should be covariant over `A`; I'm not sure about `ArrayPtrMut`.)

> The alternative would be to not have automatic unsharing with array ptrs, and it would be a footgun to us and consumers of the library?

Well, we could have both `as_array_ptr` and `as_array_ptr_mut` return a common `ArrayPtr` type (where `as_array_ptr_mut` would perform the unsharing), but, as you said, it would be easy to forget to use the `_mut` one when necessary.

src/impl_ptrs.rs
```diff
@@ -0,0 +1,172 @@
...
+
+    /// Converts to a read-only view of the array.
+    ///
+    /// **Warning** from a safety standpoint, this is equivalent to
+    /// dereferencing a raw pointer for every element in the array. You must
+    /// ensure that all of the data is valid and choose the correct lifetime.
+    #[inline]
+    pub(crate) unsafe fn deref_into_view<'a>(self) -> ArrayView<'a, A, D> {
+        ArrayView::new_(self.ptr, self.dim, self.strides)
+    }
```
bluss: This into method seems to be the most interesting to provide, so I'd go with it to be pub?

src/impl_ptrs.rs
```diff
@@ -0,0 +1,172 @@
+use dimension::stride_offset;
+use imp_prelude::*;
+use StrideShape;
+
+impl<A, D> ArrayPtr<A, D>
+where
+    D: Dimension,
+{
+    /// Create a new `ArrayPtr`.
```
bluss: What do you think about names `RawArrayView` and `RawArrayViewMut` ?

src/data_traits.rs
```diff
@@ -125,6 +135,12 @@ unsafe impl<A> DataRaw for PtrRepr<*const A> {
     private_impl!{}
 }
 
+unsafe impl<A> DataRawClone for PtrRepr<*const A> {
+    unsafe fn clone_with_ptr(&self, ptr: *mut Self::Elem) -> (Self, *mut Self::Elem) {
+        (*self, ptr)
+    }
+}
```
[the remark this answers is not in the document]
jturner314: In current `ndarray`, [`DataClone`](https://docs.rs/ndarray/0.12.0/ndarray/trait.DataClone.html) has a `Data` bound. If we want to be able to clone `ArrayPtr`/`ArrayPtrMut`, this is too restrictive given  only the [current implementation of `Clone` for `ArrayBase`](https://docs.rs/ndarray/0.12.0/ndarray/struct.ArrayBase.html#impl-Clone). Possible alternatives to the approach in this PR:

1. Don't add a `DataRawClone` trait; instead, implement `Clone` individually for `ArrayPtr` and `ArrayPtrMut`. That would be inconsistent and would prevent code with just a `DataClone` bound from being able to clone `ArrayPtr` and `ArrayPtrMut`.

   With a breaking change, though, we could get rid of `DataClone` entirely and just implement `Clone` individually for `ArcArray`, `Array`, `ArrayView`, `ArrayPtr`, and `ArrayPtrMut`. This actually seems pretty nice, but it does make writing generic code more difficult. (In user code, the bound to allow `array.clone()` would have to be `ArrayBase<S, D>: Clone` instead of `S: DataRawClone`.)

2. Add the `DataRawClone` trait and remove the `DataClone` trait. That would work but would be a breaking change.

Fwiw, `DataClone` does have a small amount of utility as the combination of `Data` and `DataRawClone`, but I agree it does seem like unnecessary complication.

src/impl_ptrs.rs
jturner314: You mean instead of `ArrayPtr` and `ArrayPtrMut`? Those names are longer, but I think I prefer them because they carry less of the `*const T`/`*mut T` connotation. The names `ArrayPtr`/`ArrayPtrMut` could be confusing because you can't create an `ArrayPtr`/`ArrayPtrMut` given an arbitrary `*const T`/`*mut T`.

Another thing to consider is how this relates to #543, in particular the "non-null" constraint. Do we want to require the `ptr` in `ArrayPtr`/`ArrayPtrMut`/`RawArrayView`/`RawArrayViewMut` to be non-null? If yes, `RawArrayView` is nice because the name doesn't carry the connotation that the pointer could be null.

src/impl_ptrs.rs
jturner314: Okay, we can make `deref_into_view` public. Would you remove `deref_view`?

src/impl_methods.rs
jturner314: Fixed.

src/data_traits.rs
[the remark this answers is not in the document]
bluss: (2) I think a breaking change is ok. We have lots of them queued? We should get on to merging them as soon as 0.12.1 is released (and I can do the releasing if you want).

I'd hope it has very little impact on users, but maybe I underestimate how much these representation traits are used?

I'd use the name `DataClone` for the new trait.

src/impl_ptrs.rs
bluss: I agree with your ideas about non-null over in that issue. We don't have a pressing need for using `NonNull<T>`.

I don't think using `NonNull` alone will solve anything about the reborrow methods though. I believe in my explanation in #412, which is not related to it being a const or mut pointer but due to the associated type `S::Data` making `S` fixed.  That said, Rust might have gotten more flexible while I was away..

jturner314: > I agree with your ideas about non-null over in that issue. But we also don't have a pressing need for using `NonNull<T>`.

What's your opinion about requiring non-nullness for the raw pointer types? I lean towards requiring non-null for the time being (but not guarantee that pointers returned from methods like `.as_ptr()` are non-null). We can always relax this constraint later.

> I don't think using `NonNull` alone will solve anything about the reborrow methods though. I believe in my explanation in #412, which is not related to it being a const or mut pointer but due to using the associated type `S::Elem` holds `S` fixed.

I just [tried it](https://play.rust-lang.org/?version=stable&mode=debug&edition=2015&gist=62fb90d1198ed563095f0227cebd1db5), and you're right. `NonNull` doesn't solve the issue, but adding an `A` type parameter to `ArrayBase` does.

bluss: yeah require nonull

src/data_traits.rs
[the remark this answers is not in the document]
jturner314: > I'd use the name `DataClone` for the new trait.

The only thing that I don't like about this is consistency of naming with the other data traits, especially `DataMut` vs. `DataRawMut`. I was naming traits without a `Data` bound `DataRaw*` and traits with a `Data` bound `Data*`.

src/impl_ptrs.rs
jturner314: Okay, I'll add the non-null requirement to the docs (and anything else I might have missed from the constraints added in #543), and I'll rename `ArrayPtr`/`ArrayPtrMut` to `RawArrayView`/`RawArrayViewMut`.

jturner314: Done.

src/impl_ptrs.rs
bluss: Yes I'd consider removing it

src/data_traits.rs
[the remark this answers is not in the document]
bluss: Good point. We still should use only one trait if we don't need 2, we could use the name DataRawClone and let DataClone be a deprecated reexport of it.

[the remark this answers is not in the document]
bluss: DataClone could be a trait alias of DataRawClone + Data? That's a new Rust feature we can transition to.

bluss:
I like the new names. It also puts the Raw types a bit outside of the list of important types with `Array*` names, so we keep in a less visible place in the docs.

src/impl_ptrs.rs
jturner314: Okay, done. The user can use `raw_view.clone().deref_into_view()` if they don't want to give up ownership of the raw view. We can add `.deref_view()` later if it turns out that it would be useful.

jturner314:
Yeah, it's nice they're sorted near the end alphabetically.

src/data_traits.rs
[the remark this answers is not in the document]
jturner314: Woah, I had no idea trait aliases were a thing. They look very convenient.

I've added a commit deprecating `DataClone`. We can switch `DataClone` to a trait alias once they're available on stable.

jturner314:
In this PR, I've created a separate `RawViewRepr` type, but I could just as easily reuse the existing `ViewRepr` type. Any thoughts?

bluss:
Seems to not matter, separate works.

src/impl_methods.rs
```diff
@@ -1126,18 +1193,36 @@ impl<A, S, D> ArrayBase<S, D> where S: Data<Elem=A>, D: Dimension
...
-        self.ensure_unique(); // for RcArray
+        self.try_ensure_unique(); // for RcArray
         self.ptr
     }
 
+    /// Return a raw view of the array.
+    #[inline]
+    pub fn as_raw_view(&self) -> RawArrayView<A, D> {
+        unsafe { RawArrayView::new_(self.ptr, self.dim.clone(), self.strides.clone()) }
+    }
```
bluss: The plain view method doesn't use `as_` in the name, it's just `view`. Can we do the same here? I guess as belongs to a convention but I don't see the benefit of it.

jturner314: Sure, that sounds good to me. Done.

src/data_traits.rs
```diff
@@ -22,92 +23,177 @@ use {
...
+/// An array representation that can be cloned and allows elements to be
+/// accessed with safe code.
 ///
 /// ***Internal trait, see `Data`.***
-pub unsafe trait DataClone : Data {
-    #[doc(hidden)]
-    /// Unsafe because, `ptr` must point inside the current storage.
-    unsafe fn clone_with_ptr(&self, ptr: *mut Self::Elem) -> (Self, *mut Self::Elem);
+#[deprecated(note="use `Data + DataRawClone` instead", since="0.12.1")]
+pub trait DataClone : Data + DataRawClone {}
```
bluss: Hard to decide, but it seems best to me to still keep `DataClone` under its current name, and let the breaking change instead be, that DataClone does not require `Data`. The code using DataClone on github that I've seen (random sample) mostly uses `DataClone + Data<Elem=X>` as bound anyway.

Ideally, we should be here. These should be internal implementation traits. Users should have better ways of being general over our arrays.

jturner314: I'm curious why you lean towards dropping the `Data` bound from `DataClone` instead of adding a `DataRawClone` trait (and keeping `DataClone: Data + DataRawClone` if desired). The name `DataClone` is a little shorter than `DataRawClone`, which is nice for code that needs to specify a `DataClone`/`DataRawClone` bound, but I don't see any other advantages to this approach. If the issue is primarily the length of the name, we could call the new trait `RawClone` instead of `DataRawClone`.

Fwiw, I'd typically write `DataClone<Elem = X>` instead of `DataClone + Data<Elem = X>`. This does come up more often with `DataMut` (example below), but I could conceivably see it happening with `DataClone` too.

```rust
extern crate ndarray;

use ndarray::prelude::*;
use ndarray::{array, Data, DataMut};

fn foo<S1, S2>(a: &ArrayBase<S1, Ix2>, b: &mut ArrayBase<S2, Ix2>)
where
    S1: Data<Elem = f32>,
    S2: DataMut<Elem = f32>, // I like this better than Data<Elem = f32> + DataMut
{
    *b += a;
}

fn main() {
    let a = array![[1., 2.], [3., 4.]];
    let mut b = array![[5., 6.], [7., 8.]];
    foo(&a, &mut b);
    println!("{}", b);
}
```

I'm against dropping the `Data` bound from `DataClone` for two reasons:

1. The biggest issue is inconsistency with the names of other traits: all the other traits named `Data*` have a `Data` bound, while the `DataRaw*` traits don't.
2. If users don't notice the change in the changelog and update their code, it would lead to safety issues for user code relying on `DataClone` implying `Data`. For example, this could happen if existing user code specifies a `DataClone<Elem = A>` bound on an array and dereferences pointers to elements in the array. The consequences of this could be pretty bad, but this coming up in practice does seem unlikely.

We could resolve reason 1 and maintain consistency of naming by dropping the `Data` bound from `DataMut`, `DataOwned`, and `DataShared` too. Then, we'd have something like this:

```rust
/// Array representation trait.
///
/// For an array that meets the invariants of the `ArrayBase` type. This trait
/// does not imply any ownership or lifetime; pointers to elements in the array
/// may not be safe to dereference.
///
/// ***Note:*** `RawData` is not an extension interface at this point.
/// Traits in Rust can serve many different roles. This trait is public because
/// it is used as a bound on public methods.
pub unsafe trait RawData : Sized {
    /// The array element type.
    type Elem;

    #[doc(hidden)]
    // This method is only used for debugging
    fn _data_slice(&self) -> Option<&[Self::Elem]>;

    private_decl!{}
}

/// Array representation trait.
///
/// For an array with writable elements.
///
/// ***Internal trait, see `RawData`.***
pub unsafe trait DataMut : RawData {
    /// If possible, ensures that the array has unique access to its data.
    ///
    /// If `Self` provides safe mutable access to array elements, then it
    /// **must** panic or ensure that the data is unique.
    #[doc(hidden)]
    fn try_ensure_unique<D>(&mut ArrayBase<Self, D>)
    where Self: Sized,
          D: Dimension;

    /// If possible, returns whether the array has unique access to its data.
    ///
    /// If `Self` provides safe mutable access to array elements, then it
    /// **must** return `Some(_)`.
    #[doc(hidden)]
    fn try_is_unique(&mut self) -> Option<bool>;
}

/// Array representation trait.
///
/// A representation that is a unique or shared owner of its data.
///
/// ***Internal trait, see `Data`.***
pub unsafe trait DataOwned : RawData {}

/// Array representation trait.
///
/// A representation that is a lightweight view.
///
/// ***Internal trait, see `Data`.***
pub unsafe trait DataShared : Clone + RawData + DataClone { }

/// Array representation trait.
///
/// An array representation that can be cloned.
///
/// ***Internal trait, see `RawData`.***
pub unsafe trait DataClone : RawData {
    #[doc(hidden)]
    /// Unsafe because, `ptr` must point inside the current storage.
    unsafe fn clone_with_ptr(&self, ptr: *mut Self::Elem) -> (Self, *mut Self::Elem);

    #[doc(hidden)]
    unsafe fn clone_from_with_ptr(&mut self, other: &Self, ptr: *mut Self::Elem) -> *mut Self::Elem {
        let (data, ptr) = other.clone_with_ptr(ptr);
        *self = data;
        ptr
    }
}

/// Array representation trait.
///
/// For an array with elements that can be accessed with safe code.
///
/// ***Internal trait, see `RawData`.***
//
// # For implementers
//
// If you implement both the `Data` and `DataMut` traits, you are guaranteeing
// that the `DataMut::try_ensure_unique` implementation always panics or
// ensures that the data is unique. You are also guaranteeing that
// `try_is_unique` always returns `Some(_)`.
pub unsafe trait Data : RawData {
    #[doc(hidden)]
    fn new(elements: Vec<Self::Elem>) -> Self
    where
        Self: DataOwned;

    /// Converts the data representation to a shared (copy on write)
    /// representation, without any copying.
    #[doc(hidden)]
    fn into_shared(self) -> OwnedRcRepr<Self::Elem>
    where
        Self: DataOwned;

    /// Converts the array to a uniquely owned array, cloning elements if necessary.
    #[doc(hidden)]
    fn into_owned<D>(self_: ArrayBase<Self, D>) -> ArrayBase<OwnedRepr<Self::Elem>, D>
    where
        Self::Elem: Clone,
        D: Dimension;

    /// Ensures that the array has unique access to its data.
    #[doc(hidden)]
    #[inline]
    fn ensure_unique<D>(self_: &mut ArrayBase<Self, D>)
    where
        Self: DataMut + Sized,
        D: Dimension
    {
        Self::try_ensure_unique(self_)
    }

    /// Returns whether the array has unique access to its data.
    #[doc(hidden)]
    #[inline]
    fn is_unique(&mut self) -> bool
    where
        Self: DataMut,
    {
        self.try_is_unique().unwrap()
    }
}
```

There are two disadvantages to this: First, the safety issue (reason 2 above) is more likely to occur with more important traits like `DataMut` missing the `Data` bound. Second, this requires explicitly writing a `Data` bound in more cases (e.g. `DataMut + Data<Elem = A>` instead of `DataMut<Elem = A>`) which isn't as convenient.

> These should be internal implementation traits. Users should have better ways of being general over our arrays.

Treating them as internal would be fine, but leaving them public doesn't seem too bad. If we treat them as internal, we'd need to add `ArrayMut`, `ArrayOwned`, etc., traits implemented on `ArrayBase` to allow users to operate in a generic way over array types.

bluss: ok, you're right :)

bluss:
After reading about the stacked borrows model and trying to catch up with what the unsafe guidelines people have been working on, this seems even more important. For example down the line, let's have an NdProducer that produces raw pointer elements, that should allow us (I think?) to get around problems like accessing uninitialized elements through a mutable reference, if that becomes an issue.

I hope that unsafe code guidelines will have some room for pragmatism and adapting to existing unsafe code practices.

jturner314:
I'd like to rename `DataRaw`/`DataRawMut`/`DataRawClone` to `RawData`/`RawDataMut`/`RawDataClone`. (I like these names better, and this makes them more consistent with `RawViewRepr`/`RawView`/`RawViewMut`.) Any objection to renaming those traits?

> After reading about the stacked borrows model and trying to catch up with what the unsafe guidelines people have been working on, this seems even more important. For example down the line, let's have an NdProducer that produces raw pointer elements, that should allow us (I think?) to get around problems like accessing uninitialized elements through a mutable reference, if that becomes an issue.

Yeah, we should be able to refactor a lot of functionality in terms of `RawView`/`RawViewMut`, especially iterators/producers, and it would be useful to provide iterators/producers over `RawView`/`RawViewMut` for use in unsafe code.

For uninitialized data specifically, I'm looking forward to [RFC 1892](https://github.com/rust-lang/rfcs/blob/master/text/1892-uninitialized-uninhabited.md), but for the time being, raw views provide a safe way to work with uninitialized data. We could even have a "raw" *owned* array type for use with uninitialized data until RFC 1892 is implemented, but I'd prefer to just wait for RFC 1892.

bluss:
New names sound good!

bluss:
:100:

PR #502, opened 2018-10-11
Refactor and improve iterators

jturner314:
This PR does the following:

* Removes lots of iterator constructor functions and replaces them with `new` methods on the iterators. This is more conventional, easier to understand, and requires fewer imports.

* Renames `OuterIterCore` to `AxisIterCore` to better reflect its purpose.

* Refactors `.split_at()` for `AxisIter` and `AxisIterMut` to eliminate the `axis_iter_split_at_impl!` macro.

* Removes the lifetime from `Baseiter` and implements iterator traits for `Baseiter`. This is mostly a conceptual simplification. It will also make it more straightforward to extend the iterators to `ArrayPtr`/`ArrayPtrMut` in #496.

* Improve docs and implement `Clone` for `LanesIter`.

### DIFF
--- a/src/impl_methods.rs
+++ b/src/impl_methods.rs
@@ -16,7 +16,6 @@ use imp_prelude::*;
 
 use arraytraits;
 use dimension;
-use iterators;
 use error::{self, ShapeError, ErrorKind};
 use dimension::IntoDimension;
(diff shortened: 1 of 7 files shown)


jturner314:
@LukeMathWalker Thanks for taking a look!

Issue #504, opened 2018-10-19
scalar_prod implementation


sebasv:
I am missing a scalar_prod function. on ArrayBase. This seems straightforward to implement (almost blindly copy/paste scalar_sum), so I will be submitting a pull request. I am not familiar enough with computer engineering to say if an unrolled_prod is a valuable addition as well, so I will just include it in the PR awaiting feedback from someone with more background!

jturner314:
> I am not familiar enough with computer engineering to say if an unrolled_prod is a valuable addition as well

If you're interested, [this is a pretty good article for background](https://docs.microsoft.com/en-us/cpp/build/reference/floating-point-optimization), even though it's talking about C++ with the MS compiler instead of Rust with LLVM. Basically, in order to allow the compiler to perform the sum/product using SIMD even with strict floating-point behavior, it's helpful to explicitly unroll and reorder the sum/product manually.

sebasv:
That seems like a very nice read, thank you!

jturner314:
Implemented in #505.

PR #505, opened 2018-10-19
Implement scalar_prod

sebasv:
Implements #504 .

### DIFF
--- a/tests/oper.rs
+++ b/tests/oper.rs
@@ -271,6 +271,26 @@ fn fold_and_sum() {
     }
 }
 
+#[test]
+fn scalar_prod() {
+    let a = Array::linspace(0.5, 2., 128).into_shape((8, 16)).unwrap();
+    assert_approx_eq(a.fold(1., |acc, &x| acc * x), a.scalar_prod(), 1e-5);
+
+    // test different strides
+    let max = 8 as Ixs;
+    for i in 1..max {
+        for j in 1..max {
+            let a1 = a.slice(s![..;i, ..;j]);
+            let mut prod = 1.;
+            for elt in a1.iter() {
+                prod *= *elt;
+            }
+            assert_approx_eq(a1.fold(1., |acc, &x| acc * x), prod, 1e-5);
+            assert_approx_eq(prod, a1.scalar_prod(), 1e-5);
+        }
+    }
+}
+
 fn range_mat(m: Ix, n: Ix) -> Array2<f32> {
     Array::linspace(0., (m * n) as f32 - 1., m * n).into_shape((m, n)).unwrap()
 }
--- a/src/numeric/impl_numeric.rs
+++ b/src/numeric/impl_numeric.rs
@@ -6,7 +6,7 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
-use std::ops::{Add, Div};
+use std::ops::{Add, Div, Mul};
 use libnum::{self, One, Zero, Float};
 use itertools::free::enumerate;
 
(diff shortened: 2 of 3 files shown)


tests/oper.rs
```diff
@@ -271,6 +271,45 @@ fn fold_and_sum() {
...
+                prod *= *elt;
+            }
+            assert_approx_eq(a1.fold(1., |acc, &x| acc * x), prod, 1e-5);
+            assert_approx_eq(prod, a1.scalar_prod(), 1e-5);
+        }
+    }
+
+    // skip a few elements
+    let max = 8 as Ixs;
+    for i in 1..max {
```
jturner314: This block of the test isn't necessary. (It's just testing the `fold` method, not `scalar_prod`, and `fold` is already tested in `fold_and_sum`.)

tests/oper.rs
```diff
@@ -271,6 +271,45 @@ fn fold_and_sum() {
     }
 }
 
+#[test]
+fn fold_and_prod() {
```
jturner314: This can be renamed to just `prod` or `scalar_prod` after removing the block I pointed out.

tests/oper.rs
```diff
@@ -271,6 +271,45 @@ fn fold_and_sum() {
...
+    let a = Array::linspace(0., 127., 128).into_shape((8, 16)).unwrap();
```
jturner314: The array of values shouldn't include zero. (The problem is that a product involving zero is always zero; it isn't a great test for arbitrary values. You could have a separate test involving zero, though.) After fixing this, you might also have to make the numbers smaller to keep the product from becoming too large.

tests/oper.rs
sebasv: Aha! I was already wondering what the point of that test was in `fold_and_sum`. Deleted

tests/oper.rs
sebasv: Good catch. After changing this I ran into another issue: multiplying 128 not-close-to-1 values results in a huge number that suffers from severe precision loss, so the order of multiplication becomes relevant, and the `fold` no longer yields something similar to `unrolled_prod`. I changed the testcase to `Array::linspace(0.5,2.0,128)` which seems a more appropriate use for `scalar_prod` anyway.

jturner314:
Will you please squash the commits into one? I don't mind squashing them myself, but then GitHub won't consider the PR merged.

(Ordinarily, I would just use GitHub's "Squash and merge", but that option is disabled for this repo, I don't have the permissions to enable it, and I haven't heard from @bluss in a while.)

sebasv:
I'll squash the commits. I also put the unrolled code in a macro, is this desired or do you want to stick with separate unrolled code for prod/sum and possible future cases?
```rust
// eightfold unrolled so that floating point can be vectorized
// (even with strict floating point accuracy semantics)
macro_rules! unrolled_fold {
    ($xs:expr, $unity:expr, $operation:expr) => {{
        let mut collected = $unity();
        let (mut p0, mut p1, mut p2, mut p3,
            mut p4, mut p5, mut p6, mut p7) =
            ($unity(), $unity(), $unity(), $unity(),
            $unity(), $unity(), $unity(), $unity());
        while $xs.len() >= 8 {
            p0 = $operation(p0, $xs[0].clone());
            p1 = $operation(p1, $xs[1].clone());
            p2 = $operation(p2, $xs[2].clone());
            p3 = $operation(p3, $xs[3].clone());
            p4 = $operation(p4, $xs[4].clone());
            p5 = $operation(p5, $xs[5].clone());
            p6 = $operation(p6, $xs[6].clone());
            p7 = $operation(p7, $xs[7].clone());

            $xs = &$xs[8..];
        }
        collected = $operation(collected.clone(), $operation(p0, p4));
        collected = $operation(collected.clone(), $operation(p1, p5));
        collected = $operation(collected.clone(), $operation(p2, p6));
        collected = $operation(collected.clone(), $operation(p3, p7));

        for i in 0..$xs.len() {
            if i >= 7 { break; }
            collected = $operation(collected.clone(), $xs[i].clone());
        }
        collected
    }}
}

/// Compute the sum of the values in `xs`
pub fn unrolled_sum<A>(mut xs: &[A]) -> A
    where A: Clone + Add<Output=A> + libnum::Zero,
{
unrolled_fold!(xs, A::zero, A::add)
}

/// Compute the product of the values in `xs`
pub fn unrolled_prod<A>(mut xs: &[A]) -> A
    where A: Clone + Mul<Output=A> + libnum::One,
{
    unrolled_fold!(xs, A::one, A::mul)
}

```

jturner314:
Sure, a macro would be nice. By the way, I just noticed that the temporary variable in `scalar_prod` is named `sum` when it would be better named `prod`.

jturner314:
Fwiw, I prefer using generic functions over macros when possible. For example:

```rust
where
    A: Clone,
    I: Fn() -> A,
    F: Fn(A, A) -> A,
{
    // eightfold unrolled so that floating point can be vectorized
    // (even with strict floating point accuracy semantics)
    let mut acc = init();
    let (mut p0, mut p1, mut p2, mut p3,
         mut p4, mut p5, mut p6, mut p7) =
        (init(), init(), init(), init(),
         init(), init(), init(), init());
    while xs.len() >= 8 {
        p0 = f(p0, xs[0].clone());
        p1 = f(p1, xs[1].clone());
        p2 = f(p2, xs[2].clone());
        p3 = f(p3, xs[3].clone());
        p4 = f(p4, xs[4].clone());
        p5 = f(p5, xs[5].clone());
        p6 = f(p6, xs[6].clone());
        p7 = f(p7, xs[7].clone());

        xs = &xs[8..];
    }
    acc = f(acc.clone(), f(p0, p4));
    acc = f(acc.clone(), f(p1, p5));
    acc = f(acc.clone(), f(p2, p6));
    acc = f(acc.clone(), f(p3, p7));

    for i in 0..xs.len() {
        if i >= 7 { break; }
        acc = f(acc.clone(), xs[i].clone())
    }
    acc
}
```

This can be called like this for a sum:

```rust
numeric_util::unrolled_fold(slc, A::zero, A::add)
```

or like this for a product:

```rust
numeric_util::unrolled_fold(slc, A::one, A::mul)
```

This generates basically the same code as the non-generic version (tested with [Compiler Explorer](https://rust.godbolt.org/) with `-C target-cpu=native -C opt-level=3`).

sebasv:
Ready for review. I agree that this does not call for a macro, unless `unrolled_dot` is to be included as well, but I really don't expect a lot more variants to show up that need to be unrolled.

jturner314:
Thanks for contributing this!

sebasv:
Thank you for the guidance! I am learning a ton more about safety and optimization.

PR #508, opened 2018-10-24
Add AsRef<[usize]> and AsMut<[usize]> bounds to Dimension trait

jturner314:
This provides a simple, public way to convert a dimension to a slice. (Before, the only way visible in the docs to do this was `dim.as_array_view().as_slice()` or `dim.as_array_view_mut().as_slice_mut()`.)

### DIFF
--- a/src/dimension/dim.rs
+++ b/src/dimension/dim.rs
@@ -62,6 +62,26 @@ pub fn Dim<T>(index: T) -> T::Dim
     index.into_dimension()
 }
 
+impl<I: ?Sized> AsRef<[usize]> for Dim<I>
+where
+    I: AsRef<[usize]>,
+{
+    #[inline]
+    fn as_ref(&self) -> &[usize] {
+        self.index.as_ref()
+    }
+}
+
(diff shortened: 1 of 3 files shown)


bluss:
Makes me uneasy. Should it be public that the dimensions are all convertible to slices?

jturner314:
> Should it be public that the dimensions are all convertible to slices?

I thought about this and couldn't come up with a good justification why not. I don't anticipate this ever *not* being true in the future, and it's useful today for users to be able to convert to slices.

Addionally, we currently provide `.as_array_view()` and `.as_array_view_mut()`, which already reveal that the dimension types are represented linearly in memory (possibly with a stride other than 1). I don't see any use for non-1 strides in the dimension representation, and even if there is a use in the future, dimension representations are small enough that converting to an `[usize; N]`/`IxDynImpl` will be cheap anyway.

bluss:
I don't think we should do this, let's look at reworking Dimension first.

jturner314:
Okay.

PR #509, opened 2018-10-24
Rename zero_index_with_ndim -> zeros, make it public, and remove zero_index

jturner314:


bluss:
This method doesn't have a nice API. But I realize that there is a need outside of ndarray to use the same ugly APIs we use inside.

jturner314:
Yeah, it's not pretty that it panics when the `ndim` argument doesn't match the type. I don't see a good way around this, though, for code that wants to operate on dimensions generically. For code that knows it's working with a particular dimension type, the `Ix0`/`Ix1`/`Ix2`/.../`IxDyn` functions are available.

Issue #506, opened 2018-10-23
Cumulative map


sebasv:
Like `numpy.cumsum()` or `numpy.cumprod()` but with `map`, `mapv`, `map_inplace`, `map_into` since unlike Numpy, in Rust we don't take a performance hit by supplying user functions. I imagine the signature to be something like 
```rust
fn cum_mapv<F: Fn(T,T) -> T>(&self<T>, f: F) -> Self// T being the element type

// cumsum example:
let a = Array::linspace(0., 100., 100).cummapv(|acc, cur| acc + cur);
```

If this will be considered for inclusion, I can write up a PR!

jturner314:
I assume you're interested in summing over only one axis? If so, this seems reasonable:

```rust
fn accumulate_axis<F>(&self, axis: Axis, f: F) -> Array<A, D>
where
    A: Clone,
    F: FnMut(&A, &A) -> A,
```

This is similar to Python's [`itertools.accumulate`](https://docs.python.org/3/library/itertools.html#itertools.accumulate). The changes from your proposal are adding an `axis` parameter, adding the `A: Clone` bound, changing the return type to `Array`, and changing `F` to take elements by reference to avoid unnecessary clones. It could be used basically the same way:

```rust
arr.accumulate_axis(Axis(1), ::std::ops::Add::add);
// or
arr.accumulate_axis(Axis(1), |acc, x| acc + x);
```

With this method, the first element of the result along each lane is always the clone of the first element of the input along that lane. I'm not quite sure what to call this method, but `accumulate_axis` seems pretty good.

Another related method would be this:

```rust
fn scan_axis<St, B, F>(&self, axis: Axis, initial_state: St, f: F) -> Array<B, D>
where
    F: FnMut(&mut St, &A) -> B,
```

This emulates [`std::iter::Iterator::scan`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.scan). For a cumulative sum, it could be used like this:

```rust
arr.scan_axis(Axis(1), 0., |acc, &x| {
    *acc += x;
    *acc
});
```

It isn't quite as concise for computing a cumulative sum/product, but it is more flexible because the first element can be different between the input and result, and the element type can change.

I think both methods are useful and would be good additions to `ndarray`.

[...]

PR #513, opened 2018-10-24
Accumulate methods

sebasv:
Implements #506 .

### DIFF
--- a/src/impl_methods.rs
+++ b/src/impl_methods.rs
@@ -1877,4 +1877,136 @@ impl<A, S, D> ArrayBase<S, D> where S: Data<Elem=A>, D: Dimension
             }
         })
     }
+
+    /// Traverse an axis, and 'cumulatively fold' over self, i.e.
+    /// return an array A where the element at index i of `axis`,
+    /// and index [..] of other axes, is the result of 
+    /// A.axis_iter(axis)
+    ///  .skip(1)
+    ///  .take(i-1)
+    ///  .fold(A.subview(axis, 0)[..], |acc, subview| f(acc, subview[..])).
+    /// 
+    /// **panics** if the dimension of `self` along `axis` is 0.
+    /// 
+    ///  # Example
+    /// ```
+    /// use ndarray::{arr1, arr2, Axis};
+    /// use std::ops::Add;
+    ///
+    /// let a = arr2(&[[1., 2.],
+    ///                [3., 4.],
+    ///                [5., 6.]]);
+    /// let accumulated = a.accumulate_axis(Axis(0), |a, b| a+b);
+    /// assert!(accumulated
+    ///     .all_close(&arr2(&[[1., 2.],
+    ///                      [4., 6.],
+    ///                      [9., 12.]]), 1e-12));
+    /// 
+    /// let b = arr1(&[1., 2., 3., 4.]);
+    /// let b_accumulated = b.accumulate_axis(Axis(0), |a, b| a*b);
+    /// assert!(b_accumulated.all_close(&arr1(&[1., 2., 6., 24.]), 1e-12));
+    /// ```
+    pub fn accumulate_axis<F>(&self, axis: Axis, mut f: F) -> Array<A, D>
+    where
+        A: Clone,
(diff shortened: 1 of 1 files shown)


src/impl_methods.rs
```diff
@@ -1877,4 +1877,136 @@ impl<A, S, D> ArrayBase<S, D> where S: Data<Elem=A>, D: Dimension
...
+        A: Clone,
+        D: ::dimension::RemoveAxis, 
+        F: FnMut(&A, &A) -> A, 
+    {
+        let mut accum = unsafe{ 
+            let mut v = Vec::with_capacity(self.len());
+            v.set_len(self.len());
+            Array::<A, D>::from_shape_vec_unchecked(self.dim(), v)
+        };
+        let mut states = self.subview(axis, 0).to_owned();
```
jturner314: This results in undefined behavior if `A` implements `Drop` and the length of axis `axis` is zero. (If the length of the axis is zero, this line will panic, which will cause the uninitialized array `accum` to be dropped, which in turn will cause each of the uninitialized elements to be dropped (which is undefined behavior).)

src/impl_methods.rs
```diff
@@ -1877,4 +1877,136 @@ impl<A, S, D> ArrayBase<S, D> where S: Data<Elem=A>, D: Dimension
...
+            let mut v = Vec::with_capacity(self.len());
+            v.set_len(self.len());
+            Array::<B, D>::from_shape_vec_unchecked(self.dim(), v)
+        };
+        let mut states = Array::<St, _>::from_elem(self.dim.remove_axis(axis), initial_state);
+        for (mut accum_i, self_i) in accum.axis_iter_mut(axis).zip(self.axis_iter(axis)) {
+            Zip::from(&mut accum_i)
+                .and(&mut states)
+                .and(&self_i)
+                .apply(|ac, st, se| *ac = f(st, se));
```
jturner314: This results in undefined behavior if `B` implements `Drop`, because `ac` is being assigned without using `ptr::write`, so its old (uninitialized value) gets dropped. We also have to worry about `f` panicking, because a panic will cause `accum` to be dropped, which will cause each of its elements to be dropped, including uninitialized elements.

src/impl_methods.rs
```diff
@@ -1877,4 +1877,136 @@ impl<A, S, D> ArrayBase<S, D> where S: Data<Elem=A>, D: Dimension
...
+                .and(&self_i)
+                .apply(|ac, st, se| {*st = f(st, se); *ac = st.clone();});
+        }
+        accum
+    }
+
+    /// Inplace version of `accumulate_axis`. See that method for more 
+    /// documentation.
+    /// 
+    /// **panics** if the dimension of `self` along `axis` is 0.
```
jturner314: It's not necessary to panic in this case. We can add an explicit check for this case and return an empty array of the correct shape.

src/impl_methods.rs
```diff
@@ -1877,4 +1877,136 @@ impl<A, S, D> ArrayBase<S, D> where S: Data<Elem=A>, D: Dimension
...
+            Array::<A, D>::from_shape_vec_unchecked(self.dim(), v)
+        };
+        let mut states = self.subview(axis, 0).to_owned();
+        accum.subview_mut(axis, 0).assign(&states);
+        for (mut accum_i, self_i) in accum.axis_iter_mut(axis).skip(1)
+        .zip(self.axis_iter(axis).skip(1)) {
+            Zip::from(&mut accum_i)
+                .and(&mut states)
+                .and(&self_i)
+                .apply(|ac, st, se| {*st = f(st, se); *ac = st.clone();});
```
jturner314: This results in undefined behavior if `A` implements `Drop`, because `ac` is being assigned without using `ptr::write`, so its old (uninitialized value) gets dropped. We also have to worry about `f` panicking, because a panic will cause `accum` to be dropped, which will cause each of its elements to be dropped, including uninitialized elements.

bluss:
Thanks for this, I'll close the PR for now, but feel free to reopen if thre are new developments! We're working on getting better fundamental building blocks in the crate to deal with uninit memory and other tools needed to implement things like this.

Issue #512, opened 2018-10-24
Implement `scalar_min` and `scalar_max` for `A: Ord`


sebasv:
Suggested by @jturner314:
I'd just do it in terms of `fold` with something like this (taking advantage of the `first` method from PR #507):

```rust
impl<A, S, D> ArrayBase<S, D>
where
    S: Data<Elem = A>,
    D: Dimension,
{
    /// Returns the minimum element, or `None` if the array is empty.
    fn scalar_min(&self) -> Option<&A>
    where
        A: Ord,
    {
        let first = self.first()?;
        Some(self.fold(first, |acc, x| acc.min(x)))
    }
}
```

We don't need to manually unroll this because the compiler does a good job automatically (checked with [Compiler Explorer](https://rust.godbolt.org/) using the `-O` compiler option).

The desired behavior for floating-point types depends on the use-case because of NaN. One option is

```rust
arr.fold(::std::f64::NAN, |acc, &x| acc.min(x))
```

which ignores NaN values. (It returns NaN only if there are no non-NaN values.) The compiler does a decent job automatically unrolling this, so we don't need to manually unroll in this case either.

_Originally posted by @jturner314 in https://github.com/bluss/ndarray/pull/505#issuecomment-432467813_

sebasv:
I haven't found a neat way yet to implement this both for `A: Ord` and `f64`, ideas are welcome!

sebasv:
Suggested api:
```rust
/// return a minimum element from `self`, ignoring unorderable elements 
/// (such as `f64::NAN`)
/// **Panics** if `self` is empty
pub fn scalar_min(&self) -> &A where A: PartialOrd;

/// return a minimum element from `self`, or `None` if `self` is empty or 
/// contains unorderable elements.
pub fn scalar_partial_min(&self) -> Option<&A> where A: PartialOrd;
```

This does not leverage the additional strictness of `A: Ord` but offers more flexibility and targets more practical use cases in my opinion.

LukeMathWalker:
It would be nice to make the API naming convention consistent with what we are working on here! - https://github.com/jturner314/ndarray-stats/issues/1

sebasv:
I like the convention `(nan)min` but these names overlook the fact that `ndarray` can be used with self-defined types (or anything non-float for that matter). Although I don't see myself ever use `ndarray` for anything else than floats, I can't speak for others. What fits best with `ndarray`s philosophy?

jturner314:
[...]

PR #514, opened 2018-10-24
implement partialord min/max

sebasv:
Implements #512 , but for `PartialOrd`. Unorderable cases such as `f64::NAN` result in an early return of `None`. Ideally there is also a variant that ignores unorderable cases, something like `scalar_max_ignore_unorderable()`, but with a more sensible name.

### DIFF
--- a/src/numeric/impl_numeric.rs
+++ b/src/numeric/impl_numeric.rs
@@ -85,6 +85,34 @@ impl<A, S, D> ArrayBase<S, D>
         }
     }  
 
+    pub fn nanmax(&self) -> Option<&A> 
+    where A: PartialOrd,
+    {
+        if self.len() == 0 {
+            None
+        }
+        else if let Some(slc) = self.as_slice_memory_order() {
+            let f = |acc, x| if acc == acc && x < acc {acc} else {x};
+            // numeric_util::unrolled_fold(slc, )
+            let mut max = self.first();
+            for item in slc.iter().skip(1) {
+                match max.partial_cmp(&Some(item)) {
+                    None => return None,
+                    Some(::std::cmp::Ordering::Less) => max = Some(item),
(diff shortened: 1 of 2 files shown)


sebasv:
Forgot to mention! This depends on #507 . Travis should pass when that PR makes it into master.

src/numeric/impl_numeric.rs
```diff
@@ -46,6 +46,84 @@ impl<A, S, D> ArrayBase<S, D>
...
+    /// 
+    /// let a = arr2(&[[1., 2.], [3., 4.]]);
+    /// 
+    /// assert_eq!(a.scalar_max(), Some(&4.));
+    /// 
+    /// let b = arr2(&[[1., f64::NAN], [3., 4.]]);
+    /// 
+    /// assert_eq!(b.scalar_max(), None);
+    /// ```
+    pub fn scalar_max(&self) -> Option<&A>
```
bluss: Any reason this is not named max?

bluss: Last I knew the plan was to rename scalar_sum to sum, so don't want it to be followed. That said, last I knew was some time ago. Commit b75d438daea2d0c5ce684fed1a5a1ebde7d4e166

sebasv: Excellent, renamed!

bluss:
Needs discussion what to do with the overlap with ndarray-stats https://docs.rs/ndarray-stats/

If this is in ndarray, I'd prefer something that's more geared towards the needs of floating point elements

thecooldrop:
Hey @bluss what do you mean, when you say that it should be more geared toward floating point elements?

sebasv:
Closing as I don't see myself picking this up anytime soon. Feel free to take over

PR #515, opened 2018-10-26
 Adjust constraints on ddof for var_axis and std_axis

jturner314:
Requiring `ddof` to not be < 0 allows us to ignore weird edge cases. Allowing `ddof` to be equal to the length of the axis allows computation of the population variance (`ddof` = 0) for a zero-length axis and computation of the sample variance (`ddof` = 1) for an axis of length one. (The result in these cases is an array of NaNs.)

Note that this is a breaking change because of the new constraint that `ddof` must not be less than zero.

### DIFF
--- a/src/numeric/impl_numeric.rs
+++ b/src/numeric/impl_numeric.rs
@@ -164,7 +164,7 @@ impl<A, S, D> ArrayBase<S, D>
     /// ```
     ///
     /// **Panics** if `ddof` is greater than or equal to the length of the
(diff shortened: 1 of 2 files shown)


LukeMathWalker:
Given that we are doing a breaking change, it might be worth it to swap the operation order in the function body: computing (and eventually panicking) on `ddof` before doing the actual computation.
This could be achieved by asking `A` to implement `FromPrimitive`.

jturner314:
That's a good idea. I've added a new commit using `FromPrimitive`.

PR #517, opened 2018-10-26
Bump to Rust 1.30, add macros to preludes, and switch macro_use to use

jturner314:
This PR adds the `array`, `azip`, `s`, and `par_azip` macros to their respective crates' preludes. Where possible, it also replaces `macro_use` with normal `use` statements.

This requires Rust 1.30.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ sudo: required
 dist: trusty
 matrix:
   include:
-    - rust: 1.27.0
+    - rust: 1.30.0
       env:
        - FEATURES='test docs'
     - rust: stable
--- a/benches/bench1.rs
+++ b/benches/bench1.rs
@@ -2,7 +2,6 @@
 #![allow(unused_imports)]
 
 extern crate test;
-#[macro_use(s, azip)]
 extern crate ndarray;
 
 use ndarray::{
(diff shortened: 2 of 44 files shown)


bluss:
Oh it's beautiful. And those things now share the same namespace (or maybe equally named things are punned and both `use`d, just like for newtype constructors).

jturner314:
Yeah, this is so much cleaner than the old `#[macro_use]` annotations.

Macros have a separate namespace, and all equally named things are imported when `use`d. (For example, `use ndarray::stack;` imports both the function and the macro.)

bluss:
The short macro name of `s` might now be yet more risky, when it comes through the prelude. I think it will work in practice though.

jturner314:
I just added another commit that makes it possible to use `s!` and `azip!` without importing them. (You can now write `ndarray::s![1..2, ..]`, for example.)

> The short macro name of `s` might now be yet more risky, when it comes through the prelude. I think it will work in practice though.

Yeah, I think it'll be fine because (1) macros are in a separate namespace from variables/functions, (2) a short name like `s` seems very unusual for a macro, and (3) macros are evaluated at compile time, so potential issues will be caught at compile time.

bluss:
Towards the future, breaking change season is here

PR #520, opened 2018-10-27
Rename scalar_prod and scalar_sum to prod and sum

jturner314:
The `scalar_prod` method hasn't been released yet, so we can freely rename it. The old `scalar_sum` method now has a `deprecated` annotation, which causes a warning when building code that uses it and is visible in the docs.

See also [this comment from @bluss](https://github.com/bluss/ndarray/pull/514#discussion_r228710415).


jturner314:
I think that `product` is a better name than `prod`, because it's clear what it means and the `Iterator` trait in the standard library uses the name [`product`](https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html#method.product).

bluss:
product is good. Good to avoid scalar_prod, that's just a confusion point vs the dot product. :slightly_smiling_face:

src/numeric/impl_numeric.rs
```diff
@@ -46,16 +46,24 @@ impl<A, S, D> ArrayBase<S, D>
         sum
     }
 
+    /// Return the sum of all elements in the array.
+    #[deprecated(note="please use `sum` instead", since="0.12.1")]
```
bluss: Micro comment but can this say it was renamed to `sum`. It's clear it's the same method, just new name.

jturner314: Good point. I've changed it to "renamed to `sum`".

Issue #524, opened 2018-11-01
Writing raw bytes &[u8] from ndarray to child stdin


jimbo1qaz:
I'm new to Rust and trying to port a program I wrote in Python, which uses Matplotlib (too slow) to generate `bytes` objects, written into a subprocess pipe.

Numpy ndarray supports conversion to bytes (https://docs.scipy.org/doc/numpy/reference/generated/numpy.ndarray.tobytes.html ).

I'm thinking of writing a quick-and-dirty line-drawing library (also too slow in Python), which loops over Rust `Array::<u8,_>`. (https://github.com/coder543/dataplotlib seems inactive). However I don't know how to pull raw bytes out of a ndarray to write to a ChildStdin.

AFAIK https://docs.rs/ndarray/0.12.0/ndarray/struct.ArrayBase.html doesn't mention `bytes` or `u8`.

https://doc.rust-lang.org/std/process/struct.ChildStdin.html and https://doc.rust-lang.org/std/io/trait.Write.html apparently only take `&[u8]`.

https://github.com/rust-ndarray/ndarray/pull/345/files ???

should I just research Nim/etc. instead?

jimbo1qaz:
Found a solution: https://docs.rs/ndarray/0.10.7/ndarray/struct.ArrayBase.html#method.as_slice etc.

```rust
let white = Array::<u8,_>::from_elem((1280,720,3), 255);
let white_s = white.as_slice().expect("oops");
```

Should this be made more clear in the documentation? I spent over 20 minutes looking for it.
- Ctrl+F `slice` returned mostly methods (including `slice()` returning `ArrayView<A, Do>` instead of slice objects.
- Ctrl+F `&[` returned mostly hardcoded input data, instead of returning slice objects.
- Ctrl+F `&[u8` returned nothing.
- Is there any IDE that would've made finding the needle in the haystack easier?

jturner314:
I've submitted a PR (#527) to add a table of conversions to/from `Vec`s and slices.

When looking for a relevant method, I often find it helpful to use the left sidebar of the docs, which has a compact listing of all of the available methods and trait implementations. (Even though there are quite a few methods, it doesn't take very long to scan the list for method names that might be relevant since there's no clutter to skip over.) You can also use the [+] and [-] buttons to expand/shrink sections. When the details are shrunk, the result is a compact listing that includes the parameter and return types.

jturner314:
#527 has been merged. Closing this issue.

PR #527, opened 2018-11-01
Add table of conversions to/from Vecs and slices

jturner314:
This improves discoverability. See also #524 for motivation.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -237,7 +237,7 @@ pub type Ixs = isize;
 /// + [Subviews](#subviews)
 /// + [Arithmetic Operations](#arithmetic-operations)
 /// + [Broadcasting](#broadcasting)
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -851,6 +851,25 @@ pub type Ixs = isize;
 /// </tr>
 /// </table>
 ///
```
bluss: Can we put a heading here, so that it's possible to find this?

I also want to disclaim the `as_slice` methods with the information that it is of course a conversion that not every array supports. (Arrays are not necessarily contiguous or in order.) I think that this information needs to be stated again, since it is non-obvious to the users that start out.

bluss:
I added comments but it doesn't need to be perfect to be merged, this is docs. Nice to have! :slightly_smiling_face:

PR #529, opened 2018-11-07
Add `ArrayBase::from_dyn()`.

boncheolgu:
I think that `from_dyn` is a missing piece, despite the presence of `into_dyn`.


jturner314:
Perhaps you're looking for [`.into_dimensionality()`](https://docs.rs/ndarray/0.12.0/ndarray/struct.ArrayBase.html#method.into_dimensionality)? I recently merged PR #511 to make conversions like this more discoverable, but the improved docs haven't made it to docs.rs yet.

boncheolgu:
Thanks!

Issue #528, opened 2018-11-04
Implement trait on Owned and View arrays


davenza:
Hi,

this could be more of a Rust question than ndarray's question, but I submit this issue because maybe some of you have solved this yet.

I have an array inside of a struct `Wrapper`. The array could be owned or view. Now, I am trying to implement a given operation that requires using `into_subview()` for both owned and view `Wrapper` structs. My take on this was:

```rust
use ndarray::{Array, ArrayBase, Axis, IxDyn, OwnedRepr, ViewRepr, Data};
use std::mem;

pub struct Wrapper<S>
    where S: Data
{
    a: ArrayBase<S, IxDyn>
}

pub trait Operation {
    fn op(&mut self, axis: Axis, index: usize);
}

impl<V> Operation for Wrapper<OwnedRepr<V>>
    where OwnedRepr<V> : Data<Elem=V>,
{
    fn op(&mut self, axis: Axis, index: usize) {
        let mut owned = mem::replace(&mut self.a,
                                     Array::from_vec(Vec::new()).into_dyn());
        owned = owned.into_subview(axis, index);
        mem::replace(&mut self.a, owned);
    }
}

impl<'a, V> Operation for Wrapper<ViewRepr<&'a V>>
    where ViewRepr<&'a V> : Data<Elem=V>,
{
    fn op(&mut self, axis: Axis, index: usize) {
        let empty_array = Array::from_vec(Vec::new()).into_dyn();
        let empty_view = empty_array.view();
        
        let mut owned = mem::replace(&mut self.a,
                                     empty_view);
        owned = owned.into_subview(axis, index);
        mem::replace(&mut self.a, owned);
    }
}
```
This code does not compile, it errors in the view implementation:

```
error[E0597]: empty_array does not live long enough                                                                                                                                                              
  --> src/mwe.rs:30:26
                                                                                                                                                                         
   |                                                                                                                                                                                                               
30 |         let empty_view = empty_array.view();                                                                                                                                                                  
   |                          ^^^^^^^^^^^ borrowed value does not live long enough                                                                                                                                 
...                                                                                                                                                                                                                
36 |     }                                                                                                                                                                                                         
   |     - borrowed value only lives until here                                                                                                                                                                    
   |                                                                                                                                                                                                               
note: borrowed value must be valid for the lifetime 'a as defined on the impl at 25:6...                                                                                                                           
  --> src/mwe.rs:25:6                                                                                                                                                                                  
[...]

PR #533, opened 2018-11-07
Add insert/remove_axis_inplace for IxDyn arrays

jturner314:
Without these methods, users must have ownership of dynamic-dimensional arrays to insert/remove axes, which is unnecessarily restrictive.

This is motivated by #528.

### DIFF
--- a/src/impl_dyn.rs
+++ b/src/impl_dyn.rs
@@ -1,0 +1,58 @@
+// Copyright 2018 bluss and ndarray developers.
+//
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+//! Methods for dynamic-dimensional arrays.
+use imp_prelude::*;
+
+/// # Methods for Dynamic-Dimensional Arrays
+impl<A, S> ArrayBase<S, IxDyn>
+where
+    S: Data<Elem = A>,
+{
+    /// Insert new array axis of length 1 at `axis`, modifying the shape and
+    /// strides in-place.
+    ///
+    /// **Panics** if the axis is out of bounds.
+    ///
+    /// ```
+    /// use ndarray::{Axis, arr2, arr3};
+    ///
+    /// let mut a = arr2(&[[1, 2, 3], [4, 5, 6]]).into_dyn();
+    /// assert_eq!(a.shape(), &[2, 3]);
+    ///
+    /// a.insert_axis_inplace(Axis(1));
+    /// assert_eq!(a, arr3(&[[[1, 2, 3]], [[4, 5, 6]]]).into_dyn());
+    /// assert_eq!(a.shape(), &[2, 1, 3]);
+    /// ```
+    pub fn insert_axis_inplace(&mut self, axis: Axis) {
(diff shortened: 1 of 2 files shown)


src/impl_dyn.rs
```diff
@@ -0,0 +1,58 @@
...
+    ///
+    /// a.remove_axis_inplace(Axis(1));
+    /// assert_eq!(a, arr1(&[1, 4]).into_dyn());
+    /// assert_eq!(a.shape(), &[2]);
+    /// ```
+    pub fn remove_axis_inplace(&mut self, axis: Axis) {
+        let len = self.len_of(axis);
+        assert_ne!(len, 0, "Length of removed axis must be nonzero.");
+        self.dim = self.dim.remove_axis(axis);
+        self.strides = self.strides.remove_axis(axis);
```
bluss: I guess removing an axis can never make us lose the first element (at index [0, 0, ..]) right, so we don't need to recompute the array's pointer here?

bluss:
Remove is surprising, maybe just make it clear to the user that it's like an in place slice, we hide some elements and can't get them back. Weird for owned arrays, but we already have the same thing in slicing, so it makes sense.

jturner314:
> Remove is surprising, maybe just make it clear to the user that it's like an in place slice, we hide some elements and can't get them back.

I can certainly document this, but I agree it's still weird.

I've always thought that [`.remove_axis()`](https://docs.rs/ndarray/0.12.0/ndarray/struct.ArrayBase.html#method.remove_axis) on `ArrayBase` was a pretty weird method because the name doesn't clearly describe what it's doing. The fact that it panics when the axis has length 0 also makes me uncomfortable. `.remove_axis(Axis(x))` is equivalent to `.into_subview(Axis(x), 0)` anyway, so I'm actually in favor of removing `.remove_axis()` entirely.

The question remains, though, what to do with `.remove_axis_inplace()` for `IxDyn` arrays. One idea is the following:

* Rename the existing `.subview_inplace()` method on `ArrayBase` to `.narrow_axis()` or something.
* Add a `.subview_inplace()` method for `IxDyn` arrays that actually removes the axis. (Don't add a `.remove_axis_inplace()` method.)

The problem with this is that changing `.subview_inplace()` like this is a breaking change in behavior that would cause problems that wouldn't be caught by the compiler, since both behaviors (keeping the axis or removing it) have the same types for `IxDyn` arrays.

A workaround for the difficult-to-catch breaking change issue is the following. I prefer this naming scheme anyway because it's more similar to the `slice`/`slice_axis` methods:

* Rename `.subview()` to `.index_axis()`
* Rename `.subview_mut()` to `.index_axis_mut()`
* Rename `.subview_inplace()` to `.narrow_axis()`/`.select_one()`/`.select_single()`/`.select_index()`/etc.
* Rename `.into_subview()` to `.index_axis_move()`
* Add `.index_axis_inplace()` to `ArrayBase<S, IxDyn>`

What do you think?

src/impl_dyn.rs
jturner314: That's correct. We need to assert that the axis length is nonzero for a similar reason.

bluss:
Sounds good! I guess `index_axis` makes more sense too (the major connection we have is to the axis iters that produce the same views).

jturner314:
See #537, which does the renames mentioned above, except using `.collapse_axis()` instead of `.narrow_axis()` and additionally renaming `.slice_inplace()` to `.slice_collapse()`.

jturner314:
Closing in favor of #537.

PR #534, opened 2018-11-11
Add from_slice methods for 1-D views

jturner314:
These are convenience methods so that you can write e.g. `ArrayView1::from_slice(s)` instead of `ArrayView1::from_shape(s.len(), s)`.


bluss:
These should already be implemented as From trait impls, see https://docs.rs/ndarray/0.12.0/ndarray/type.ArrayView.html#method.from

I think we can skip this to not duplicate the API? If not that, then reuse the same implementation.

PR #535, opened 2018-11-11
Add into_scalar method for Array0

jturner314:
Other possible names: `into_first`, `into_inner`, `unwrap`.

Edit: I just realized the implementation is broken and I need to fix it. Don't merge this yet!

Edit2: Fixed now! (See the first part of the `array0_into_scalar` test for the case where my original implementation was broken.)

### DIFF
--- a/src/impl_owned_array.rs
+++ b/src/impl_owned_array.rs
@@ -1,5 +1,44 @@
 use imp_prelude::*;
 
+/// Methods specific to `Array0`.
+///
+/// ***See also all methods for [`ArrayBase`]***
+///
+/// [`ArrayBase`]: struct.ArrayBase.html
+impl<A> Array<A, Ix0> {
+    /// Returns the single element in the array without cloning it.
+    ///
+    /// ```
+    /// use ndarray::{arr0, Array0};
+    ///
+    /// // `Foo` doesn't implement `Clone`.
+    /// #[derive(Debug, Eq, PartialEq)]
+    /// struct Foo;
+    ///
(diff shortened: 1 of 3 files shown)


LukeMathWalker:
This would avoid using the `[()]` which I think is quite confusing/difficult to find out. Good addition!

src/impl_owned_array.rs
```diff
@@ -1,5 +1,41 @@
...
+            // Any index in the `Vec` is fine since all elements are identical.
+            self.data.0.remove(0)
+        } else {
+            // Find index in the `Vec` corresponding to `self.ptr`.
+            let first = self.ptr as usize;
+            let base = self.data.0.as_ptr() as usize;
+            let index = (first - base) / size;
+            debug_assert_eq!((first - base) % size, 0);
+            // Remove the element at that index and return it.
+            self.data.0.remove(index)
```
bluss: Gnarly but it works fine. Maybe a comment to explain that the len of the Vec might be bigger than the array's current len

jturner314: Okay, I've added a explanatory comment. If someone changes the implementation in the future, the test will also help catch this edge case.

jturner314:
Do you think it would make sense to add `.into_first()` for all `Array`s (since we have `.first()` and `.first_mut()` methods), or should we just have `.into_scalar()` for `Array0`?

jturner314:
Never mind. I think `.into_scalar()` for `Array0` makes sense because `.into_first()` for general `Array`s would have to return an `Option`.

PR #537, opened 2018-11-12
Rework subview methods and other related methods

jturner314:
This PR does the following:

* Rename `subview_inplace` to `collapse_axis` (deprecating the old name).
* Rename `subview_mut` to`index_axis_mut` (deprecating the old name).
* Rename `into_subview` to `index_axis_move` (deprecating the old name).
* Rename private function `do_sub` to `do_collapse_axis`.
* Rename `subview` to `index_axis` (deprecating the old name).
* Improve docs.
* Rename `slice_inplace` to `slice_collapse` (deprecating the old name).
* Add `insert_axis_inplace` and `index_axis_inplace` for `IxDyn` arrays.
* Deprecate `remove_axis`.

These changes make method names more uniform, especially between subviews and slicing, and provide new functionality for `IxDyn` arrays.

Possible alternative names for "collapse" include "narrow" and "restrict".

After a while with `slice_inplace` deprecated, we can remove it, then add a `slice_inplace` method for `IxDyn` arrays that removes axes like `slice_move` does.

What do you think?

### DIFF
--- a/examples/axis_ops.rs
+++ b/examples/axis_ops.rs
@@ -47,7 +47,7 @@ fn main() {
     }
     a.swap_axes(0, 1);
     a.swap_axes(0, 2);
-    a.slice_inplace(s![.., ..;-1, ..]);
+    a.slice_collapse(s![.., ..;-1, ..]);
     regularize(&mut a).ok();
 
     let mut b = Array::<u8, _>::zeros((2, 3, 4));
@@ -64,6 +64,6 @@ fn main() {
     for (i, elt) in (0..).zip(&mut a) {
         *elt = i;
     }
-    a.slice_inplace(s![..;-1, ..;2, ..]);
+    a.slice_collapse(s![..;-1, ..;2, ..]);
     regularize(&mut a).ok();
 }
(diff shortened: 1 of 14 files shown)


bluss:
Nice reorganization.

bluss:
Great. I don't have the release overview at the moment, if we have breaking changes in flight already or not. Is it time to start merging them? I can make point releases before then if we need to.

jturner314:
One point of interest about the way I've implemented `insert_axis_inplace` and `index_axis_inplace` in this PR: the `dim` and `strides` are temporarily inconsistent. This is fine, though, even if `self.strides.insert_axis(axis)` or `self.strides.remove_axis(axis)` panics, because nothing can observe the inconsistency since `ArrayBase` doesn't implement `Drop` and `&mut ArrayBase<S, D>` is `!UnwindSafe`.

On a related note, I'm currently in the process of clarifying the invariants that methods on `ArrayBase` must uphold (and when the invariants need to be upheld) and updating the array creation methods to enforce the invariants. (This will also fix a couple of edge cases that currently result in undefined behavior.) I'll submit a PR when I'm done.

jturner314:
I just rebased the commits off the latest master and updated the `array0_into_scalar` test added in #535.

jturner314:
Is this ready to merge? Everything looks good to me.

bluss:
Sure!

PR #543, opened 2018-11-17
Clarify invariants and fix constructors and is_standard_layout

jturner314:
Various method and iterator implementations in `ndarray` implicitly made some assumptions that were not enforced. For example,

* In a number of places, methods/iterators violated the safety constraints on [`.offset()`](https://doc.rust-lang.org/std/primitive.pointer.html#method.offset). For example, `AxisIterCore` offsets the pointer along the axis of iteration even if one of the other axes is zero-length. As a result, before this PR, this program would have undefined behavior:

  ```rust
  extern crate ndarray;

  use ndarray::prelude::*;

  fn main() {
      let a = ArrayView2::<i32>::from_shape((0, 3), &[1]).unwrap();
      for v in a.axis_iter(Axis(1)) {
          println!("{:?}", v);
      }
  }
  ```

  The stride of axis 1 is 1, so `AxisIter` calls `ptr.offset(0 * 1)`, `ptr.offset(1 * 1)`, and `ptr.offset(2 * 1)`. The first and second offsets are fine because they're "in bounds or one byte past the end of the same allocated object", but the last offset is undefined behavior.

  This PR states the invariant that offsets along all axes must be safe, even if there are zero-length axes, and it checks this property in array constructors.

* In some places, implementations made the assumption that the axis lengths would always be less than or equal to `isize::MAX`, such as in the [`stride_offset` function in the `dimension` module](https://github.com/rust-ndarray/ndarray/blob/245cfbb031eb8c59981ea8639c109b83d9ba8954/src/dimension/mod.rs#L33-L35). However, it is possible to create a `Vec`/slice with length greater than `isize::MAX` of zero-size elements, and the implementation of `size_checked` allowed creation of arrays with axis lengths greater than `isize::MAX` (e.g. `ArrayView2::<i32>::from_shape((isize::MAX as usize + 1, 0), &[]).unwrap()`. I have trouble finding a place where this would cause undefined behavior, but these cases could cause unexpected panics in debug mode due to overflow, and the fact that I've seen code making this invalid assumption makes me uncomfortable that there may be a soundness issue I haven't found.

  This PR changes the constructors to enforce that the product of non-zero axis lengths must not exceed `isize::MAX`.

This PR explicitly states the invariants that `ArrayBase` must uphold and methods can depend on. It also adds the necessary checks to enforce these invariants when arrays are created. The invariants are carefully designed so that slices/subviews/reshapes/etc. preserve them, so checks are needed only in the constructors and not in method/iterator implementations.

This PR does make some changes to the behavior of constructors, the default strides for empty arrays, and `is_standard_layout` in some edge cases. However, I would still consider it a backwards-compatible change because it's enforcing assumptions that were made in previous versions that should have been checked. In other words, it's a bug fix.

A few notes about possible alternatives:

* We could drop the requirement that it must be safe to offset the pointer along all axes even in empty arrays. However, this would mean we'd need to add explicit checks for empty arrays in things like `AxisIter` and `.subview_inplace()`. These checks would be easy-to-forget, which is problematic because missing a check could result in undefined behavior. I much prefer performing the necessary checks in constructors so that all the other method/iterator implementations can be simple.
* We could drop the requirement that the product of non-zero axis lengths must fit in `isize`. This would be slightly more flexible because, for example, it be possible possible to create a 1-D array of `usize::MAX` zero-size elements. However, this additional flexibility doesn't have practical applications and would require us to explicitly check for overflow in a lot more places.
* We could drop the requirement that the pointer must be non-null even for empty arrays. There may even be some advantage to doing this because it would make `ArrayPtr`/`ArrayPtrMut` in #496 more flexible. However, method implementations already make this assumption (e.g. [`.as_slice()`](https://github.com/rust-ndarray/ndarray/blob/245cfbb031eb8c59981ea8639c109b83d9ba8954/src/impl_methods.rs#L1079)), allowing null pointers isn't significantly more useful than allowing dangling pointers, and forbidding null pointers will allow us to switch `ArrayBase.ptr` to use `NonNull` in the future (which should remove the need for "reborrow" methods (#412)).


src/array_serialize.rs
```diff
@@ -74,7 +74,10 @@ impl<A, S, D> Decodable for ArrayBase<S, D>
             let elements = try!(
                 d.read_struct_field("data", 2, |d| {
                     d.read_seq(|d, len| {
-                        if len != dim.size() {
+                        let dim_size = dim
+                            .size_checked()
+                            .ok_or_else(|| d.error("overflow calculating array size"))?;
+                        if len != dim_size {
```
bluss: Nice! This reminds me, it's more than time to drop rustc-serialize I think :)

src/arraytraits.rs
```diff
@@ -216,11 +216,17 @@ pub const ARRAY_FORMAT_VERSION: u8 = 1u8;
...
+/// **Panics** if the slice length is greater than `isize::MAX`.
 impl<'a, A, Slice: ?Sized> From<&'a Slice> for ArrayView<'a, A, Ix1>
     where Slice: AsRef<[A]>
 {
     fn from(slice: &'a Slice) -> Self {
         let xs = slice.as_ref();
+        assert!(
+            xs.len() <= ::std::isize::MAX as usize,
+            "Slice length must fit in `isize`.",
+        );
```
bluss: Shouldn't this be conditional and we only check if size_of A is zero?

bluss: It would be optimal to have this be a "zero cost" conversion in the regular case

src/arraytraits.rs
```diff
@@ -242,11 +248,17 @@ impl<'a, A, S, D> From<&'a ArrayBase<S, D>> for ArrayView<'a, A, D>
...
+/// **Panics** if the slice length is greater than `isize::MAX`.
 impl<'a, A, Slice: ?Sized> From<&'a mut Slice> for ArrayViewMut<'a, A, Ix1>
     where Slice: AsMut<[A]>
 {
     fn from(slice: &'a mut Slice) -> Self {
         let xs = slice.as_mut();
+        assert!(
+            xs.len() <= ::std::isize::MAX as usize,
+            "Slice length must fit in `isize`.",
+        );
```
bluss: Same size_of A == 0 comment

src/dimension/dimension_trait.rs
```diff
@@ -83,9 +83,27 @@ pub trait Dimension : Clone + Eq + Debug + Send + Sync + Default +
...
+    ///
+    /// Returns `None` if the product of non-zero axis lengths overflows
+    /// `isize`.
+    ///
+    /// If `dim.size_checked()` returns `Some(size)`, a `Vec` is created of
+    /// length `size`, and `strides` are created with `self.default_strides()`
+    /// or `self.fortran_strides()`, then the invariants are met to construct
+    /// an owned `Array` from the `Vec`, `dim`, and `strides`. (See the docs of
+    /// `can_index_slice_not_custom` for a slightly more general case.)
     fn size_checked(&self) -> Option<usize> {
```
bluss: I don't really see why this is needed, yet

src/free_functions.rs
```diff
@@ -88,16 +88,14 @@ pub fn aview1<A>(xs: &[A]) -> ArrayView1<A> {
 }
 
 /// Create a two-dimensional array view with elements borrowing `xs`.
+///
+/// **Panics** if the product of non-zero axis lengths or the numer of bytes in
```
bluss: I don't see why this would be in bytes, actually. "numer" → "number".

src/impl_constructors.rs
```diff
@@ -32,17 +32,25 @@ impl<S, A> ArrayBase<S, Ix1>
...
     /// ```rust
     /// use ndarray::Array;
     ///
     /// let array = Array::from_vec(vec![1., 2., 3., 4.]);
     /// ```
     pub fn from_vec(v: Vec<A>) -> Self {
+        assert!(
+            v.len() <= ::std::isize::MAX as usize,
+            "Length must fit in `isize`.",
+        );
```
bluss: Same size_of A == 0 check

bluss:
I agree with your thoughts about the tradeoffs. We don't need to prioritize zero size elements much at all, and I'll look if we can do this without making docs and code so much more complicated.

src/impl_constructors.rs
```diff
@@ -275,51 +289,52 @@ impl<S, A, D> ArrayBase<S, D>
...
         // debug check for issues that indicates wrong use of this constructor
-        debug_assert!(match dimension::can_index_slice(&v, &dim, &strides) {
-            Ok(_) => true,
-            Err(ref e) => match e.kind() {
-                ErrorKind::OutOfBounds => false,
-                ErrorKind::RangeLimited => false,
-                _ => true,
-            }
-        });
+        debug_assert!(dimension::can_index_slice(&v, &dim, &strides).is_ok());
```
bluss: Is this now correct? There were some negative stride issues we couldn't check properly before.

src/impl_constructors.rs
```diff
@@ -275,51 +289,52 @@ impl<S, A, D> ArrayBase<S, D>
...
+    /// Creates an array from a vector and interpret it according to the
+    /// provided shape and strides. (No cloning of elements needed.)
+    ///
+    /// The caller must ensure that the following conditions are met:
+    ///
+    /// 1. The ndim of `dim` and `strides` must be the same.
+    ///
+    /// 2. The product of non-zero axis lengths must not exceed `isize::MAX`.
+    ///
+    /// 3. For axes with length > 1, the stride must be nonnegative.
```
bluss: Not sure about why we require this

src/lib.rs
```diff
@@ -898,15 +902,128 @@ pub type Ixs = isize;
...
+// allow accessing elements that should not be possible to access).
+//
+// Method/function implementations can rely on these constraints being upheld.
+// The constraints can be temporarily violated within a method/function
+// implementation since `ArrayBase` doesn't implement `Drop` and `&mut
+// ArrayBase` is `!UnwindSafe`, but the implementation must not call
+// methods/functions on the array while it violates the constraints.
+//
+// Users of the `ndarray` crate cannot rely on these constraints because they
+// may change in the future.
```
bluss: Whole block is super

src/dimension/dimension_trait.rs
bluss: Ok, this seems good.

Isn't there an option to let this method just work with the overflow check and then check the isize::MAX requirement in constructors?

src/arraytraits.rs
jturner314: That's a good question, and the answer isn't entirely clear. The problem is that while `Vec<T>` and `Box<[T]>` promise never to allocate more than `isize::MAX` bytes, it's unclear whether slices created in other ways are allowed to contain more than `isize::MAX` bytes. If slices with non-ZST elements are allowed to have length greater than `isize::MAX` or contain more than `isize::MAX` bytes, we need to perform this check even for non-ZST elements.

The docs for [`std::slice::from_raw_parts`](https://doc.rust-lang.org/std/slice/fn.from_raw_parts.html) don't say anything about a maximum length or number of bytes, which makes me think that slices with non-ZST elements longer than `isize::MAX` or containing more than `isize::MAX` bytes may be allowed. However, [the current implementation of indexing for slices in the standard library](https://doc.rust-lang.org/src/core/slice/mod.rs.html#2012-2020) *does* assume that the slice contains no more than `isize::MAX` bytes. (It doesn't check whether the offset in bytes will overflow `isize`, but it calls [`slice.as_ptr().add(index)`](https://doc.rust-lang.org/src/core/slice/mod.rs.html#2035) anyway even though the offset in bytes not overflowing `isize` is a safety constraint for [`add`](https://doc.rust-lang.org/std/primitive.pointer.html#method.add).)

If you'd prefer, I wouldn't mind assuming, for the time being, that slices can contain no more than `isize::MAX` bytes since the standard library makes this assumption. Then, the check would only be necessary for ZST elements, and I could simplify the `can_index_slice_not_custom` function in the `dimension` module.

I haven't created an issue yet about the "more than `isize::MAX` bytes in slices" question, but a related issue is rust-lang/rust#54857, which discusses the safety of offsetting pointers from empty slices/`Vec`s by zero bytes.

bluss: Thanks. I'd go with assuming slices can't be longer than isize::MAX. It goes hand in hand with what you are working on in this PR, IMO

src/impl_constructors.rs
jturner314: If any axes with length > 1 have negative strides, moving along those axes would result in offsetting the array's pointer backwards outside the `Vec`. (`from_shape_vec_unchecked` is implemented in terms of `from_vec_dim_stride_unchecked`, which [creates the array's pointer from `v.as_mut_ptr()`](https://docs.rs/ndarray/0.12.0/src/ndarray/impl_constructors.rs.html#324).) We could allow negative strides if we allow the user to specify the array's `ptr` (or, equivalently, the offset of `ptr` relative to `v.as_mut_ptr()`).

src/impl_constructors.rs
jturner314: See my comment below. If we want to allow negative strides, we need to provide some way for the user to specify the `ptr` since it's always different from `vec.as_mut_ptr()` when there are negative strides for axes with length > 1.

src/arraytraits.rs
jturner314: Okay, I'll add a commit making that assumption.

jturner314:
By the way, one thing that surprised me was that before this PR, `can_index_slice` required that strides be strictly positive for axes with length > 0 (or zero for length = 0). I didn't see anything that would require this more strict constraint, so I changed the implementation to allow any stride for axes with length <= 1.

src/free_functions.rs
jturner314: With the assumption that slices cannot contain more than `isize::MAX` bytes, the "numer of bytes" condition can be removed. Good catch on the typo :).

src/array_serialize.rs
jturner314: Yeah, I was thinking the same thing. :)

src/impl_constructors.rs
jturner314: Good catch. This case doesn't even require the extra assumption.

src/dimension/dimension_trait.rs
jturner314: Oh, whoops. I didn't realize that `size_checked` was a public method.

Initially, I didn't do this because I thought `size_checked` was private, and I noticed that all places but one where `.size_checked()` was called needed to perform the `isize::MAX` check too. (The only exception is [this line in `indices_iter_f`](https://github.com/rust-ndarray/ndarray/blob/master/src/indexes.rs#L215).)

I'll change it back to the original behavior and add a separate checking function for internal use.

src/impl_constructors.rs
bluss: Oh great point, I just never considered that about these constructors. Thanks for the explanation.

bluss:
@jturner314 ok, so the argument for allowing more general strides for zero length axes would be that they can come up in normal operation somehow? Otherwise it seems good to ban them

jturner314:
> the argument for allowing more general strides for zero length axes would be that they can come up in normal operation somehow?

Here's an example demonstrating that arbitrary strides can occur for zero-length axes in normal operation:

```rust
extern crate ndarray;

use ndarray::{array, s, Axis};

fn main() {
    let mut a = array![[1, 2, 3], [4, 5, 6]];
    a.invert_axis(Axis(0));
    println!("a =\n{:?}", a);
    let s = a.slice(s![0..0, ..]);
    println!("s =\n{:?}", s);
}
```

It prints:

```
a =
[[4, 5, 6],
 [1, 2, 3]] shape=[2, 3], strides=[-3, 1], layout=Custom (0x0)
s =
[[]] shape=[0, 3], strides=[-3, 1], layout=Custom (0x0)
```

The stride of axis 0 is `-3` in the view `s`.

We could still ban them in constructors, but I don't see a good reason to if they can occur in normal operation.

bluss:
@jturner314 sounds good

jturner314:
I think I've resolved all of the comments, and it finally passes CI. :)

> I'll look if we can do this without making docs and code so much more complicated.

Making the assumption that slices contain at most `isize::MAX` bytes simplified things quite a bit, but it still is fairly complicated. I'd love any suggestions to simplify things. Writing this PR, designing the invariants, and correctly checking all the invariants was very difficult; there were a whole bunch of edge cases to worry about (zero-size types, arrays with some zero-length axes, arrays with all zero-length axes, arrays with zero axes, overflow in shape, overflow in offset counts, maintaining invariants after slicing/subviews/reshapes, etc.). I'm pretty satisfied with the result but would welcome improvements.

bluss:
Monumental effort, it's great! I'll try my hand at adding a comment more

src/free_functions.rs
```diff
@@ -88,16 +89,13 @@ pub fn aview1<A>(xs: &[A]) -> ArrayView1<A> {
...
-    let data = unsafe {
-        slice::from_raw_parts(xs.as_ptr() as *const A, cols * rows)
-    };
-    let dim = Ix2(rows, cols);
-    unsafe {
-        ArrayView::from_shape_ptr(dim, data.as_ptr())
-    }
+    let data = unsafe { slice::from_raw_parts(xs.as_ptr() as *const A, cols * rows) };
+    ArrayView::from_shape(Ix2(rows, cols), data).unwrap()
 }
```
bluss: I think this function should only have a check if `size_of::<V>() == 0` and no possible panic otherwise. In the case when `size_of::<V>() > 0`, we have an in-memory slice which we know Rust supports and can index into, and thus we can too. I think we can infer that our conditions are met.

bluss: Same for aview2_mut

bluss:
ready to merge when you think so

PR #544, opened 2018-11-17
Update ndarray-rand for rand 0.6

bluss:
This prepares a new release of ndarray-rand, just for this bump


bluss:
This is breaking only for ndarray-rand, and we can update it at any time. In fact I think we should strive to update it quickly whenever ndarray or rand upgrades

jturner314:
Yeah, I think it would be good to go ahead merge this PR and release a new version of `ndarray-rand`.

PR #546, opened 2018-11-17
fold axis: Include panic on axis oob in doc

bluss:
We didn't properly check the axis parameter to this method, now we panic if the axis is out of bounds (before we had partial coverage of debug assertions).

Added new test file for higher order functions like fold_axis, since the existing `tests/array.rs` is very large.

I am adamant to remove and reduce overhead for panic checks and formatting of panic messages. Maybe that's something to look at a different time, for example if the code size overhead can be minimized.

### DIFF
--- a/src/impl_methods.rs
+++ b/src/impl_methods.rs
@@ -1826,12 +1826,18 @@ impl<A, S, D> ArrayBase<S, D> where S: Data<Elem=A>, D: Dimension
     /// function and initial value `init`.
     ///
     /// Return the result as an `Array`.
+    ///
+    /// **Panics** if `axis` is out of bounds.
     pub fn fold_axis<B, F>(&self, axis: Axis, init: B, mut fold: F)
         -> Array<B, D::Smaller>
         where D: RemoveAxis,
               F: FnMut(&B, &A) -> B,
               B: Clone,
     {
+        if axis.index() >= self.ndim() {
+            panic!("ndarray: axis {} out of bounds in array of dim {} in fold_axis",
+                   axis.index(), self.ndim());
+        }
         let mut res = Array::from_elem(self.raw_dim().remove_axis(axis), init);
         for subview in self.axis_iter(axis) {
             res.zip_mut_with(&subview, |x, y| *x = fold(x, y));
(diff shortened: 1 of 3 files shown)


bluss:
Reluctantly probably a breaking change, but minor

jturner314:
Good catch on this bug!

> I am adamant to remove and reduce overhead for panic checks and formatting of panic messages. Maybe that's something to look at a different time, for example if the code size overhead can be minimized.

What do you mean? I wouldn't expect there'd be any overhead in the non-panicking case other than the `>=` comparison.

```rust
assert!(
    axis.index() >= self.ndim(),
    "ndarray: axis {} out of bounds in array dim {} in fold_axis",
    axis.index(),
    self.ndim(),
);
```

src/impl_methods.rs
```diff
@@ -1826,12 +1826,18 @@ impl<A, S, D> ArrayBase<S, D> where S: Data<Elem=A>, D: Dimension
...
+    ///
+    /// **Panics** if `axis` is out of bounds.
     pub fn fold_axis<B, F>(&self, axis: Axis, init: B, mut fold: F)
         -> Array<B, D::Smaller>
         where D: RemoveAxis,
               F: FnMut(&B, &A) -> B,
               B: Clone,
     {
+        if axis.index() >= self.ndim() {
+            panic!("ndarray: axis {} out of bounds in array dim {} in fold_axis",
```
jturner314:
```suggestion
            panic!("ndarray: axis {} out of bounds in array with ndim={} in fold_axis",
```

or

"ndarray: axis {} out of bounds in array of dim {} in fold_axis"

bluss:
@jturner314 There is some kind of code size overhead, but sure, it can be hard to quantify.

src/impl_methods.rs
bluss: Thanks, using `array of dim`

bluss:
Oh we actually do panic already, axis_iter "checks" the parameter for us. So then we'd only need this change if we want a better panic message. I'll update the PR.

bluss:
Merging with the changes (now mostly doc, test, debug assert update)

PR #550, opened 2018-11-19
Fix .axis_chunks_iter/_mut() for zero stride and zero sized types

### LINKED ISSUE #542: AxisChunksIter is broken for non-exact chunks with zero-size elements
Example:

```rust
extern crate ndarray;

use ndarray::{array, s};
use ndarray::prelude::*;

fn main() {
    let a = array![(), (), ()];
    let chunks: Vec<_> = a.axis_chunks_iter(Axis(0), 2).collect();
    assert_eq!(chunks, vec![a.slice(s![0..2]), a.slice(s![2..])])
}
```

The example results in this:

```
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `[[()] shape=[1], strides=[1], layout=C | F (0x3), [()] shape=[1], strides=[1], layout=C | F (0x3)]`,
 right: `[[(), ()] shape=[2], strides=[1], layout=C | F (0x3), [()] shape=[1], strides=[1], layout=C | F (0x3)]`', src/main.rs:9:5
```

The problem is that the shape of each chunk is determined by [comparing the current pointer to the pointer of the last chunk](https://github.com/rust-ndarray/ndarray/blob/245cfbb031eb8c59981ea8639c109b83d9ba8954/src/iterators/mod.rs#L1073), but for ZSTs, these pointers are always equal.

bluss:
Two special cases were identified, first by @jturner314, that break the
logic that picks out the last and uneven chunk in axis_chunks_iter/_mut;
we can't rely on a specific pointer value to recognize this chunk, as
the cases in the new test show.

Instead use the indices and length stored in the iterator.

Also add a new public method `fn stride_of(&self, Axis) -> isize` which was first added to clean up the style of the axis chunks iter code.

Fixes #542

### DIFF
--- a/src/iterators/mod.rs
+++ b/src/iterators/mod.rs
@@ -1040,8 +1040,7 @@ impl<'a, A, D: Dimension> NdProducer for AxisIterMut<'a, A, D>
 /// See [`.axis_chunks_iter()`](../struct.ArrayBase.html#method.axis_chunks_iter) for more information.
 pub struct AxisChunksIter<'a, A: 'a, D> {
     iter: AxisIterCore<A, D>,
-    /// Starting length of the iterator (number of chunks)
-    original_len: usize,
+    n_whole_chunks: usize,
     /// Dimension of the last (and possibly uneven) chunk
     last_dim: D,
     life: PhantomData<&'a A>,
@@ -1052,7 +1051,7 @@ clone_bounds!(
     AxisChunksIter['a, A, D] {
         @copy {
             life,
-            original_len,
+            n_whole_chunks,
         }
         iter,
         last_dim,
@@ -1091,15 +1090,15 @@ fn chunk_iter_parts<A, D: Dimension>(v: ArrayView<A, D>, axis: Axis, size: usize
         ptr: v.ptr,
     };
 
-    (iter, iter_len, last_dim)
+    (iter, n_whole_chunks, last_dim)
 }
 
 impl<'a, A, D: Dimension> AxisChunksIter<'a, A, D> {
     pub(crate) fn new(v: ArrayView<'a, A, D>, axis: Axis, size: usize) -> Self {
-        let (iter, len, last_dim) = chunk_iter_parts(v, axis, size);
+        let (iter, n_whole_chunks, last_dim) = chunk_iter_parts(v, axis, size);
         AxisChunksIter {
             iter: iter,
-            original_len: len,
+            n_whole_chunks: n_whole_chunks,
             last_dim: last_dim,
             life: PhantomData,
         }
@@ -1111,11 +1110,11 @@ macro_rules! chunk_iter_impl {
         impl<'a, A, D> $iter<'a, A, D>
             where D: Dimension
         {
-            fn get_subview(&self, iter_item: Option<*mut A>, is_last: bool)
+            fn get_subview(&self, iter_item: Option<*mut A>, is_uneven: bool)
                 -> Option<$array<'a, A, D>>
             {
                 iter_item.map(|ptr| {
-                    if !is_last {
+                    if !is_uneven {
                         unsafe {
                             $array::new_(ptr,
                                          self.iter.inner_dim.clone(),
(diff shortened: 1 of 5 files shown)


bluss:
This brings pointer `.wrapping_offset` to mind in two ways, one it can be used to safely offset outside the object (with optimization caveats), but that's useful for stepping in chunks to eschew indices.

And secondly the way the slice iterator uses `offset` for non-ZST and `wrapping_offset` for ZST to use the pointer as a counter.

src/iterators/mod.rs
```diff
@@ -1050,22 +1050,22 @@ clone_bounds!(
...
 );
 
 fn chunk_iter_parts<A, D: Dimension>(v: ArrayView<A, D>, axis: Axis, size: usize)
-    -> (AxisIterCore<A, D>, *mut A, D)
+    -> (AxisIterCore<A, D>, usize, D)
 {
     let axis = axis.index();
     let axis_len = v.shape()[axis];
     let size = if size > axis_len { axis_len } else { size };
     let last_index = axis_len / size;
```
jturner314: I realize this isn't part of the PR, but `last_index` shouldn't really be named `last_index`, because it's the last index only in the non-evenly-divisible case. (Consider e.g. `axis_len == 3` and `size == 1` -> `last_index == 3`.) I'd consider inlining the division in the computation of `iter_len` since `last_index` isn't used anywhere else.)

src/iterators/mod.rs
```diff
@@ -1050,22 +1050,22 @@ clone_bounds!(
...
             life,
-            last_ptr,
+            original_len,
         }
         iter,
         last_dim,
     }
 );
 
 fn chunk_iter_parts<A, D: Dimension>(v: ArrayView<A, D>, axis: Axis, size: usize)
```
jturner314: Will you please add some documentation?

```rust
/// Computes the information necessary to construct an iterator over chunks
/// along an axis, given a `view` of the array, the `axis` to iterate over, and
/// the chunk `size`.
///
/// Returns an axis iterator with the correct stride to move between chunks,
/// the number of chunks, and the shape of the last chunk.
```

src/iterators/mod.rs
```diff
@@ -1040,7 +1040,7 @@ impl<'a, A, D: Dimension> NdProducer for AxisIterMut<'a, A, D>
 /// See [`.axis_chunks_iter()`](../struct.ArrayBase.html#method.axis_chunks_iter) for more information.
 pub struct AxisChunksIter<'a, A: 'a, D> {
     iter: AxisIterCore<A, D>,
-    last_ptr: *mut A,
+    original_len: usize,
```
jturner314: Will you please add a comment describing what this is?

```rust
/// Length of iterator before starting. (It's necessary to preserve the length
/// because `self.iter.next_back()` changes `self.iter.len`.)
```

or

```rust
/// Number of chunks.
```

jturner314:
> And secondly the way the slice iterator uses `offset` for non-ZST and `wrapping_offset` for ZST to use the pointer as a counter.

The way `::std::slice::Iter` is implemented is clever. I think we could use the same approach if desired. The stride = 0 case could be handled the same as the ZST case, but unfortunately the check for size = 0 in `.next()` wouldn't be eliminated by the compiler as easily as the `size_of::<T>() == 0` check.

bluss:
I added an internal method (that it makes sense to make public eventually I guess?) of `fn stride_of(axis) -> isize` just to match len_of for code style cleanup purposes. Then I saw we already use this in NdProducer, and fortunately the definition for arrays matches exactly..

bluss:
Added your comments, thanks for them

bluss:
Mega microoptimization (can it ever matter?) I wonder if it makes sense to have the original_len be an inert value like usize::MAX if there is no uneven chunk, that way for even chunk splits the branch with the last chunk is never taken (so branch prediction is always right and not suddenly wrong on the last branch).

jturner314:
> I added an internal method (that it makes sense to make public eventually I guess?) of `fn stride_of(axis) -> isize`

I've wanted this method in the past, and I'm in favor of making it public.

> I wonder if it makes sense to have the original_len be an inert value like usize::MAX if there is no uneven chunk, that way for even chunk splits the branch with the last chunk is never taken (so branch prediction is always right and not suddenly wrong on the last branch).

At first glance, you could achieve the same result with the following:

* Instead of storing `original_len`, store `n_whole_chunks`.
* Instead of passing `is_last` to `.get_subview()`, pass `is_uneven`.

This might even be slightly easier to understand than the current version of this PR.

bluss:
Imperceptible difference in benchmarks of course, but that's easier to read so let's go, that's great :)

bluss:
Thanks for the comments. I'm trying to get stuff ready for 0.12.1

PR #553, opened 2018-11-20
Prepare 0.12.1

jturner314:
I think we're ready to release 0.12.1.


LukeMathWalker:
I'd say that the changelog is quite impressive!

PR #559, opened 2018-11-23
Special case is_standard_layout for 1D

bluss:
This method is called a lot, and it seems like it is sometimes too big
to optimize out like it should, even for the 1D case. Put in a special
cases for 1D.

This improves the dot_extended benchmark significantly.


bluss:
Special cases are not very nice for generality or ease of refactoring, but that the is_standard_layout method even for 1D should be too complicated to compile down into two conditionals, that's not good either

bluss:
Ok, even the 2D case was wrong, so it's removed for now.

Issue #265, opened 2017-01-20
Broadcast Illustration


bluss:
This one is really good, ndarray should adopt this (just need to change the code for each illu)

http://www.astroml.org/book_figures/appendix/fig_broadcast_visual.html

bluss:
draft. Doesn't look good in github, fine in vim.

```rust
array![0, 1, 2] + 1

┌───┬───┬───┐   ┌───┐┄┄┄┬┄┄┄┐   ┌───┬───┬───┐
│ 0 │ 1 │ 2 │ + │ 1 │ 1 ┊ 1 ┊ = │ 1 │ 2 │ 3 │
└───┴───┴───┘   └───┘┄┄┄┴┄┄┄┘   └───┴───┴───┘

array![[1, 1, 1],
       [1, 1, 1],
       [1, 1, 1]] + array![0, 1, 2]

┌───┬───┬───┐   ┌───┬───┬───┐   ┌───┬───┬───┐
│ 1 │ 1 │ 1 │   │ 0 │ 1 │ 2 │   │ 1 │ 2 │ 3 │
├───┼───┼───┤   └───┴───┴───┘   ├───┼───┼───┤
│ 1 │ 1 │ 1 │ + ┊ 0 ┊ 1 ┊ 2 ┊ = │ 1 │ 2 │ 3 │
├───┼───┼───┤   ├┄┄┄ ┄┄┄ ┄┄┄┤   ├───┼───┼───┤
│ 1 │ 1 │ 1 │   ┊ 0 ┊ 1 ┊ 2 ┊   │ 1 │ 2 │ 3 │
└───┴───┴───┘   └┄┄┄┴┄┄┄┴┄┄┄┘   └───┴───┴───┘

array![0, 1, 2].into_shape((3, 1)).unwrap() + array![0, 1, 2]

┌───┐┄┄┄┬┄┄┄┐   ┌───┬───┬───┐   ┌───┬───┬───┐
│ 0 │ 0 ┊ 0 ┊   │ 0 │ 1 │ 2 │   │ 0 │ 1 │ 2 │
├───┤┄┄┄ ┄┄┄┤   └───┴───┴───┘   ├───┼───┼───┤
│ 1 │ 1 ┊ 1 ┊ + ┊ 0 ┊ 1 ┊ 2 ┊ = │ 1 │ 2 │ 3 │
[...]

PR #565, opened 2018-12-01
Use box drawing art for illustrating an example for ArrayView::split_at

bluss:
This is a proof of concept.

Temporarily hosting the docs on github pages [see the ArrayView::split_at example there.](http://rust-ndarray.github.io/ndarray/temporary/box-art-doc/ndarray/type.ArrayView.html#method.split_at)

Box drawings are relatively easy to produce, easy to distribute and
maintain in all places rustdoc renders, readable in plain text and
when rendered, and they have a consistent style.

The biggest positive is that it enables us to produce many
illustrations.

They have the drawback that they restrict us to mostly 2-D (but that makes
for easier to understand illustrations anyway), and it doesn't look very
impressive.


bluss:
High priority items we can illustrate:

- Broadcasting, see box drawings #265 
- axis iter, lanes iter, outer iter

@jturner314 What do you think? Too ugly and unprofessional or pragmatic? Using boxes is not exactly flexible, can't color anything, but the boxes by themselves look surprisingly sharp.

jturner314:
Text has a lot of practical advantages – easy to edit, doesn't require any special software to edit, works nicely with version control, integrates well into existing docs, and makes it easy to maintain a consistent style. The best alternative would be SVG (which can be edited with Inkscape and included directly as an image), but then the images would be in separate files from the docs, hosting them would be more difficult, they wouldn't be as easy to edit/diff, and keeping the style consistent would be more difficult. Plain text doesn't look as nice as SVG images and is somewhat limiting, but IMO the practical advantages outweigh the disadvantages.

By the way, I do have a few minor suggestions about the sample drawing:

1. I'd put the indices of the vertical axis closer to the matrix to make it easier to see that they match up to the edges between boxes (see below).
2. I'd suggest making the elements include their indices to more easily see how they correspond to the indices on the edges:

   ```rust
   ///          ──▶ Axis(1)
   ///         ┌─────┬─────┬─────┬─────┐ 0
   ///       │ │ a₀₀ │ a₀₁ │ a₀₂ │ a₀₃ │
   ///       ▼ ├─────┼─────┼─────┼─────┤ 1
   ///  Axis(0)│ a₁₀ │ a₁₁ │ a₁₂ │ a₁₃ │
   ///         ├─────┼─────┼─────┼─────┤ 2
   ///         │ a₂₀ │ a₂₁ │ a₂₂ │ a₂₃ │
   ///         └─────┴─────┴─────┴─────┘ 3 ↑
   ///         0     1     2     3     4 ← possible split_at indices.
   ```

   or

   ```rust
   ///          ──▷ Axis(1)
   ///         ┌────┬────┬────┬────┐ 0
   ///       │ │ 00 │ 01 │ 02 │ 03 │
   ///       ▽ ├────┼────┼────┼────┤ 1
   ///  Axis(0)│ 10 │ 11 │ 12 │ 13 │
   ///         ├────┼────┼────┼────┤ 2
   ///         │ 20 │ 21 │ 22 │ 23 │
   ///         └────┴────┴────┴────┘ 3 ↑
   ///         0    1    2    3    4 ← possible split_at indices.
   ```

3. I'd suggest making the horizontal axis arrow longer to be more similar to the vertical axis arrow (see above).

4. I'd consider replacing the `v` and `>` with triangles, which I think look a little nicer (see above).

bluss:
Rebased & resolved conflict

bluss:
Let's go, one less PR in the queue :)

PR #567, opened 2018-12-06
Update ndarray to Rust 2018 edition

jturner314:
Most of the changes are adding a `crate` prefix to `use` statements and changing `try!` to `?`. I also renamed `libnum` to `num_traits` because `cargo fix` was wanting to add a `crate` prefix, e.g. `use crate::libnum::Whatever`, and I didn't see any reason for the `num-traits` crate to be renamed to `libnum` anyway.

This PR changes only the `ndarray` crate; it doesn't change the `ndarray-rand` crate.

Edit: Also note that I've just run `cargo fix --edition` and made some manual improvements. I haven't run `cargo fix --edition-idioms`, because according to the docs, `--edition-idioms` isn't ready yet.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ sudo: required
 dist: trusty
 matrix:
   include:
-    - rust: 1.30.0
+    - rust: 1.31.0
       env:
        - FEATURES='test docs'
     - rust: stable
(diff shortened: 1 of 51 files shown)


hellow554:
Is there a `0.13` planned for the version bumb? Or do you still want to push it as `0.12`?

bluss:
Several breaking changes merged and coming, so it's a bump

Issue #501, opened 2018-10-11
Calling slice on a dynamic size array


kali:
Hey folks,

I'm struggling to find a way to call slice() on a dynamic ranked array, and can find a way to make it work.

```
extern crate ndarray;
use ndarray::*;

fn main() {
    let array: ArrayD<f32> = Array::zeros(vec![8, 8, 8]);
    foo(&array.view());
}

fn foo<'a>(a: &'a ArrayViewD<'a, f32>) -> ArrayViewD<'a, f32> {
    let slicer: Vec<SliceOrIndex> = (0..a.ndim())
        .map(|_| SliceOrIndex::Slice {
            start: 2,
            end: Some(6),
            step: 1,
        }).collect();
    let slice = SliceInfo::new(slicer).unwrap();
    a.slice(slice)
}
```

The gist of the issue is the type of the `AsRef<[SliceOrIndex]>` in  (`slicer` in the example).

```
   |
13 |     a.slice(slice)
   |             ^^^^^ expected reference, found struct `ndarray::SliceInfo`
   |
   = note: expected type `&ndarray::SliceInfo<[ndarray::SliceOrIndex], ndarray::Dim<ndarray::IxDynImpl>>`
              found type `ndarray::SliceInfo<std::vec::Vec<ndarray::SliceOrIndex>, _>`
```

I can't see how to make the typechecker happy here, and I can't instantiate `SliceInfo<[ndarray::SliceOrIndex]>` because `[ndarray::SliceOrIndex]` is not sized. Am I missing something obvious ?

jturner314:
The solution is to call `.as_ref()` manually on `slice` to perform the conversion, like this:

```rust
a.slice(slice.as_ref())
```

This calls this `AsRef` implementation ([`SliceInfo` docs](https://docs.rs/ndarray/0.12.0/ndarray/struct.SliceInfo.html)):

```rust
impl<T, D> AsRef<SliceInfo<[SliceOrIndex], D>> for SliceInfo<T, D>
where
    T: AsRef<[SliceOrIndex]>,
    D: Dimension,
```

This is a somewhat ugly part of the API. I have a plan to fix this but haven't gotten around to it yet.

Fwiw, for this particular example, it's simpler to use a series of `.slice_axis()` or `.slice_axis_inplace()` calls, like this:

```rust
fn foo<'a>(a: &'a ArrayViewD<'a, f32>) -> ArrayViewD<'a, f32> {
    let mut view = a.clone();
[...]

PR #570, opened 2018-12-10
Add support for inserting new axes while slicing

### LINKED ISSUE #895: trait MultiSlice is undocumented
It seems like we forget to pub use the MultiSlice?

See: https://docs.rs/ndarray/0.14.0/ndarray/index.html?search=multislice

jturner314:
This is equivalent to NumPy's [`np.newaxis`](https://docs.scipy.org/doc/numpy/reference/constants.html#numpy.newaxis) feature.

A few notes:

* Adding a `CanSlice` trait has the additional benefit that `.as_ref()` is no longer needed when slicing dynamic-dimensional arrays. (See #501.)
* I haven't made `CanSlice` public because I don't see any reason (except for docs) to expose it.
* It would be possible to use something like `_` instead of an expression that evaluates to `NewAxis` to signal a new axis in the `s![]` macro.

Fixes #895

### DIFF
--- a/blas-tests/tests/oper.rs
+++ b/blas-tests/tests/oper.rs
@@ -6,7 +6,7 @@ extern crate num_traits;
 use ndarray::linalg::general_mat_mul;
 use ndarray::linalg::general_mat_vec_mul;
 use ndarray::prelude::*;
-use ndarray::{AxisSliceInfo, Ix, Ixs, SliceInfo};
+use ndarray::{AxisSliceInfo, Ix, Ixs};
 use ndarray::{Data, LinalgScalar};
 
 use approx::{assert_abs_diff_eq, assert_relative_eq};
@@ -432,7 +432,7 @@ fn scaled_add_3() {
 
                 {
                     let mut av = a.slice_mut(s![..;s1, ..;s2]);
-                    let c = c.slice(&SliceInfo::<_, IxDyn, IxDyn>::new(cslice).unwrap());
+                    let c = c.slice(&*cslice);
 
                     let mut answerv = answer.slice_mut(s![..;s1, ..;s2]);
                     answerv += &(beta * &c);
--- a/blas-tests/tests/oper.rs
+++ b/blas-tests/tests/oper.rs
@@ -432,7 +432,7 @@ fn scaled_add_3() {
 
                 {
                     let mut av = a.slice_mut(s![..;s1, ..;s2]);
-                    let c = c.slice(SliceInfo::<_, IxDyn>::new(cslice).unwrap().as_ref());
+                    let c = c.slice(&SliceInfo::<_, IxDyn, IxDyn>::new(cslice).unwrap());
 
                     let mut answerv = answer.slice_mut(s![..;s1, ..;s2]);
                     answerv += &(beta * &c);
--- a/blas-tests/tests/oper.rs
+++ b/blas-tests/tests/oper.rs
@@ -6,8 +6,8 @@ extern crate num_traits;
 use ndarray::linalg::general_mat_mul;
 use ndarray::linalg::general_mat_vec_mul;
 use ndarray::prelude::*;
+use ndarray::{AxisSliceInfo, Ix, Ixs, SliceInfo};
 use ndarray::{Data, LinalgScalar};
-use ndarray::{Ix, Ixs, SliceInfo, SliceOrIndex};
 
 use approx::{assert_abs_diff_eq, assert_relative_eq};
 use defmac::defmac;
@@ -420,11 +420,11 @@ fn scaled_add_3() {
                 let mut answer = a.clone();
                 let cdim = if n == 1 { vec![q] } else { vec![n, q] };
                 let cslice = if n == 1 {
-                    vec![SliceOrIndex::from(..).step_by(s2)]
+                    vec![AxisSliceInfo::from(..).step_by(s2)]
                 } else {
                     vec![
-                        SliceOrIndex::from(..).step_by(s1),
-                        SliceOrIndex::from(..).step_by(s2),
+                        AxisSliceInfo::from(..).step_by(s1),
+                        AxisSliceInfo::from(..).step_by(s2),
                     ]
                 };
 
--- a/src/dimension/mod.rs
+++ b/src/dimension/mod.rs
@@ -7,7 +7,7 @@
 // except according to those terms.
 
 use crate::error::{from_kind, ErrorKind, ShapeError};
-use crate::slice::CanSlice;
+use crate::slice::SliceArg;
 use crate::{AxisSliceInfo, Ix, Ixs, Slice};
 use num_integer::div_floor;
 
@@ -599,8 +599,8 @@ fn slice_min_max(axis_len: usize, slice: Slice) -> Option<(usize, usize)> {
 /// Returns `true` iff the slices intersect.
 pub fn slices_intersect<D: Dimension>(
     dim: &D,
-    indices1: &impl CanSlice<D>,
-    indices2: &impl CanSlice<D>,
+    indices1: &impl SliceArg<D>,
+    indices2: &impl SliceArg<D>,
 ) -> bool {
     debug_assert_eq!(indices1.in_ndim(), indices2.in_ndim());
     for (&axis_len, &si1, &si2) in izip!(
--- a/src/dimension/mod.rs
+++ b/src/dimension/mod.rs
@@ -19,6 +19,7 @@ pub use self::dim::*;
 pub use self::dimension_trait::Dimension;
 pub use self::dynindeximpl::IxDynImpl;
 pub use self::ndindex::NdIndex;
+pub use self::ops::DimAdd;
 pub use self::remove_axis::RemoveAxis;
 
 use crate::shape_builder::Strides;
@@ -36,6 +37,7 @@ pub mod dim;
 mod dimension_trait;
 mod dynindeximpl;
 mod ndindex;
+mod ops;
 mod remove_axis;
 
 /// Calculate offset from `Ix` stride converting sign properly
--- a/src/dimension/mod.rs
+++ b/src/dimension/mod.rs
@@ -601,7 +601,11 @@ pub fn slices_intersect<D: Dimension>(
     indices2: &impl CanSlice<D>,
 ) -> bool {
     debug_assert_eq!(indices1.in_ndim(), indices2.in_ndim());
-    for (&axis_len, &si1, &si2) in izip!(dim.slice(), indices1.as_ref(), indices2.as_ref()) {
+    for (&axis_len, &si1, &si2) in izip!(
+        dim.slice(),
+        indices1.as_ref().iter().filter(|si| !si.is_new_axis()),
+        indices2.as_ref().iter().filter(|si| !si.is_new_axis()),
+    ) {
         // The slices do not intersect iff any pair of `AxisSliceInfo` does not intersect.
         match (si1, si2) {
             (
@@ -649,6 +653,7 @@ pub fn slices_intersect<D: Dimension>(
                     return false;
                 }
             }
+            (AxisSliceInfo::NewAxis, _) | (_, AxisSliceInfo::NewAxis) => unreachable!(),
         }
     }
     true
@@ -720,7 +725,7 @@ mod test {
     };
     use crate::error::{from_kind, ErrorKind};
     use crate::slice::Slice;
-    use crate::{Dim, Dimension, Ix0, Ix1, Ix2, Ix3, IxDyn};
+    use crate::{Dim, Dimension, Ix0, Ix1, Ix2, Ix3, IxDyn, NewAxis};
     use num_integer::gcd;
     use quickcheck::{quickcheck, TestResult};
 
@@ -994,17 +999,45 @@ mod test {
 
     #[test]
     fn slices_intersect_true() {
-        assert!(slices_intersect(&Dim([4, 5]), s![.., ..], s![.., ..]));
-        assert!(slices_intersect(&Dim([4, 5]), s![0, ..], s![0, ..]));
-        assert!(slices_intersect(&Dim([4, 5]), s![..;2, ..], s![..;3, ..]));
-        assert!(slices_intersect(&Dim([4, 5]), s![.., ..;2], s![.., 1..;3]));
+        assert!(slices_intersect(
+            &Dim([4, 5]),
+            s![NewAxis, .., NewAxis, ..],
+            s![.., NewAxis, .., NewAxis]
+        ));
+        assert!(slices_intersect(
+            &Dim([4, 5]),
+            s![NewAxis, 0, ..],
+            s![0, ..]
+        ));
+        assert!(slices_intersect(
+            &Dim([4, 5]),
+            s![..;2, ..],
+            s![..;3, NewAxis, ..]
+        ));
+        assert!(slices_intersect(
+            &Dim([4, 5]),
+            s![.., ..;2],
+            s![.., 1..;3, NewAxis]
+        ));
         assert!(slices_intersect(&Dim([4, 10]), s![.., ..;9], s![.., 3..;6]));
     }
 
     #[test]
     fn slices_intersect_false() {
-        assert!(!slices_intersect(&Dim([4, 5]), s![..;2, ..], s![1..;2, ..]));
-        assert!(!slices_intersect(&Dim([4, 5]), s![..;2, ..], s![1..;3, ..]));
-        assert!(!slices_intersect(&Dim([4, 5]), s![.., ..;9], s![.., 3..;6]));
+        assert!(!slices_intersect(
+            &Dim([4, 5]),
+            s![..;2, ..],
+            s![NewAxis, 1..;2, ..]
+        ));
+        assert!(!slices_intersect(
+            &Dim([4, 5]),
+            s![..;2, NewAxis, ..],
+            s![1..;3, ..]
+        ));
+        assert!(!slices_intersect(
+            &Dim([4, 5]),
+            s![.., ..;9],
+            s![.., 3..;6, NewAxis]
+        ));
     }
 }
--- a/src/dimension/mod.rs
+++ b/src/dimension/mod.rs
@@ -7,6 +7,7 @@
 // except according to those terms.
 
 use crate::error::{from_kind, ErrorKind, ShapeError};
+use crate::slice::CanSlice;
 use crate::{AxisSliceInfo, Ix, Ixs, Slice};
 use num_integer::div_floor;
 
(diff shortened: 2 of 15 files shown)


bluss:
Exciting! I'll look closer later

src/slice.rs
```diff
@@ -103,55 +112,66 @@ pub enum SliceOrIndex {
...
-            &SliceOrIndex::Index(_) => true,
+            &AxisSliceInfo::Index(_) => true,
+            _ => false,
+        }
+    }
+
+    /// Returns `true` if `self` is a `NewAxis` value.
+    pub fn is_new_axis(&self) -> bool {
+        match self {
+            &AxisSliceInfo::NewAxis => true,
```
bluss: style nit: drop the `&`

(of course hoping to be able to provide some more valuable feedback than that :)

src/slice.rs
```diff
@@ -282,188 +303,323 @@ impl_sliceorindex_from_index!(isize);
...
-pub struct SliceInfo<T: ?Sized, D: Dimension> {
-    out_dim: PhantomData<D>,
+pub struct SliceInfo<T: ?Sized, Di: Dimension, Do: Dimension> {
+    in_dim: PhantomData<Di>,
+    out_dim: PhantomData<Do>,
     indices: T,
 }
 
-impl<T: ?Sized, D> Deref for SliceInfo<T, D>
+impl<T: ?Sized, Di, Do> Deref for SliceInfo<T, Di, Do>
```
bluss: I suppose this could have been my idea in the start, but names like `Din` and `Dout` would help readability

src/slice.rs
```diff
@@ -282,188 +303,323 @@ impl_sliceorindex_from_index!(isize);
...
-    /// consistent with `indices`.
+    /// If you call this method, you are guaranteeing that `in_dim` and
+    /// `out_dim` are consistent with `indices`.
     #[doc(hidden)]
-    pub unsafe fn new_unchecked(indices: T, out_dim: PhantomData<D>) -> SliceInfo<T, D> {
+    pub unsafe fn new_unchecked(
+        indices: T,
+        in_dim: PhantomData<Di>,
+        out_dim: PhantomData<Do>,
+    ) -> SliceInfo<T, Di, Do> {
```
bluss: This function indent, paranthesis and brace style, what is this, is this what rustfmt prefers?

src/slice.rs
```diff
@@ -282,188 +303,323 @@ impl_sliceorindex_from_index!(isize);
...
+    /// `Di::NDIM.unwrap()`. Otherwise, the value is calculated by iterating
+    /// over the `AxisSliceInfo` elements.
+    pub fn in_ndim(&self) -> usize {
+        Di::NDIM.unwrap_or_else(|| {
+            self.indices
+                .as_ref()
+                .iter()
+                .filter(|s| !s.is_new_axis())
+                .count()
+        })
```
bluss: I think if let / match is much easier to read than unwrap_or_else.

src/impl_methods.rs
```diff
@@ -321,9 +321,9 @@ where
     ///
     /// **Panics** if an index is out of bounds or step size is zero.<br>
     /// (**Panics** if `D` is `IxDyn` and `info` does not match the number of array axes.)
-    pub fn slice_mut<Do>(&mut self, info: &SliceInfo<D::SliceArg, Do>) -> ArrayViewMut<A, Do>
+    pub fn slice_mut<I>(&mut self, info: &I) -> ArrayViewMut<A, I::OutDim>
```
bluss: impl Trait would be ideal here as a user interface, but it seems like we can't access the associated type in any way then.

bluss:
@jturner314 DimAdd looks like a fine idea

src/slice.rs
```diff
@@ -67,32 +67,41 @@ impl Slice {
     }
 }
 
-/// A slice (range with step) or an index.
+/// Token to represent a new axis in a slice description.
+///
+/// See also the [`s![]`](macro.s!.html) macro.
+pub struct NewAxis;
```
bluss: derive(Copy, Clone, Debug) I guess, add others if needed (doubtful)

src/slice.rs
```diff
@@ -103,55 +112,66 @@ pub enum SliceOrIndex {
...
-copy_and_clone!{SliceOrIndex}
+copy_and_clone!{AxisSliceInfo}
 
-impl SliceOrIndex {
+impl AxisSliceInfo {
     /// Returns `true` if `self` is a `Slice` value.
     pub fn is_slice(&self) -> bool {
         match self {
-            &SliceOrIndex::Slice { .. } => true,
+            &AxisSliceInfo::Slice { .. } => true,
```
bluss: style nit: drop the `&`

blas-tests/tests/oper.rs
```diff
@@ -433,19 +433,19 @@ fn scaled_add_3() {
...
+                        AxisSliceInfo::from(..).step_by(s2),
                     ]
                 };
 
                 let c = range_mat64(n, q).into_shape(cdim).unwrap();
 
                 {
                     let mut av = a.slice_mut(s![..;s1, ..;s2]);
-                    let c = c.slice(SliceInfo::<_, IxDyn>::new(cslice).unwrap().as_ref());
+                    let c = c.slice(&SliceInfo::<_, IxDyn, IxDyn>::new(cslice).unwrap());
```
bluss: This needs to be fixed (in this PR or another), we can't design a library where `SliceInfo::<_, IxDyn, IxDyn>::new(cslice).unwrap()` is a normal thing to say.

I don't keep up with `IxDyn`.

src/slice.rs
```diff
@@ -103,55 +112,66 @@ pub enum SliceOrIndex {
...
-    /// Returns a new `SliceOrIndex` with the given step size (multiplied with
+    /// Returns a new `AxisSliceInfo` with the given step size (multiplied with
     /// the previous step size).
     ///
     /// `step` must be nonzero.
     /// (This method checks with a debug assertion that `step` is not zero.)
     #[inline]
     pub fn step_by(self, step: isize) -> Self {
-        debug_assert_ne!(step, 0, "SliceOrIndex::step_by: step must be nonzero");
+        debug_assert_ne!(step, 0, "AxisSliceInfo::step_by: step must be nonzero");
```
bluss: I guess we want a debug assert here, that we have a slice variant

src/slice.rs
```diff
@@ -163,6 +183,7 @@ impl fmt::Display for SliceOrIndex {
                     write!(f, ";{}", step)?;
                 }
             }
+            AxisSliceInfo::NewAxis => write!(f, "NewAxis")?,
```
bluss: stupid but new idea I had was to use `stringify!(NewAxis)` etc when doing this. It's like half a level cleaner code for refactoring.

bluss:
We have existing and new concepts and the dimensionality of each:

- Slice (struct) 1D
- AxisSliceInfo (enum) 1D
- SliceInfo (struct) nD
- CanSlice (trait) nD
- NewAxis 1D

1D concepts are for a single axis and nD concepts pertain to the whole array.

NewAxis already has the perfect name, so it's fine. We again need to bring some order to this and give them the best names. I'll come back to this.

src/slice.rs
jturner314: Yes, it's the `rustfmt` style when the function/method params don't fit on one line. (See the [style guide](https://github.com/rust-lang/rfcs/blob/master/style-guide/items.md#function-definitions).)

src/slice.rs
jturner314: Yes, I agree. I have a tendency to use the methods on `Option` too much when pattern matching would work.

src/dimension/mod.rs
```diff
@@ -529,20 +530,24 @@ fn slice_min_max(axis_len: usize, slice: Slice) -> Option<(usize, usize)> {
 #[doc(hidden)]
 pub fn slices_intersect<D: Dimension>(
     dim: &D,
-    indices1: &D::SliceArg,
-    indices2: &D::SliceArg,
+    indices1: &impl CanSlice<D>,
+    indices2: &impl CanSlice<D>,
```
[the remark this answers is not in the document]
jturner314: Yeah, it's nice not to have an extra two type parameters.

src/impl_methods.rs
jturner314: Yeah, that's the issue. We could use `impl CanSlice<D>` for `slice_collapse`/`slice_inplace`, though. IMO the `I` param is nice for consistency with the other slicing methods, but I'd be fine with `impl CanSlice<D>` too.

src/slice.rs
jturner314: You mean panic in debug mode if `self` is not the `AxisSliceInfo::Slice` variant? That's probably a good idea. We might as well panic in release mode, too, because we have to perform the check regardless so there's no extra performance cost.

jturner314:
I've rebased this off the latest `master`. I think this is ready to merge, assuming the various type/trait names are fine. (See @bluss's comment above.)

jturner314:
Rebased again to resolve conflict in `lib.rs` in slicing example due to #913.

bluss:
Docs is a good reason to make CanSlice visible in the docs. Users depend on clicking on it to understand what it is. (MultiSlice should be visible by the same logic).

Now, we don't have much of a module hierarchy in this crate. We're starting to miss it just to hide away lesser used things - suggestions welcome (with that said, I like the mostly-flat namespace).

src/impl_methods.rs
```diff
@@ -441,33 +455,35 @@ where
...
+                    axis += 1;
                 }
-                SliceOrIndex::Index(index) => {
+                AxisSliceInfo::Index(index) => {
                     let i_usize = abs_index(self.len_of(Axis(axis)), index);
-                    self.collapse_axis(Axis(axis), i_usize)
+                    self.collapse_axis(Axis(axis), i_usize);
+                    axis += 1;
                 }
+                AxisSliceInfo::NewAxis => {}
```
bluss, line 484: What would happen if it was an error instead?

src/slice.rs
```diff
@@ -240,220 +260,342 @@ impl From<RangeFull> for Slice {
...
+        AxisSliceInfo::NewAxis
+    }
+}
+
+/// A type that can slice an array of dimension `D`.
+///
+/// This trait is unsafe to implement because the implementation must ensure
+/// that `D`, `Self::OutDim`, `self.in_dim()`, and `self.out_ndim()` are
+/// consistent with the `&[AxisSliceInfo]` returned by `self.as_ref()`.
+pub unsafe trait CanSlice<D: Dimension>: AsRef<[AxisSliceInfo]> {
```
bluss, line 311: should probably use the private trait marker because we don't want it as an extension surface (or would that be useful?)

src/impl_methods.rs
```diff
@@ -340,9 +340,9 @@ where
     ///
     /// **Panics** if an index is out of bounds or step size is zero.<br>
     /// (**Panics** if `D` is `IxDyn` and `info` does not match the number of array axes.)
-    pub fn slice<Do>(&self, info: &SliceInfo<D::SliceArg, Do>) -> ArrayView<'_, A, Do>
+    pub fn slice<I>(&self, info: &I) -> ArrayView<'_, A, I::OutDim>
     where
-        Do: Dimension,
+        I: CanSlice<D>,
```
bluss, line 345: Potentially we'd be better off with this trait being named `SliceArg`, `SliceInfo` or something similar, direct and descriptive.

src/slice.rs
```diff
@@ -240,220 +260,342 @@ impl From<RangeFull> for Slice {
...
+/// that if `Din` is a fixed dimension type (`Ix0`, `Ix1`, `Ix2`, etc.), the
+/// `SliceInfo` instance can still be used to slice an array with dimension
+/// `IxDyn` as long as the number of axes matches.
 ///
 /// [`.slice()`]: struct.ArrayBase.html#method.slice
 #[derive(Debug)]
 #[repr(C)]
-pub struct SliceInfo<T: ?Sized, D: Dimension> {
-    out_dim: PhantomData<D>,
+pub struct SliceInfo<T: ?Sized, Din: Dimension, Dout: Dimension> {
```
bluss, line 375: If we do this, does that actually help? At least we simplify the awful `SliceInfo<_, IxDyn, IxDyn>` invocation

```suggestion
pub struct SliceInfo<T: ?Sized, Din: Dimension, Dout: Dimension = Din> {
```

src/slice.rs
```diff
@@ -240,220 +260,342 @@ impl From<RangeFull> for Slice {
...
+/// The type `T` is typically `[AxisSliceInfo; n]`, `[AxisSliceInfo]`, or
+/// `Vec<AxisSliceInfo>`. The type `Din` is the dimension of the array to be
+/// sliced, and `Dout` is the output dimension after calling [`.slice()`]. Note
+/// that if `Din` is a fixed dimension type (`Ix0`, `Ix1`, `Ix2`, etc.), the
+/// `SliceInfo` instance can still be used to slice an array with dimension
+/// `IxDyn` as long as the number of axes matches.
 ///
 /// [`.slice()`]: struct.ArrayBase.html#method.slice
 #[derive(Debug)]
 #[repr(C)]
```
bluss, line 374: We could have `repr(transparent)` here now I think

blas-tests/tests/oper.rs
jturner314: I just pushed a better solution. If we implement `CanSlice<IxDyn>` for `[AxisSliceInfo]`, then there's no need for the user to create a `SliceInfo` instance here; they can just write `&*cslice`. (Optionally, we could also implement `CanSlice<IxDyn>` for `Vec<AxisSliceInfo>` to avoid the need for the `*`.)

We could also consider implementing `CanSlice<IxDyn>` for `[Slice]` and `CanSlice<Dim<[usize; N]>>` for `[Slice; N]`.

I do think this test is better written by slicing individual axes with `.slice_axis*()` rather than combining the slicing information and calling `.slice()`.

src/slice.rs
jturner314: IMO, this change would be too confusing for the small gain in conciseness. I added a comment about the `SliceInfo::<_, IxDyn, IxDyn>` above with a good solution. If we implement `CanSlice<IxDyn>` for `[AxisSliceInfo]`, there really isn't much reason to ever create a `SliceInfo<_, IxDyn, IxDyn>` instance.

src/slice.rs
jturner314: Good point; I've made the change.

src/slice.rs
```diff
@@ -240,220 +260,357 @@ impl From<RangeFull> for Slice {
...
     fn deref(&self) -> &Self::Target {
         &self.indices
     }
 }
 
-impl<T, D> SliceInfo<T, D>
+impl<T, Din, Dout> SliceInfo<T, Din, Dout>
 where
-    D: Dimension,
+    T: AsRef<[AxisSliceInfo]>,
```
bluss, line 406: This raises my `AsRef` concern - an unfaithful/evil impl of `AsRef` can return different slices at different times, we can't trust `AsRef` for `unsafe` code blocks - I think it means the check we do in `SliceInfo::new` is not effectively guarding the invariant we need because `.as_ref()` can later return a different slice. 🙁  This is one of my pet peeves - can't trust `AsRef` for unsafe code, unfortunately.

jturner314: Oh, good point. I've changed `SliceInfo::new` to be `unsafe` and added `TryFrom` implementations for a whitelisted set of types (slices, `Vec`, and fixed-length arrays) which we should be able to depend on having a correct `AsRef` impl.

It would be nice to make `CanSlice` and the `SliceInfo` constructors safe, i.e. move the responsibility for the necessary checks to the slicing methods. Modifying the `.slice*()` methods would be fairly simple. (The primary change would be to make sure that they call `.as_ref()` only once.) However, modifying the implementation of multi-slicing would be trickier.

src/impl_methods.rs
jturner314: That's probably better than just ignoring it silently. I've made that change to the PR. Other options include:

- Panic if `info` contains `NewAxis`.
- Prevent this at compile time. (Add another type parameter to `SliceInfo` to track whether or not it contains `NewAxis`, and add another trait like `CanSlice`, but specifically for `slice_collapse`. We'll have to add another type parameter to support ellipsis anyway, so maybe the additional complexity of this option wouldn't be too bad.)

src/slice.rs
jturner314: Okay, I've added the private marker to `CanSlice` and `MultiSlice` and exposed them in the public API.

src/impl_methods.rs
jturner314: Yeah, I don't really know what to call it. Updating the list from your earlier comment, we have the following:

- `Slice` (struct) 1D
- `AxisSliceInfo` (enum) 1D
- `SliceArg` (trait) 1D, used internally by `s![]` macro and hidden from the docs
- `SliceInfo` (struct) nD
- `CanSlice` (trait) nD
- `MultiSlice` (trait) nD
- `NewAxis` (struct) 1D

It would be nice for the following to have similar names:

- `CanSlice` and `MultiSlice`
- `AxisSliceInfo` and `SliceInfo`

`SliceArg`/`MultiSliceArg` would work. I've also considered `Slicer`/`MultiSlicer`. `SliceInfo`/`MultiSliceInfo` would be fine, but we'd need to rename the `SliceInfo` struct and probably `AxisSliceInfo`, too.

Do you have a preference?

jturner314:
I just added a commit refining the `DimAdd` trait. In particular, it removes the `Dimension` bound on `DimAdd`, and adds a number of common `DimAdd` bounds to `Dimension`. This is similar to [the trick @bluss proposed for the `BroadcastShape` trait](https://github.com/rust-ndarray/ndarray/pull/898#issuecomment-765797544) in #898.

One interesting application for this is that we can now describe twice the dimensionality of any `D: Dimension` as `<D as DimAdd<D>>::Out`, which would be useful for adding `.into_view()` methods to `Windows`, `ExactChunks`, and `ExactChunksMut` where the "outer" axes of the view are the iteration axes and the "inner" axes of the view are the window/chunk axes.

We could also consider removing the `Larger` associated trait from `Dimension` because it's equivalent to `<D as DimAdd<Ix1>>::Out`.

Another naming question -- should the associated type be named `Out` or `Output`? I originally called it `Out` for brevity, but now I'm leaning towards `Output` for consistency with `std::ops`. [`nalgebra` also uses `Output` for the comparable associated type.](https://docs.rs/nalgebra/0.24.1/nalgebra/base/dimension/trait.DimAdd.html#associatedtype.Output)

bluss:
Yes, please use Output, it's a pretty wide-spread convention including closure traits etc.

src/impl_methods.rs
bluss: I think Slicer/MultiSlicer sound good then (the SliceArg pair would also works) because SliceInfo sounds good as a concrete struct.

bluss:
(I think the back-and-forth of reviewing is awkward, but we still have one more thing.. looks good so far).

I think `slice_collapse` is better with either silently ignoring NewAxis or panicking on NewAxis. Panicking on indexing errors is ok, that's what we do in this crate so far.  Returning an error after performing part of the slicing, rubs me (a little) the wrong way - all or nothing would be better? But Panic probably the easy way out.

src/impl_methods.rs
jturner314: I've changed it to a panic, based on @bluss's comment below.

src/impl_methods.rs
jturner314: After thinking about it some more, `SliceArg`/`MultiSliceArg` sound a little more natural to me than `Slicer`/`MultiSlicer` in this context, so I've renamed `SliceArg` to `SliceNextDim`, `CanSlice` to `SliceArg`, and `MultiSlice` to `MultiSliceArg`.

jturner314:
I think I've made all the changes @bluss suggested. I could clean up the Git history somewhat if that would be useful and worth the effort, but otherwise, I think this is ready to merge.

bluss:
Rebased to fix merge conflicts

bluss:
Thanks a lot :tada:

bluss:
With this done, it's time to release 0.15. Just need to know if there are any last-minute cleanups needed or necessary.

jturner314:
I'm glad this is merged! :tada: I'm not aware of anything critical that needs to be done before the release. #919 and #936 are the two things that come to mind, but they could wait until 0.15.1 if necessary.

Issue #584, opened 2019-01-22
Incorrect output of dot


sebasv:
I noticed very peculiar behavior of the dot product when the `ndarray-linalg` package is included in a project:

### Code to replicate:

```rust
extern crate ndarray;
#[cfg(test)]
mod tests {
    use ndarray::prelude::*;
    #[test]
    fn test_dgemv() {
        let a = arr2(&[[1.], [2.]]);
        let b = arr1(&[1., 2.]);
        println!("{:?}", a.t().dot(&b));
    }
}

```

### Output:
```
Intel MKL ERROR: Parameter 7 was incorrect on entry to cblas_dgemv.
[0.0] shape=[1], strides=[1], layout=C | F (0x3), const ndim=1
```

### dependencies:
```toml
[dependencies]
ndarray = "0.12.1"

[dependencies.ndarray-linalg]
version = "0.10"
features = ["intel-mkl"]
```

Note the lack of a panic. If I remove the `ndarray-linalg` dependency the output is as expected. 
IMO `ndarray` should not outsource the dot product to BLAS when another package requires BLAS, but that is secondary to the uncaught error and faulty returned value.

sebasv:
I noticed the problem does not persist if I manually broadcast the shapes to match:
```rust
        println!("{:?}", &b.broadcast((1, 2)).unwrap().dot(&a));
```
```
[[5.0]] shape=[1, 1], strides=[1, 1], layout=C (0x1), const ndim=2
```

bluss:
Thanks for the reproducer. While the method is called dot I guess in long form we need to call this operation matrix-vector multiplication (hence "gemv"), and it's avoided by making it a matrix-matrix multiplication in your workaround. (It dispatches based on dimensionality and not axis length).

bluss:
I'm not sure we expect a panic here. Still need to fix the gemv call though.

First thing I thought about were the shapes, if the row vs column was an issue, not sure what the expected panic is?

`arr2(&[[1.], [2.]])` is a 2D array so it is an 2 × 1 matrix, transposed it's a 1 × 2 matrix.

`arr1(&[1., 2.])` is a 1D array so it can be a "row" or "column" depending on use. As the right hand operand of gemv, it is a row matrix, so in ndarray it has the shape 2 but a as a matrix in this operation it has the shape 2 × 1. Thus the matrix multiplication 1 × 2 by 2 × 1 is compatible with an 1 × 1 output.

[...]

PR #585, opened 2019-01-22
Fix blas mat-vec multiplication on array with only 1 nontrivial dimension

sebasv:
First attempt at fixing #584

### DIFF
--- a/src/linalg/impl_linalg.rs
+++ b/src/linalg/impl_linalg.rs
@@ -588,8 +588,12 @@ pub fn general_mat_vec_mul<A, S1, S2, S3>(alpha: A,
                     {
                         let a_trans = CblasNoTrans;
                         let a_stride = match layout {
-                            CBLAS_LAYOUT::CblasRowMajor => a.strides()[0].max(k as isize) as blas_index,
-                            CBLAS_LAYOUT::CblasColMajor => a.strides()[1].max(m as isize) as blas_index,
+                            CBLAS_LAYOUT::CblasRowMajor => {
+                                a.strides()[0].max(k as isize) as blas_index
+                            }
+                            CBLAS_LAYOUT::CblasColMajor => {
+                                a.strides()[1].max(m as isize) as blas_index
+                            }
                         };
 
                         let x_stride = x.strides()[0] as blas_index;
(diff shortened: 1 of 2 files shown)


sebasv:
The failing tests are behavior intended by this fix, needs discussion if this is desired behavior.

src/linalg/impl_linalg.rs
```diff
@@ -588,8 +588,8 @@ pub fn general_mat_vec_mul<A, S1, S2, S3>(alpha: A,
                     {
                         let a_trans = CblasNoTrans;
                         let a_stride = match layout {
```
jturner314: It would be nice to add a comment, e.g.

```rust
                        // Determine stride between rows or columns. Note that the stride is
                        // adjusted to at least `k` or `m` to handle the case of a matrix with a
                        // trivial (length 1) dimension, since the stride for the trivial dimension
                        // may be arbitrary.
```

bluss:
Using the maintainer push to PR to update.

I'll add your comment @jturner314 jturner314 but not in that way, because it causes a textual conflict.

bluss:
@jturner314 @sebasv do you remember what the failing test comment was about? I guess it's obsolete now, because the tests pass and this looks ready to merge to me.

jturner314:
> @jturner314 @sebasv do you remember what the failing test comment was about? I guess it's obsolete now, because the tests pass and this looks ready to merge to me.

I'm not sure. Maybe it's referring to the `mat_vec_product_1d` test added by the PR? (That would be surprising because the test was added after @sebasv made that comment, but maybe he just hadn't pushed the test yet?) That test fails before the fix in this PR.

bluss:
ok must be obsolete, thanks :)

Thanks so much @sebasv for reporting and fixing! (It's me who renamed the test and the PR to avoid using the term "dot product" here.)

sebasv:
Just for completeness, if I recall correctly there were two tests not related to this PR which failed because they called a dot product on a transposed array with a trivial dimension, which messed up the strides. If no tests fail now, this was probably resolved with another bugfix. @bluss @jturner314  thank you for your reviews and for merging the fix!

PR #591, opened 2019-02-27
Improve docs of .raw_dim(), .shape(), and .strides()

jturner314:
This should help prevent a common source of confusion, even among relatively experienced `ndarray` users: using `.shape()` where `.raw_dim()` is necessary. This has come up multiple times in the past.

Please let me know if the wording could be improved.

### DIFF
--- a/src/impl_methods.rs
+++ b/src/impl_methods.rs
@@ -102,18 +102,28 @@ where
 
     /// Return the shape of the array as a slice.
     ///
-    /// Note that you probably don't want to use this to create an
-    /// array of the same shape as another array because creating an
-    /// array with e.g. [`Array::zeros()`](ArrayBase::zeros) using a
-    /// shape of type `&[usize]` results in a dynamic-dimensional
-    /// array. For example, if `arr` is of type `Array<A, D>`, then
-    /// `Array::zeros(arr.shape())` returns an instance of type
-    /// `Array<B, IxDyn>`, not `Array<B, D>`.
+    /// Note that you probably don't want to use this to create an array of the
+    /// same shape as another array because creating an array with e.g.
+    /// [`Array::zeros()`](ArrayBase::zeros) using a shape of type `&[usize]`
+    /// results in a dynamic-dimensional array. If you want to create an array
+    /// that has the same shape and dimensionality as another array, use
+    /// [`.raw_dim()`](ArrayBase::raw_dim) instead:
     ///
-    /// If you want to create an array that has the same shape and
-    /// dimensionality as another array, use
-    /// [`.raw_dim()`](ArrayBase::raw_dim) instead, e.g.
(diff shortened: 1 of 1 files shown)


LukeMathWalker:
Yes please :heart:

src/impl_methods.rs
```diff
@@ -83,17 +83,43 @@ where
...
+    /// let a = Array::from_elem((2, 3), 5.);
+    ///
+    /// // Create an array of zeros that's the same shape and dimensionality as `a`.
+    /// let b = Array::<f64, _>::zeros(a.raw_dim());
+    /// ```
     pub fn raw_dim(&self) -> D {
         self.dim.clone()
     }
 
     /// Return the shape of the array as a slice.
```
LukeMathWalker: What would be really useful here is a doctest example. Something like
```rust
let a = array![
    [1, 2, 3, 4],
    [3, 4, 2, 1],
];
let shape = a.shape();
assert_eq!(shape, vec![2, 4].as_slice());
let b = ArrayBase::zeros(shape);
let c = Array::zeros(a.raw_dim());
assert_ne!(b, c);
assert_eq!(b, c.into_dyn());
```

bluss:
@jturner314 @LukeMathWalker  Is it possible you want to join maintenance team @LukeMathWalker? I'm away more than not, so I'm really not helping unfortunately. If you drive development you can also make decisions independently from me. (That's how it should be if I don't show up.)

LukeMathWalker:
I'd be more than happy to join the maintenance team @bluss: I have learned most of the non-trivial things  I know about Rust using/messing up/trying to expand `ndarray` and putting that code out for review by you and @jturner314 - it would be exciting to get more involved with the project, giving something back. 
I can't say I am familiar with the deepest and darkest corners of `ndarray`, but the goal is getting there :stuck_out_tongue:

jturner314:
@bluss @LukeMathWalker I've added @LukeMathWalker as a GitHub team member and collaborator. Welcome to the team! You've been doing great work on `ndarray-stats` and `ndarray`, and I'm glad to have you with us!

src/impl_methods.rs
jturner314: I've added an example based on what you provided. Is that better?

By the way, the reason why I didn't include `assert_ne!(b, c)` is that `PartialEq` is not implemented for two arrays of different dimensionality. Ideally, we'd support comparing `Array<A, D>` with `ArrayD<A>` for any `D: Dimension`, but that's not possible until Rust has specialization, because it would overlap the existing impl.

LukeMathWalker: It looks pretty neat to me - do we want to go ahead and merge this PR?

jturner314: Okay, I've merged it. Thanks for taking a look.

PR #601, opened 2019-03-20
Improve documentation for sum_axis and mean_axis

JP-Ellis:
Instead of using square initial matrices, use 2x3 matrices to make it clearer in
the example which Axis is calculated over.


jturner314:
This is great, thanks!

PR #610, opened 2019-03-25
Add docs about nested `Vec`s/`Array`s

### LINKED ISSUE #609: Improve documentation or add methods for conversion from nested `Vec`/`ArrayBase`
A somewhat frequently-requested feature (e.g. #197, #590, #607) is to provide methods to create `Array` instances from nested `Vec` and/or `ArrayBase` instances (e.g. `Array2<A>` from `Vec<Vec<A>>` or `Array3<A>` from `Vec<Array2<A>>`).

Note that it's generally advisable to avoid nested `Vec`/`ArrayBase` types like this because they require extra heap allocations, can scatter data all over memory, cause unnecessary indirection, fail to enforce consistent shape within the nested `Vec`s/`ArrayBase`s, and are generally more difficult to work with.

I think we should:

* Improve the documentation to explain why these nested data structures are best avoided and the best way to handle the use cases (primarily constructing an `Array` by appending rows).
* For users who need to use nested data structures anyway, either add the methods to perform the conversions or suggest how they could perform the conversion manually. (`data.iter().flatten().cloned().collect::<Vec<_>>()` followed by `from_shape_vec` seems pretty clean if the user is confident that the nested `Vec`s/`ArrayBase`s are all the same length/shape).

jturner314:
Fixes #609 (assuming we don't want to actually add conversion methods).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -939,7 +939,8 @@ pub type Ixs = isize;
 /// <sup><a name="req_contig">3</a></sup>Works only if the array is contiguous.
 ///
 /// The table above does not include all the constructors; it only shows
-/// conversions to/from `Vec`s/slices. See below for more constructors.
+/// conversions to/from `Vec`s/slices. See
+/// [below](#constructor-methods-for-owned-arrays) for more constructors.
 ///
 /// [ArrayView::reborrow()]: type.ArrayView.html#method.reborrow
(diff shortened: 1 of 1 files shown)


LukeMathWalker:
I think this is a good addition - if we later decide to provide some new constructor from nested collections we can amend it as required.

PR #611, opened 2019-03-27
Add accumulate_axis_inplace method and implement NdProducer for RawArrayView/Mut

jturner314:
This `accumulate_axis_inplace` method is a non-cloning version of the `accumulate_axis_method` proposed in #513.

### DIFF
--- a/src/impl_methods.rs
+++ b/src/impl_methods.rs
@@ -2268,9 +2268,9 @@ where
         if self.len_of(axis) <= 1 {
             return;
         }
-        let mut prev = self.raw_view();
+        let mut curr = self.raw_view_mut(); // mut borrow of the array here
+        let mut prev = curr.raw_view(); // derive further raw views from the same borrow
         prev.slice_axis_inplace(axis, Slice::from(..-1));
-        let mut curr = self.raw_view_mut();
         curr.slice_axis_inplace(axis, Slice::from(1..));
         // This implementation relies on `Zip` iterating along `axis` in order.
         Zip::from(prev).and(curr).apply(|prev, curr| unsafe {
--- a/src/impl_methods.rs
+++ b/src/impl_methods.rs
@@ -2231,4 +2231,60 @@ where
             })
         }
     }
+
+    /// Iterates over pairs of consecutive elements along the axis.
+    ///
+    /// The first argument to the closure is an element, and the second
+    /// argument is the next element along the axis. Iteration is guaranteed to
+    /// proceed in order along the specified axis, but in all other respects
+    /// the iteration order is unspecified.
+    ///
+    /// # Example
+    ///
+    /// For example, this can be used to compute the cumulative sum along an
+    /// axis:
+    ///
+    /// ```
+    /// use ndarray::{array, Axis};
+    ///
+    /// let mut arr = array![
+    ///     [[1, 2], [3, 4], [5, 6]],
+    ///     [[7, 8], [9, 10], [11, 12]],
+    /// ];
+    /// arr.accumulate_axis_inplace(Axis(1), |&prev, curr| *curr += prev);
+    /// assert_eq!(
+    ///     arr,
+    ///     array![
+    ///         [[1, 2], [4, 6], [9, 12]],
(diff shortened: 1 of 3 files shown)


bluss:
Amazing, we'll need this to improve upon .. so many things, including how we handle uninitalized values. I'll need to remind myself to come back to this.

src/impl_methods.rs
```diff
@@ -2138,4 +2138,60 @@ where
...
+        S: DataMut,
+    {
+        if self.len_of(axis) <= 1 {
+            return;
+        }
+        let mut prev = self.raw_view();
+        prev.slice_axis_inplace(axis, Slice::from(..-1));
+        let mut curr = self.raw_view_mut();
+        curr.slice_axis_inplace(axis, Slice::from(1..));
+        // This implementation relies on `Zip` iterating along `axis` in order.
```
bluss: Good point, I guess we'll think about if that's something we can guarantee. Probably not publically.

bluss:
Rebased to resolve the conflict. I had to reformat and fix some things for NonNull. I added a commit on top with what I think would be the careful way to use raw view. I haven't studied the unsafe coding guidelines that well, so this is just my rough understanding.

bluss:
Updating because RawView + Zip should be an awesome ally in solving fundamentals, like how to do more operations in place or without excessive copying, or just to handle uninitialized data better.

bluss:
The format checker wants to ruin the formatting of the arrays in tests, and I'd like to make it less strict :/

src/impl_methods.rs
```diff
@@ -2231,4 +2231,60 @@ where
...
+    /// arr.accumulate_axis_inplace(Axis(1), |&prev, curr| *curr += prev);
+    /// assert_eq!(
+    ///     arr,
+    ///     array![
+    ///         [[1, 2], [4, 6], [9, 12]],
+    ///         [[7, 8], [16, 18], [27, 30]],
+    ///     ],
+    /// );
+    /// ```
+    pub fn accumulate_axis_inplace<F>(&mut self, axis: Axis, mut f: F)
```
bluss: This method is a neat demonstration of an array Zip-ing with itself. It seems a bit niche but we can of course include it. But the question of course goes out, how do we generalize this. Apart from the raw pointer Zip, of course?

src/impl_methods.rs
```diff
@@ -2231,4 +2231,60 @@ where
...
+    {
+        if self.len_of(axis) <= 1 {
+            return;
+        }
+        let mut curr = self.raw_view_mut(); // mut borrow of the array here
+        let mut prev = curr.raw_view(); // derive further raw views from the same borrow
+        prev.slice_axis_inplace(axis, Slice::from(..-1));
+        curr.slice_axis_inplace(axis, Slice::from(1..));
+        // This implementation relies on `Zip` iterating along `axis` in order.
+        Zip::from(prev).and(curr).apply(|prev, curr| unsafe {
```
bluss: This is a particularly simple Zip - the prev and curr pointer are always just the stride of `axis` apart, and in that sense it could have been just a unary loop, with one pointer `prev` and another `prev + offset`. If the Zip only compiles to the equivalent, all the better, then we have a nice way to express exactly that.

bluss:
Thanks for this, the ndproducer impls will be very useful going forward. And I guess we should have discussions about how to use it correctly (and eventually produce docs for that when we know :slightly_smiling_face: )

src/impl_methods.rs
jturner314: I think we really need to wait until Rust has generic associated types before we can write good self-Zip abstractions. We can do things like this today:

```rust
    /// Folds over the mutable windows of the specified size.
    pub fn windows_mut_fold<E, B, F>(&mut self, window_size: E, init: B, f: F)
    where
        E: IntoDimension<Dim = D>,
        F: for<'a> FnMut(B, ArrayViewMut<'a, A, D>) -> B,
        S: DataMut,
    {
        // ...
    }
```

but without generic associated types, we can't abstract this into a `WindowsMut` producer that implements a trait, because the item type (which is an associated type of the trait) has a temporary lifetime.

Issue #579, opened 2019-01-07
Support zero-length `axis` in `.map_axis/_mut()`


jturner314:
The implementation of `.map_axis/_mut()` panics when the specified `axis` is zero-length. Ideally, we'd just apply `mapping` to empty views instead.

Example:

```rust
extern crate ndarray;

use ndarray::prelude::*;

fn main() {
    let a = Array2::<f32>::zeros((0, 2));
    a.map_axis(Axis(0), |_| {});
}
```

andrei-papou:
Hi @jturner314 . I can work on this one if you don't mind.

jturner314:
@andrei-papou Sure, please feel free.

andrei-papou:
@jturner314 could you please point me to an example of creating an empty array? Thanks a lot.

jturner314:
The simplest way to create a 1-dimensional, zero-length view in `map_axis` would be `unsafe { ArrayView::new_(std::ptr::NonNull::<A>::dangling().as_ptr(), Ix1(0), Ix1(0)) }`. See [this commentary](https://github.com/rust-ndarray/ndarray/blob/47b26916b242f1dd5ba92703c64e8c7e3d55caaa/src/lib.rs#L1130-L1136).

andrei-papou:
@jturner314 thanks. please take a look at the PR https://github.com/rust-ndarray/ndarray/pull/612

jturner314:
Fwiw, I realized that a cleaner (safe) way to create a zero-length one-dimensional view is `ArrayView1::from(&[])`.

jblondin:
PR was merged. Is anything still remaining here, or can this issue be closed?

jturner314:
Fixed by #612. Thanks for following up.

PR #612, opened 2019-03-29
Support zero-length `axis` in `.map_axis/_mut()`

andrei-papou:
PR for https://github.com/rust-ndarray/ndarray/issues/579


src/impl_methods.rs
```diff
@@ -2096,7 +2105,16 @@ where
         let view_stride = self.strides.axis(axis);
         // use the 0th subview as a map to each 1d array view extended from
         // the 0th element.
-        self.index_axis_mut(axis, 0).map_mut(|first_elt: &mut A| {
+        let mut first_subview = if view_len != 0 {
+            self.index_axis_mut(axis, 0)
+        } else {
+            let smaller_zero_dim = <D::Smaller as Dimension>::zeros(self.dim.ndim() - 1);
+            let ptr = std_ptr::NonNull::<A>::dangling().as_ptr();
+            unsafe {
```
LukeMathWalker: I'd like a comment here to explain why this is a safe thing to do (same in the other spot).

andrei-papou: @LukeMathWalker is it be enough to point to [this](https://github.com/rust-ndarray/ndarray/blob/master/src/lib.rs#L1108) section in the `ArrayBase` docstring?
> The only exceptions are if the array is empty or the element
> type is zero-sized. In these cases, `ptr` may be dangling, but it must
> still be safe to [`.offset()`] the pointer along the axes.

LukeMathWalker: I'd suggest to extract it in a function, so that we can reuse code across the two calling sites and keep a better eye on it. It's probably better to copy the relevant section over, to have it close to the code.

andrei-papou: Ready

src/impl_methods.rs
```diff
@@ -2068,7 +2068,12 @@ where
         let view_stride = self.strides.axis(axis);
         // use the 0th subview as a map to each 1d array view extended from
         // the 0th element.
-        self.index_axis(axis, 0).map(|first_elt| {
+        let first_subview = if view_len != 0 {
+            self.index_axis(axis, 0)
+        } else {
+            ArrayView::<A, D::Smaller>::new_empty()
```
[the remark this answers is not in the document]
andrei-papou: @jturner314 correct me if I wrong, but array with shape `&[3, 0, 4]` may not contain any elements since the total number of elements is `3 * 0 * 4 = 0`. Example:
```rust
let a = Array3::from_elem((3, 0, 4), 1.0);
assert_eq!(a.is_empty(), true);
```
But array of shape `&[3, 4]` should contain 12 elements. Where should we take 12 more elements from in this case?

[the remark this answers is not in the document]
andrei-papou: Thanks, I got the idea

tests/array.rs
```diff
@@ -2002,6 +2002,35 @@ fn test_map_axis() {
...
+    // Test zero-length axis case
+    let arr = Array3::<f32>::zeros((3, 0, 4));
+    let mut counter = 0;
+    let result = arr.map_axis(Axis(1), |x| {
+        assert_eq!(x.shape(), &[0]);
+        counter += 1;
+        counter
+    });
+    assert_eq!(result.shape(), &[3, 4]);
+    assert_eq!(result, arr2(&[
```
jturner314: Overall, this test is good, but we don't guarantee the iteration order of `map_axes`, so the elements of `result` are not guaranteed to be arranged in this way. Something like

```rust
itertools::assert_equal(result.iter().cloned().sorted(), 1..=3*4);
```

would work to verify that the elements are correct while ignoring their ordering.

(For access to `.sorted()`, make sure to `use itertools::Itertools` and add `features = ["use_std"]` to the `itertools` dependency in `Cargo.toml`.)

andrei-papou: Thanks, I've updated the test accordingly.

andrei-papou:
@jturner314 any more comments on the changes? the issues above are fixed

jturner314:
Looks good to me. Thanks!

PR #614, opened 2019-04-03
Forward more iterator methods

### LINKED ISSUE #555: Forward more iterator methods for element iterators
The `Iter` and `IterMut` are composed of a contiguous case and a general case, and because of this we should forward most (non-adaptor) methods to the underlying iterators. The existing `fold` and `rfold` methods serve as examples. No new custom implementations of the methods are needed in ElementsBase or Baseiter.

The ultimate ones to forward are `try_fold` and `try_rfold`, but they need to wait for having stable signatures (all traits in the signature).

Good ones to forward:

- nth
- collect
- all, any
- find, find_map

Some don't matter because they use fold that we already forward: count, max, min, max_by etc, sum, product

andrei-papou:
Fixes #555

Forwarded the following `Iter` and `IterMut` methods to the `inner` iterator:
- `nth`
- `collect`
- `all`
- `any`
- `find`
- `find_map`


andrei-papou:
@jturner314  I've removed `Self` bounds and forwarded all the methods you've mentioned except `rposition`. I'm not sure if we can forward it since it requires `DoubleEndedIterator` bound on `Self`, but `DoubleEndedIterator` is only implemented for `Iter<'a, A, Ix1>`. I've tried the following solutions:
1. Add more specific `Iterator` impl for `Iter<'a, A, Ix1>`. It does not work because this impl overlaps with the general one for `Iter<'a, A, D>`.
2. Add special marker trait (for example `FlatDimension`), impl it only for `Ix1` and add a bound `D: FlatDimension` to the `rposition` where clause. This also does not work because Rust does not allow method bound to be stricter than trait bound.

PR #615, opened 2019-04-10
Implement Zip::all()

mneumann:
Test if every element of the iterator matches a predicate.

This is useful, for instance, to test if two matrices are elementwise
"similar" (within a given threshold) without modifying the matrices.false

### DIFF
--- a/src/zip/mod.rs
+++ b/src/zip/mod.rs
@@ -738,6 +738,22 @@ macro_rules! map_impl {
                 })
             }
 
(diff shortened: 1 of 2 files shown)


jturner314:
Once this is merged, I'll rewrite #581 in terms of `.all()`.

LukeMathWalker:
Happy to merge this @jturner314 :+1:

jturner314:
@LukeMathWalker Sure, please go ahead.

Issue #532, opened 2018-11-07
Implement ascontiguousarray() or contiguous() method


Rikorose:
See https://docs.scipy.org/doc/numpy/reference/generated/numpy.ascontiguousarray.html or https://pytorch.org/docs/stable/tensors.html#torch.Tensor.contiguous

A method like this is e.g. needed when passing &ndarray.as_slice() to an external api.

andrei-papou:
@LukeMathWalker could you please take a look at the [PR](https://github.com/rust-ndarray/ndarray/pull/616)? Thanks.

Issue #390, opened 2017-11-30
Resolve into_shape() limitations


jturner314:
The [`.into_shape()`](https://docs.rs/ndarray/0.10.13/ndarray/struct.ArrayBase.html#method.into_shape) method works in some cases, but compared to NumPy's [`reshape()`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.reshape.html#numpy.reshape) has the following limitations:

1. It only works for arrays with contiguous memory layout.
2. The result depends on the layout of the data in memory. (Calling `.into_shape()` on an array with Fortran memory layout will give a different result from calling `.into_shape()` on the logically equivalent array with C memory layout.)
3. All of the new axis lengths need to be known. (NumPy has the nice ability to use `-1` to indicate that an axis length should be inferred.)

I'd like to add a method or methods to resolve at least limitations 1 and 2.

In order to handle arrays without contiguous memory layout, the data needs to be cloned in some cases. If the input array owns its data, this is simple – just return a new owned array. However, if the input array is an `ArrayView`, then you need to return something like `Cow`. If the input array is an `ArrayViewMut`, you need some way to return a mutable view with the correct shape but make sure that any changes to that view are copied back to the original underlying data.

A prototype of one approach is at jturner314/rust-ndarray@0157df867bc7638a42cad60cc7b1463edf31a722. This introduces a few new types, including `ArrayViewTemp` which handles the temporary array if the input is an `ArrayView`, and `ArrayViewMutTemp`, which handles the case where the input is an `ArrayViewMut` (using `Drop` to make sure the data is written from the temporary array back to the original underlying data).

It's worth noting that this type of problem (needing a temporary array in some cases) also occurs for operations other than reshaping. For example, I'm working on a similar approach for `ndarray-linalg` to convert arrays/views to Fortran layout to pass to LAPACK. If something like `ArrayViewTemp` and `ArrayViewMutTemp` get added to `ndarray`, I can reuse those types in `ndarray-linalg` for the changing-layout problem.

What do you think?

Edit: After thinking about this a little more, I realized:

1. Instead of wrapping arrays in `ArrayViewTemp` and `ArrayViewMutTemp` enums, it would be cleaner to create new reprs for the `S` type parameter in `ArrayBase` that implement `Data`/`DataMut`.
2. It might be possible for `drop_hook` in `ArrayViewMutTempRepr` to own the view of the original data so that the `view` field and `Do` type parameter could be removed.

[...]

PR #616, opened 2019-04-12
As standard layout method

andrei-papou:
PR for https://github.com/rust-ndarray/ndarray/issues/532

Added `as_contiguous` method to `ArrayBase` which accepts array by reference returns a copy of it in standard layout.


LukeMathWalker:
This is quite needed to simplify the story around manipulation of strides, nice.
It should also be fairly easy to approach with some property-based testing using `quickcheck`, asserting that a random array returns true when probed using `is_standard_layout`.
We should check these classes of arrays:
- the weird ones (0 dimensional, multidimensional with some 0-length axes, empty multidimensional arrays);
- C and F layout for random multidimensional arrays;
- non-contiguous multidimensional arrays (as you are doing already in the existing unit test).

Another important piece is guarantees, to be added in the method documentation: what happens if I pass in an array that is already owned and arranged with standard layout? Do you still create a copy or do you return the same array? 
Any guarantee we promise should be tested, to check for regressions.

jturner314:
I think this functionality is important, but I'd like to refine the return type.

> what happens if I pass in an array that is already owned and arranged with standard layout? Do you still create a copy or do you return the same array?

Based on the current signature of the method (`&ArrayBase -> Array`), the only option is to copy it, which is unfortunate. As a result, I'd prefer to change the method to

```rust
fn as_standard_layout(&self) -> ArrayCow<'_, A, D>
```

to avoid copying the data when the array is already in the correct layout.

`ArrayCow<'a, A, D>` would be a new type for `as_standard_layout` and similar methods analogous to `Cow` from the standard library. We'll need to add a new `CowRepr` type and implement the appropriate `*Data*` traits for it (see `data_traits.rs`). See #390 for some related discussion. I can provide more guidance if that would be helpful.

andrei-papou:
I've reworked the implementation to use the copy-on-write strategy. It is still a draft, but could you please take a look and tell me whether I am on a right track? Thanks a lot.

jturner314:
Yes, it looks like you're on the right track. I would change the wording a little (rename `Temp` to `Owned` and `.as_contiguous()` to `.as_standard_layout()`), but otherwise it looks good on first glance.

tests/array.rs
```diff
@@ -1891,3 +1891,66 @@ fn array_macros() {
     let empty2: Array<f32, Ix2> = array![[]];
     assert_eq!(empty2, array![[]]);
 }
+
+#[test]
+fn test_as_contiguous() {
```
LukeMathWalker: Can you break this in separate test functions? It makes it easier to see what is broken (i.e. you run all the tests instead of stopping at the first failing one).

andrei-papou: I've split that large test function into more granular test functions under new module.

andrei-papou:
Seems like I should also write some test for the new `ArrayCow` type. Should I test only its custom methods? Or maybe also `*Data*` trait implementations? I need a bit of guidance here. Thanks.

jturner314:
I created andrei-papou/ndarray#4 with some refinements.

I'd like to test the following for `ArrayCow` (for both the `View` and `Owned` variants) with non-standard layout (permuted axes as well as axis inversions and steps other than 1):

* Mutate an element. For the `View` variant, make sure that the array changes to the `Owned` variant and that the element values are correct. For the `Owned` variant, make sure that `array.as_ptr()` is the same before and after (which indicates that the data was not cloned), and make sure that the element values are correct.

* Call `.clone()` on the array. Make sure that the data variant (`View` vs. `Owned`), element values, shape, and strides are the same in the original and the clone.

* Use `.clone_from()` to clone from one `ArrayCow` to another, both when they initially have the same variant and when the variants are initially different. Make sure that the data variant (`View` vs. `Owned`), element values, shape, and strides in the destination array match the source.

* Call `.into_owned()`. For the `View` variant, make sure that the array changes to the `Owned` variant and that the element values are correct. For the `Owned` variant, make sure that `array.as_ptr()` is the same before and after (which indicates that the data was not cloned), and make sure that the element values are correct.

These tests should be sufficient to check the behavior of `CowRepr`.

Other than tests, the only remaining piece is documentation (normal doc comments as well as mentioning `ArrayCow` in the `ArrayBase` and `ndarray_for_numpy_users` documents).

Would you be willing to move the `ArrayCow` stuff into a separate PR?

andrei-papou:
@jturner314 Thanks for your refinements and guidance. As you suggested, I've moved all the `ArrayCow` stuff to the separate PR https://github.com/rust-ndarray/ndarray/pull/632 and added some tests there.

src/impl_methods.rs
```diff
@@ -14,14 +14,14 @@ use itertools::{izip, zip};
...
 
-use crate::arraytraits;
 use crate::dimension;
 use crate::dimension::IntoDimension;
 use crate::dimension::{
     abs_index, axes_of, do_slice, merge_axes, size_of_shape_checked, stride_offset, Axes,
 };
 use crate::error::{self, ErrorKind, ShapeError};
 use crate::zip::Zip;
+use crate::{arraytraits, CowArray};
```
jturner314: This change isn't necessary. `CowArray` is included in `imp_prelude`.

src/impl_methods.rs
```diff
@@ -1218,6 +1218,23 @@ where
         D::is_contiguous(&self.dim, &self.strides)
     }
 
+    pub fn as_standard_layout(&self) -> CowArray<'_, A, D>
```
jturner314: This method needs documentation.

src/impl_methods.rs
```diff
@@ -1218,6 +1218,23 @@ where
...
+        S: Data<Elem = A>,
+        A: Clone,
+    {
+        if self.is_standard_layout() {
+            CowArray::from(self.view())
+        } else {
+            let v = self.iter().cloned().collect::<Vec<A>>();
+            let owned_array: Array<A, D> = unsafe {
+                // Safe because we use shape and content of existing array here.
+                ArrayBase::from_shape_vec_unchecked(self.raw_dim(), v)
```
jturner314: This makes me a bit uncomfortable, especially because I don't know if in the future it will be possible for users to specialize the implementation of `FromIterator<T> for Vec<T>`. Let's add an assertion that the length of `v` is correct:

```rust
let v: Vec<A> = self.iter().cloned().collect();
let dim = self.dim.clone();
assert_eq!(v.len(), dim.size());
let owned_array: Array<A, D> = unsafe {
    // Safe because the shape and element type are from the existing array
    // and the strides are the default strides.
    Array::from_shape_vec_unchecked(dim, v)
};
CowArray::from(owned_array)
```

tests/array.rs
```diff
@@ -1970,6 +1970,77 @@ fn array_macros() {
     assert_eq!(empty2, array![[]]);
 }
 
+#[cfg(test)]
+mod as_standard_layout_tests {
```
jturner314: A lot of duplicate code can be removed from these tests with the following function:

```rust
fn test_as_standard_layout_for<S, D>(orig: ArrayBase<S, D>)
where
    S: Data,
    S::Elem: Clone + Debug + PartialEq,
    D: Dimension,
{
    let orig_is_standard = orig.is_standard_layout();
    let out = orig.as_standard_layout();
    assert!(out.is_standard_layout());
    assert_eq!(out, orig);
    assert_eq!(orig_is_standard, out.is_view());
}
```

Then, the body of each test is just calling `test_as_standard_layout_for` with an array.

tests/array.rs
```diff
@@ -1970,6 +1970,77 @@ fn array_macros() {
...
+    use super::*;
+
+    #[test]
+    fn test_f_layout() {
+        let shape = Ix2(2, 2).strides(Ix2(1, 2));
```
jturner314: Fwiw, this can be written `(2, 2).strides((1, 2))` or, since it's Fortran layout in this case, `(2, 2).f()`.

andrei-papou:
@jturner314 I've updated the PR according to your comments. Any more suggestions on this?

jturner314:
I added a little more to the docs and merged the the PR. Thanks for working on this and the previous PR! This is very useful functionality that's been necessary for quite a while.

PR #617, opened 2019-04-12
Add `logspace` constructor

JP-Ellis:
This pull requests adds the `logspace` constructor for Array1 that works in the same way as `linspace` but creates logarithmically spaced floats.

This implementation supports both positive and negative intervals, though
(obviously) not intervals which cross `0.0`.

### DIFF
--- a/src/impl_constructors.rs
+++ b/src/impl_constructors.rs
@@ -112,11 +112,9 @@ impl<S, A> ArrayBase<S, Ix1>
     /// use ndarray::{Array, arr1};
     ///
     /// let array = Array::logspace(1e0, 1e3, 4);
-    /// println!("{}", array);
     /// assert!(array.all_close(&arr1(&[1e0, 1e1, 1e2, 1e3]), 1e-5));
     ///
     /// let array = Array::logspace(-1e3, -1e0, 4);
-    /// println!("{}", array);
     /// assert!(array.all_close(&arr1(&[-1e3, -1e2, -1e1, -1e0]), 1e-5));
     /// ```
     pub fn logspace(start: A, end: A, n: usize) -> Self
--- a/src/impl_constructors.rs
+++ b/src/impl_constructors.rs
@@ -10,18 +10,18 @@
 //!
 //!
 
-use num_traits::{Zero, One, Float};
+use num_traits::{Float, One, Zero};
 use std::isize;
 use std::mem;
 
-use crate::imp_prelude::*;
-use crate::StrideShape;
 use crate::dimension;
-use crate::linspace;
 use crate::error::{self, ShapeError};
-use crate::indices;
+use crate::imp_prelude::*;
 use crate::indexes;
+use crate::indices;
 use crate::iterators::{to_vec, to_vec_mapped};
+use crate::StrideShape;
+use crate::{linspace, logspace};
 
 /// # Constructor Methods for Owned Arrays
 ///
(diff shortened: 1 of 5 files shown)


LukeMathWalker:
Thanks for working on this @JP-Ellis!
It definitely looks like a piece of functionality that we were missing.

To move this forward it would be nice to have tests for these new functions, testing both happy and unhappy paths :+1:

JP-Ellis:
@LukeMathWalker All done, I have added tests within a `tests` submodule of the new `logspace` module.

src/impl_constructors.rs
```diff
@@ -101,6 +101,30 @@ impl<S, A> ArrayBase<S, Ix1>
...
+    /// end]` with `n` elements logarithmically spaced. `A` must be a floating
+    /// point type.
+    ///
+    /// **Panics** if `n` is greater than `isize::MAX`.
+    ///
+    /// ```rust
+    /// use ndarray::{Array, arr1};
+    ///
+    /// let array = Array::logspace(1e0, 1e3, 4);
+    /// println!("{}", array);
```
LukeMathWalker: I think we should remove the println lines (both here and below) from the doctest - you are showing what is inside the array with the assertion.

src/logspace.rs
```diff
@@ -0,0 +1,168 @@
...
+    use super::logspace;
+    use crate::{arr1, Array1};
+
+    #[test]
+    fn valid() {
+        let array: Array1<_> = logspace(1e0, 1e3, 4).collect();
+        assert!(array.all_close(&arr1(&[1e0, 1e1, 1e2, 1e3]), 1e-5));
+
+        let array: Array1<_> = logspace(-1e3, -1e0, 4).collect();
+        assert!(array.all_close(&arr1(&[-1e3, -1e2, -1e1, -1e0]), 1e-5));
```
LukeMathWalker: Another useful case to add here is start < end I'd say.

jturner314:
I haven't had a chance to review this in detail, but at first glance, I had a couple of thoughts:

* We should name this `geomspace` instead of `logspace` to be consistent with NumPy and Matlab. ([`logspace` in NumPy](https://docs.scipy.org/doc/numpy/reference/generated/numpy.logspace.html#numpy.logspace) and [Matlab](https://www.mathworks.com/help/matlab/ref/logspace.html) takes the exponent bounds as parameters, while [`geomspace` in NumPy](https://docs.scipy.org/doc/numpy/reference/generated/numpy.geomspace.html#numpy.geomspace) takes the actual value bounds as parameters, like this PR.)

* I'm interested that the implementation of this PR multiplies by a step for each element. In contrast, NumPy's `geomspace` is [implemented in terms of `logspace`](https://github.com/numpy/numpy/blob/v1.16.1/numpy/core/function_base.py#L288-L427), which is [implemented by using `linspace` as an exponent](https://github.com/numpy/numpy/blob/v1.16.1/numpy/core/function_base.py#L187-L280). I would guess that directly computing `base ^ (log_start + log_step * i)` for each element (like NumPy) would be more accurate than `(base ^ log_start) * step * step * ⋯ * step` (this PR) due to accumulated errors, but I'm not really sure. There's also some advantage to maintaining consistency with NumPy, but that's not necessarily important. Any thoughts on this?

LukeMathWalker:
> * We should name this `geomspace` instead of `logspace` to be consistent with NumPy and Matlab. ([`logspace` in NumPy](https://docs.scipy.org/doc/numpy/reference/generated/numpy.logspace.html#numpy.logspace) and [Matlab](https://www.mathworks.com/help/matlab/ref/logspace.html) takes the exponent bounds as parameters, while [`geomspace` in NumPy](https://docs.scipy.org/doc/numpy/reference/generated/numpy.geomspace.html#numpy.geomspace) takes the actual value bounds as parameters, like this PR.)

If we decide to keep compatibility with their naming convention, then I'd advice to add a `logspace`-like method to our API as well. `geomspace` picks its name from geometric progression, but for searchability purposes I think most people looking for such a functionality (and I put myself in there) would expect something named after a logarithm, like `logspace`. If we provide both, I think we should be fine.

> * I'm interested that the implementation of this PR multiplies by a step for each element. In contrast, NumPy's `geomspace` is [implemented in terms of `logspace`](https://github.com/numpy/numpy/blob/v1.16.1/numpy/core/function_base.py#L288-L427), which is [implemented by using `linspace` as an exponent](https://github.com/numpy/numpy/blob/v1.16.1/numpy/core/function_base.py#L187-L280). I would guess that directly computing `base ^ (log_start + log_step * i)` for each element (like NumPy) would be more accurate than `(base ^ log_start) * step * step * ⋯ * step` (this PR) due to accumulated errors, but I'm not really sure. There's also some advantage to maintaining consistency with NumPy, but that's not necessarily important. Any thoughts on this?

I have the same gut feeling that `base ^ (log_start + log_step * i)` would lead to a more precise estimate, but no theoretical/practical example to back it up. We could run a quick check comparing errors for an (long) exact sequence.

JP-Ellis:
Great, so I've renamed the `logspace` function to `geomspace`, and I've added a new `logspace` function to be in line with NumPy and Matlab.

Additionally, I've changed the way each step is calculated to avoid a possible small error from accumulating.

Before merging this, I would recommend that this pull request be squashed into one commit.

LukeMathWalker:
It looks good to me - happy to squash and merge :rocket:

PR #619, opened 2019-04-19
Add link to #rust-sci in README

jturner314:
It's helpful to have a place for users to chat. The link is to #rust-sci in the Mibbit client (the same client used for the "Mozilla IRC" link on https://www.rust-lang.org/community). Mibbit is nice because it provides a way for people to join the chat using their browser without even having to register.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -12,15 +12,19 @@ __ https://docs.rs/ndarray/
 |build_status|_ |crates|_ |matrix-chat|_ |irc|_
 
 .. |build_status| image:: https://github.com/rust-ndarray/ndarray/workflows/Continuous%20integration/badge.svg?branch=master
+   :alt: CI build status
 .. _build_status: https://github.com/rust-ndarray/ndarray/actions
 
 .. |crates| image:: http://meritbadge.herokuapp.com/ndarray
+   :alt: ndarray at crates.io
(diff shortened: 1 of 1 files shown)


jturner314:
The Mozilla IRC server is [shutting down in the near future](http://exple.tive.org/blarg/2019/04/26/synchronous-text/), so we'll need to choose something other than #rust-sci on irc.mozilla.org. It looks like #rust-sci is migrating to oftc.net ([link to web chat](https://webchat.oftc.net?channels=rust-sci)), but let's wait a little while for things to stabilize before adding it to the README.

ZuseZ4:
@bluss Shall we merge this?
It's not a breaking change, but I see no reason to keep it open, since it's just a doc change.

bluss:
That channel has moved to matrix entirely, so I guess we'd like to link to matrix.

jturner314:
I've changed this PR to provide a link to the Matrix room and the bridged IRC room on OFTC. (It's useful to provide a link to the IRC web chat so that users who don't want to go to the effort of creating a Matrix account can still ask a quick question.) [Click here to preview how GitHub renders the proposed README.](https://github.com/jturner314/ndarray/tree/irc-link)

bluss (approved):

Issue #622, opened 2019-04-24
[de-feature request] tuples for const-defined dimensions


swfsql:
Hello, I'm new to this ecosystem here- and thanks a lot for your work on it!
As I was using the `ndarray` crate - it may be just me but - I thought that "what if we always use a tuple to indicate a shape's dimension?".

I mean, this is already how it works most of the time, except for the `Ix1` and the dynamic case.
That is, the only case, for compile-time defined number of dimensions, that doesn't fall into what I said above was the `Ix1` case, where a single `usize` is used as a `Pattern` to indicate a 1-dimensional array.

I propose a de-featuring (feature removal) so that `Ix1` case would also require a 1-sized tuple, similarly to the `Ix2` (`Ix3`, etc) cases.

This would simply increase verbosity whenever a `Ix1` were to be used, so.. well.. maybe this [proposal] would be taking a step into the wrong direction.
(I'll be PRing for a reference since this would require simple changes after all).

jturner314:
Hi @swfsql, and welcome to the ecosystem!

Will you please provide more information about the motivation for doing this? I generally find it convenient to use `5` instead of `(5,)` and haven't experienced any problems with this. If you want to use `(usize,)` as the type instead of `usize`, that's supported too.

swfsql:
Thanks @jturner314 ,
I initially got motivated when I called `.dim()` I expected a `(usize,)` return type - as an inexperienced user - since I was used to understand tuple length as dimension quantity after using the crate for a while.

This specific change - the return type of `dim()` - doesn't implies the additional change about building dimension information - about using `usize` or `(usize,)` for dimension construction - but I thought that for symmetry, if one change happened, the other should follow.

I hope this helps clarifying - and I'd like to restate that this is a pretty subjective ergonomic opinion from an inexperienced user!

jturner314:
Oh, okay, you're talking about `Ix1::Pattern`. I misunderstood. (I should have read the PR.)

Now that Rust supports destructuring fixed-size arrays in addition to tuples, I actually think we should change all the `Dimension::Pattern` types for const-dimensions from tuples to fixed-size arrays, so we'd have

* `Ix0::Pattern = [usize; 0]`
* `Ix1::Pattern = [usize; 1]`
* `Ix2::Pattern = [usize; 2]`
* ...
* `Ix6::Pattern = [usize; 6]`
* `IxDyn::Pattern = IxDyn`

[...]

PR #623, opened 2019-04-24
Remove usize as dimension pattern

swfsql:
Removes the feature that an usize may be used as a pattern for a one-dimensional shape.

- Closes https://github.com/rust-ndarray/ndarray/issues/622
- The documentation wasn't extensively reviewed.


swfsql:
Closed as a shift into arrays (as mentioned in the issue by jturner314) makes a better design

PR #608, opened 2019-03-25
Run rustfmt and add to travis

cafeclimber:
Thought this was pretty low hanging fruit, but results in a...nasty looking commit (#597)

PR #628, opened 2019-05-04
Add Code Coverage

JP-Ellis:
This pull request's primary goal is to add code coverage, which can be seen at [codecov](https://codecov.io/gh/JP-Ellis/ndarray) and [coveralls](https://coveralls.io/github/JP-Ellis/ndarray) (note that former only displays stats from the `master` branch on the front page, but you can see stats from commits at https://codecov.io/gh/JP-Ellis/ndarray/commits).

In addition, this pull requests also adds `cargo clippy` and `cargo fmt`.  Failure to pass these does *not* cause the build to break, but serves as a warning.

I also took the slight liberty to change some of the features relating to testing.  Nightly features are kept behind the `nightly` flag (which has no other dependencies), and `cargo test` automatically enables the `cfg(test)` flag making the `tst` feature somewhat redundant.  Tests which are specific to certain features (such as `blas`) can be enabled by invoking `cargo` with the `--feature blas` flag.

Lastly, the first build does take a long time due to the `cargo travis` installation; however, I have enabled caching with travis so that subsequent runs can complete substantially faster.

EDIT:  Forgot to mention that I've added the flags `-D warnings` and `-D missing_docs` which causes the build to fail if there are any warnings or missing documentation on public facing elements.  Fortunately, there were very few such warnings which I've fixed (albeit temporarily in the case of the documentation).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,13 +2,16 @@ language: rust
 
 sudo: required
 
-rust:
-  - 1.32.0
-  - stable
-  - beta
-  - nightly
-
 matrix:
+  include:
+    - rust: 1.32.0
+      env: FEATURES="blas serde-1 docs"
+    - rust: stable
+      env: FEATURES="blas serde-1 docs"
+    - rust: beta
+      env: FEATURES="blas serde-1 docs"
+    - rust: nightly
+      env: FEATURES="blas serde-1 docs nightly"
   allow_failures:
     - rust: beta
     - rust: nightly
@@ -16,7 +19,6 @@ matrix:
 env:
   global:
     - RUSTFLAGS="-D warnings"
-    - FEATURES="blas serde-1 docs"
 
 cache: cargo
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,34 +1,45 @@
 language: rust
-# use trusty for newer openblas
+
 sudo: required
-dist: trusty
+
+rust:
+  - 1.32.0
+  - stable
+  - beta
+  - nightly
+
 matrix:
-  include:
-    - rust: 1.32.0
-      env:
-       - FEATURES='test docs'
-       - RUSTFLAGS='-D warnings'
-    - rust: stable
-      env:
-       - FEATURES='test docs'
-       - RUSTFLAGS='-D warnings'
+  allow_failures:
     - rust: beta
-      env:
-       - FEATURES='test docs'
-       - CHANNEL='beta'
-       - RUSTFLAGS='-D warnings'
     - rust: nightly
-      env:
-       - FEATURES='test docs'
-       - CHANNEL='nightly'
+
 env:
   global:
-    - HOST=x86_64-unknown-linux-gnu
-    - CARGO_INCREMENTAL=0
+    - RUSTFLAGS="-D warnings"
+    - FEATURES="blas serde-1 docs"
+
+cache: cargo
+
 addons:
   apt:
+    update: true
     packages:
       - libopenblas-dev
       - gfortran
+      - libcurl4-openssl-dev
+      - libelf-dev
+      - libdw-dev
+      - cmake
+      - gcc
+      - binutils-dev
+      - libiberty-dev
+      - libgsl0-dev
+
+before_script:
+  - ./ci/before_script.sh
+
 script:
-  - ./scripts/all-tests.sh "$FEATURES" "$CHANNEL"
+  - ./ci/script.sh
+
+after_success:
+  - ./ci/after_success.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,17 @@ addons:
     packages:
       - libopenblas-dev
       - gfortran
-      - kcov
+      # kcov dependencies
+      - cmake
+      - g++
+      - pkg-config
+      - jq
+      - libcurl4-openssl-dev
+      - libelf-dev
+      - libdw-dev
+      - binutils-dev
+      - libiberty-dev
+
 
 before_script:
   - ./ci/before_script.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,14 +28,7 @@ addons:
     packages:
       - libopenblas-dev
       - gfortran
-      - libcurl4-openssl-dev
-      - libelf-dev
-      - libdw-dev
-      - cmake
-      - gcc
-      - binutils-dev
-      - libiberty-dev
-      - libgsl0-dev
+      - kcov
 
 before_script:
   - ./ci/before_script.sh
--- a/build.rs
+++ b/build.rs
@@ -1,10 +1,8 @@
-///
-/// This build script emits the openblas linking directive if requested
-///
+//! This build script emits the openblas linking directive if requested
 
 fn main() {
     println!("cargo:rerun-if-changed=build.rs");
-    if cfg!(feature = "test-blas-openblas-sys") {
+    if cfg!(feature = "blas") {
         println!("cargo:rustc-link-lib={}=openblas", "dylib");
     }
 }
--- a/ci/after_success.sh
+++ b/ci/after_success.sh
@@ -7,22 +7,14 @@ set -o nounset
 # Exit on any error
 set -o errexit
 
-KCOV_BIN="./target/kcov-master/build/src/kcov"
-
 run_kcov() {
     # Run kcov on all the test suites
-    cargo coverage --all --tests --benches --no-default-features
-    mv target/kcov target/kcov-no-default-features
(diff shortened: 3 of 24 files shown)


LukeMathWalker:
Nice!
A couple of questions:
- there is an open PR for `cargo fmt` (#608) and I'd like to keep the same spirit: if the formatting check fails, I think we should fail the build.
- I have seen you have added in a lot of "TODO: document" where I assume `clippy` was complaining about missing docs. Wouldn`t it be better to keep those missing docs warnings, so that we actually go and fix those things? :stuck_out_tongue:

JP-Ellis:
* Regarding `cargo fmt`, I completely agree that if formatting fails the build should fail.  I'm happy to wait for #628 to be merged first and then this pull request will add code coverage.
* I forgot the mention in the original post, but I have also added the flags `-D warnings` and `-D missing_docs` which causes compilation to fail if there are any warnings or undocumented (public) functions/structs/traits.  Since the code coverage is only executed after everything else succeeds, I had to add `/// TODO: Document` in various places as a temporary fix so that CI can pass and run the code coverage.  

  If you would like, I'm happy to add a commit reverting these changes just before we merge this.
* `clippy` is set at the moment to *not* cause the build the fail.  This is because there are certain (rare) cases where `clippy` gets it wrong, but one particularly relevant to `ndarray` is the use of the slice macro `s![...]`.

LukeMathWalker:
> * I forgot the mention in the original post, but I have also added the flags `-D warnings` and `-D missing_docs` which causes compilation to fail if there are any warnings or undocumented (public) functions/structs/traits.  Since the code coverage is only executed after everything else succeeds, I had to add `/// TODO: Document` in various places as a temporary fix so that CI can pass and run the code coverage.
>   If you would like, I'm happy to add a commit reverting these changes just before we merge this.

I'd prefer to see those reverted for now - we can open another PR to properly fill in the missing gaps and then start failing builds if docs are missing. 
I have merged #635, so if you can re-run the formatter we should be ready to go.

JP-Ellis:
The `-D warnings` flag was easy to fix as there was only one issue (the use of a deprecated feature in one of the examples or benches).  So if that's alright, I'll leave that in the pull request.

I'll disable `-D missing_docs` for now and open up another issue regarding `missing_docs`.

LukeMathWalker:
I'd like to go ahead and merge this - do you want to have a look first @jturner314 @termoshtt?

ci/after_success.sh
```diff
@@ -0,0 +1,60 @@
...
+
+# We only need to run the coverage suite once
+COVERAGE_RUN=false
+KCOV_BIN="./target/kcov-master/build/src/kcov"
+
+run_kcov() {
+    # Run kcov on all the test suites
+    if [[ $COVERAGE_RUN != "true" ]]; then
+        cargo coverage --all --tests --benches --no-default-features
+        mv target/kcov target/kcov-no-default-features
```
jturner314: Do we need to move the coverage results like this? The docs for the `cargo coverage` command seem to indicate that it automatically merges all results together into `target/kcov` when run multiple times, but I may be misinterpreting the docs.

ci/after_success.sh
```diff
@@ -0,0 +1,60 @@
...
+        cargo coverage --all --tests --benches --features "$FEATURES"
+        mv target/kcov target/kcov-features
```
jturner314: The commands above don't run all the tests. (They miss the ones in `serialization-tests`, `blas-tests`, and `numeric-tests`.)

ci/script.sh
```diff
@@ -0,0 +1,34 @@
...
+
+# Ensure the code is correctly formatted.
+check_format() {
+    cargo fmt --all -- --check
+}
+
+# Run the test suite.
+check_tests() {
+    cargo test --all --examples --tests --benches --no-default-features
+    cargo test --all --examples --tests --benches --features "$FEATURES"
```
jturner314: IIUC, the commands above don't run all the tests. (They miss the ones in `serialization-tests`, `blas-tests`, and `numeric-tests`.)

As I see it, we have two options:

1. Keep the existing `serialization-tests`, `blas-tests`, and `numeric-tests` crates and run their tests in the scripts.

2. Merge them into the `ndarray` crate, adding the necessary optional dependencies and creating the necessary test-specific features in `ndarray`, similar to how I'm suggesting to handle the `test-blas-openblas-src` feature.

I think option 2 is a bit cleaner, but either is fine with me. Either way, we need to make sure that we run these tests.

max-sixty:
@jd-ellis I'm happy to help on integrating the test workspaces, if you can find some minimal version of this to merge in

.travis.yml
```diff
@@ -1,38 +1,44 @@
 language: rust
-# use trusty for newer openblas
+
 sudo: required
-dist: trusty
+
+rust:
```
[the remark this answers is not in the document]
JP-Ellis: This is going to be a little annoying to do as some of the code checking dependencies require a newer version.  Would it be acceptable if code coverage is disable on 1.31.0?

ci/after_success.sh
```diff
@@ -0,0 +1,60 @@
...
+        return
+    fi
+
+    run_kcov
+
+    bash <(curl -s https://codecov.io/bash) -s target/kcov
+    echo "Uploaded code coverage to codecov.io"
+}
+
+coverage_coveralls() {
```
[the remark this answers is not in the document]
JP-Ellis: I had both as it isn't anymore complicated, but I can certainly just have one.

ci/script.sh
JP-Ellis: I created an issue about specifically this in #637.  I agree that keeping these scripts separate isn't nice and integrating them I think would be the best option.

parallel/Cargo.toml
```diff
@@ -13,10 +13,11 @@ keywords = ["data-structure", "multidimensional", "parallel", "concurrent"]
 categories = ["data-structures", "science", "concurrency"]
 
 [dependencies]
-rayon = { version = "1.0" }
-ndarray = { version = "0.12.0", path = "../" }
+rayon = "1.0"
+ndarray = { version = "0.12.0", path = "../", features = ["rayon"]}
```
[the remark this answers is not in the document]
JP-Ellis: If I remember correctly, it wasn't running the tests correctly without the rayon feature being enabled (but it was a little while ago that I did this).

.travis.yml
```diff
@@ -1,38 +1,44 @@
...
+  - stable
+  - beta
+  - nightly
```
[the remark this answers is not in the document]
JP-Ellis: Yes, that is true.  I thought I was doing that, but I'll double check.

build.rs
```diff
@@ -1,10 +1,8 @@
-///
-/// This build script emits the openblas linking directive if requested
-///
+//! This build script emits the openblas linking directive if requested
 
 fn main() {
     println!("cargo:rerun-if-changed=build.rs");
-    if cfg!(feature = "test-blas-openblas-sys") {
+    if cfg!(feature = "blas") {
```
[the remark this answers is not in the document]
JP-Ellis: Ok, I was misunderstanding how this works (and I'm not sure I fully understand what you would like).  My reasoning was that if seems weird to have a feature called `test-foo`.  If anything, should we not have a `foo` feature and then run these tests of `cfg!(test) && cfg!(feature = "foo")`?

ci/after_success.sh
JP-Ellis: That's odd; the `--all` flag should run all tests from all workspaces.  Could this be left as-is and once this pull request in merged, I'll work on #637?

ci/after_success.sh
JP-Ellis: Unfortunately yes.  `cargo coverage` merges results from one run, but not from several different runs with different features unfortunately.

JP-Ellis:
I've rebased this pull request on top of master.  I've also incorporated some of the changes mentioned above, including:
- Added a missing `nightly` features on the nightly build
- Only build `cargo-travis` on stable, and only run code coverage on stable.
- Remove part of the code relating to coveralls.

Let me know if there's anything else that should be added?

ci/after_success.sh
bluss: numeric-tests is explicitly not in the workspace, so that it can use its own profile/compilation settings. It can be added if that can be achieved another way

JP-Ellis: I don't see really see anything special about `numeric-tests` which can't be included as part of #637.  If it is because of a specific combination of features, then we can easily add them by adding to the Travis-CI test matrix.  If there's something more subtle, perhaps this can be addressed in #637?

PR #629, opened 2019-05-04
Bump some dependencies

Eijebong:
This helps avoiding quite a few duplicates in ndarray and update itertools to avoid duplicates in crates depending on ndarray


LukeMathWalker:
Good hygiene - thanks for opening a PR!

PR #630, opened 2019-05-05
Improve linspace, range, logspace, geomspace

jturner314:
See the individual commit messages for details.

The only visible change to the API is that `geomspace` now returns `None` instead of panicking in the case that the `start` and `end` have different signs or are zero. This helps to make the user more aware of this constraint and allows the user to recover in the case of failure.


LukeMathWalker:
It looks good - returning `None` is better than panicking. We should probably review the whole crate and see where we could improve error handling.

jturner314:
> It looks good - returning `None` is better than panicking.

Thanks for reviewing it.

> We should probably review the whole crate and see where we could improve error handling.

Yeah, that would be a good idea.

PR #631, opened 2019-05-06
Use approx for approximate comparisons in tests

jturner314:
This allows us to get rid of the custom comparison functions.


LukeMathWalker:
It looks good to me.

jturner314:
Thanks for reviewing it.

PR #632, opened 2019-05-06
CowArray feature

andrei-papou:


tests/array.rs
```diff
@@ -1891,3 +1891,119 @@ fn array_macros() {
...
+    fn test_is_view() {
+        let arr: Array<i32, Ix2> = array![[1, 2], [3, 4]];
+        let arr_cow = ArrayCow::<i32, Ix2>::from(arr.view());
+        assert!(arr_cow.is_view());
+        let arr_cow = ArrayCow::<i32, Ix2>::from(arr.clone());
+        assert!(!arr_cow.is_view());
+    }
+
+    #[test]
+    fn test_is_owned() {
```
nilgoyette: It's only a suggestion, but I would merge `test_is_view` and `test_is_owned` because one is simply the reverse of the other,

andrei-papou:
@nilgoyette thanks for your notes, I've updated the PR.

andrei-papou:
@jturner314 any feedback on this? Once it's merged, I will be able to update https://github.com/rust-ndarray/ndarray/pull/616 .

jturner314:
I'm sorry for the delay. Life has been busy recently. I've created andrei-papou#7 with the following:

* Apply the tests to arrays with non-standard layouts.
* Add/improve documentation.
* Minor implementation changes.
* Rename `ArrayCow` to `CowArray`. The name `CowArray` seems a little more natural to me, and it seems to fit better with `ArcArray`/`RawArrayView`/`RawArrayViewMut`. It's also nice that alphabetic sorting moves `CowArray` after e.g. `ArrayView`.

Once that's merged, I think we're good-to-go on this PR.

jturner314:
Yay! I've been wanting this feature for quite a while. Thanks for working on it!

SuperFluffy:
Excellent, this brings me one step closer to being able to pass in slices containing arrays, like `&[&ArrayBase<S, D>], where S: Data, D: Dim`.

I believe that right now, due to monomorphisation, I wouldn't actually be able to pass in a mixture of `Array` and `ArrayView`, e.g. `&[&Array2<f64>, &ArrayView2<f64>, &Array2<f64>]`.

I have to play around with it, but I don't think this feature will completely fix this either, because the trait bound `fn f<T: Into<CowArray>>(myarrays: &[&T])` won't work for mixed `T`, again due to monomorphization.

Might a `Deref<Target=ArrayView> for Array` help?

andrei-papou:
@SuperFluffy I'm not sure `&[&Array2<f64>, &ArrayView2<f64>, &Array2<f64>]` will work with `Deref<Target=ArrayView> for Array` because you still need all the members of a slice to be of the same type. `&Array2<f64>` and `&ArrayView2<f64>` are completely different types from the compiler's point of view.
IMHO what you need in your case is dynamic dispatch, but it is not available for `ArrayBase`. You cannot write `ArrayBase<dyn Data, dyn Dim>` just because `S` and `D` type arguments of `ArrayBase` are `Sized` by defaults.

jturner314:
Right, all the elements in a slice need to have the same type. @SuperFluffy It seems like you want to be able to do something like this, but for `ArrayView` and `Array` instead of `&[A]` and `Vec<A>`?

```rust
let slice = &[1, 2, 3][..];
let vec = vec![4, 5, 6];
let collection = &[slice, &vec];
```

Note that this doesn't work with

```rust
let collection = &[&slice, &vec];
```

which is the statement analogous to `let collection = &[&view, &owned];`.

An implementation of `Deref<Target = ArrayView>` for `ArrayBase` would be pretty dangerous because we'd effectively have to bitwise re-interpret a portion of the `ArrayBase` as an `ArrayView`. (I think it would be possible to do correctly if we make `ArrayBase` `#[repr(C)]`, but it's pretty tricky.) 

For arrays, why not just call `.view()` on all the arrays and create a slice of `ArrayView`s (all the same type)? Like this:

```rust
let collection = &[arr1.view(), arr2.view()];
```

If you need mutable views, then use `.view_mut()` instead.

PR #635, opened 2019-05-09
Rust cargo-fmt

termoshtt:
Revival of #608


LukeMathWalker:
It seems one bit skipped your `cargo fmt` command - if you can fix it then we can go ahead and merge it (and unblock #628)

termoshtt:
Fix format (with 1.31.0 toolchain)

PR #636, opened 2019-05-10
Fix slice offset computation for special cases

jturner314:
With the old implementation, this resulted in undefined behavior in release mode and a panic in debug mode:

```rust
let mut arr = Array2::<i32>::zeros((5, 5));
arr.slice_axis_inplace(Axis(0), Slice::new(0, Some(0), -1));
```

as did this:

```rust
let mut arr = Array2::from_shape_vec((1, 1).strides((10, 1)), vec![5]).unwrap();
arr.slice_axis_inplace(Axis(0), Slice::new(1, Some(1), 1));
```

Now, both examples operate correctly.


jturner314:
Thanks for reviewing this @LukeMathWalker!

Issue #89, opened 2016-02-26
Meta Issue: Support for parallelized/blocked algorithms


kernelmachine:
What are your thoughts on implementing something similar to http://dask.pydata.org/en/latest/ on top of ndarrays? I suspect parallelized computations on submatrices should be pretty natural to do in the Rust framework, and it seems you've already created sub-array view functions. Do you agree?

(Community Edits below)

----

Actionable sub issues:

- [x] Send/Sync splittable array views are already present
- [x] Implement Rayon parallel iterator traits for Axis Iter #248  #252
- [x] Implement Rayon parallel iterator traits for element iterators / for Array itself #252
- [x] alternative methods of collecting an axis_iter to ndarray matrix #249
- [x] Parallel Iter for AxisChunksIter
- [x] Parallel support for Array::map_inplace #288 
- [x] Parallel support for Array::map -> Array
- [x] Parallel lock step function application (Zip) #288

bluss:
The goal is absolutely to be able to support a project like that. Iterators already provide chunking in inner_iter, outer_iter, axis_iter, axis_chunks_iter and their mut counterparts. We also want to add just a few more `split_at` like interfaces to support easy chunking like this.

bluss:
Integrating with https://github.com/nikomatsakis/rayon would be pretty exciting too.

pegasos1:
Yup, that's exactly my thought! Would love to work on this, if you're
interested in collaborating.

pegasos1:
Also on the subject of integrations, I've been writing a crate that wraps Lapack/BLAS with high level, easy to use functions, inspired by the hmatrix library in Haskell. Focus is on compile-time, descriptive error checking, enumerated matrix types, and an easy interface. I wrote my own (simple) matrix representation for the project, but it actually seems way better to build the crate on top of ndarray. 

How actively are you working on the BLAS integration that I see on the docs? Would love to exchange notes.

bluss:
Not very actively, but it's the thing I must solve now. Not sure if ndarray wants to continue with rblas or use more raw blas bindings.

One problem is specialization, i.e. how to dispatch to use BLAS for element types f32, f64 while still supporting other array element types. Rust will learn specialization down the line, but what it looks now, we can do some dispatch using `Any` instead. Which is fine, it just adds that `Any` bound.

bluss:
Note that `Any` allows static (compile time) dispatch on the element type.

bluss:
[...]

PR #639, opened 2019-05-23
Parallel Iterator for AxisChunksIter

nitsky:
This PR implements rayon parallelization for AxisChunksIter. See #89.

### DIFF
--- a/parallel/tests/rayon.rs
+++ b/parallel/tests/rayon.rs
@@ -7,6 +7,8 @@ use ndarray_parallel::prelude::*;
 
 const M: usize = 1024 * 10;
 const N: usize = 100;
+const CHUNK_SIZE: usize = 100;
+const N_CHUNKS: usize = (M + CHUNK_SIZE - 1) / CHUNK_SIZE;
 
 #[test]
 fn test_axis_iter() {
@@ -52,4 +54,33 @@ fn test_regular_iter_collect() {
     }
     let v = a.view().into_par_iter().map(|&x| x).collect::<Vec<_>>();
     assert_eq!(v.len(), a.len());
+}
+
+#[test]
+fn test_axis_chunks_iter() {
+    let mut a = Array2::<f64>::zeros((M, N));
+    for (i, mut v) in a.axis_chunks_iter_mut(Axis(0), CHUNK_SIZE).enumerate() {
+        v.fill(i as _);
+    }
+    assert_eq!(a.axis_chunks_iter(Axis(0), CHUNK_SIZE).len(), N_CHUNKS);
+    let s: f64 = a
+        .axis_chunks_iter(Axis(0), CHUNK_SIZE)
+        .into_par_iter()
+        .map(|x| x.sum())
+        .sum();
+    println!("{:?}", a.slice(s![..10, ..5]));
+    assert_eq!(s, a.sum());
+}
+
+#[test]
+fn test_axis_chunks_iter_mut() {
+    let mut a = Array::linspace(0., 1.0f64, M * N)
+        .into_shape((M, N))
+        .unwrap();
+    let b = a.mapv(|x| x.exp());
+    a.axis_chunks_iter_mut(Axis(0), CHUNK_SIZE)
+        .into_par_iter()
+        .for_each(|mut v| v.mapv_inplace(|x| x.exp()));
+    println!("{:?}", a.slice(s![..10, ..5]));
+    assert!(a.all_close(&b, 0.001));
 }
--- a/parallel/src/lib.rs
+++ b/parallel/src/lib.rs
@@ -64,6 +64,29 @@
 //! }
 //! ```
 //!
+//! ## Axis chunks iterators
+//!
+//! Use the parallel `.axis_chunks_iter()` to process your data in chunks.
+//!
+//! ```
+//! extern crate ndarray;
+//!
+//! use ndarray::Array;
+//! use ndarray::Axis;
+//! use ndarray_parallel::prelude::*;
+//!
+//! fn main() {
+//!     let a = Array::linspace(0., 63., 64).into_shape((4, 16)).unwrap();
+//!     let mut shapes = Vec::new();
+//!     a.axis_chunks_iter(Axis(0), 3)
+//!      .into_par_iter()
+//!      .map(|chunk| chunk.shape().to_owned())
+//!      .collect_into_vec(&mut shapes);
+//!
+//!     assert_eq!(shapes, [vec![3, 16], vec![1, 16]]);
+//! }
+//! ```
+//!
 //! ## Zip
 //!
 //! Use zip for lock step function application across several arrays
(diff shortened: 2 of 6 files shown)


AaronRocinante:
I happen to need the same feature! When are we going to merge this?

parallel/tests/rayon.rs
```diff
@@ -33,6 +35,35 @@ fn test_axis_iter_mut() {
...
+    for (i, mut v) in a.axis_chunks_iter_mut(Axis(0), CHUNK_SIZE).enumerate() {
+        v.fill(i as _);
+    }
+    assert_eq!(a.axis_chunks_iter(Axis(0), CHUNK_SIZE).len(), N_CHUNKS);
+    let s: f64 = a
+        .axis_chunks_iter(Axis(0), CHUNK_SIZE)
+        .into_par_iter()
+        .map(|x| x.sum())
+        .sum();
+    println!("{:?}", a.slice(s![..10, ..5]));
```
nilgoyette: Is there a reason why you always print a slice of the array? You have 4 tests and you do it 4 times. I don't see the gain in the context of a test.

src/iterators/mod.rs
```diff
@@ -1312,6 +1352,26 @@ impl<'a, A, D: Dimension> AxisChunksIterMut<'a, A, D> {
             life: PhantomData,
         }
     }
+    pub fn split_at(self, index: usize) -> (Self, Self) {
```
nilgoyette: I see that this 19-lines block is exactly identical to the other `split_at` above. This may be "normal" (I'm not a Rust expert at all), but it looks wrong! Is there any tool to avoid this? Templating `split_chunk_iter` to return the right types?

parallel/tests/rayon.rs
nitsky: I copied the tests for AxisIter, which include this println. `cargo test` captures stdout for passing tests, so this will only print if the test fails.

src/iterators/mod.rs
nitsky: I believe your idea would work, but this is my first contribution to ndarray, so I chose to mirror the existing code style for AxisIter.

AaronRocinante: @nilgoyette I believe it can be done using trait inheritance. `AxisIterMut` can be a sub trait of `AxisIter`

LukeMathWalker: I think we can get there with a more lightweight solution: we can have a generic private function and just call it in each of the method. What do you think @nitsky?

src/iterators/mod.rs
```diff
@@ -1207,6 +1207,26 @@ fn chunk_iter_parts<A, D: Dimension>(
...
+fn split_chunk_iter<A, D: Dimension>(
+    iter: AxisIterCore<A, D>,
+    n_whole_chunks: usize,
+    last_dim: D,
+    index: usize,
+) -> (
+    (AxisIterCore<A, D>, usize, D),
+    (AxisIterCore<A, D>, usize, D),
+) {
+    let left_n_whole_chunks = index;
```
jturner314: This implementation isn't quite right in the case where the axis is not evenly divisible by the chunk size and the specified `index` is equal to the number of chunks. Here's a test that fails with this implementation:

```rust
#[test]
fn axis_chunks_split_at() {
    let mut a = Array2::<usize>::zeros((11, 3));
    a.iter_mut().enumerate().for_each(|(i, elt)| *elt = i);
    for source in &[
        a.slice(s![..1, ..]),
        a.slice(s![..5, ..]),
        a.slice(s![..10, ..]),
        a.slice(s![..11, ..]),
    ] {
        let chunks_iter = source.axis_chunks_iter(Axis(0), 5);
        let all_chunks: Vec<_> = chunks_iter.clone().collect();
        let n_chunks = chunks_iter.len();
        assert_eq!(n_chunks, all_chunks.len());
        for index in 0..=n_chunks {
            let (left, right) = chunks_iter.clone().split_at(index);
            assert_eq!(&all_chunks[..index], &left.collect::<Vec<_>>()[..]);
            assert_eq!(&all_chunks[index..], &right.collect::<Vec<_>>()[..]);
        }
        assert_panics!({
            chunks_iter.split_at(n_chunks + 1);
        });
    }
}
```

One way to fix the implementation is this:

```rust
fn split_chunk_iter<A, D: Dimension>(
    iter: AxisIterCore<A, D>,
    n_whole_chunks: usize,
    last_dim: D,
    index: usize,
) -> (
    (AxisIterCore<A, D>, usize, D),
    (AxisIterCore<A, D>, usize, D),
) {
    // Note: `index` is checked to be `<= iter.len` in `iter.split_at(index)`.
    if index > n_whole_chunks {
        // In this case, the entire iterator stays in the left piece; the right
        // piece has length zero.
        let (left, right) = iter.split_at(index);
        debug_assert_eq!(right.len, 0);
        (
            (left, n_whole_chunks, last_dim.clone()),
            (right, 0, last_dim),
        )
    } else {
        // In this case, the right iterator contains the last chunk (and
        // possibly more chunks before it).
        let left_n_whole_chunks = index;
        let right_n_whole_chunks = n_whole_chunks - left_n_whole_chunks;
        let left_last_dim = iter.inner_dim.clone();
        let right_last_dim = last_dim;
        let (left, right) = iter.split_at(index);
        (
            (left, left_n_whole_chunks, left_last_dim),
            (right, right_n_whole_chunks, right_last_dim),
        )
    }
}
```

IMO, a cleaner way to implement this is to keep track not of `n_whole_chunks` but of the index corresponding to the partial chunk, and then implement `.split_at()` like this:

```rust
    pub fn split_at(self, index: usize) -> (Self, Self) {
        let (left, right) = self.iter.split_at(index);
        (
            AxisChunksIter {
                iter: left,
                partial_chunk_index: self.partial_chunk_index,
                partial_chunk_dim: self.partial_chunk_dim.clone(),
                life: self.life,
            },
            AxisChunksIter {
                iter: right,
                partial_chunk_index: self.partial_chunk_index,
                partial_chunk_dim: self.partial_chunk_dim,
                life: self.life,
            },
        )
    }
```

Once #669 is merged, I can provide a more complete suggestion.

jturner314: I've created #691 to add (hopefully correct) `.split_at()` implementations. Once that is merged, it should be straightforward to update this PR.

nitsky: @jturner314 great thanks! I'll update after that is merged.

nitsky:
@jturner314 this is ready for review now. Thanks for taking care of the implementation in #691.

bluss:
Hey - sorry for the late notice.

From ndarray 0.13, the crate inside ./parallel is deprecated - it will be removed, as soon as we have made the deprecated-marking release of it when ndarray 0.13 itself goes live.

The place where this change needs to be made is inside src/parallel and inside tests/par_*.rs

bluss:
Maybe there's a point to making the change both places? It makes for a nicer deprecation, but the two modules have already diverged, so I don't think we need to pursue that.

nitsky:
Hi @bluss, I believe this PR includes changes in both places, with the parallel iterator declared in parallel/src/par.rs and in src/parallel/par.rs. I believe the tests and documentation appear in both places as well. Can you double check and let me know if I’m mistaken?

bluss:
@nitsky It looks fine, it changes in both places even if one would be enough. We can take it from here.

nitsky:
Okay, thanks!!

bluss:
Thank you for this :slightly_smiling_face:

Issue #262, opened 2017-01-11
`as_slice` vs `into_slice`: inconsistent naming?


SuperFluffy:
Is there some inconsistency in the naming of `as_slice` (implemented for all `ArrayBase`), and `into_slice` (implemented for those `ArrayBase` where the data is a `ViewRepr`)?

The functions do exactly the same.

Since `as_slice` seems to apply to views as well, I believe `into_slice` could be removed?

bluss:
They have different behaviour, and they exist because that borrowing behaviour has been useful.

Given `&'a ArrayView<'b, T>`, `as_slice` makes a `&'a [T]`. Given `ArrayView<'b, T>`, into_slice makes a `&'b [T]`

SuperFluffy:
Alright, thanks for the clarification, this makes sense. I didn't notice that in `pub fn into_slice(&self) -> Option<&'a [A]>` `self` will have a different lifetime than `'b`.

bluss:
Np. some cleanup is needed here and there. I think we need more rather than fewer into_slice kind of methods for views, unfortunately. (For example .into_axis_iter)

PR #642, opened 2019-06-13
Some Clippy lints

max-sixty:
I'm trying to get familiar with the repo and thought this was a reasonable way to start.

It's not complete - there are a lot of `immediately dereferencing a reference` around `s!` invocations (some I explicitly ignored), and some around using `assert_eq` on f32 & f64.

It's not yet possible to set Clippy rules for a whole workspace - it's required to do crate-by-crate - so there are some standard `ignore` pieces copied and pasted. I'm not sure how to do this for the test path which don't have a `lib.rs` or `mod.rs`.


src/lib.rs
```diff
@@ -1552,8 +1558,8 @@ where
     {
         if let Some(slc) = self.as_slice_memory_order_mut() {
             // FIXME: Use for loop when slice iterator is perf is restored
-            for i in 0..slc.len() {
-                f(&mut slc[i]);
+            for x in slc.iter_mut() {
```
max-sixty: Are the performance issues solved now? This comment was from several years ago

max-sixty:
I've ended up pushing through and either correcting them all or ignoring them.

A couple I need help on - they're ignored for the moment, but my rust understanding wasn't enough to solve the issue in a reasonable way.
And a few need a review - I'm sure my solutions won't be optimal is some of the cases.

src/linalg/impl_linalg.rs
```diff
@@ -78,7 +78,7 @@ where
         let mut sum = A::zero();
         for i in 0..self.len() {
             unsafe {
-                sum = sum.clone() + self.uget(i).clone() * rhs.uget(i).clone();
+                sum = sum + *self.uget(i) * *rhs.uget(i);
```
max-sixty: Unsafe so needs a review

max-sixty:
Hi team - any thoughts here? Would be great to get some of this merged before we start getting conflicts

Happy to split into an 'easy' PR and one that requires more discussion if anything requires that

src/impl_views.rs
```diff
@@ -137,6 +137,9 @@ where
 
     /// Return the array’s data as a slice, if it is contiguous and in standard order.
     /// Return `None` otherwise.
+    // TODO: Should this be `to_slice` if taking a reference?
```
jturner314: Yes, I agree `to_slice` is a better name. (See also #262.) Will you please create a separate PR for this?

src/indexes.rs
```diff
@@ -140,6 +140,7 @@ impl<D: Dimension + Copy> NdProducer for Indices<D> {
     private_impl! {}
 
     #[doc(hidden)]
+    #[allow(clippy::clone_on_copy)]
```
jturner314: I'd prefer to remove the `.clone()` instead of ignoring the Clippy lint.

src/layout/mod.rs
```diff
@@ -57,5 +57,5 @@ impl Layout {
     }
 }
 
-pub const CORDER: u32 = 1 << 0;
+pub const CORDER: u32 = 1;
```
jturner314: IMO, it's clearer to leave it as-is or use this instead (which satisfies Clippy):

```rust
pub const CORDER: u32 = 0b01;
pub const FORDER: u32 = 0b10;
```

src/linalg/impl_linalg.rs
jturner314: Looks good to me. (It's just changing `.clone()` calls to copies.)

src/numeric_util.rs
```diff
@@ -48,11 +48,11 @@ where
 
     // make it clear to the optimizer that this loop is short
     // and can not be autovectorized.
-    for i in 0..xs.len() {
+    for (i, x) in xs.iter().enumerate() {
```
jturner314: I assume this is fine, but we need to check that it still compiles to clean assembly.

src/numeric_util.rs
```diff
@@ -99,13 +99,13 @@ where
     sum = sum + (p2 + p6);
     sum = sum + (p3 + p7);
 
-    for i in 0..xs.len() {
+    for (i, x) in xs.iter().enumerate() {
```
jturner314: Same here.

tests/iterators.rs
```diff
@@ -555,11 +561,11 @@ fn test_fold() {
     a += 1;
     let mut iter = a.iter();
     iter.next();
-    assert_eq!(iter.fold(0, |acc, &x| acc + x), a.sum() - 1);
+    assert_eq!(iter.sum::<i32>(), a.sum() - 1);
```
jturner314: The `.sum()` implementation for `i32` *does* call `.fold()`, but since `test_fold` is intended to test `.fold()`, I think we should be explicit here (leave it as-is).

tests/iterators.rs
```diff
@@ -555,11 +561,11 @@ fn test_fold() {
...
 
     let mut a = Array0::<i32>::default(());
     a += 1;
-    assert_eq!(a.iter().fold(0, |acc, &x| acc + x), 1);
+    assert_eq!(a.iter().sum::<i32>(), 1);
```
jturner314: Same here.

max-sixty:
Amazing, thanks for all the comments and explanations. Helpful for the PR and like a free Rust class! 

I'll merge your changes, review your comments, and make any final changes

src/numeric_util.rs
max-sixty: Great - how do you generally do this? I tried running `cargo asm ndarray::numeric_util::unrolled_fold --rust` but it couldn't find the function. (I can look more but figured you might know off-hand)

src/dimension/axes.rs
```diff
@@ -62,6 +62,9 @@ impl AxisDescription {
     pub fn stride(self) -> Ixs {
         self.2
     }
+    pub fn is_empty(self) -> bool {
```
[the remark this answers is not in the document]
max-sixty: Yeah we get this warning:

```
warning: item `dimension::axes::AxisDescription` has a public `len` method but no corresponding `is_empty` method
  --> src/dimension/axes.rs:49:1
   |
49 | / impl AxisDescription {
50 | |     /// Return axis
51 | |     #[inline(always)]
52 | |     pub fn axis(self) -> Axis {
...  |
67 | |     // }
68 | | }
   | |_^
   |
   = note: #[warn(clippy::len_without_is_empty)] on by default
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#len_without_is_empty
```

Agree with your point, added the `allow`

max-sixty:
Great - after you did most of the follow ups (thanks for the notes though, I find the responses v helpful), we're almost there: need to ensure there are no performance issues re https://github.com/rust-ndarray/ndarray/pull/642#discussion_r296337112, and then two potential flups in separate PRs

src/impl_constructors.rs
```diff
@@ -62,6 +64,9 @@ where
     /// let array = Array::from_iter((0..5).map(|x| x * x));
     /// assert!(array == arr1(&[0, 1, 4, 9, 16]))
     /// ```
+    // FIXME: I don't yet understand why we can't rely on FromIterator
```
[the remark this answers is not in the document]
LukeMathWalker: I agree with you @jturner314 - when possible, we should rely on the existing conversion traits in the standard library. 
More often than not, the compiler will point to the missing `use` statement, so I don't an issue there. Now that `TryFrom` is stable, we should be able to handle most conversions with stdlib traits.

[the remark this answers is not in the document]
max-sixty: PR for this here: https://github.com/rust-ndarray/ndarray/pull/648

max-sixty:
Anything else here? Only remaining item is https://github.com/rust-ndarray/ndarray/pull/642#discussion_r296337112, if that's still a concern

alex179ohm:
master branch fails running `cargo clippy` on both stable and nightly versions:
```
cargo +nightly --version
cargo 1.37.0-nightly (807429e1b 2019-06-11)
cargo +stable --version
cargo 1.35.0 (6f3e9c367 2019-04-04)
```
most notable errors are:
https://github.com/rust-ndarray/ndarray/blob/beb537c61bba7f23d95ef01cc2bbe84f28228e8e/src/data_traits.rs#L136
the `since` field must contain semver compliant version and 0.13 seems not to be. (there is a 0.13 version?)

https://github.com/rust-ndarray/ndarray/blob/beb537c61bba7f23d95ef01cc2bbe84f28228e8e/src/dimension/dimension_trait.rs#L133
and
https://github.com/rust-ndarray/ndarray/blob/beb537c61bba7f23d95ef01cc2bbe84f28228e8e/src/dimension/dimension_trait.rs#L159
actually, this loops never loops and maybe `if let Some(rs) = it.next()` it's enough to set the first element to 1.

max-sixty:
Yes @alex179ohm - I think those are covered by this PR, is that correct?

alex179ohm:
@max-sixty, no
Unfortunately, the nice-semver error still remain until the 0.13 will be released

max-sixty:
>  Unfortunately, the nice-semver error still remain until the 0.13 will be released

I think this wants a semver version, so `0.13.0`, not necessarily a released version. That change is in this PR

max-sixty:
If [this](https://github.com/rust-ndarray/ndarray/pull/642#discussion_r296337112) is still an open issue, shall I remove that change so we can merge? I'm cognizant that this touches a lot of code and keen to avoid merge conflicts

alex179ohm:
@max-sixty All clippy warnings are fixed on your clippy branch? If so it could be fine add clippy on the CI/travis build pipeline

max-sixty:
>  If so it could be fine add clippy on the CI/travis build pipeline
Ask and ye shall receive: https://github.com/rust-ndarray/ndarray/pull/642/files?file-filters%5B%5D=.sh#diff-407dae7942292c08ef6b0f9b2b8c2677R17

LukeMathWalker:
I agree that this PR touches a lot of code and it would be better to get it merged sooner rather than later (even more so because several people try to have a go at it, and we end up with duplicate effort).
I know @jturner314 cannot put much time in `ndarray` right now, so I would propose to set aside https://github.com/rust-ndarray/ndarray/pull/642#discussion_r296337112 until we have time to go back at it using a `deny` statement, so that we can merge the rest. What do you think @max-sixty?

max-sixty:
Great, done @LukeMathWalker

LukeMathWalker:
Merged @max-sixty :+1: Can you open an issue to track the remaining lints?
@alex179ohm - should I close the other PR or are there additional changes there that we need to review?

max-sixty:
Great - @LukeMathWalker - here: https://github.com/rust-ndarray/ndarray/pull/657

Let me know if I missed anything

src/numeric_util.rs
jturner314: I like using Matt Godbolt's Compiler Explorer (https://godbolt.org/) for standalone functions like this. (Make sure to switch the language to Rust.) Note that in order to get any output, you need to have a public, non-generic function. So, you could use this as the input:

```rust
pub fn foo(xs: &[f32]) -> f32 {
    unrolled_fold(xs, || 0., std::ops::Add::add)
}

where
    A: Clone,
    I: Fn() -> A,
    F: Fn(A, A) -> A,
{
    ...
}
```

to get the assembly for `foo`. You can specify the compiler options in the upper part of the right pane. It's worth trying `-O` or, for maximum optimization, `-C opt-level=3 -C target-cpu=native`.

I'm not very familiar with `cargo asm`, but I tried the command you suggested and got an error too. The error message for `cargo asm` says that the function must be present in the final binary. An easy way to ensure this is to add a function that is not generic and is part of the public API. So, for example, you could add this to `lib.rs`:

```rust
pub fn unrolled_sum_f32(xs: &[f32]) -> f32 {
    numeric_util::unrolled_fold(xs, num_traits::Zero::zero, std::ops::Add::add)
}
```

and then use `cargo asm ndarray::unrolled_sum_f32 --rust`.

The results I got in Compiler Explorer showed that the assembly for

```rust
    for (i, x) in xs.iter().enumerate() {
        if i >= 7 {
            break;
        }
        acc = f(acc.clone(), x.clone())
    }
```

was comparable to that for

```rust
    for i in 0..xs.len() {
        if i >= 7 {
            break;
        }
        acc = f(acc.clone(), xs[i].clone())
    }
```

so switching from the range version to the iterator version sounds good to me.

src/lib.rs
jturner314: The commit adding the comment (86415f780b6ef70be4eb314fab36c3230ee6ac29) says

> Use counted loop in unordered_foreach_mut
>
> This improves the generated code for the inner loop we use for
> .assign_scalar() and other array operations with a single scalar.
>
> The old code would generate something that looked like it assumed the
> element being written to and the iterator's pointer could alias another. This
> new version is unrolled and vectorized to a larger degree, but it does
> still not produce memset for some inputs like rust could do in january 2016.
>
> name                     before ns/iter  after ns/iter    diff ns/iter
> diff %
> assign_scalar_2d_corder  1,631           868                      -763 -46.78%
> assign_scalar_2d_cutout  1,704           1,332                    -372 -21.83%
> assign_scalar_2d_forder  1,628           869                      -759 -46.62%
> assign_zero_2d_corder    1,602           855                      -747 -46.63%
> assign_zero_2d_cutout    1,703           1,335                    -368 -21.61%
> assign_zero_2d_forder    1,618           858                      -760 -46.97%
>
> before: Using the slice iterator
> after: using indexed loop.

I tried switching the implementation to use `for_each` and got the same results for these benchmarks as the range-based implementation, so I think that would work well. (I didn't try a `for` loop.)

```rust
        if let Some(slc) = self.as_slice_memory_order_mut() {
            slc.iter_mut().for_each(f);
        } else {
            for row in self.inner_rows_mut() {
                row.into_iter_().fold((), |(), elt| f(elt));
            }
        }
```

Issue #449, opened 2018-05-11
Incompatible shapes error could give more information.


gilescope:
Incompatible shapes error could give more information about the dimentions that do and don't match.
It would be great to see if it's an off by one error or whether it's a fundermental shape missmatch.

Would be great to see (10,20,1) != expected shape (20,10,1) as an error message.

I'm guessing shape mismatches will happen quite frequently for users as they are developing so excellent errors will speed up the dev process.

Thoughts?

PR #644, opened 2019-06-20
Ndarray with Failure

rochamatcomp:
Ndarray with Failure. Added detailed error messages #449.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,6 +45,10 @@ matrixmultiply = { version = "0.2.0" }
 # Use via the `serde-1` crate feature!
 serde = { version = "1.0", optional = true }
 
+# Error-handling
+failure = "0.1"
+failure_derive = "0.1"
+
 [dev-dependencies]
 defmac = "0.2"
 quickcheck = { version = "0.8", default-features = false }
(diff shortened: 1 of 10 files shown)


jturner314:
I hate to reject this outright because it looks like you've put a lot of work into it, but I'm strongly opposed to replacing `Error` with `Fail` for a few reasons:

1. `failure` is ["experimental"](https://docs.rs/failure/0.1.5/failure/) and ["currently evolving"](https://github.com/rust-lang-nursery/failure#evolution), so while it may be fine for an end-user application, I don't think we should use it for a library like `ndarray`.
2. The standard library's `Error` is being updated based on the `failure` experiment to resolve most of the issues with the old version of `Error`. ([RFC 2504](https://github.com/rust-lang/rfcs/blob/master/text/2504-fix-error.md)) In stable, the [`.source()`](https://doc.rust-lang.org/std/error/trait.Error.html#method.source) method has been added, [`.cause()`](https://doc.rust-lang.org/std/error/trait.Error.html#method.cause) is deprecated, and [`.description()`](https://doc.rust-lang.org/std/error/trait.Error.html#method.description) is "soft-deprecated".
3. It's annoying to have to call [`.compat()`](https://docs.rs/failure/0.1.5/failure/trait.Fail.html#method.compat) to get something that implements `Error`.

What is your root objective of this PR? (There are lots of places for improvement in `ndarray`'s error handling. Is your objective primarily to add more detail to the errors? If so, we can do so while continuing to use the standard library's `Error` trait.)

rochamatcomp:
OK, it was good to know more about shape errors.

My objective is to add more detail to the errors indeed. How do you think we can do this with the standard library's `Error` trait?

PR #646, opened 2019-06-21
rename `into_slice` on ArrayView `to_slice` 

max-sixty:
Deprecated the existing method

from https://github.com/rust-ndarray/ndarray/pull/642#discussion_r296072708

PR #656, opened 2019-07-07
Fix docs for ArrayViewMut -> &mut [A] conversion

jturner314:
This line was accidentally changed in #646.


max-sixty:
Ah, this was my fault, thanks for fixing

LukeMathWalker:
I should have caught it, my bad.

PR #657, opened 2019-07-11
Clippy lint left over from #642

max-sixty:
ref https://github.com/rust-ndarray/ndarray/pull/642#discussion_r296337112 - need to check we get good assembly. Happy to help but not sure how to do this, and initial attempt not successful.


max-sixty:
Closing in favor of #665

Issue #637, opened 2019-05-13
Integrate Test Workspaces/Crates


JP-Ellis:
While working on another pull requests, I noticed that the workspaces seem to be often neglected as people don't look into them, and it made me question why we have them in the first place.

In particular, three of them contain only tests, which makes me wonder whether they should really be in a workspace of their own, or whether these tests should be migrated into the main `tests` directory.  The particular workspaces in question are:
- `blas-tests` which should be fine to have in the `tests` directory and gated behind the `blas` feature
- `numeric-tests`
- `serialization-tests`

I'm happy to work on a pull requests to remove these three workspaces and move the top-level `tests` directory.

max-sixty:
I was just about to lodge the same issue. Any thoughts?

max-sixty:
(I think strictly this should be "integrate test _crates_" but the meaning is clear)

LukeMathWalker:
@jturner314 left comments about this in https://github.com/rust-ndarray/ndarray/pull/628 - we could probably start from there to sort this out @max-sixty

bluss:
They should not be in the main tests - they are compiled separately for various reasons, for example blas-tests to use its own blas-enabling settings and to not slow down the regular development cycle ( you don't have to compile it by default if it's not in the main tests).

Please respect the contributors' time and don't bloat the default dev dependencies too much.

JP-Ellis:
Apologies if you found my pull respect as disrespectful.

If the main concern is how long it takes to build the various dev dependencies, then fair enough, there isn't much that can be done at this stage.  Feel free to close this issue as "won't fix / intended".

bluss:
This is forceful but it's about the code changes, it's not personal. Respecting the contributors time is a design constraint, like, let's make it a joy to work on and one of the goals we can have to do that, is to limit compile time needed to work on changes.

I am of two minds - it's good that ndarray is developing a lot outside of my comfort zone. But I still see that there is nobody to really provide a real vision if I don't - it needs more than stewardship. Making the crate slow to compile would mean I would be a type of contributor to jump ship.

JP-Ellis:
Like I said, I completely understand and you are free to close this issue as "won't fix / intended".  I agree that if it takes too long to compile ndarray, or if the test suite takes too long to build every time, then people might not contribute as much.  I opened this issue not knowing the reason behind the (seemingly arbitrary) separation of certain tests into workspaces.

[...]

PR #662, opened 2019-07-18
Use 2018 edition for ndarray-rand

max-sixty:
Can do this more broadly once we've decided on https://github.com/rust-ndarray/ndarray/issues/637


LukeMathWalker:
Agreed, but let's get what we can in the mean time. Thanks for picking this up!

PR #659, opened 2019-07-17
Upgrade to rand 0.7.0

max-sixty:
Would close https://github.com/rust-ndarray/ndarray/issues/658

I'm not sure what the policy is on upgrading dependencies; posting as draft

Issue #658, opened 2019-07-17
ndarray-rand doesn't work with rand 0.7


ngoldbaum:
The test program in the ndarray-rand documentation doesn't compile with rand 0.7:

main.rs:
```
use ndarray::Array;
use ndarray_rand::RandomExt;
use rand::distributions::Uniform;

fn main() {
    let a = Array::random((2, 5), Uniform::new(0., 10.));
    println!("{:8.4}", a);
}
```

cargo.toml:
```
[package]
name = "rand-test"
version = "0.1.0"
authors = ["Nathan Goldbaum <nathan12343@gmail.com>"]
edition = "2018"

[dependencies]
ndarray = { version = "0.12", features = ["blas"] }
ndarray-rand = "0.9"
rand = "0.7"
```

compile error:
```
    Updating crates.io index
   Compiling rand-test v0.1.0 (/home/goldbaum/Documents/rand-test)
error[E0277]: the trait bound `rand::distributions::uniform::Uniform<{float}>: rand::distributions::Distribution<_>` is not satisfied
 --> src/main.rs:6:13
  |
6 |     let a = Array::random((2, 5), Uniform::new(0., 10.));
  |             ^^^^^^^^^^^^^ the trait `rand::distributions::Distribution<_>` is not implemented for `rand::distributions::uniform::Uniform<{float}>`
  |
  = note: required by `ndarray_rand::RandomExt::random`

error: aborting due to previous error

For more information about this error, try `rustc --explain E0277`.
error: Could not compile `rand-test`.
```

It works if I specify `rand = "0.6"` in cargo.toml.

Apologies if this issue is documented already and I've missed it.

ngoldbaum:
perhaps i'm hitting https://github.com/rust-lang/rust/issues/22750. I guess if ndarray-rand doesn't support rand 0.7 that should be encoded in ndarray-rand's cargo.toml?

ngoldbaum:
Ah I see it does specify 0.6 explicitly:

https://github.com/rust-ndarray/ndarray/blob/98bf3b6605208b0498bfbcf0f302510ade870e81/ndarray-rand/Cargo.toml#L15

LukeMathWalker:
I had the exact same issue a couple of hours ago - I am little bit surprised by `cargo`'s behavior here, because `ndarray-rand` does indeed specify `0.6`.

I guess, overall, we should aim to keep close to the latest release of `rand` - #659 goes in the right direction.

[...]

Issue #530, opened 2018-11-07
ndarray-rand


DevinBayly:
The random function from the ndarray_rand::RandomExt trait seems to go unrecognized when following the test example on the repo.

```//lib.rs // also tried with the 2015 edition flag and that's not the prob
extern crate rand;
extern crate ndarray;
extern crate ndarray_rand;



#[cfg(test)]
mod tests {
    use rand::distributions::Range;
    use ndarray::Array;
    use ndarray_rand::RandomExt;

    #[test]
    fn it_works() {
        assert_eq!(2 + 2, 4);
        let (mm, nn) = (5, 5);
        for m in 0..mm {
            for n in 0..nn {
                let a = Array::random((m, n), Range::new(0., 2.));
                assert_eq!(a.shape(), &[m, n]);
                assert!(a.iter().all(|x| *x <= 2.));
                assert!(a.iter().all(|x| *x >= 0.));
            }
        }
    }
}
```
cargo test  generates this compilation error
```
   Compiling randarr v0.1.0 (/tmp/randarr)                                                                                                            
error[E0599]: no function or associated item named `random` found for type `ndarray::ArrayBase<ndarray::OwnedRepr<_>, _>` in the current scope        
  --> src/lib.rs:19:25                                                                                                                                
   |                                                                                                                                                  
19 |                 let a = Array::random((m, n), Range::new(0., 2.));                                                                               
   |                         ^^^^^^^^^^^^^ function or associated item not found in `ndarray::ArrayBase<ndarray::OwnedRepr<_>, _>`                    
                                                                                                                                                      
error: aborting due to previous error                                                                                                                 
                                                                                                                                                      
For more information about this error, try `rustc --explain E0599`.                                                                                   
error: Could not compile `randarr`. 
```

jturner314:
I'm not able to reproduce the issue. When I run `cargo test` in the `ndarray-rand` crate, all the tests pass. When I create a project with the following contents:

* `Cargo.toml`

  ```toml
  [package]
  name = "foo"
  version = "0.1.0"
  
  [dependencies]
  ndarray = "0.12"
  ndarray-rand = "0.8"
  rand = "0.5"
  ```

* `src/main.rs`

  ```rust
  extern crate rand;
  extern crate ndarray;
  extern crate ndarray_rand;
  
  use rand::distributions::Range;
  use ndarray::Array;
  use ndarray_rand::RandomExt;
  
  fn main() {
      assert_eq!(2 + 2, 4);
[...]

PR #663, opened 2019-07-20
Re-export `rand` distributions

LukeMathWalker:
This is what I meant in #659 @max-sixty. 

I have tested it on my [linear regression example](https://github.com/rust-ndarray/ndarray-linalg/blob/for-display-purposes/examples/linear_regression.rs) - it runs without any issue, even if I explicitly specify `rand = 0.7` in `Cargo.toml`.


jturner314:
I'm in favor of trying to prevent confusion like #530 and #658 from ocurring again, but I'm not sure that this is the right approach. If we do this, should we also re-export the `Rng` trait used by `RandomExt::random_using`? If we do that, should we re-export RNG creation routines?

Here are a few options:

1. Re-export all of `rand`. This works okay for stuff in `rand`, but it could be confusing when people want to use other `rand`-dependent crates too (e.g. `rand_isaac` for the ISAAC RNGs).

2. Use the `rename-dependency` trick I proposed in #554 (to support conversions between `ndarray`'s types and multiple versions of `nalgebra`) to support multiple versions of `rand`. So, we'd have `RandomExt_0_6`, `RandomExt_0_7`, etc. In addition to the benefit of supporting multiple versions, the names of the traits would make it clear which version of `rand` would be required to use them.

3. Update `ndarray-rand` to `rand` 0.7 and clearly indicate in the docs that it uses `rand` 0.7.

Personally, I prefer option 3 because it's the simplest and because I don't see much benefit to supporting multiple versions of `rand`, but I'm open to other options.

Fwiw, when I use `ndarray-rand`, I pretty much always pick a specific RNG and seed it from a known value so that I get reproducible results. This means that I need to `use` a RNG from a separate crate (e.g. `rand_isaac::Isaac64Rng`), the `rand::SeedableRng` trait, the relevant `Distribution` type, and `ndarray_rand::RandomExt`.

LukeMathWalker:
I was focusing on proving that the approach actually worked to solve #658, but I do agree with your list when it comes to the options we have to move forward.

I would personally go for both 1. and 3. - at least until `rand` reaches 1.0.
We can vendor it in a `ndarray_rand::rand` submodule and it would provide a confusion-free experience for most of the people using the crate. As far as I can see, the latest `rand` also includes the Isaac RNG, so you would not need any additional dependency for a basic workflow with reproducible results. I don't see too many disadvantages.

On 21 Jul 2019, at 22:22, Jim Turner <notifications@github.com<mailto:notifications@github.com>> wrote:

I'm in favor of trying to prevent confusion like #530<https://github.com/rust-ndarray/ndarray/issues/530> and #658<https://github.com/rust-ndarray/ndarray/issues/658> from ocurring again, but I'm not sure that this is the right approach. If we do this, should we also re-export the Rng trait used by RandomExt::random_using? If we do that, should we re-export RNG creation routines?

Here are a few options:

  1.  Re-export all of rand. This works okay for stuff in rand, but it could be confusing when people want to use other rand-dependent crates too (e.g. rand_isaac for the ISAAC RNGs).

  2.  Use the rename-dependency trick I proposed in #554<https://github.com/rust-ndarray/ndarray/pull/554> (to support conversions between ndarray's types and multiple versions of nalgebra) to support multiple versions of rand. So, we'd have RandomExt_0_6, RandomExt_0_7, etc. In addition to the benefit of supporting multiple versions, the names of the traits would make it clear which version of rand would be required to use them.

  3.  Update ndarray-rand to rand 0.7 and clearly indicate in the docs that it uses rand 0.7.

Personally, I prefer option 3 because it's the simplest and because I don't see much benefit to supporting multiple versions of rand, but I'm open to other options.

Fwiw, when I use ndarray-rand, I pretty much always pick a specific RNG and seed it from a known value so that I get reproducible results. This means that I need to use a RNG from a separate crate (e.g. rand_isaac::Isaac64Rng), the rand::SeedableRng trait, the relevant Distribution type, and ndarray_rand::RandomExt.

jturner314:
> I would personally go for both 1. and 3. - at least until `rand` reaches 1.0. We can vendor it in a `ndarray_rand::rand` submodule and it would provide a confusion-free experience for most of the people using the crate.

Okay, that sounds good to me. Thanks for looking into this.

jturner314:
Should we re-export `rand_distr` in addition to `rand` since most of `rand::distributions` is deprecated?

LukeMathWalker:
> Should we re-export `rand_distr` in addition to `rand` since most of `rand::distributions` is deprecated?

I'd say it makes sense to do so.
I am out in the woods these days, so I won't be able to fix this until I am back home (August 4th). If we want to bundle it with #659 and cut a release sooner than that feel free to take over :sweat_smile:

LukeMathWalker:
Ok, this should be ready now:
- I re-exported `rand-distr` and adapted the docs;
- I added a new example to `random_using` showing the workflow with a seeded rng (`Isaac64`)
- I converted the `README` to Markdown format for `crates.io` consumption (with an example)
- I moved the changelog to a separate file, keeping the last release changelog in the `README.md`

Once this gets merged I think we could cut a release (or should we wait on #670 and just release everything at once?).

bluss:
Seems reasonable, although I don't see a written motivation for why we should depend on rand_distr (since it's an optional extra), but it seems reasonable.

An important note though, with `rename-dependency` if it works well, we could move rand support directly into ndarray itself, since we can risk taking on a public dependency - since we can add feature flags for new versions of rand.

It's still a burden and a risk - we risk having to support many rand versions per ndarray version.

If it would work, ndarray 0.13 (next version) could be released with a feature called rand_0_7, and one could add a rand_0_8 feature later during the ndarray 0.13.x cycle if required. It's a bit messy, and I'm not sure users would enjoy to have traits called `ndarray::RandomExt07` or similar? But it's hard to avoid the naming problem if we have multiple parallel possible implementations.

One way to avoid the parallel names would be if we could "key" the trait by a type or trait from rand. Like the trait `ndarray::RandomExt<T>` would automatically be different if it was parameterized by a type `T` that's from rand 0.7, rand 0.8 etc respectively.

LukeMathWalker:
> It's still a burden and a risk - we risk having to support many rand versions per ndarray version.

I am little scared by this - I don't have good visibility over `rand` and how "far" they are from a stable 1.0 release. We might end up piling up a lot of different implementations without gaining too much from it.
It might be simpler to be start out supporting only a single version and then see how that plays out.

bluss:
When I envison the parallel rand features, it just doesn't work out for our use case. We'd need some way to make it work transparently without resorting to parallel RandExt07, RandExt08 traits etc, that would just move the problem to a different place.

Oh and i noticed that ndarray-rand is still not published with a new version.. What's holding this back? Is it just that the releasers are not paying attention?

bluss:
Ok, so I initially agree with jturner that reexporting is not a great solution.

The only way I can see it as a good solution is if you have a project that's very ndarray focused, and only consuming rand through ndarray itself, and no other ways?

ndarray-rand/src/lib.rs
```diff
@@ -50,13 +67,12 @@ where
     /// overflows usize.
     ///
     /// ```
-    /// extern crate rand;
     /// extern crate ndarray;
     /// extern crate ndarray_rand;
```
bluss: Let's remove all the extern crate lines, they aren't needed

jturner314:
Everything looks good to me except for the `extern crate` lines that @bluss pointed out (and the similar `extern crate` lines elsewhere).

LukeMathWalker:
> I have a hard time deciding, but I think this is an ok solution, when I realize that using the reexports is optional. Users can use the crate directly if they want, and hopefully this will be clear.

In the end, it's up to the user to use it, we are not forcing it on them.
I'll fix the extern crate lines, so that we can move forward with it.

ndarray-rand/README.md
```diff
@@ -0,0 +1,68 @@
...
+Constructors for randomized arrays: `rand`'s integration with `ndarray`.
+
+Example
+=======
+
+Generate a 2-dimensional array with shape `(2,5)` and elements drawn from a uniform distribution
+over the `(0., 10.)` interval:
+
+```rust
+extern crate ndarray;
```
max-sixty:
```suggestion
```

ndarray-rand/README.md
```diff
@@ -0,0 +1,68 @@
...
+extern crate ndarray_rand;
```
max-sixty:
```suggestion
```

ndarray-rand/README.md
LukeMathWalker: Forgot the README :man_facepalming: Thank you for spotting it :+1:

ndarray-rand/src/lib.rs
```diff
@@ -6,33 +6,50 @@
...
-use rand::{thread_rng, Rng, SeedableRng};
+use crate::rand::distributions::Distribution;
+use crate::rand::rngs::SmallRng;
+use crate::rand::{thread_rng, Rng, SeedableRng};
 
 use ndarray::ShapeBuilder;
 use ndarray::{ArrayBase, DataOwned, Dimension};
 
+/// [`rand`](https://docs.rs/rand/0.7.0/rand/), re-exported for convenience and version-compatibility.
+pub mod rand {
```
bluss: Might be my ignorance, I thought we had pub extern crate for re-exporting, but I'm unsure what difference it makes?

LukeMathWalker: Never heard of `pub extern crate <...>`! 
Digging around, it seems they have exactly the same effect ( https://github.com/nix-rust/nix/pull/694 ). 
Using `mod` has the plus of being more explicit in my eyes, given that it uses the "standard"  visibility mechanism of the 2018 edition.

bluss:
Nice :) so I published the next version of [ndarray-rand 0.10](https://docs.rs/ndarray-rand/0.10), so we finally have that out the door, sorry for the delay in that.

ndarray-rand might need to be careful about guaranteeing specific minor versions since we don't track or check it perfectly, when it's embedded in this repo. But it's such a small crate it might not matter.

PR #666, opened 2019-07-22
cargo fix --edition-idioms

max-sixty:
Looks like it's almost exclusively `extern crate` and anonymous lifetimes

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -93,9 +93,6 @@
 //! * If you have experience with NumPy, you may also be interested in
 //!   [`ndarray_for_numpy_users`](doc/ndarray_for_numpy_users/index.html).
 
-#[cfg(feature = "blas")]
-extern crate cblas_sys;
-
 #[cfg(feature = "docs")]
 pub mod doc;
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -93,6 +93,9 @@
 //! * If you have experience with NumPy, you may also be interested in
 //!   [`ndarray_for_numpy_users`](doc/ndarray_for_numpy_users/index.html).
 
+#[cfg(feature = "blas")]
+extern crate cblas_sys;
+
 #[cfg(feature = "docs")]
 pub mod doc;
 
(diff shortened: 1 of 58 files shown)


max-sixty:
Only done on the main crate given uncertainties around integrating test crates

src/lib.rs
```diff
@@ -107,16 +107,6 @@ extern crate blas_src;
 #[cfg(feature = "blas")]
```
jturner314: I think it should be possible to remove all these other `extern crate` lines too. To check this, remove them, and then make sure that e.g. `cargo build --features="approx"` passes.

tests/array.rs
```diff
@@ -7,9 +7,8 @@
     clippy::float_cmp
 )]
 
-extern crate defmac;
-extern crate itertools;
-extern crate ndarray;
+use itertools;
```
jturner314: Are these `use` lines necessary?

tests/stacking.rs
```diff
@@ -1,4 +1,4 @@
-extern crate ndarray;
+use ndarray;
```
jturner314: Is this `use` necessary?

tests/windows.rs
```diff
@@ -4,8 +4,7 @@
     clippy::unreadable_literal,
     clippy::many_single_char_names
 )]
-extern crate itertools;
-extern crate ndarray;
+use itertools;
```
jturner314: Is this `use` necessary?

src/lib.rs
```diff
@@ -93,30 +93,11 @@
...
-extern crate serde;
-
-#[cfg(feature = "rayon")]
-extern crate rayon;
-
-#[cfg(feature = "approx")]
-extern crate approx;
-
 #[cfg(feature = "blas")]
 extern crate blas_src;
```
max-sixty: Not sure why, but these blas `extern crate`s seem to be required...?

jturner314: Huh, based on the error messages, it seems to have something to do with linking. The blas crates link to non-Rust code, so I assume that's what makes them different. Leaving in the `extern crates` is fine since they seem to be necessary to compile successfully.

Issue #664, opened 2019-07-20
Use `sccache` to speed up CI


LukeMathWalker:
[`sccache`](https://github.com/mozilla/sccache) is a nice tool from Mozilla that can used to reduce building times for Rust crates. 
Combining it with [Travis' caching support ](https://docs.travis-ci.com/user/caching/) it should be possible to shorten our CI build times.

@apiraino has written a useful [gist](https://gist.github.com/apiraino/5027c045b0a7f937032ac6af975e4724) where he uses `sscache` on CircleCI - it should not be too difficult to adapt it to Travis.

LukeMathWalker:
We tried in #668, but it didn't work out - little to no benefit. Closing this :disappointed:

PR #668, opened 2019-07-22
[WIP] TravisCI: add sccache to the pipeline

apiraino:
working on #664 

Hi, this PR enables `sccache` on your Rust builds.

I don't see a huge speed up to be honest (In [my test builds](https://travis-ci.com/apiraino/ndarray/builds/120052210) I can only shave off ~10 minutes), I'm still trying to figure out [why I have a very low rate of cache hits](https://github.com/mozilla/sccache/issues/484).

ref: https://docs.travis-ci.com/user/languages/rust/#dependency-management

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: rust
 # use trusty for newer openblas
 sudo: required
 dist: trusty
+cache: cargo
 matrix:
   include:
     - rust: 1.31.0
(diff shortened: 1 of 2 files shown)


apiraino:
Following [this issue](https://github.com/mozilla/sccache/issues/484)

apiraino:
ok, so it looks like that using `sccache` won't be a huge win for `ndarray` because the dependencies are not greatly cacheable.

@LukeMathWalker If you want to keep / merge / discard this PR, feel free to do so. The benefits are not life changing, but probably merging won't hurt  ¯\\_(ツ)_/¯

LukeMathWalker:
I guess we will have to give up - I actually noticed degraded performance on some of my trials due to the overhead of fetching/saving in the cache; Travis' machines are probably not the best too...

PR #669, opened 2019-07-22
Fix axis iterators

jturner314:
While reviewing #639, I came across some bugs in the existing implementations of `AxisIter/Mut` and `AxisChunksIter/Mut`. Specifically,

* The `.split_at()` and `NdProducer` implementations of `AxisIterCore` didn't handle partially consumed iterators properly.

* `AxisChunksIter` and `AxisChunksIterMut` didn't properly handle the chunk size being zero or the axis length being zero.

I've fixed the bugs and added tests. (Most of the tests fail before this PR, but they all pass after this PR.)

### DIFF
--- a/src/impl_methods.rs
+++ b/src/impl_methods.rs
@@ -1005,7 +1005,7 @@ where
     /// The last view may have less elements if `size` does not divide
     /// the axis' dimension.
     ///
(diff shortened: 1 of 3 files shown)


jturner314:
Thanks for reviewing this @bluss!

Issue #670, opened 2019-07-23
0.13 release


rth:
The last release (0.12.1) was 8 month ago and in looks like a lot of development has happened since. 

I was wondering if there were plans to make a new release in the near future? Couldn't find any existing issues or milestones about it.

Personally, I am interested in using https://github.com/rust-ndarray/ndarray/pull/581 for approximate equality tests, but it's not part of any release yet..

Thanks for your great work on ndarray!

LukeMathWalker:
I'd say it makes sense to start thinking about a new release. 
There are several PRs in flight at the moment, but I don't see any major new functionality piece being worked on. Could we perhaps set a date in the nearby future to wrap up the existing work and review what has been done before releasing? (Given that we have more than one breaking change already in master)

End of August/beginning of September sounds reasonable to me. What do you think @jturner314?

jturner314:
Sure, that seems reasonable to me. How about tentatively Sep. 1?

At a minimum, we should merge the PRs labeled [`bug`](https://github.com/rust-ndarray/ndarray/pulls?q=is%3Apr+is%3Aopen+label%3Abug) and PR #633 before releasing 0.13, and we should decide which of the open [`breaking-change`](https://github.com/rust-ndarray/ndarray/pulls?q=is%3Apr+is%3Aopen+label%3Abreaking-change) PRs to merge. #627 is postponed. It would be good to finish up #626 and merge it. PR #570 would be nice, but could wait until 0.14 (and I plan to add support for ellipses while slicing, which will require further breaking changes, so it would be nice to do all the breaking changes at once). All the other open PRs can be part of 0.13.1 if necessary.

bluss:
I would be excited to finally have matrixmultiply 0.2 improvements in the user's hands.

LukeMathWalker:
Let's do a checkpoint, just to see where we are at.

[...]

PR #671, opened 2019-07-23
Move changelog to a separate file

rth:
Hoping to help with https://github.com/rust-ndarray/ndarray/issues/670, this moves the changelog from the README to a separate file (`RELEASES.md` similarly to [rust repo](https://github.com/rust-lang/rust/blob/master/RELEASES.md) as it is getting quite long.

Also,
 - converts .rst to .md
 - Improve section titles (include date) for newer releases
 - Add some of the changes for 0.13 but clearly there is much more of those that can be addressed in a separate PR
 - Add subsections for 0.13 to better see new feature, API changes etc.


LukeMathWalker:
I think this was long overdue given the size of our README :sweat_smile: 
I'd like to keep the latest release changelog in the README though (0.13 - to be released) - what do you think @rth?

rth:
> I'd like to keep the latest release changelog in the README though (0.13 - to be released

Then maybe have it repeated both in the changelog and the readme? If there is a changelog, users are likely to look at it for the latest release as well?

LukeMathWalker:
> > I'd like to keep the latest release changelog in the README though (0.13 - to be released
> 
> Then maybe have it repeated both in the changelog and the readme? If there is a changelog, users are likely to look at it for the latest release as well?

Duplicating it seems to be the best solution, I agree.

LukeMathWalker:
Merging - we can then add the changelog for the latest release on the README before cutting `0.13`.

PR #673, opened 2019-07-23
ENH Add Array2::from_diag

### LINKED ISSUE #672: Add from_diag constructor for Array2
Array.diag can returns the diagonal of an array, however I couldn't find the associated constructor i.e. to construct an array given it's diagonal.

In numpy this can be done with the [`np.diag`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.diag.html) function.

rth:
Adds a `Array2::from_diag` method to create 2D arrays from a diagonal.

Closes #672

### DIFF
--- a/src/impl_constructors.rs
+++ b/src/impl_constructors.rs
@@ -171,6 +171,20 @@ where
         }
         eye
     }
+
+    /// Create a 2D matrix from its diagonal
+    pub fn from_diag(diag: &ArrayBase<S, Ix1>) -> Self
+    where
+        S: DataMut,
+        A: Clone + Zero + One,
+    {
+        let mut eye = Self::zeros((diag.len(), diag.len()));
+        for (idx, val) in diag.iter().enumerate() {
+            eye[[idx, idx]] = val.clone();
+        }
+        eye
+    }
+
 }
 
 #[cfg(not(debug_assertions))]
--- a/src/impl_constructors.rs
+++ b/src/impl_constructors.rs
@@ -173,16 +173,16 @@ where
     }
 
     /// Create a 2D matrix from its diagonal
-    pub fn from_diag(diag: &ArrayBase<S, Ix1>) -> Self
+    pub fn from_diag<S2>(diag: &ArrayBase<S2, Ix1>) -> Self
     where
+        A: Clone + Zero,
         S: DataMut,
-        A: Clone + Zero + One,
+        S2: Data<Elem = A>,
     {
-        let mut eye = Self::zeros((diag.len(), diag.len()));
-        for (idx, val) in diag.iter().enumerate() {
-            eye[[idx, idx]] = val.clone();
-        }
-        eye
+        let n = diag.len();
+        let mut arr = Self::zeros((n, n));
+        arr.diag_mut().assign(&diag);
+        arr
     }
 }
 
--- a/src/impl_constructors.rs
+++ b/src/impl_constructors.rs
@@ -179,11 +179,9 @@ where
     /// ```rust
     /// use ndarray::{Array2, arr1, arr2};
     ///
-    /// # #[cfg(feature = "approx")] {
     /// let diag = arr1(&[1, 2]);
     /// let array = Array2::from_diag(&diag);
     /// assert_eq!(array, arr2(&[[1, 0], [0, 2]]));
-    /// # }
     pub fn from_diag<S2>(diag: &ArrayBase<S2, Ix1>) -> Self
     where
         A: Clone + Zero,
--- a/src/impl_constructors.rs
+++ b/src/impl_constructors.rs
@@ -182,6 +182,7 @@ where
     /// let diag = arr1(&[1, 2]);
     /// let array = Array2::from_diag(&diag);
     /// assert_eq!(array, arr2(&[[1, 0], [0, 2]]));
+    /// ```
     pub fn from_diag<S2>(diag: &ArrayBase<S2, Ix1>) -> Self
     where
         A: Clone + Zero,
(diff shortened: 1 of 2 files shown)


src/impl_constructors.rs
```diff
@@ -171,6 +171,20 @@ where
...
+
+    /// Create a 2D matrix from its diagonal
+    pub fn from_diag(diag: &ArrayBase<S, Ix1>) -> Self
+    where
+        S: DataMut,
+        A: Clone + Zero + One,
+    {
+        let mut eye = Self::zeros((diag.len(), diag.len()));
+        for (idx, val) in diag.iter().enumerate() {
+            eye[[idx, idx]] = val.clone();
```
rth: Maybe something like,
```rust
for (a_ii, val) in eye.diag_mut().iter().zip(diag) {
   *a_ii = val.clone();
}
```
would have been better, but I haven't managed to make that work (haven't searched very long). The error was,
```
error[E0594]: cannot assign to `*a_ii` which is behind a `&` reference
   --> src/impl_constructors.rs:183:13
    |
182 |         for (a_ii, val) in eye.diag_mut().iter().zip(diag) {
    |              ---- help: consider changing this to be a mutable reference: `&mut A`
183 |             *a_ii = val.clone()
    |             ^^^^^ `a_ii` is a `&` reference, so the data it refers to cannot be written
```

src/impl_constructors.rs
```diff
@@ -171,6 +171,20 @@ where
         }
         eye
     }
+
+    /// Create a 2D matrix from its diagonal
+    pub fn from_diag(diag: &ArrayBase<S, Ix1>) -> Self
+    where
+        S: DataMut,
+        A: Clone + Zero + One,
```
rth: Not fully sure if this is correct.

src/impl_constructors.rs
jturner314: You can fix the error by changing `.iter()` to `.iter_mut()`, but a faster and more concise implementation is to use `.assign()`:

```rust
    pub fn from_diag<S2>(diag: &ArrayBase<S2, Ix1>) -> Self
    where
        A: Clone + Zero,
        S: DataMut,
        S2: Data<Elem = A>,
    {
        let n = diag.len();
        let mut arr = Self::zeros((n, n));
        arr.diag_mut().assign(diag);
        arr
    }
```

tests/array.rs
```diff
@@ -1954,6 +1954,14 @@ fn test_array_clone_same_view() {
     assert_eq!(a, b);
 }
 
+#[test]
+fn test_array2_from_diag() {
```
jturner314: It would be good to also include a test for when `diag` is zero-length. (In my experience writing code for `ndarray`, it's common to have bugs related to zero-size arrays, so it's useful to have tests for those cases.)

src/impl_constructors.rs
```diff
@@ -171,6 +171,19 @@ where
         }
         eye
     }
+
+    /// Create a 2D matrix from its diagonal
+    pub fn from_diag(diag: &ArrayBase<S, Ix1>) -> Self
```
jturner314: `from_diag` should be generic over the storage type of `diag`, so that you could e.g. create an `Array2` from a `ArrayView1` diagonal.

src/impl_constructors.rs
```diff
@@ -171,6 +171,19 @@ where
         }
         eye
     }
+
+    /// Create a 2D matrix from its diagonal
```
jturner314: We should add a note about the panic conditions, and it would be nice to have an example:

```rust
/// Create a 2D matrix from its diagonal.
///
/// **Panics** if `diag.len() * diag.len()` would overflow `isize`.
///
/// ```rust
/// // example here
/// ```
```

src/impl_constructors.rs
rth: I wasn't sure what I need to change to make the above comment possible.

rth:
Thanks for the detailed review @jturner314 ! I added an example and panic conditions as you suggested.

The only remaining point is https://github.com/rust-ndarray/ndarray/pull/673#discussion_r306321841 that I'm not sure how to address. You previously marked it as resolved, so I don't know if some action is needed on my part.

src/impl_constructors.rs
jturner314: I was just talking about adding the `S2` type parameter, which you've already done.

src/impl_constructors.rs
```diff
@@ -171,6 +171,30 @@ where
...
+    ///
+    /// **Panics** if `diag.len() * diag.len()` would overflow `isize`.
+    ///
+    /// ```rust
+    /// use ndarray::{Array2, arr1, arr2};
+    ///
+    /// # #[cfg(feature = "approx")] {
```
jturner314: Is this attribute necessary? I think it should be possible to remove it.

rth: @jturner314 if you meant `cfg`, approx is an optional dependency, in `Cargo.toml` as far as I can tell,
```
approx = { version = "0.3.2", optional = true }
```
and that's what other examples are doing?

jturner314: Sorry, I wasn't clear. You're correct that `approx` is an optional dependency/feature, and some of the other examples have a `#[cfg(feature = "approx")]` annotation because they use the `approx` feature. In this case, the example isn't using `approx`, so the lines `    /// # #[cfg(feature = "approx")] {` and `    /// # }` shouldn't be necessary. (You should be able to remove them and have `cargo test` still pass.)

rth: Aww, yes, sorry, I forgot I was no longer using float equality tests there. Thanks for your patience. Should be fixed now.

rth:
Thanks again for the review @jturner314 !

PR #675, opened 2019-07-25
Remove some unsafe code

alex179ohm:
I tried to remove some unsafe code from ArrayBase without apparently touching the overall performance.
btw: for a more accurate performance estimation I started on rewriting all the benches with [criterion](https://github.com/bheisler/criterion.rs) (running diff between different benches is getting a bit boring ^^).

### DIFF
--- a/src/impl_constructors.rs
+++ b/src/impl_constructors.rs
@@ -249,7 +249,7 @@ where
         let shape = shape.into_shape();
         let size = size_of_shape_checked_unwrap!(&shape.dim);
         let v = vec![elem; size];
-        unsafe { Self::from_shape_vec_unchecked(shape, v) }
+        Self::from_shape_vec_unchecked(shape, v)
     }
 
     /// Create an array with zeros, shape `shape`.
@@ -285,7 +285,7 @@ where
         let shape = shape.into_shape();
         let size = size_of_shape_checked_unwrap!(&shape.dim);
         let v = to_vec((0..size).map(|_| A::default()));
-        unsafe { Self::from_shape_vec_unchecked(shape, v) }
+        Self::from_shape_vec_unchecked(shape, v)
     }
 
     /// Create an array with values created by the function `f`.
@@ -303,11 +303,11 @@ where
         let _ = size_of_shape_checked_unwrap!(&shape.dim);
         if shape.is_c {
             let v = to_vec_mapped(indices(shape.dim.clone()).into_iter(), f);
-            unsafe { Self::from_shape_vec_unchecked(shape, v) }
+            Self::from_shape_vec_unchecked(shape, v)
         } else {
             let dim = shape.dim.clone();
             let v = to_vec_mapped(indexes::indices_iter_f(dim), f);
-            unsafe { Self::from_shape_vec_unchecked(shape, v) }
+            Self::from_shape_vec_unchecked(shape, v)
         }
     }
 
@@ -363,7 +363,7 @@ where
                 return Err(error::incompatible_shapes(&Ix1(v.len()), &dim));
             }
         }
-        unsafe { Ok(Self::from_vec_dim_stride_unchecked(dim, strides, v)) }
+        Ok(Self::from_vec_dim_stride_unchecked(dim, strides, v))
     }
 
     /// Creates an array from a vector and interpret it according to the
@@ -387,7 +387,8 @@ where
     ///
     /// 5. The strides must not allow any element to be referenced by two different
     ///    indices.
-    pub unsafe fn from_shape_vec_unchecked<Sh>(shape: Sh, v: Vec<A>) -> Self
+    //pub unsafe fn from_shape_vec_unchecked<Sh>(shape: Sh, v: Vec<A>) -> Self
+    pub fn from_shape_vec_unchecked<Sh>(shape: Sh, v: Vec<A>) -> Self
     where
         Sh: Into<StrideShape<D>>,
     {
@@ -395,7 +396,8 @@ where
         Self::from_vec_dim_stride_unchecked(shape.dim, shape.strides, v)
     }
 
-    unsafe fn from_vec_dim_stride_unchecked(dim: D, strides: D, mut v: Vec<A>) -> Self {
+    //unsafe fn from_vec_dim_stride_unchecked(dim: D, strides: D, mut v: Vec<A>) -> Self {
+    fn from_vec_dim_stride_unchecked(dim: D, strides: D, mut v: Vec<A>) -> Self {
         // debug check for issues that indicates wrong use of this constructor
         debug_assert!(dimension::can_index_slice(&v, &dim, &strides).is_ok());
         ArrayBase {
--- a/src/arraytraits.rs
+++ b/src/arraytraits.rs
@@ -140,7 +140,7 @@ where
                 "Length must fit in `isize`.",
             );
         }
-        unsafe { Self::from_shape_vec_unchecked(v.len() as Ix, v) }
+        Self::from_shape_vec_unchecked(v.len() as Ix, v)
     }
 }
 
--- a/src/data_traits.rs
+++ b/src/data_traits.rs
@@ -206,12 +206,10 @@ where
         if self_.dim.size() <= self_.data.0.len() / 2 {
             // Create a new vec if the current view is less than half of
             // backing data.
-            unsafe {
-                *self_ = ArrayBase::from_shape_vec_unchecked(
-                    self_.dim.clone(),
-                    self_.iter().cloned().collect(),
-                );
-            }
+            *self_ = ArrayBase::from_shape_vec_unchecked(
+                self_.dim.clone(),
+                self_.iter().cloned().collect(),
+            );
             return;
         }
         let rcvec = &mut self_.data.0;
--- a/src/free_functions.rs
+++ b/src/free_functions.rs
@@ -54,7 +54,7 @@ macro_rules! array {
 
 /// Create a zero-dimensional array with the element `x`.
 pub fn arr0<A>(x: A) -> Array0<A> {
-    unsafe { ArrayBase::from_shape_vec_unchecked((), vec![x]) }
+    ArrayBase::from_shape_vec_unchecked((), vec![x])
 }
 
 /// Create a one-dimensional array with elements from `xs`.
--- a/src/impl_methods.rs
+++ b/src/impl_methods.rs
@@ -186,12 +186,10 @@ where
         S: Data,
     {
         if let Some(slc) = self.as_slice_memory_order() {
-            unsafe {
-                Array::from_shape_vec_unchecked(
-                    self.dim.clone().strides(self.strides.clone()),
-                    slc.to_vec(),
-                )
-            }
+            Array::from_shape_vec_unchecked(
+                self.dim.clone().strides(self.strides.clone()),
+                slc.to_vec(),
+            )
         } else {
             self.map(|x| x.clone())
         }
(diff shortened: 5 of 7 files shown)


src/impl_constructors.rs
```diff
@@ -387,15 +387,17 @@ where
...
     where
         Sh: Into<StrideShape<D>>,
     {
         let shape = shape.into();
         Self::from_vec_dim_stride_unchecked(shape.dim, shape.strides, v)
     }
 
-    unsafe fn from_vec_dim_stride_unchecked(dim: D, strides: D, mut v: Vec<A>) -> Self {
+    //unsafe fn from_vec_dim_stride_unchecked(dim: D, strides: D, mut v: Vec<A>) -> Self {
+    fn from_vec_dim_stride_unchecked(dim: D, strides: D, mut v: Vec<A>) -> Self {
```
max-sixty: Is this actually safe, though?

alex179ohm: The only unsafe code implemented in `from_vec_dim_stride_unchecked` is there:
```
        ArrayBase {
            ptr: v.as_mut_ptr(),
            data: DataOwned::new(v),
            strides,
            dim,
        }
```
and more precisely at `v.as_mut_ptr()`.
IMO there are no reasons to makes this function unsafe.
`from_vec_dim_stride_unchecked` is just used internally, maybe more detailed documentation for developers is better than making this unsafe.
even more `as_mut_ptr()` is extensively used on all the code base without signing the functions as unsafe.
Yes, there are differences from using a function and hold an unsafe pointer inside a struct that cannot be destroyed, the latest must be carefully managed, but all the tests are ok, (maybe we can write more specific tests to check the correctness of its implementation).
IMO making the function unsafe not makes the function safer than making it safe. (except for performance, but there is no evidence of performance regression).

jturner314: There seems to be some confusion about the purpose of `unsafe`. [This section of the nomicon](https://doc.rust-lang.org/nomicon/safe-unsafe-meaning.html) provides a good explanation:

> The `unsafe` keyword has two uses: to declare the existence of contracts the compiler can't check, and to declare that a programmer has checked that these contracts have been upheld.
>
> You can use `unsafe` to indicate the existence of unchecked contracts on *functions* and *trait declarations*. On functions, `unsafe` means that users of the function must check that function's documentation to ensure they are using it in a way that maintains the contracts the function requires. On trait declarations, `unsafe` means that implementors of the trait must check the trait documentation to ensure their implementation maintains the contracts the trait requires.
>
> You can use `unsafe` on a block to declare that all unsafe actions performed within are verified to uphold the contracts of those operations. For instance, the index passed to `slice::get_unchecked` is in-bounds.
>
> You can use `unsafe` on a trait implementation to declare that the implementation upholds the trait's contract. For instance, that a type implementing `Send` is really safe to move to another thread.

Unfortunately, the `unsafe` keyword on a function can be somewhat confusing because it conflates both uses of the `unsafe` keyword: (1) it indicates that callers of the function must uphold the contract specified in the documentation, and (2) it effectively wraps the function body in a `unsafe {}` block.

In the case of `from_vec_dim_stride_unchecked` and `from_shape_vec_unchecked`, it is true that the bodies do not do any unsafe operations, so `unsafe` is unnecessary as far as use (2) is concerned. However, the `unsafe` keyword on these functions is needed for use (1): to indicate that the caller must uphold certain unchecked contracts. The contract that callers must uphold when calling `from_shape_vec_unchecked` is described in [its docs](https://docs.rs/ndarray/0.12.1/ndarray/struct.ArrayBase.html#method.from_shape_vec_unchecked). (See the list of five conditions.)

From an API design perspective, the key when deciding which functions/trait declarations to mark `unsafe` is described by the nomicon:

> The need for all of this separation boils down a single fundamental property of Safe Rust:
>
> **No matter what, Safe Rust can't cause Undefined Behavior.**

If we remove the `unsafe` keyword from `from_shape_vec_unchecked`, then this is one possible way to cause undefined behavior in safe code:

```rust
let empty_data = Vec::new();
let non_empty_shape = (10, 10);
let bad_array = Array2::<i32>::from_shape_vec_unchecked(non_empty_shape, empty_data);
println!("{}", bad_array[(5, 5)]);
```

(`bad_array[(5, 5)]` is an out-of-bounds read, so it's undefined behavior.)

All code must uphold certain constraints on instances of the `ArrayBase` type in order to avoid undefined behavior. (Method implementations rely on these constraints to be upheld when performing unsafe operations, so all code must uphold these constraints.) These constraints are described in a [comment on `ArrayBase`](https://github.com/rust-ndarray/ndarray/blob/483200bccfb903a31a3befbb737d4a907862f8dc/src/lib.rs#L1099-L1210). Safe functions/methods must perform the necessary checks in order to ensure these constraints are upheld. Functions/methods marked as `unsafe` must specify in their documentation what the caller needs to do to make sure the constraints are upheld.

In this way, we provide an API to the user that is guaranteed not to cause undefined behavior (except for any bugs in `ndarray`'s implementation) while needing few safety checks. The user doesn't have to worry about any safety issues if they stick to safe functions; and if they call any unsafe functions, they just have to read the docs for those functions and uphold their contracts. They don't have to know anything about the internal implementation of `ndarray` to be safe.

Note that adding/removing the `unsafe` keyword doesn't have any direct performance impact. The usual reason why `unsafe` is associated with performance is sometimes you can use `unsafe` to avoid safety checks associated with safe code. For example, `ArrayBase` provides both a safe `get` method and an unsafe `uget` method. The safe `get` method performs a check to make sure that the specified index is in-bounds. In contrast, the `uget` method relies on the caller to make sure that the specified index is in-bounds. If the caller can *guarantee* that an index is in-bounds, they can call `uget` in an `unsafe {}` block to avoid the performance cost of the bounds check. Just wrapping a `get` call in an `unsafe {}` block has no effect; the bounds check is still performed.

alex179ohm: > In this way, we provide an API to the user that is guaranteed not to cause undefined behavior.

And I was surprised discovering that this kind of danger method it's exposed to the end-users.
The target audience of this crate maybe it's different from others crate, usually, the end-uses comes from an academic background and its supposed they are not regular rust developers.
As I could look around, end-users of this crate starts using the rust programming language because attracted by what the rust language has to offer in terms of performance and safety.
And safety and simplicity are what the end-users expect from a high-level library.
Maybe they don't expect to find unsafe functions exposed on a high-level crate as ndarray.
And at least end-users are so scared by the crates which extensively use unsafe code on their internal code.


```rust
let empty_data = Vec::new();
let non_empty_shape = (10, 10);
let bad_array = Array2::<i32>::from_shape_vec_unchecked(non_empty_shape, empty_data);
println!("{}", bad_array[(5, 5)]);
```
this could be intended as andefined behavior or as error, and it's supposed that a end-users could not instantiate ArrayBase with unsafe functions.
the same could happen with a function witch it's not marked _unsafe_:
```rust
let a = Array2::<i32>::zeros((2, 2));
println!("{}" a[(5, 5)]);
```
the two panics as is expected they does, the difference is that `from_shape_vec_unchecked` should not be available to the end-users.
The most close example wich come to mind is the Vec standard library:
```rust
let v = Vec::<i32>::new();
println!("{}", v[1]);
```
it panics as expected and it's not a undefined behavior.
you can argues that the relative example could be something like:
```rust
let v = Vec::<i32>::new();
let v2 = unsafe {
        Vec::from_raw_parts(
                v.as_mut_ptr(),
                2,
                3);
}
println!("{}", v[2]);
```
and you are completely right! but the context is completely different.
first:
`from_raw_parts` is not extensively implemented in the Vec crate as `from_shape_vec_unchecked` is in ndarray,
second:
Vec needs `capacity` and `len` to allocate the elements on the fly or to allocate the memory when the vector is created, and the undefined behavior is intended as a serious memory issue on allocation, ArrayBase doesn't manage allocations as Vec does.
but at least and more important:
Vec is located in the standard library, so peoples assume that its unsafe code is reviewed by an entire community, and take the reasonable risk to use the library even if contains unsafe code.
ndarray, is a high-level crate, and even if you, bluss, max and luke (I just picked the main contributors) are very good developers, the users don't take the same risk as they do with a crate inside the std namespace. (It's arguable, I know, but it's exactly what happen).
At the moment `#[forbid(unsafe_code)]` is starting to be extensively used.

My efforts to remove some unsafe code from the crate had to be intended in this way, and not as a trolling or unwary move.

So please consider removing some unsafe code that is not strictly necessary.

As an early-contributor I want to propose three solutions that come to mind at this moment:

- mark the function pub(crate) and remove the unsafe mark.
pros:
could be a reasonable way because the code involved seems stable and has not been touched in nearly two years and maybe there are no reasons to touch it.
cons:
as you rightly argued could generate a lot of problems when a contributor implements it without taking care of possible implications.
( the gun is under the control and every shoot is reviewed by the maintainers).

- remove completely the functions, and paste the 5 lines of code where `from_shape_vec_unchecked` is implemented around the crate documenting what this code does (for contributor only inside the functions involved).
pros:
the code seems stable enough (some code is two years old), numbers of function stacks are reduced and the number of unsafe blocks inside the crate is reduced too.
cons:
some spaghetti code.
(no gun, no shot, no foot broken).

- mark the function pub(crate), remove the unsafe code and pass it as a weird macro witch forces the contributors to read its documentation when is called.
pros:
the same as the first solution.
the solution partially complains of the rusty rule 'if it compiles, it works`.
cons:
the macro is very weird and looks really bad.
(the gun is locked and if someone wants to shot their foot need to previously unlock the gun):
```rust
fn from_shape_vec_unchecked(a: Sh, b: Vec<T>) -> ArrayBase {};
macro_rules! arraybase_from_shape_vec {
        ($a:expr, $b:expr) =>  {
                compile_error!("You tried to implement a very dangerous macro\n
                                please read the documentation at: https://doc.html\n
                                and then you can call the macro as explained below:\n
                                arraybase_from_shape_vec!(a, b, checked);");
        };
        ($a:expr, $b:expr, checked) => { ArrayBase::from_shape_vec_unchecked($a, $b) };
};
arraybase_from_shape_vec!(s, v); // compiler error.
arraybase_from_shape_vec!(s, v, checked) // ok
```

P.S. Someone noticed me that sometimes my comments seem presumptuous, or rude, but it's absolutely not my intention, just takes it as no-uk-european bad english. Maybe adopting the rust [code of conduct](https://www.rust-lang.org/policies/code-of-conduct) and add a properly contribution section on the README, could help in this way, the new contributors are aware of what is the correct behavior expected by the mantainers.

max-sixty: There's a lot here but let me try and address the central point. 

FWIW I'm new to rust and learning by contributing to a few OSS libraries - I'm new to this too.

> You tried to implement a very dangerous macro\n
                                please read the documentation at: http s://doc.html\n
                                and then you can call the macro as explained below:\n

This is basically the role of `unsafe`! By marking some functions at `unsafe`, users can't cause undefined behavior without being aware of the risks.

_Marking a function as unsafe doesn't cause the code to be any less safe per se_ - rather it a) tells the compiler that it should _allow_ undefined behavior and b) tells users they the compiler will allow undefined behavior, so they're responsible for checking the code

Then library developers can expose that functionality with a combination of safe functions (e.g. `arraybase_from_shape_vec`) and checks (e.g. [these checks](https://github.com/rust-ndarray/ndarray/pull/675/files#diff-33d16d5beafe2f3ea47855086180dd86R1431) )

If that's unclear, maybe read up on `unsafe` in the documentation. 

I'm impressed with your enthusiasm to contribute - maybe @LukeMathWalker or @jturner314 could suggest a couple of easier issues to start with?

bluss:
I'd like to avoid criterion - it depends on what level of increases in compilation time contributors want to deal with, though.

For benchmark comparisons, I use cargo-benchcmp

bluss:
Thanks to those that reviewed and explained. I'll close the PR, since the unsafe functions and unsafe code blocks were all quite important, and we'd like to keep them :slightly_smiling_face: .

Issue #661, opened 2019-07-18
ArcArray1


max-sixty:
There are some references to [`RcArray1` & `RcArray2`](https://github.com/rust-ndarray/ndarray/blob/master/serialization-tests/tests/serialize.rs#L29), causing some warnings. I was planning to put in a PR to move these to ArcArray, but ArcArray1 doesn't exist.

Should we create them? Or move specializations up a level?

LukeMathWalker:
In the end, it's syntactic sugar and I think we should keep them (if nothing else, for consistency with `Array1`, `ArrayView1`, etc.). What do you mean by moving specializations up a level?

max-sixty:
Should we create `ArcArray1` & `ArcArray2`? Currently we only have a deprecated `RcArray1` & `RcArray2`

LukeMathWalker:
Yes, I think we should.

max-sixty:
@alex197ohm if you're looking for a good issue to begin with, this could be interesting

PR #680, opened 2019-08-02
Deny warnings in CI (except for nightly channel)

jturner314:
I was surprised to read in #661 that there are warnings that we haven't fixed. This PR makes the build fail if there are any compiler warnings, except on the nightly channel.

It's tempting to add `#![deny(warnings)]` to the top of `lib.rs`, but when working on `ndarray`, I don't really want warnings to get in the way of development. So, I'm proposing just denying warnings in CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,13 +7,16 @@ matrix:
     - rust: 1.32.0
       env:
        - FEATURES='test docs'
+       - RUSTFLAGS='-D warnings'
     - rust: stable
       env:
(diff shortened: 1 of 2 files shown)


jturner314:
CI failed on the first commit of this PR because there were warnings that need to be fixed. This was the desired result. I've now added another commit to fix the warnings, so CI should pass.

LukeMathWalker:
I'd say it makes to limit it to CI - merging :+1:

PR #684, opened 2019-08-15
Add .fold() method to Zip

jturner314:
This is just for convenience to avoid having to deal with `FoldWhile`.

### DIFF
--- a/src/zip/mod.rs
+++ b/src/zip/mod.rs
@@ -755,8 +755,44 @@ macro_rules! map_impl {
             /// Apply a fold function to all elements of the input arrays,
             /// visiting elements in lock step.
             ///
-            /// The fold continues while the return value is a
-            /// `FoldWhile::Continue`.
+            /// # Example
+            ///
+            /// The expression `tr(AᵀB)` can be more efficiently computed as
+            /// the equivalent expression `∑ᵢⱼ(A∘B)ᵢⱼ` (i.e. the sum of the
+            /// elements of the entry-wise product). It would be possible to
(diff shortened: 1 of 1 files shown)


src/zip/mod.rs
```diff
@@ -755,8 +755,44 @@ macro_rules! map_impl {
...
+            /// let sum_prod_zip = Zip::from(&a).and(&b).fold(0, |acc, a, b| acc + a * b);
+            ///
+            /// assert_eq!(sum_prod_nonzip, sum_prod_zip);
+            /// ```
+            pub fn fold<F, Acc>(mut self, acc: Acc, mut function: F) -> Acc
+            where
+                F: FnMut(Acc, $($p::Item),*) -> Acc,
+            {
+                self.apply_core(acc, move |acc, args| {
+                    let ($($p,)*) = args;
```
max-sixty: @jturner314 for my education: I hadn't seen this syntax outside macros. Do you have any reference to docs explaining what's going on?

Jvanrhijn: It is inside a macro, starts at line [736](https://github.com/rust-ndarray/ndarray/blob/27c010dbe3b3efc2cdd62e2f25d44da07ef97c8e/src/zip/mod.rs#L736). I had to do a double take there :smile:

max-sixty: Ah! Thanks :)

LukeMathWalker:
This makes the API nicer to use - I don't see any downside, we can go ahead and merge.

jturner314:
Thanks for reviewing this @LukeMathWalker!

PR #686, opened 2019-08-16
Implement .nth_back() for iterators

jturner314:
The `.nth_back()` method was added to the `DoubleEndedIterator` trait in Rust 1.37 (released today). Providing an implementation for `Baseiter` and forwarding it for `Iter/Mut` improves performance.

### DIFF
--- a/tests/iterators.rs
+++ b/tests/iterators.rs
@@ -565,6 +565,65 @@ fn test_fold() {
 }
 
 #[test]
+fn test_nth_back() {
+    let mut a: Array1<i32> = (0..256).collect();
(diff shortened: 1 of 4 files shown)


tests/iterators.rs
```diff
@@ -564,6 +564,65 @@ fn test_fold() {
     assert_eq!(a.iter().fold(0, |acc, &x| acc + x), 1);
 }
 
+#[test]
```
LukeMathWalker: For the sake of readability, would it make sense to split these checks in multiple test cases?

jturner314: Yeah, I agree. I've split it up.

LukeMathWalker:
Merging :+1:

Issue #652, opened 2019-07-01
Kronecker product


ethanhs:
Myself (and another user on the rust discord) have run into the need for the kronecker product of two matrices.

I currently have code like:
```rust
fn kron(a: &Array2<f64>, b: &Array2<f64>) -> Array2<f64> {
    let dima = a.shape()[0];
    let dimb = b.shape()[0];
    let dimout = dima * dimb;
    let mut out = Array2::zeros((dimout, dimout));
    for (mut chunk, elem) in out.exact_chunks_mut((dimb, dimb)).into_iter().zip(a.iter()) {
        chunk.assign(&(*elem * b));
    }
    out
}
```
(note: in my code I only work on unitary matrices, so this is not general in a few ways)

It would be nice to have a general implementation of this in ndarray (numpy as `np.kron`).

Jvanrhijn:
I myself have run into a the need for an outer product of two vectors, which is of course a special case of the kronecker product. Looks like this used to exist in [older versions of ndarray_linalg](https://docs.rs/ndarray-linalg/0.2.1/x86_64-apple-darwin/ndarray_linalg/vector/fn.outer.html), but was removed at some point.

On a related note, numpy also has support for some other outer operations, i.e. if `B` is a binary operation between two elements of a field, and `a` and `b` are vectors over this field, then `B(a, b)[i, j] == B(a[i], b[j])`. This would be quite nice to have in ndarray as well. something like

~~~rust
fn outer_operation<T>(op: impl Fn(T, T) -> T, a: &Array1<T>, b: &Array1<T>) -> Array2<T>
~~~

LukeMathWalker:
@termoshtt can you provide the rationale behind the removal of `outer` from `ndarray-linalg`?

Kronecker product looks like a good addition - I have added the `good first issue` label to the issue and linked it in #597. If you fancy submitting a PR we can work on it from there :+1:

notmgsk:
After some failed attempts I was able to come up with a working generic implementation of `kron`, starting from the snippet above (@ethanhs)

```
use ndarray::{arr2, Array2, LinalgScalar};
use num_complex::Complex64;

pub fn kron<T>(a: &Array2<T>, b: &Array2<T>) -> Array2<T>
where
    T: LinalgScalar,
{
    let dima = a.shape()[0];
    let dimb = b.shape()[0];
    let dimout = dima * dimb;
    let mut out = Array2::zeros((dimout, dimout));
[...]

PR #690, opened 2019-08-20
[WIP] Outer product implementation

krtab:
Hi !
This is my first contribution to ndarray ! :)
This deals with issue #652.

There are two proposed function here, more or less general. I think that a lot of choices can be made (should the `f` function be able to work on chunks or not), and efficiency wise, I am not too sure about what I have. Anyhow, I think it is a nice starting point, so let me know what you think about it, whether interface or implementation wise.

Also I didn't really know where to put it, so if you want to move the code somewhere else, please do.

### DIFF
--- a/src/linalg/impl_linalg.rs
+++ b/src/linalg/impl_linalg.rs
@@ -9,7 +9,7 @@
 use crate::imp_prelude::*;
 use crate::numeric_util;
 
-use crate::{LinalgScalar, Zip};
+use crate::{ErrorKind, LinalgScalar, Zip};
 
 use std::any::TypeId;
 use std::mem::MaybeUninit;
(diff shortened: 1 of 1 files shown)


src/linalg/impl_linalg.rs
```diff
@@ -815,3 +815,129 @@ mod blas_tests {
...
+
+#[allow(dead_code)]
+fn kron_to_dyn<Sa, I, Sb, T>(a: &ArrayBase<Sa, IxDyn>, b: &ArrayBase<Sb, I>) -> Array<T, IxDyn>
+where
+    T: Copy,
+    Sa: Data<Elem = T>,
+    Sb: Data<Elem = T>,
+    I: Dimension,
+    T: crate::ScalarOperand,
+    for<'a> &'a ArrayBase<Sb, I>: std::ops::Mul<T, Output = Array<T, I>>,
```
bluss: This seems inefficient since it's using a multiplication that produces an Array result, can we use a reusable scratch array for this or some other way to make this mostly in-place?

Maybe using Zip to make the assignment directly without using assign

krtab: Hi !

Yes indeed, this is not the most efficient. Yet, I haven't really found a function allowing me to directly perform a multiplication into a specified array/buffer. A possibility would be to use `MulAssign` but then I would have to add a pass to copy the first matrix and then mulassign each chunk, which I'm not sure would be more efficient.

> Maybe using Zip to make the assignment directly without using assign

Not sure of what you mean here.

Let me know what you think, and thanks for your review.

krtab: Hey @bluss, I'm allowing myself to bump the discussion, do you have any opinion, let me know if something isn't clear.

krtab:
This new API is both easier to use and also takes into account the comments of @bluss. Apart from function names, I'm satisfied with it. Let me know if you have any comments, and I'll start writing documentation!

krtab:
@LukeMathWalker @jturner314 I'm not sure @bluss  is around anymore, do you have any input ?

bluss:
I won't be around much, but I'm trying to get 0.13.1 done when I'm here, and maybe get started on the next release.

Why do we have both generic dim and dynamic dim functions? If you replace the dimension type parameter with IxDyn, both functions have the same type signature. That should mean we only need one function of each kind.

Your methods are not callable by crate users, because they are not public. I suppose that tells us this is a draft PR, and that's ok.

bluss:
In general, should this be in ndarray-linalg or here? I think as much as possible can go in ndarray-linalg, but have we defined the boundaries? Since this is more about the dimensionality of the operation and less about the linalg, maybe the ndarray crate is a good fit(?).  Maybe @LukeMathWalker knows

src/linalg/impl_linalg.rs
```diff
@@ -815,3 +815,132 @@ mod blas_tests {
...
+
+    let res_ndim = std::cmp::max(a.ndim(), b.ndim());
+    let res_dim: Vec<Ix> = a_shape_iter
+        .zip(b_shape_iter)
+        .take(res_ndim)
+        .map(|(x, y)| x * y)
+        .collect();
+
+    unsafe {
+        let mut res: ArrayD<T> = ArrayBase::uninitialized(res_dim);
```
bluss, line 843: If you have time, this can be ported to `Array::maybe_uninit` now. Look into using `Zip::apply_assign_into` - see example in tests?  I'd still use `T: Copy` as the restriction, as long as we don't handle drop of partially filled arrays on panics.

src/linalg/impl_linalg.rs
```diff
@@ -815,3 +815,132 @@ mod blas_tests {
...
+    a: &ArrayBase<Sa, I>,
+    b: &ArrayBase<Sb, I>,
+    f: F,
+) -> Array<T, I>
+where
+    T: Copy,
+    Sa: Data<Elem = T>,
+    Sb: Data<Elem = T>,
+    I: Dimension,
+    F: Fn(T, T) -> T,
```
bluss, line 876: Using FnMut is strictly more powerful, and we should do that, unless we have some other goal in mind. Parallelization? Then it needs to be Send/Sync and so on, so maybe it's not that.

krtab:
Hi @bluss,

Thanks for your comments.

> Why do we have both generic dim and dynamic dim functions? If you replace the dimension type parameter with IxDyn, both functions have the same type signature. That should mean we only need one function of each kind.

I made it so because in general, if you take the kronecker product of two matrices, one with n dimensions and the other with m dimensions, the dimensions of the result will be max(n,m). As the "max" operation is not implemented on types having the Dimension Trait, we have to loose the dimensionality information at the type-level. But the special case were both matrices have the same dimensionality is quite common, and allows us to preserve the dimensionality info at the type-level, hence my choice of two implementations.

> Your methods are not callable by crate users, because they are not public. I suppose that tells us this is a draft PR, and that's ok.

Yes, in general I think it should be up to you the crate maintainers to decide the API
I have implemented the modifications for both `MaybeUninit`, `FnMut` and `apply_assign_into`.

> I'd still use T: Copy as the restriction, as long as we don't handle drop of partially filled arrays on panics.

I'm not sure I fully understand what you mean.

In general, I think you should double check my unsafe code, I'm pretty new to this so I'm not so confident it is ok.

bluss:
In one case the dimensionality inputs are D and D (where D is a type parameter for the dimension) and the output dimensionality is D.

For the dynamic case, the dim of the inputs are IxDyn, IxDyn and the output has type IxDyn, so it looks like it can be handled by the first case (substitute D = IxDyn and they are the same). There might be something I'm missing.

src/linalg/impl_linalg.rs
```diff
@@ -823,3 +824,126 @@ mod blas_tests {
...
+where
+    T: Copy,
+    Sa: Data<Elem = T>,
+    Sb: Data<Elem = T>,
+    I: Dimension,
+    F: FnMut(T, T) -> T,
+{
+    let mut res_dim = a.raw_dim();
+    let mut res_dim_view = res_dim.as_array_view_mut();
+    res_dim_view *= &b.raw_dim().as_array_view();
```
bluss, line 886: Obviously I kind of love this, that we can use array methods on dimensions 🙂.

The unsafe code looks good, we only need to prove that all elements are assigned to, and they will be if the exact chunks don't leave any uneven remainder. And that looks good to me, B's shape evenly divides the result's shape, right?

bluss: Ah no, we also have to check the product. We need to use a saturating multiplication, otherwise we can overflow, and then the above doesn't hold?

src/linalg/impl_linalg.rs
bluss: Nice!

krtab:
> In one case the dimensionality inputs are D and D (where D is a type parameter for the dimension) and the output dimensionality is D.
> 
> For the dynamic case, the dim of the inputs are IxDyn, IxDyn and the output has type IxDyn, so it looks like it can be handled by the first case (substitute D = IxDyn and they are the same). There might be something I'm missing.
> 
> Edit: (Reading more) seems like it will work, both operations are really the same, just adapting to the dimensionality. Making one interface with type parameter D will work, but it might be more tricky to code - and be like `if <dynamic dimensional> { ... } else { .. } `

It is indeed quite tricky to code : I think the problem can be seen as: what should the type of `res_dim` be:
- If we are working with dynamic dimensions we can collect into `Vec<Ix>` which is `IntoDimension<Dim=IxDyn>`, this allows us to leverage all the std iterator functions and even perform the checked multiplication that is needed as you pointed out.
- If we are working with static Dimension, then res_dim should have the same type as the input ones, which is `[Ix,n]` (or a n-tuple of `Ix` which has the right `IntoDimension`). But we can't `collect()` into an `[Ix,n]` so that's why I had used the array multiplication trick. Unfortunately this trick won't allow the overflow check so we will have to do without it.

I ended up with the current solution which could replace the zeros with some uninitialized mem but that would require unsafe code I think.

I can also try to optimize a bit for the case where the number of dimensions are the same.

bluss:
I think both problems can be solved 😸 . Since I've written most of ndarray I have done a fair bit of weird dimension hacks. Not sure if it needs to be solved right now, I can't commit that much time right now, otherwise I'd do it.

Creating a new dimension value: `D::zeros(n)`. Iterating it and filling it with its values by using `.slice_mut()` (shared ref iteration, `.slice()`).

src/linalg/impl_linalg.rs
```diff
@@ -823,3 +824,190 @@ mod blas_tests {
...
+                Some(n) => *res_dim_elem = n,
+                None => return Err(ErrorKind::Overflow),
+            }
+        }
+        res_dim
+    };
+
+    // Reshape input arrays to compatible shapes
+    let a_reshape = a.view().into_shape(a_shape_completed).unwrap();
+    let b_reshape = b.view().into_shape(b_shape_completed.clone()).unwrap();
```
bluss, line 889: into_shape is unfortunately not general enough to always succeed here :( I'm sorry if I have pushed you over to this solution with into_shape. We'll need to think about what we can do, because we know the completed shape is compatible.

src/linalg/impl_linalg.rs
```diff
@@ -823,3 +824,190 @@ mod blas_tests {
...
+    //Create and fill the result array
+    let mut res: Array<MaybeUninit<T>, D> = ArrayBase::maybe_uninit(res_dim);
+    let res_chunks = res.exact_chunks_mut(b_shape_completed);
+    Zip::from(res_chunks)
+        .and(a_reshape)
+        .apply(|res_chunk, &a_elem| {
+            Zip::from(&b_reshape)
+                .apply_assign_into(res_chunk, |&b_elem| MaybeUninit::new(f(a_elem, b_elem)))
+        });
+    // This is safe because the exact chunks covered exactly the res
```
bluss, line 900: nice

src/linalg/impl_linalg.rs
krtab: What's the issue with it? The layout constraint?

bluss: It only supports c/f-contiguous arrays, unfortunately.

bluss: The shape completed code can be much simpler I guess. Make views `av` and `bv` and use `.insert_axis()` until they both have the same number of axes.

bluss: hm no, insert_axis changes the type, that's no good :(

bluss: One can check if it's a dyn dimension - then convert to explicitly typed `ArrayD` and convert back to `Array<_, D>` later with `.into_dimensionality()`? Is that truly the best way? Not sure.

krtab:
I'm closing this as too old. If anyone reworks on this feel free to let me know.

PR #691, opened 2019-08-20
Add .split_at() methods for AxisChunksIter/Mut

jturner314:
This adds `.split_at()` methods for `AxisChunksIter` and `AxisChunksIterMut`. Once this is merged, it will be straightforward to implement #639 in terms of `.split_at()`.

### DIFF
--- a/src/iterators/mod.rs
+++ b/src/iterators/mod.rs
@@ -1182,9 +1182,13 @@ impl<'a, A, D: Dimension> NdProducer for AxisIterMut<'a, A, D> {
 /// See [`.axis_chunks_iter()`](../struct.ArrayBase.html#method.axis_chunks_iter) for more information.
 pub struct AxisChunksIter<'a, A, D> {
     iter: AxisIterCore<A, D>,
-    n_whole_chunks: usize,
-    /// Dimension of the last (and possibly uneven) chunk
-    last_dim: D,
+    /// Index of the partial chunk (the chunk smaller than the specified chunk
+    /// size due to the axis length not being evenly divisible). If the axis
+    /// length is evenly divisible by the chunk size, this index is larger than
+    /// the maximum valid index.
+    partial_chunk_index: usize,
+    /// Dimension of the partial chunk.
+    partial_chunk_dim: D,
     life: PhantomData<&'a A>,
 }
 
@@ -1193,10 +1197,10 @@ clone_bounds!(
     AxisChunksIter['a, A, D] {
         @copy {
             life,
-            n_whole_chunks,
+            partial_chunk_index,
         }
         iter,
-        last_dim,
+        partial_chunk_dim,
     }
 );
 
@@ -1233,12 +1237,9 @@ fn chunk_iter_parts<A, D: Dimension>(
     let mut inner_dim = v.dim.clone();
     inner_dim[axis] = size;
 
-    let mut last_dim = v.dim;
-    last_dim[axis] = if chunk_remainder == 0 {
-        size
-    } else {
-        chunk_remainder
-    };
+    let mut partial_chunk_dim = v.dim;
+    partial_chunk_dim[axis] = chunk_remainder;
+    let partial_chunk_index = n_whole_chunks;
 
     let iter = AxisIterCore {
         index: 0,
@@ -1249,16 +1250,16 @@ fn chunk_iter_parts<A, D: Dimension>(
         ptr: v.ptr,
     };
 
-    (iter, n_whole_chunks, last_dim)
+    (iter, partial_chunk_index, partial_chunk_dim)
 }
 
 impl<'a, A, D: Dimension> AxisChunksIter<'a, A, D> {
     pub(crate) fn new(v: ArrayView<'a, A, D>, axis: Axis, size: usize) -> Self {
-        let (iter, n_whole_chunks, last_dim) = chunk_iter_parts(v, axis, size);
+        let (iter, partial_chunk_index, partial_chunk_dim) = chunk_iter_parts(v, axis, size);
         AxisChunksIter {
             iter,
-            n_whole_chunks,
-            last_dim,
+            partial_chunk_index,
+            partial_chunk_dim,
             life: PhantomData,
         }
     }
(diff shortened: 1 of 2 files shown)


src/iterators/mod.rs
```diff
@@ -1182,9 +1195,13 @@ impl<'a, A, D: Dimension> NdProducer for AxisIterMut<'a, A, D> {
...
 pub struct AxisChunksIter<'a, A, D> {
     iter: AxisIterCore<A, D>,
-    n_whole_chunks: usize,
-    /// Dimension of the last (and possibly uneven) chunk
-    last_dim: D,
+    /// Index of the partial chunk (the chunk smaller than the specified chunk
+    /// size due to the axis length not being evenly divisible). If the axis
+    /// length is evenly divisible by the chunk size, this index is larger than
+    /// the maximum valid index.
+    partial_chunk_index: usize,
```
LukeMathWalker: Would it be beneficial to rephrase this as an `Option`, to make it clearer that we might (or might not) have a partial chunk? Something along the lines of:
```rust
pub struct AxisChunksIter<'a, A, D> {
   iter: AxisIterCore<A, D>,
   partial_chunk: Option<PartialChunk>,
   life: PhantomData<&'a A>
}

struct PartialChunk {
   partial_chunk_index: usize,
   partial_chunk_dim: D
}
```

src/iterators/mod.rs
```diff
@@ -1270,30 +1284,49 @@ macro_rules! chunk_iter_impl {
...
+                (
+                    Self {
+                        iter: left,
+                        partial_chunk_index: self.partial_chunk_index,
+                        partial_chunk_dim: self.partial_chunk_dim.clone(),
+                        life: self.life,
+                    },
+                    Self {
+                        iter: right,
+                        partial_chunk_index: self.partial_chunk_index,
```
bluss: I haven't read the whole code unfortunately (what's not visible in the diff) - why doesn't this partial_chunk_index require adjusting - the right part of the iter now starts at `index`, so I'd expect this to be offset by `- index`?

jturner314: Here's an example:

```rust
use ndarray::prelude::*;

fn main() {
    let a: Array1<i32> = (0..13).collect();
    let mut iter = a.axis_chunks_iter(Axis(0), 3);
    iter.next();  // skip the first element so that we consider a partially-consumed iterator
    println!("before_split = {:#?}", iter);
    let (left, right) = iter.split_at(2);
    println!("left = {:#?}", left);
    println!("right = {:#?}", right);
}
```

which gives the output

```text
before_split = AxisChunksIter {
    iter: AxisIterCore {
        index: 1,
        end: 5,
        stride: 3,
        inner_dim: [3],
        inner_strides: [1],
        ptr: 0x00005634af728b40,
    },
    partial_chunk_index: 4,
    partial_chunk_dim: [1],
    life: PhantomData,
}
left = AxisChunksIter {
    iter: AxisIterCore {
        index: 1,
        end: 3,
        stride: 3,
        inner_dim: [3],
        inner_strides: [1],
        ptr: 0x00005634af728b40,
    },
    partial_chunk_index: 4,
    partial_chunk_dim: [1],
    life: PhantomData,
}
right = AxisChunksIter {
    iter: AxisIterCore {
        index: 3,
        end: 5,
        stride: 3,
        inner_dim: [3],
        inner_strides: [1],
        ptr: 0x00005634af728b40,
    },
    partial_chunk_index: 4,
    partial_chunk_dim: [1],
    life: PhantomData,
}
```

We can visualize the situation like this:

```
               0 1 2 3 4
before split:    ^      |
after split:     ^  |^  |
```

The `^`s represent the `index`es and the `|`s represent the `end`s of the iterators. (The `|`s appear just before the corresponding `end` indices.) There are 4 full chunks (indices `0..=3`) and 1 partial chunk (index `4`). Note that all indices are relative to the start of the axis, so any given index value represents the same location before and after the split. This is why `partial_chunk_index` is the same before and after splitting. Before splitting, the index of the partial chunk is `4`, and it stays `4` in the split pieces. (The left piece will never actually reach index `4` since its `end` is `3`; that's the desired behavior.)

src/iterators/mod.rs
jturner314: I don't think it makes sense to use *both* the `Option` variant and the value of `partial_chunk_index` to represent whether or not there's a partial chunk. (The biggest reason is that I prefer data structures where there's a single source of truth, rather than having to keep multiple things in sync. There might also be a small performance cost to accessing `partial_chunk_index` through the `Option` (since accessing it requires checking whether the `Option` is the `Some` variant), but we'd need to test to determine if that would really be noticeable.) IMO, putting the fields in an `Option` would be additional complication over the current approach without much benefit.

It would be reasonable to eliminate `partial_chunk_index` and just use the `Option` variant to represent the presence of a partial chunk, like this:

```rust
pub struct AxisChunksIter<'a, A, D> {
   iter: AxisIterCore<A, D>,
   partial_chunk: Option<D>,
   life: PhantomData<&'a A>
}
```

or to always store the shape of the last chunk (regardless of whether or not it's a partial chunk):

```rust
pub struct AxisChunksIter<'a, A, D> {
   iter: AxisIterCore<A, D>,
   last_chunk_dim: D,
   life: PhantomData<&'a A>
}
```

These approaches have two disadvantages since they rely on checking whether the iterator is at its end to handle the partial chunk instead of checking whether the current index is equal to `partial_chunk_index`:

1. `.split_at()` needs to check whether or not the partial chunk is in the left piece and determine `partial_chunk` or `last_chunk_dim` of the left piece accordingly. (The partial chunk is in the left piece when `index == self.iter.len()`.)

2. `.next_back()` needs to set `partial_chunk` to `None` or `last_chunk_dim` to `self.iter.inner_dim` each time it's called.

So, I'd rather keep the current approach and add more comments if necessary to make it clear.

src/iterators/mod.rs
bluss: Thanks. If the only use of index is counting up to the partial_chunk_index, it makes total sense.

jturner314: `index` is also used in `AxisIterCore` (which `AxisChunksIter` wraps) to compute the pointer of each element/chunk and to check for the end of the iterator; see `AxisIterCore`'s implementation of `.next()`. (`.split_at()` on `AxisIterCore` doesn't change the `ptr` value; `ptr` always corresponds to the start of the axis. This was part of #669.)

jturner314:
Thanks for reviewing this @bluss! I generally like to get a review from someone before merging, but thanks for the vote of confidence. I'm comfortable merging my own PRs without a review when necessary.

Issue #688, opened 2019-08-18
into_scalar() for 0-dimensional array views


Palladinium:
`Array0` has the handy `into_scalar()` method, which from what I gather roughly equates to calling `arr.index(Ix0())`.

Unfortunately, this isn't available for array views, so we're stuck using the more obscure and verbose syntax for `ArrayView0` and `ArrayViewMut0`. It looks to me like it'd just be a matter to move the `into_scalar()` impl up to `ArrayBase`, letting the view representation choose the return type.

Would it be a breaking change? How difficult would it be to implement?

bluss:
`arr.index(Ix0())` is the syntax that looks obscure and verbose to me, I'm not sure why you'd use that.

The standard way of accessing an element in a 0-dim array is to use a 0-dim index and using indexing notation:

```rust
// Alternative A
let element = array[[]];

// Alternative B
let element = array[()];
```

jturner314:
I think @Palladinium may be referring to the `.index()` method from the `IndexLonger` trait. It's currently the best way to get an `&'a A` from an `ArrayView0<'a, A, D>`. (Note the matching lifetimes.)

I do think it would make sense to add `.into_scalar()` methods to `ArrayView0` and `ArrayViewMut0` for convenience. (Those types specifically, not all of `ArrayBase<S, Ix0>`.) It would not be a breaking change, and it would be easy to implement.

LukeMathWalker:
I opened a PR to close this issue: #700

PR #700, opened 2019-09-08
Add into_scalar for ArrayView0 and ArrayViewMut0

LukeMathWalker:
This addresses #688.

I have taken the chance to re-organize `impl_views`, converting it into a folder sub-module to then split it into smaller files with a functionality-related name. 
I think this makes it easier to navigate the crate (and we should probably do the same for other very big files we have in the repository) but I am ok to revert the changes if you believe otherwise.


src/impl_views/conversions.rs
```diff
@@ -56,6 +57,36 @@ where
...
+    /// // `Foo` doesn't implement `Clone`.
+    /// #[derive(Debug, Eq, PartialEq)]
+    /// struct Foo;
+    ///
+    /// let array: Array0<Foo> = arr0(Foo);
+    /// let view = array.view();
+    /// let scalar: &Foo = view.into_scalar();
+    /// assert_eq!(scalar, &Foo);
+    /// ```
+    pub fn into_scalar(self) -> &'a A {
```
bluss: the ArrayView doesn't need to be consumed to allow this (infuriating difference between shared and mut for IndexLonger). I guess one could just as well make this a &self method, it makes difference for non-copy ArrayViews?

bluss:
Looks good. Continual splitting is good I think. Impl_views isn't even large in ndarray :smile:

src/impl_views/conversions.rs
LukeMathWalker: I thought about it, but then I opted to keep the signature consistent with the other two (`Array`/`ArrayViewMut`). Given that it's named `into_*`, it sounded preferable to consume the input.
On the other hand, taking `&self` makes it more convenient for `ArrayView` because you can keep using `self` afterwards.

I don't know, I would lean on the side of symmetry in this case.

bluss: ok, that works.

bluss:
Tip, use the https://help.github.com/en/articles/closing-issues-using-keywords names. With "Fixes #688." it closes the issue automatically. I prefer to put these in the PR text (not in the commit log because rebasing and re-pushing causes spam on the issue).

LukeMathWalker:
Nifty trick, thanks!

Issue #698, opened 2019-09-04
Rename genrows()?


bluss:
`.genrows()` has an unfortunate name. It's supposedly the *generalized rows* of an array. The reason for the tricky name, was that `.rows()` was taken - by the row count.

We could rename `rows` →`row_count` and `genrows` → `rows` to solve this, ideally with deprecations, so in a two version cycle.

`rows()` is a more natural name for the ndproducer and iterable for rows, for example because the user doesn't have to care that it's “generalized” rows if they only use 2D arrays — they don't need to know and it doesn't matter.

The main problem with the `genrows` name is that it is non-intuitive to guess to type and guess what it means.

Note that we chose the `.rows()` name for the row count to match the sprs crate, but I think we can choose to go our own way here, if we have a good reason.

An alternative to the `row_count` name is `nrows()` — anyone have a survey of what other crates prefer apart from sprs?

NOTE: The same story applies to `gencolumns`.

jturner314:
I agree that `.genrows()` has an unfortunate name. It's funny, I've always read it as "generate rows" instead of "generalized rows", which confused me because none of the other iterator/producer methods had the `gen` prefix. `.rows()` does seem like a better name.

Other crates:

* `nalgebra`: `nrows` and `ncols` are the counts. (`rows` and `columns` are slicing operations. `row_iter` and `column_iter` are the iterators.)

* `rulinalg`: `rows` and `cols` are the counts. (`row_iter` and `col_iter` are the iterators.)

* `sprs`: `rows` and `cols` are the counts.

My preference for the counts is `nrows` and `ncols`/`ncolumns` because these names are clear while being shorter than `*_count`.

LukeMathWalker:
> I agree that `.genrows()` has an unfortunate name. It's funny, I've always read it as "generate rows" instead of "generalized rows"

Same here :sweat_smile: 

> My preference for the counts is `nrows` and `ncols`/`ncolumns` because these names are clear while being shorter than `*_count`.

:+1:

PR #701, opened 2019-09-08
Add .nrows(), .ncols() and old names .rows(), .cols() are now deprecated

bluss:
.rows()/.cols() are renamed to nrows/ncols respectively, to make space
for putting the row iterable/ndproducer in this place instead.

The deprecated methods will be removed in the release after they were
first deprecated.

These new names and the deprecations are the first step towards renaming genrows/gencolumns  #698

### DIFF
--- a/examples/life.rs
+++ b/examples/life.rs
@@ -59,7 +59,7 @@ fn iterate(z: &mut Board, scratch: &mut Board) {
 
 fn turn_on_corners(z: &mut Board) {
     let n = z.nrows();
-    let m = z.cols();
+    let m = z.ncols();
     z[[1, 1]] = 1;
     z[[1, m - 2]] = 1;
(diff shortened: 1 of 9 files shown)


examples/life.rs
```diff
@@ -58,7 +58,7 @@ fn iterate(z: &mut Board, scratch: &mut Board) {
 }
 
 fn turn_on_corners(z: &mut Board) {
-    let n = z.rows();
+    let n = z.nrows();
     let m = z.cols();
```
LukeMathWalker: This `cols` call escaped your purge @bluss :eyes:

jturner314: I fixed that one and another instance I found.

bluss:
:star2: teamwork :1st_place_medal: Thanks!

PR #703, opened 2019-09-09
ndarray-rand documentation fixes

bluss:
Update examples in documentation to

1. skip extern crate lines which are not needed now
2. Use rand_distr since using the code in the old examples would give deprecation warnings

### DIFF
--- a/ndarray-rand/src/lib.rs
+++ b/ndarray-rand/src/lib.rs
@@ -50,10 +50,6 @@ where
     /// overflows usize.
     ///
     /// ```
-    /// extern crate rand;
-    /// extern crate ndarray;
-    /// extern crate ndarray_rand;
-    ///
     /// use rand::distributions::Uniform;
     /// use ndarray::Array;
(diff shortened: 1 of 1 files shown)


bluss:
This is in preparation of imminently releasing the version bump of ndarray-rand which supports ndarray 0.12 and rand 0.7.

I'm pushing this through after being a bit "offended" at seeing that ndarray-rand was still not published with a new version. I guess I should direct this to myself. But also to make a good example.. let's release ndarray-rand often..

Talking about my activity in the repo, not just this PR of course - hear me out - these spurts of activity on github are not something I can or want to do, even if I love it. I'll try to make regular contributing a habit, but it too easily gets very intense, so I have to back off a bit.

bluss:
Makes #659 available

bluss:
Since it conflicts with #663, I'll wait for that

bluss:
Superseded by the rand_distr PR

PR #707, opened 2019-09-11
Use alternate form flag (#) to disable overflow with ellipsis

### LINKED ISSUE #705: Provide a way to print the whole array
#606 implemented some sweet ellipsizing for arrays, but there's no way to print the whole array and we need to provide for it.

We don't use the alternate flag (#) for anything at the moment, so we could use this for forcing the full display.

I also think that the default ellipsizing should be far more generous than just three elements on each size of the fold - a larger number would be convenient for most users.

jturner314:
Currently, there is no way to print all the elements of an array if one of the axis lengths exceeds the limit before overflowing with an ellipsis. This PR makes it so that if the alternate form flag (`#`) is specified, the axis length limit is disabled. It also simplifies the array formatting implementation.

Fixes #705.

### DIFF
--- a/src/arrayformat.rs
+++ b/src/arrayformat.rs
@@ -11,53 +11,54 @@ use std::fmt;
 
 const PRINT_ELEMENTS_LIMIT: Ix = 3;
 
-fn format_1d_array<A, S, F>(
-    view: &ArrayBase<S, Ix1>,
+const ELLIPSIS: &str = "...";
+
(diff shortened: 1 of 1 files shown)


bluss:
Thanks! Do you think we can increase the limit? For 1D/2D arrays the current limit seems a bit low, but I know the number of elements we show scales up pretty hard for high dimensional arrays.

PR #713, opened 2019-09-15
Increase limits for overflow in formatting

bluss:
Use a total element count limit, before abbreviating the format output.
This means "small" arrays are always printed in full.

Use separate limits for the last axis, the next to last axis and the
other axes before that.

The last two axes are printed like a matrix of elements which is why
they show more elements.

The other axes are "stacked" and printed by repeating the inner two axes
(like stacked 2D arrays). For this reason they are harder to read
already, and we use a lower limit.

Limits:

- Array is printed in full below 500 elements
- Max row length otherwise: 11
- Max column length otherwise: 11
- Max other axis length otherwise: 6.

These are further fixes on top of #705 

Future features can be added to make the limits user configurable, but it is not
a priority to do that before the next release.


bluss:
What's everyone's thoughts on this. It's tweaked but similar to jturner's idea in #705

bluss:
This should be a draft PR, because tests and formatting remain to be updated.

LukeMathWalker:
> What's everyone's thoughts on the limits? It's tweaked but similar to jturner's idea in #705

It looks good to me - the only one I am slightly worried about is printing the array in full if we have less than 500 elements. 
Consider printing a 2D array with shape `(1, 499)` in a terminal... Not so nice.
Should we perhaps consider a lower number (e.g. `100`), given that we are allowing the user to print the array in full anyway using `#` if they want to?

bluss:
numpy thinks it is nice to print up to 1000 elements, so I'm not sure where to draw the line. Although of course, they line wrap if the lines get long. On the outset I think it doesn't matter if it's 100 or 500 - it will be longer than current linewidth anyway.

bluss:
I used cargo bloat to look at the reduction of code size for the last two commits. If someone is interested in formatting performance, maybe serialization is the right place to look then, but otherwise we can come back to optimizing it.

I wouldn't mind changing the rules or tweaking the numbers later. If we expose formatting parameters to users, it might be a good time to make the rules simple so that it uses few params.

PR #717, opened 2019-09-19
Add multi_slice_* methods (supports flat tuples only)

### LINKED ISSUE #687: move semantics for multislice!()

jturner314:
This is an alternative implementation of #716 that supports only flat tuples of slicing information. This means that it's no longer possible to perform arbitrarily many slices in a single call. The implementations are also limited only to the cases `(SliceInfo, SliceInfo, ...)` and `(&SliceInfo, &SliceInfo, ...)`, so slicing with e.g. `(SliceInfo, &SliceInfo, ...)` is not supported.

Despite the fact that the trait implementations are more limited than #716, I prefer this version because the implementation is simpler, it's easier to understand, and the unsafe code is more contained.

The implementation is done, but the tests need improvement.

Fixes #687. Closes #716.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -475,6 +475,13 @@ pub type Ixs = isize;
 /// [`.slice_move()`]: #method.slice_move
 /// [`.slice_collapse()`]: #method.slice_collapse
 ///
+/// It's possible to take multiple simultaneous *mutable* slices with the
+/// [`.multi_slice_mut()`] or (for [`ArrayViewMut`] only)
+/// [`.multi_slice_move()`].
+///
+/// [`.multi_slice_mut()`]: #method.multi_slice_mut
+/// [`.multi_slice_move()`]: type.ArrayViewMut.html#method.multi_slice_move
+///
 /// ```
 /// extern crate ndarray;
 ///
@@ -525,6 +532,20 @@ pub type Ixs = isize;
 ///                [12, 11, 10]]);
 /// assert_eq!(f, g);
 /// assert_eq!(f.shape(), &[2, 3]);
+///
+/// // Let's take two disjoint, mutable slices of a matrix with
+/// //
+/// // - One containing all the even-index columns in the matrix
+/// // - One containing all the odd-index columns in the matrix
+/// let mut h = arr2(&[[0, 1, 2, 3],
+///                    [4, 5, 6, 7]]);
+/// let (s0, s1) = h.multi_slice_mut((s![.., ..;2], s![.., 1..;2]));
+/// let i = arr2(&[[0, 2],
+///                [4, 6]]);
+/// let j = arr2(&[[1, 3],
+///                [5, 7]]);
+/// assert_eq!(s0, i);
+/// assert_eq!(s1, j);
 /// }
 /// ```
 ///
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -475,7 +475,7 @@ pub type Ixs = isize;
 /// [`.slice_move()`]: #method.slice_move
 /// [`.slice_collapse()`]: #method.slice_collapse
 ///
-/// It's possible to take multiple simultaneous *mutable* slices with the
+/// It's possible to take multiple simultaneous *mutable* slices with
 /// [`.multi_slice_mut()`] or (for [`ArrayViewMut`] only)
 /// [`.multi_slice_move()`].
 ///
--- a/src/slice.rs
+++ b/src/slice.rs
@@ -643,7 +643,7 @@ where
     /// The type of the slices created by `.multi_slice_move()`.
     type Output;
 
-    /// Slice the raw view into multiple raw views, and dereference them.
+    /// Split the view into multiple disjoint slices.
     ///
     /// **Panics** if performing any individual slice panics or if the slices
     /// are not disjoint (i.e. if they intersect).
--- a/src/slice.rs
+++ b/src/slice.rs
@@ -660,20 +660,6 @@ where
     fn multi_slice_move(&self, _view: ArrayViewMut<'a, A, D>) -> Self::Output {}
 }
 
-impl<'a, A, D, Do0> MultiSlice<'a, A, D> for (SliceInfo<D::SliceArg, Do0>,)
-where
-    A: 'a,
-    D: Dimension,
-    D::SliceArg: Sized,
-    Do0: Dimension,
-{
-    type Output = (ArrayViewMut<'a, A, Do0>,);
-
-    fn multi_slice_move(&self, view: ArrayViewMut<'a, A, D>) -> Self::Output {
-        (view.slice_move(&self.0),)
-    }
-}
-
 impl<'a, A, D, Do0> MultiSlice<'a, A, D> for (&SliceInfo<D::SliceArg, Do0>,)
 where
     A: 'a,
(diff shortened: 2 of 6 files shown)


src/lib.rs
```diff
@@ -473,6 +473,13 @@ pub type Ixs = isize;
 /// [`.slice_move()`]: #method.slice_move
 /// [`.slice_collapse()`]: #method.slice_collapse
 ///
+/// It's possible to take multiple simultaneous *mutable* slices with the
+/// [`.multi_slice_mut()`] or (for [`ArrayViewMut`] only)
+/// [`.multi_slice_move()`].
+///
+/// [`.multi_slice_mut()`]: #method.multi_slice_mut
+/// [`.multi_slice_move()`]: type.ArrayViewMut.html#method.multi_slice_move
```
bluss: I don't know about you, but I don't mind transitioning to nightly-only rustdoc links (because they render correctly in docs.rs).. Importance of rendering the links when the user generates docs on stable?

src/slice.rs
```diff
@@ -629,3 +630,135 @@ macro_rules! s(
...
+where
+    A: 'a,
+    D: Dimension,
+{
+    type Output = ();
+
+    fn multi_slice_move(&self, _view: ArrayViewMut<'a, A, D>) -> Self::Output {}
+}
+
+impl<'a, A, D, Do0> MultiSlice<'a, A, D> for (SliceInfo<D::SliceArg, Do0>,)
```
bluss: Do we need both the with-& and without impl? It seems like we don't need both for s![].

src/slice.rs
```diff
@@ -629,3 +630,135 @@ macro_rules! s(
...
+        impl<'a, A, D, $($Do,)*> MultiSlice<'a, A, D> for ($(SliceInfo<D::SliceArg, $Do>,)*)
+        where
+            A: 'a,
+            D: Dimension,
+            D::SliceArg: Sized,
+            $($Do: Dimension,)*
+        {
+            type Output = ($(ArrayViewMut<'a, A, $Do>,)*);
+
+            fn multi_slice_move(&self, view: ArrayViewMut<'a, A, D>) -> Self::Output {
```
bluss: If we keep both these impls, one of them could just forward to the other, I think?

src/slice.rs
```diff
@@ -629,3 +630,135 @@ macro_rules! s(
...
+                #[allow(non_snake_case)]
+                let ($($Do,)*) = self;
+
+                let shape = view.raw_dim();
+                assert!(!impl_multislice_tuple!(@intersects_self &shape, ($($Do,)*)));
+
+                let raw_view = view.into_raw_view_mut();
+                unsafe {
+                    ($(raw_view.clone().slice_move($Do).deref_into_view_mut(),)*)
```
bluss: So rather less important, but if we peel off one `$Do` from the end of the repeat, we can save one clone() call. And as it happens there are always at least 2 $Do.

src/lib.rs
jturner314: That would be fine, although I'm having trouble getting nightly-style links to work correctly. The following don't work:

```markdown
[`.multi_slice_mut()`]
```
```markdown
[`multi_slice_mut()`]
```

If I add an explicit destination, `multi_slice_mut` works fine, but I'm not sure how to make `multi_slice_move` work:

```markdown
This works:
[`.multi_slice_mut()`]

[`.multi_slice_mut()`]: ArrayBase::multi_slice_mut()


This doesn't:
[`.multi_slice_move()`]

[`.multi_slice_move()`]: ArrayViewMut::multi_slice_move()
```

src/slice.rs
jturner314: We don't need both if the user only uses `s![]` to construct slicing information, because `s![]` always evaluates to a reference. However, it is possible to create owned `SliceInfo` instances directly, so supporting tuples of them is convenient for the user. If a user has a `(SliceInfo, SliceInfo)` instance, they could always call `.multi_slice_mut((&tuple.0, &tuple.1))`, but that's not as nice as `.multi_slice_mut(tuple)`. Or, if they have two `SliceInfo` instances, they could call `.multi_slice_mut((&a, &b))`, but that's not quite as nice as `.multi_slice_mut((a, b))`.

I wouldn't be strongly opposed to removing support for multi-slicing with `(SliceInfo, SliceInfo)`, because I doubt it would be used often. However, there isn't much cost to keeping it in the PR another than a little more code and maybe slightly longer compile time.

src/slice.rs
jturner314: Yeah, that's a good idea. I've made the change.

src/slice.rs
jturner314: This is possible, but unfortunately the obvious approach (using a matcher like `($($but_last:ident,)* $last:ident,)`) doesn't work due to rust-lang/rust#24827/rust-lang/rust#42838. The best solution I see is to build the expression in macro parameters, like this:

```rust
// ...
                let raw_view = view.into_raw_view_mut();
                unsafe {
                    impl_multislice_tuple!(@build_slice_expr raw_view, ($($Do,)*), ())
                }
            }
        }
    };
    (@build_slice_expr $raw_view:ident, ($head:expr,), ($($slice:expr,)*)) => {
        ($($slice,)* $raw_view.slice_move($head).deref_into_view_mut(),)
    };
    (@build_slice_expr $raw_view:ident, ($head:expr, $($tail:expr,)*), ($($slice:expr,)*)) => {
        impl_multislice_tuple!(
            @build_slice_expr $raw_view, ($($tail,)*),
            (
                $($slice,)*
                $raw_view.clone().slice_move($head).deref_into_view_mut(),
            )
        )
    };
// ...
```

but I'm not sure it's worth the additional complexity, especially since that complexity is in an `unsafe` block. Maybe there's a better way to do it?

jturner314: I found a better way. (Just separate into "but last" and "last" in the original macro call; don't make the macro try to do it.) I've updated the PR.

src/slice.rs
jturner314: Fwiw, in the future, I'd like to change the parameter in all the slicing methods from `&SliceInfo<D::SliceArg, Do>` to `T: CanSlice<D>`. This simplifies handling of const versus dyn dimension types, avoids the need for extra type parameters on slicing methods when we add support for ellipses in slicing, and allows us to be generic over whether the slicing info is owned (`SliceInfo`) or a reference (`&SliceInfo`). This last advantage would also avoid the need for separate `MultiSlice` implementations for tuples of `SliceInfo` and tuples of `&SliceInfo`. (We'd implement `MultiSlice` for `(T1, T2, ...)` where `T1: CanSlice<D>, T2: CanSlice<D>, ...`.)

bluss: I lean towards you ain't gonna need it and having 95% of the functionality for just the one impl. Yes, I fear I might lean towards a foolish stinginess sometimes. (So I need guidance on how to productively make the crate both simple to understand and fast to compile.)

jturner314: Okay, I've updated the PR.

jturner314:
Thanks for reviewing this, @bluss. After your approval, I fixed a minor issue in the docs, added more tests, rebased on the latest `master`, and merged it.

bluss:
Great, now we have reason to go for 0.13.1 soon and I can manage the release.

I expect the work after that to be for 0.13.2, further non-breaks.

bluss:
I'm sorry that I haven't been around to manage 0.13.1, but it would be ok if others had. That said, maybe I can do it now.

PR #721, opened 2019-09-20
Prepare for 0.13.0 release

jturner314:
This updates the version numbers to 0.13.0. Once this is merged, I think we're ready to release 0.13.0!

cc #670

PR #722, opened 2019-09-20
Update `par_azip!()` syntax to match `azip!()`

bluss:
Update the syntax to match. First refactor `azip!()` slightly, so that we can reuse it wholesale from `par_azip`.

cc #670 
cc #721

### DIFF
--- a/src/parallel/zipmacro.rs
+++ b/src/parallel/zipmacro.rs
@@ -13,10 +13,12 @@
 /// This is a version of the [`azip`] macro that requires the crate feature
 /// `rayon` to be enabled.
 ///
+/// See the [`azip`] macro for more details about the macro syntax!
+///
 /// This example:
 ///
 /// ```rust,ignore
-/// par_azip!(mut a, b, c in { *a = b + c })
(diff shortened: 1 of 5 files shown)


src/parallel/zipmacro.rs
```diff
@@ -47,60 +49,13 @@
...
-    (@parse [index => $a:expr, $($aa:expr,)*] $t1:tt in $t2:tt) => {
-        $crate::par_azip!(@finish ($crate::Zip::indexed($a)) [$($aa,)*] $t1 in $t2)
-    };
-    // Build Zip Rule (no index)
-    (@parse [$a:expr, $($aa:expr,)*] $t1:tt in $t2:tt) => {
-        $crate::par_azip!(@finish ($crate::Zip::from($a)) [$($aa,)*] $t1 in $t2)
-    };
-    // Build Finish Rule (both)
-    (@finish ($z:expr) [$($aa:expr,)*] [$($p:pat,)+] in { $($t:tt)*}) => {
-        use $crate::parallel::prelude::*;
```
bluss: Didn't notice it before, but `par_apply` is an inherent method nowadays and no trait use is needed.

PR #724, opened 2019-09-22
Add lane sampling to ndarray-rand

LukeMathWalker:
While working on https://github.com/rust-ndarray/ndarray-examples/pull/2 I realised that a sampling method could actually fit quite nicely in `ndarray-rand`.

I'd like to relax the `Copy` bound on `A`, I need to look deeper into `select`'s implementation.

### DIFF
--- a/ndarray-rand/Cargo.toml
+++ b/ndarray-rand/Cargo.toml
@@ -16,6 +16,7 @@ keywords = ["multidimensional", "matrix", "rand", "ndarray"]
 [dependencies]
 ndarray = { version = "0.13", path = ".." }
 rand_distr = "0.2.1"
+quickcheck = { version = "0.9", default-features = false, optional = true }
 
 [dependencies.rand]
 version = "0.7.0"
@@ -23,6 +24,7 @@ features = ["small_rng"]
 
 [dev-dependencies]
 rand_isaac = "0.2.0"
+quickcheck = { version = "0.9", default-features = false }
 
 [package.metadata.release]
 no-dev-version = true
--- a/ndarray-rand/src/lib.rs
+++ b/ndarray-rand/src/lib.rs
@@ -29,12 +29,15 @@
 //! that the items are not compatible (e.g. that a type doesn't implement a
 //! necessary trait).
 
-use crate::rand::distributions::Distribution;
+use crate::rand::distributions::{Distribution, Uniform};
 use crate::rand::rngs::SmallRng;
+use crate::rand::seq::index;
 use crate::rand::{thread_rng, Rng, SeedableRng};
 
-use ndarray::ShapeBuilder;
+use ndarray::{Array, Axis, RemoveAxis, ShapeBuilder};
 use ndarray::{ArrayBase, DataOwned, Dimension};
+#[cfg(feature = "quickcheck")]
+use quickcheck::{Arbitrary, Gen};
 
 /// [`rand`](https://docs.rs/rand/0.7), re-exported for convenience and version-compatibility.
 pub mod rand {
@@ -59,9 +62,9 @@ pub mod rand_distr {
 /// low-quality random numbers, and reproducibility is not guaranteed. See its
 /// documentation for information. You can select a different RNG with
 /// [`.random_using()`](#tymethod.random_using).
-pub trait RandomExt<S, D>
+pub trait RandomExt<S, A, D>
 where
-    S: DataOwned,
+    S: DataOwned<Elem = A>,
     D: Dimension,
 {
     /// Create an array with shape `dim` with elements drawn from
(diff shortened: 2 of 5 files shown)


bluss:
Relaxing the Copy bound on select is a bigger task, I think I'd like to help (but feel free to work on it if you get to it). I consider it a bigger task, since it entails handling copying partial arrays and being able to tear down and drop the partial array if something panics during this operation. Definitely something it's about time to handle.

jturner314:
> Relaxing the Copy bound on select

Fwiw, `nditer` can do this. See [this example](https://github.com/jturner314/nditer/blob/de85403c526183f0c8b9ea1409f3e71415589f53/src/lib.rs#L546-L559); just replace `x * x` with `x.clone()`. Right now, this is implemented using external iteration (since it relies on `Vec`'s `from_iter` implementation), but I'll probably change it to internal iteration in the future. I really need to finish up `nditer` and officially release it...

The key idea is to collect the data into a `Vec` in whatever order you're iterating over it, and then convert that into an array with the correct shape and strides. `nditer` attempts to choose the best iteration order. However, the simplest (but somewhat slow) implementation of `select` would be something like this:

```rust
use ndarray::prelude::*;
use ndarray::RemoveAxis;

fn select<A, D>(arr: ArrayView<'_, A, D>, axis: Axis, indices: &[usize]) -> Array<A, D>
where
    A: Clone,
    D: Dimension + RemoveAxis,
{
    let data = indices
        .iter()
        .flat_map(|&index| arr.index_axis(axis, index).iter())
        .cloned()
        .collect();

    // TODO: Check that new length doesn't overflow `isize`.
    let mut dim = arr.raw_dim();
    dim[axis.index()] = indices.len();

    // TODO: Compute strides. `axis` is the outermost axis (largest stride),
    // and the remaining axes are in standard order.
    let strides = unimplemented!();

    Array::from_vec_dim_stride_unchecked(dim, strides, data)
}
```

LukeMathWalker:
> Relaxing the Copy bound on select is a bigger task, I think I'd like to help (but feel free to work on it if you get to it). I consider it a bigger task, since it entails handling copying partial arrays and being able to tear down and drop the partial array if something panics during this operation. Definitely something it's about time to handle.
> 
> The choice to use Copy to implement something that's fast and safe was a good one at the time.

I'll keep this mind, I was planning to have a look this weekend most likely.
I would suggest to evaluate/merge this independently of the `select` overhaul, then it's just a matter of dropping the bound here when that lands.

ndarray-rand/src/lib.rs
```diff
@@ -141,6 +246,39 @@ where
...
+        A: Copy,
+        D: RemoveAxis,
+    {
+        let indices: Vec<_> = if with_replacement {
+            let distribution = Uniform::from(0..self.len_of(axis));
+            (0..n_samples).map(|_| distribution.sample(rng)).collect()
+        } else {
+            sample(rng, self.len_of(axis), n_samples).into_vec()
+        };
+        self.select(Axis(0), &indices)
```
bluss: should not be Axis(0)

bluss: Now, maybe axis should be a parameter to the quickcheck tests

ndarray-rand/Cargo.toml
```diff
@@ -23,6 +23,7 @@ features = ["small_rng"]
 
 [dev-dependencies]
 rand_isaac = "0.2.0"
+quickcheck = { version = "0.8", default-features = false }
```
bluss: We can upgrade to quickcheck 0.9 I think - but, as a separate task from this PR?

ndarray-rand/src/lib.rs
LukeMathWalker: Rookie mistake, and having it as a test parameter would have indeed caught it.

bluss:
Sorry for the unstructured critique, but how do we decide if it's best with separate methods for independent sampling and sampling with replacement, or to have them in one? Looks like a logical addition anyhow.

LukeMathWalker:
> Sorry for the unstructured critique, but how do we decide if it's best with separate methods for independent sampling and sampling with replacement, or to have them in one? Looks like a logical addition anyhow.

I don't have a strong preference in either direction.
If we make them separate, I can see we will end up with quite verbose names.
If we fold them in a single one (as in this PR), then there is an option that you have to specify every time. But maybe it's good, but you get to think about it?

ndarray-rand/src/lib.rs
```diff
@@ -29,11 +29,12 @@
 //! that the items are not compatible (e.g. that a type doesn't implement a
 //! necessary trait).
 
-use crate::rand::distributions::Distribution;
+use crate::rand::distributions::{Distribution, Uniform};
 use crate::rand::rngs::SmallRng;
+use crate::rand::seq::index::sample;
```
jturner314: I'd shorten this to `use crate::rand::seq::index`, and in the code use `index::sample`. Just seeing `sample` in the code is somewhat confusing because it's not clear what "sample" means.

LukeMathWalker:
I lean towards a single method with a parameter - I changed it from a `bool` to an enum (`SamplingStrategy`).

I have added a `quickcheck` feature-flag to provide an `Arbitrary` implementation for it (something I'd like to do for `ndarray` as well at a certain point) and I realized we were not running `ndarray-rand` tests in CI - hence I changed the script to make sure we do.

LukeMathWalker:
I am planning to start working on another convenience function for `ndarray-rand` (`shuffle_axis`, similar to [`SliceRandom::shuffle`](https://docs.rs/rand/0.7.2/rand/seq/trait.SliceRandom.html#tymethod.shuffle) from `rand`) - I am going to need the changes I have already done in this PR (adding an extra type parameter to `RandomExt`, fixing CI for `ndarray-rand`, etc.).

What is missing for this PR to be merged? Should I branch from here or should I branch from master? Just conscious of wasting time in merge conflicts :sweat_smile:

bluss:
This needs rebase/fixup :)  Otherwise it looks good — I won't have much opinions on ndarray-rand. So I'd say, branch off from this branch, but not in the state it is now.

quickcheck as a public dependency is not something I'd like to have for ndarray.

LukeMathWalker:
I was going for a squash and merge, but rebasing sounds good as well :) 

> quickcheck as a public dependency is not something I'd like to have for ndarray.

Could you elaborate?
My idea/proposal is mostly a result of the orphan rule - as far as I understand it, it makes it impossible to provide an `Arbitrary` implementation for `ndarray` outside of the `ndarray` crate (or `quickcheck` itself).
`quickcheck` seems to have become the de facto standard for property-based testing in Rust right now and I don't feel that a feature-flagged `quickcheck` dependency would be much different from our current feature-flagged `serde` dependency for serialization.

bluss:
I can't elaborate all the way, but if we had quickcheck as a pub dep, then rand is a pub dep of ndarray, and ndarray-rand would be pointless, because the separation would be gone. And sure, we have talked about rand as pub dep as a possibility through dependency rename.

Another facet is that serde is 1.0 and has been for a while, and it's a more important feature and a feature that doesn't have alternatives.

PR #725, opened 2019-09-23
Merge in ndarray-parallel 0.9.1 release

bluss:
This was the last release of ndarray-parallel, just to mark it deprecated (it still uses ndarray 0.12 etc).
Merge in the last changes if possible, before deleting it.


bluss:
ndarray-parallel 0.9.1 was released from this branch, which is directly after the parallel-0.9.0 tag.

PR #728, opened 2019-09-24
Faster version of from_shape_fn and speed up ndarray-rand

bluss:
Add new constructor from_shape_simple_fn

This is like from_shape_fn, but avoids the overhead of iterating the
indices for the elements - instead elements are just visited in memory
order.

Then, use from_shape_simple_fn in ndarray-rand, since
it doesn't care for indices, it generates items randomly.

Quick benchmarks (the existing ndarray-rand benches) show improvements, especially in the simpler uniform distribution:

```
 name     before ns/iter  after ns/iter  diff ns/iter   diff % 
 norm_f32     247,189     233,467          -13,722   -5.55% 
 norm_f64     239,662     223,620          -16,042   -6.69% 
 uniform_f32  95,626      59,371           -36,255  -37.91%
```

### DIFF
--- a/ndarray-rand/src/lib.rs
+++ b/ndarray-rand/src/lib.rs
@@ -246,7 +246,7 @@ where
         R: Rng + ?Sized,
         Sh: ShapeBuilder<Dim = D>,
     {
-        Self::from_shape_fn(shape, |_| dist.sample(rng))
+        Self::from_shape_fn_memory_order(shape, |_| dist.sample(rng))
     }
 
     fn sample_axis(&self, axis: Axis, n_samples: usize, strategy: SamplingStrategy) -> Array<A, D>
--- a/ndarray-rand/src/lib.rs
+++ b/ndarray-rand/src/lib.rs
@@ -246,7 +246,7 @@ where
         R: Rng + ?Sized,
         Sh: ShapeBuilder<Dim = D>,
     {
-        Self::from_shape_fn_memory_order(shape, |_| dist.sample(rng))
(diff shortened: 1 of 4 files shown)


bluss:
I idly wonder if we can just turn this around and implement internal iteration - `fold` for the indices iterators, and use that instead. It's still unlikely to generalize as well as this.

bluss:
One can also see that the `ndarray-rand` crate has no substance beyond this line:

```rust
Self::from_shape_fn(shape, |_| dist.sample(rng))
```

The whole thing is a wrapper for that + a method with a default for the `rng`. :slightly_smiling_face:

bluss:
I don't really like the name `from_shape_fn_memory_order` but can't think of a better one. `from_shape_fn_linear` ? It's likely to cause confusion. `from_shape_fn_linear_index`?

jturner314:
I think this is a good idea, and the implementation looks good. I agree, though, that `from_shape_fn_memory_order` needs a better name, since the existence of a method named `from_shape_fn_memory_order` suggests that `from_shape_fn` doesn't iterate in memory order. Maybe `from_shape_fn_offset` to indicate the the closure's parameter is the memory offset from the first element? Alternatively, we could rename `from_shape_fn` to `from_shape_indexed_fn` and add a new `from_shape_fn` method that doesn't pass anything to the closure.

bluss:
Now it has the name `from_shape_simple_fn`. It more or less sacrifices the index as parameter - now the closure is called with no parameter, so that it has an easier to understand name or explanation.  We can do it either way, for me.

PR #729, opened 2019-09-24
Simplify Zip::all implementation

bluss:
We notice FoldWhile::Continue/Done is already a boolean itself and the
payload bool is redundant.

### DIFF
--- a/src/zip/mod.rs
+++ b/src/zip/mod.rs
@@ -934,14 +934,14 @@ macro_rules! map_impl {
             pub fn all<F>(mut self, mut predicate: F) -> bool
                 where F: FnMut($($p::Item),*) -> bool
             {
-                self.apply_core(true, move |_, args| {
+                !self.apply_core((), move |_, args| {
                     let ($($p,)*) = args;
                     if predicate($($p),*) {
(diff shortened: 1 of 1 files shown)


bluss:
Tangentially, for future method additions to zip I'd vote for a high threshold of utility. We don't want to reproduce `Iterator` or `Itertools` on `Zip`.

jturner314:
Good catch. I didn't notice that.

> We don't want to reproduce `Iterator` or `Itertools` on `Zip`.

`Zip` is currently the only efficient way to iterate over arrays except for `map*`/`fold*` on single arrays. I would think that we'd want to provide most of the `Iterator` methods that make sense for `Zip`. (The remaining ones I see are `any`, `find`, `find_map`, and `position`.)

bluss:
I'm not sure we should be building on top of Zip in that way, while the whole thing seems to be unfinished. I'd avoid building too much add-ons to make it easier to change Zip itself.

Secondly, we can't continue duplicating methods through the 6 arities like this, so we need a different model for that, unfortunately. It seems like the practice of duplicating stuff with macros is part of why the crate takes so long to build.

Issue #723, opened 2019-09-20
ndarray build time / compile time


bluss:
I just wanted to share a neat graph. (This is a new feature in cargo - run with `-Ztimings` and it outputs this!)

[HTML build time for ndarray + deps](http://gistpreview.github.io/?f57a987580d861d1c84ca162cae0b0ad)  I used a rayon benchmark as testcase, so it's a release compile, and my laptop struggles a bit, so it's slowish :-)

You can set the min unit time to 1 second to get an easier to overview time chart of the build.

Here we can look at how the build time of the crate is broken down, obviously we only decide about the direct deps and not transitive deps of any crates we don't control.

bluss:
Also, look at that sweet pipelining (when cargo starts building the next crate in the chain before the codegen of the dependency is finished).

max-sixty:
Awesome chart. Would it be worth looking at the non-dev dependencies only too? e.g. I see quickcheck uses up a bunch of time there

jturner314:
Very cool. This feature is really helpful for figuring out how to reduce build times. I noticed that it also shows the edges of the dependency graph associated with a crate if you hover over that crate. I think it's interesting that the vast majority of the time spent compiling `ndarray` itself is in things other than code generation; it would be helpful to know more specifically where that time is spent.

bluss:
Acrichto wrote about this today too https://internals.rust-lang.org/t/exploring-crate-graph-build-times-with-cargo-build-ztimings/10975?u=dtolnay

PR #730, opened 2019-09-25
Remove itertools as a dependency

bluss:
We use `izip!()` as the only major feature from itertools, and can
remove it to save build time and two small crates as deps.

`izip!()` is license compatible and originally written by @krdln
(Michał Krasnoborski) and @bluss in itertools, so we copy it into the
crate.

The free functions zip(i, j) and enumerate(i) are just style
preferences, written by me, and simple to copy.

Itertools is still used as a dev-dependency.

### DIFF
--- a/src/layout/layoutfmt.rs
+++ b/src/layout/layoutfmt.rs
@@ -8,7 +8,6 @@
 
 use super::Layout;
 use super::LayoutPriv;
-use itertools::Itertools;
 
 const LAYOUT_NAMES: &[&str] = &["C", "F"];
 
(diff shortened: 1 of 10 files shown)


src/layout/layoutfmt.rs
```diff
@@ -17,20 +16,18 @@ use std::fmt;
...
+                    if let Some(name) = LAYOUT_NAMES.get(i) {
+                        write!(f, "{}", name)
                     } else {
-                        f(&format_args!("0x{:x}", i))
-                    })
-            )
-        }?;
+                        write!(f, "{:#x}", i)
+                    }
+                })?;
```
bluss: With `.try_fold()` the `Itertools::format` method is no longer as indispensible as it was.

bluss:
It's no major thing, but looking at it, we didn't use much of itertools, and removing it speeds up the build by a tiny amount and removes itertools and either from the default dependency tree.

cc #723

jturner314:
I haven't had a chance to review this PR in detail, but I do think it's a good idea, considering how little code `ndarray` uses from `itertools`.

PR #733, opened 2019-09-27
Implement fold for IndicesIter

jturner314:
This improves performance somewhat.

#728 made me curious about this. On my machine, this PR improves the `norm_f32` benchmark by 8%, the `norm_f64` benchmark by 9%, and the `uniform_f32` benchmark by 2%. I haven't compared it to #728 directly for improving `ndarray-rand` performance; I would guess that #728 would be better. Regardless, this PR is an improvement to `IndicesIter`.

### DIFF
--- a/src/indexes.rs
+++ b/src/indexes.rs
@@ -80,6 +80,25 @@ where
         };
         (l, Some(l))
     }
+
+    fn fold<B, F>(self, init: B, mut f: F) -> B
+    where
+        F: FnMut(B, D::Pattern) -> B,
+    {
+        let IndicesIter { mut index, dim } = self;
+        let inner_axis = dim.ndim() - 1;
+        let inner_len = dim[inner_axis];
+        let mut acc = init;
+        while let Some(mut ix) = index {
+            // unroll innermost axis
+            while ix[inner_axis] < inner_len {
+                acc = f(acc, ix.clone().into_pattern());
+                ix[inner_axis] += 1;
+            }
+            index = dim.next_for(ix);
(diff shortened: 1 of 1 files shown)


src/indexes.rs
```diff
@@ -80,6 +80,25 @@ where
         };
         (l, Some(l))
     }
+
+    fn fold<B, F>(self, init: B, mut f: F) -> B
+    where
+        F: FnMut(B, D::Pattern) -> B,
+    {
+        let IndicesIter { mut index, dim } = self;
+        let inner_axis = dim.ndim() - 1;
```
bluss: We need to handle ndim zero somehow

jturner314: Good catch. I should have tested this more thoroughly.

jturner314: I also found a bug for `ndim > 2`: `.next_for()` requires requires the input index to be in-bounds, but my original implementation didn't keep the index in-bounds. I've added more tests, and I think everything should be fixed now. I'm still getting the same performance improvement in the fixed version.

bluss: No worries. It's the tiniest things. And shows how much tests we need to cover it all..

bluss:
There is more that we can do - there always is, and there are other things that are much more important! All the repetition is frustrating (The other things are corresponding rfold, and corresponding on IterIndicesF, but it's really not important right now. This fold is what gets the foremost use.)

bluss:
Thanks for this, I'll go ahead and merge

PR #734, opened 2019-09-29
Simplify ArrayView construction from NonNull<T> and add RawView .cast() method

bluss:
1. Add ArrayView/Mut::new constructor from NonNull pointers (for internal use)

This saves some unwrapping/rewrapping in NonNull, by simpliy copying
`self.ptr` along.

For the array views, the previous new_ constructor still remains for use in
all places we make array views directly from raw pointers.

2. Add `.cast::<B>()` method on raw views. Matches NonNull::cast.

Add methods for raw view casts, this makes it easier to build upon the
raw array view functionality.

Of course, adding more capabilities for the unsafe APIs opens for users
to make mistakes, but we need these capabilities ourselves for
developing features for ndarray.

The reason for this to exist is for the `ArrayViewMut<T>` →
`ArrayView<Cell<T>>` transformation demonstrated in the tests. However, we
don't add a method for this directly, at least not yet. The reason is
that we'd like to have a type that is exactly like `Cell<T>` but also
supports the arithmetic ops.

### DIFF
--- a/src/impl_raw_views.rs
+++ b/src/impl_raw_views.rs
@@ -1,3 +1,4 @@
+use std::mem;
 use std::ptr::NonNull;
 
 use crate::dimension::{self, stride_offset};
@@ -111,6 +112,32 @@ where
 
         (left, right)
     }
+
+    /// Cast the raw pointer of the raw array view to a different type
+    ///
+    /// **Panics** if element size is not compatible.
+    ///
+    /// Lack of panic does not imply it is a valid cast. The cast works the same
+    /// way as regular raw pointer casts.
+    ///
+    /// While this method is safe, for the same reason as regular raw pointer
+    /// casts are safe, access through the produced raw view is only possible
+    /// in an unsafe block or function.
+    pub fn cast<B>(self) -> RawArrayView<B, D> {
+        assert_eq!(
+            mem::size_of::<B>(),
+            mem::size_of::<A>(),
+            "size mismatch in raw view cast"
+        );
+        let ptr = self.ptr.cast::<B>();
+        debug_assert!(
+            is_aligned(ptr.as_ptr()),
+            "alignment mismatch in raw view cast"
+        );
+        /* Alignment checked with debug assertion: alignment could be dynamically correct,
+         * and we don't have a check that compiles out for that. */
+        unsafe { RawArrayView::new(ptr, self.dim, self.strides) }
+    }
 }
 
 impl<A, D> RawArrayViewMut<A, D>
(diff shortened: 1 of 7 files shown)


src/impl_raw_views.rs
```diff
@@ -105,6 +112,32 @@ where
...
+
+    /// Cast the raw pointer of the raw array view to a different type
+    ///
+    /// **Panics** if element size is not compatible.
+    ///
+    /// Lack of panic does not imply it is a valid cast. The cast works the same
+    /// way as regular raw pointer casts.
+    ///
+    /// While this method is safe, for the same reason as regular raw pointer
+    /// casts are safe, accessing the produced raw view is only possible
```
bluss:
```suggestion
    /// casts are safe, access through the produced raw view is only possible
```

src/impl_raw_views.rs
```diff
@@ -207,9 +244,35 @@ where
...
```
bluss:
```suggestion
    /// casts are safe, access through the produced raw view is only possible
```

bluss:
@jturner314 I'd like your review on this, because there might be facets of this I have overlookd, that mean this should be an unsafe function. We have the size assert as a firm check, since it's a check that compiles out when it passes, that's simple to do.

src/impl_raw_views.rs
```diff
@@ -207,9 +244,35 @@ where
...
+            mem::size_of::<B>(),
+            mem::size_of::<A>(),
+            "size mismatch in raw view cast"
+        );
+        let ptr = self.ptr.cast::<B>();
+        debug_assert!(
+            is_aligned(ptr.as_ptr()),
+            "alignment mismatch in raw view cast"
+        );
+        /* Alignment checked with debug assertion: alignment could be dynamically correct,
```
jturner314: Same thing here -- the alignment check is a safety issue, so it should be `assert!`, not `debug_assert!`.

src/impl_views/constructors.rs
```diff
@@ -235,17 +247,24 @@ where
...
         ArrayViewMut {
             data: ViewRepr::new(),
-            ptr: nonnull_debug_checked_from_ptr(ptr),
+            ptr,
             dim,
             strides,
         }
     }
+
+    /// Create a new `ArrayView`
```
jturner314: Typo: This should be `ArrayViewMut` (same for the docs of `new` above).

tests/raw_views.rs
```diff
@@ -0,0 +1,86 @@
...
+}
+
+#[test]
+fn raw_view_cast_zst() {
+    struct Zst;
+
+    let a = Array::<(), _>::default((250, 250));
+    let b: RawArrayView<Zst, _> = a.raw_view().cast::<Zst>();
+    assert_eq!(a.shape(), b.shape());
+    assert_eq!(a.as_ptr() as *const u8, b.as_ptr() as *const u8);
```
jturner314: I'd write this as

```rust
assert_eq!(a.as_ptr().cast::<()>(), b.as_ptr().cast::<()>());
```

or

```rust
assert_eq!(a.as_ptr() as usize, b.as_ptr() as usize);
```

but that's just personal preference.

src/impl_raw_views.rs
```diff
@@ -105,6 +112,32 @@ where
...
```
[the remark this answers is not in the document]
bluss: Where does the alignment requirement come from, and doesn't it only matter if we dereference?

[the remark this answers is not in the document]
jturner314: I'm referring to [this constraint on `ArrayBase`](https://github.com/rust-ndarray/ndarray/blob/ecb7643d908ee929b6109f0974560a6bfa42db82/src/lib.rs#L1109-L1110). From what I can tell, you're right -- the only time alignment actually matters for `RawArrayView/Mut` is when dereferencing a pointer or reading/writing a pointed-to value in some way (i.e. most of the functions in `std::ptr`).

It sounds like the best option is to remove the alignment restriction on `RawArrayView/Mut`. All we'd need to do is update the list of restrictions in that comment above `ArrayBase`, the docs for `RawArrayView/Mut::from_shape_ptr`, and the docs for `deref_into_view/_mut`.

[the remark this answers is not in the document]
jturner314: I've created #738 to remove the alignment restriction on `RawArrayView/Mut`.

Issue #667, opened 2019-07-22
Add dstack, vstack, and hstack


ngoldbaum:
In the vein of #75 which added `stack`, it would be really nice if there were a way to easily convert e.g. `Vec<Array2<D>>` into an `Array3<D>`. Current I'm doing something like this:

```
let imshape = data[0].shape()
let mut activation: Array3<f64> = Array3::zeros((n_batch, imshape[0], imshape[1]));
for (i, mm) in data.iter().enumerate() {
    activation.slice_mut(s![i, .., ..]).assign(&m);
}
```

It would be nice to not have to do this manually, it would also be clearer to people who are used to numpy (IMO).

I think this is equivalent to relaxing the restriction that `axis` in [`ndarray::stack`](https://docs.rs/ndarray/0.12.1/ndarray/fn.stack.html) has to be an existing axis.

jturner314:
What we call `stack`, NumPy calls [`concatenate`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.concatenate.html), and we don't have an equivalent of NumPy's [`stack`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.stack.html). I actually like NumPy's naming (`concatenate` along existing axis, `stack` along new axis) better than `ndarray`'s.

Any thoughts on renaming the existing `stack` function/macro to `concatenate` or `concat` and adding a `stack` function that stacks along a new axis?

LukeMathWalker:
It sounds good to me - it would also help for issues like #650.

bluss:
@jturner314 Sounds good

andrei-papou:
Seems like @bluss is a bit busy so could one of you guys @LukeMathWalker @jturner314 please take a look at the [PR](https://github.com/rust-ndarray/ndarray/pull/735) changes? Thanks a lot!

TheButlah:
New user of ndarray here - coming from python, i expected `ndarray::stack` to have the same semantics as `np.stack`. I would like to also put a vote in for changing the name to `ndarray::concatenate`.

Also, is there a `ndarray` alternative for `np.stack` now? Trying to find the function I should be using, but not seeing one.

ngoldbaum:
No, that’s all in #735, which appears to have stalled.

Luthaf:
[...]

PR #735, opened 2019-10-01
Stack and concatenate

andrei-papou:
Issue: https://github.com/rust-ndarray/ndarray/issues/667

Renamed `stack` to `concatenate` to follow the behaviour of `numpy` and also added `stack` function. Examples:

##### concatenate
```rust
use ndarray::{arr2, Axis, concatenate};

let a = arr2(&[[2., 2.], [3., 3.]]);
assert!(
     concatenate(Axis(0), &[a.view(), a.view()])
     == Ok(arr2(&[[2., 2.], [3., 3.], [2., 2.], [3., 3.]]))
);
```

##### stack
```rust
use ndarray::{arr2, arr3, stack, Axis};

let a = arr2(&[[2., 2.],
               [3., 3.]]);
assert!(
    stack![Axis(0), a, a]
    == arr3(&[[[2., 2.],
               [3., 3.]],
              [[2., 2.],
               [3., 3.]]])
);
```

### DIFF
--- a/src/slice.rs
+++ b/src/slice.rs
@@ -5,8 +5,9 @@
 // <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
+use crate::dimension::slices_intersect;
 use crate::error::{ErrorKind, ShapeError};
-use crate::Dimension;
+use crate::{ArrayViewMut, Dimension};
 use std::fmt;
 use std::marker::PhantomData;
 use std::ops::{Deref, Range, RangeFrom, RangeFull, RangeInclusive, RangeTo, RangeToInclusive};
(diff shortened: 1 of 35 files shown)


bluss:
Nice, this is the right direction. I think it would be best to make this in the shape of  a non breaking change to start with so that it can be included soon.

src/stacking.rs
```diff
@@ -76,26 +73,45 @@ where
...
+    D::Larger: RemoveAxis,
+{
+    if let Some(ndim) = D::NDIM {
+        if axis.index() > ndim {
+            return Err(from_kind(ErrorKind::OutOfBounds));
+        }
+    }
+    let arrays: Vec<ArrayView<A, D::Larger>> =
+        arrays.into_iter().map(|a| a.insert_axis(axis)).collect();
+    concatenate(axis, &arrays)
```
bluss: Implementation strategy wise, can't we take the old `stack`, put that into an internal function. Give it a parameter that tells it whether to remove an axis or not. It can be a type parameter, since it influences the return type (the dim of the returned array).

That way both stack and concatenate can use this internal function and we don't need the intermediate vector here in concatenate

andrei-papou: Looks good! I'll try to implement it this way.

bluss:
This branch will eventually need to be rebased to skip the merges in the history (see commit view). The blast back to #1, #5 etc in ndarray history is fun, but we don't want to have those merges there.

Also, "this is the right direction" isn't supposed to sound like "nice try". It's supposed to say.. This is the right **destination**, this is how we want to name those functions. To be pragmatic I think we'd like to have this in 0.13.x, and then as a non-breaking change in the first round (preparing to put the functions in their right places in the next version).

andrei-papou:
@bluss sorry for the delay, the month is pretty busy. I've refactored the new function to get rid of the redundant allocation. Also I renamed the functions in the backward compatible way. Thanks.

src/stacking.rs
```diff
@@ -76,14 +73,57 @@ where
...
+
+/// Concatenate arrays along the given axis.
 ///
 /// Uses the [`stack`][1] function, calling `ArrayView::from(&a)` on each
 /// argument `a`.
 ///
 /// [1]: fn.stack.html
 ///
-/// ***Panics*** if the `stack` function would return an error.
+/// ***Panics*** if the `concatenate` function would return an error.
```
LukeMathWalker: `concatenate` does not exist yet, we shouldn't refer to it here.

LukeMathWalker:
If we want to introduce this in `0.13.x`, we should probably:
- mark the existing `stack` function as deprecated, suggesting to use `stack_new_axis`;
- introduce `concatenate` and `concatenate!`, which underneath are just calling `stack` for now;
- introduce `stack_new_axis` which will become the "new" stack.

What do you think? @andrei-papou 
Always difficult to change what a name means :/

andrei-papou:
> * mark the existing `stack` function as deprecated, suggesting to use `stack_new_axis`;

@LukeMathWalker shouldn't we suggest to use `concatenate` instead? `concatenate` is a new name for the current `stack` function while `stack_new_axis` behaves in a different way.

LukeMathWalker:
Yes, you are right :sweat_smile:

andrei-papou:
@LukeMathWalker ready, but I had to suppress a few deprecation warnings: we still need `stack` function in `concatenate` implementation, in `lib.rs` and in tests.

andrei-papou:
@bluss @LukeMathWalker could you please do another round of review? Looks like all the comments have been resolved quite a while ago. Thanks.

src/stacking.rs
```diff
@@ -76,7 +77,79 @@ where
...
+where
+    A: Copy,
+    D: RemoveAxis,
+{
+    stack(axis, arrays)
+}
+
+pub fn stack_new_axis<A, D>(
+    axis: Axis,
+    arrays: Vec<ArrayView<A, D>>,
```
bluss, line 111: `stack_new_axis` and `concatenate` should agree on if they take the arrays as a slice or a Vec, they should have the same interface if they can. Avoiding cloning of the array views seems like a minor issue, but it can be decided either way (slice or vec)

src/stacking.rs
```diff
@@ -76,7 +77,79 @@ where
...
+#[allow(deprecated)]
+pub fn concatenate<A, D>(axis: Axis, arrays: &[ArrayView<A, D>]) -> Result<Array<A, D>, ShapeError>
+where
+    A: Copy,
+    D: RemoveAxis,
+{
+    stack(axis, arrays)
+}
+
+pub fn stack_new_axis<A, D>(
```
bluss, line 109: Missing doc comment :)

src/stacking.rs
```diff
@@ -109,3 +182,71 @@ macro_rules! stack {
         $crate::stack($axis, &[ $($crate::ArrayView::from(&$array) ),* ]).unwrap()
```
bluss, line 182: The stack macro should also be deprecated

src/stacking.rs
```diff
@@ -76,7 +77,79 @@ where
...
+    let mut res_dim = common_dim.insert_axis(axis);
+
+    if arrays.iter().any(|a| a.raw_dim() != common_dim) {
+        return Err(from_kind(ErrorKind::IncompatibleShape));
+    }
+
+    res_dim.set_axis(axis, arrays.len());
+
+    // we can safely use uninitialized values here because they are Copy
+    // and we will only ever write to them
```
bluss, line 135: This looks good, it follows what we do in stack. We'll revise handling of uninit data, but it's not in scope for this PR. (#796 )

src/stacking.rs
```diff
@@ -76,7 +77,79 @@ where
...
+    let size = res_dim.size();
+    let mut v = Vec::with_capacity(size);
+    unsafe {
+        v.set_len(size);
+    }
+    let mut res = Array::from_shape_vec(res_dim, v)?;
+
+    res.axis_iter_mut(axis)
+        .zip(arrays.into_iter())
```
bluss, line 144: this zip seems to show that we would be fine with just using a slice of `ArrayView` as input, we don't need to use them by value

src/slice.rs
```diff
@@ -544,62 +544,62 @@ impl_slicenextdim_larger!((), Slice);
 #[macro_export]
 macro_rules! s(
     // convert a..b;c into @convert(a..b, c), final item
-    (@parse $dim:expr, [$($stack:tt)*] $r:expr;$s:expr) => {
+    (@parse $dim:expr, [$($concatenate:tt)*] $r:expr;$s:expr) => {
```
bluss, line 547: This rename of the variable inside the macro is an incidental mistake, and should not be in the PR

bluss:
I have rebased your commit history and pushed to a new branch: See https://github.com/rust-ndarray/ndarray/pull/new/ap/stack-and-concatenate

Please use it as a starting point - it removes all the merges in the commit history of this PR. We accept some unclean histories, but this needed to be cleaned up before merging.

There are some commits that have the title "Changelog:" please edit those, and make sure the first line is a summary of what the commit does. See my commits in the repo, if you want an example. Thanks :)

src/slice.rs
bluss: Oh I removed this in my rebase-fix of the branch. No other changes in that branch. :)

PR #738, opened 2019-10-02
Remove alignment restriction on RawArrayView/Mut

jturner314:
This makes `.cast()` safe without a check for alignment.

This might be a breaking change because we [documented](https://docs.rs/ndarray/0.13.0/ndarray/type.RawArrayView.html) that `RawArrayView/Mut` had alignment requirements. I suppose it depends on how that statement could be interpreted. Someone could think that it would mean that, given any `RawArrayView`, it would be safe to call `.deref_into_view()` without ensuring that the pointer was aligned. IMO, though, since we didn't say exactly what the alignment restrictions were, no one should have made that assumption.

### DIFF
--- a/src/impl_raw_views.rs
+++ b/src/impl_raw_views.rs
@@ -33,8 +33,8 @@ where
     ///
     /// Unsafe because caller is responsible for ensuring all of the following:
     ///
-    /// * `ptr` must be non-null and aligned, and it must be safe to
-    ///   [`.offset()`] `ptr` by zero.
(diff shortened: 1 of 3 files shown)


bluss:
Thanks a lot. Fixed up to use `as` for raw pointer casts, the cast method on raw pointers was a Rust 1.37 or 1.38 feature apparently (and it was a bit earlier on NonNull).

bluss:
I don't see it as a breaking change, this refinement of detail seems reasonable.

jturner314:
Okay, sounds good. Thanks for fixing up the PR for Rust 1.37.

PR #739, opened 2019-10-02
Simplify LinalgScalar trait bounds

### LINKED ISSUE #737: LinalgScalar does not have to imply Sub and Div
Recall the definition of `LinalgScalar`:
https://github.com/rust-ndarray/ndarray/blob/eb82c93f0147df38e061597221ece3627f119a60/src/linalg_traits.rs#L18

However, no algorithms involving `LinalgScalar` requires the scalar to be invertible in the additive group and multiplicative semigroup, in this crate at the very least. Therefore, `Sub` and `Div` trait bounds are not used. It would be more flexible if `Sub` and `Div` trait bounds are relaxed.

In addition, `Add<Self, Output=Self>` and `Mul<Self, Output=Self>` is already implied by `num::Zero` and `num::One`. These trait bounds can be relaxed, too.

dingxiangfei2009:
Fix #737 

This PR relaxes the trait bounds of `LinalgScalar`, since `Sub` and `Div` traits are unused in matrix multiplication routines in this crate. In addition, `Add` and `Mul` trait bounds are already implied by `num::Zero` and `num::One` trait bounds.

### DIFF
--- a/src/linalg_traits.rs
+++ b/src/linalg_traits.rs
@@ -8,7 +8,6 @@
 use crate::ScalarOperand;
 use num_traits::{Float, One, Zero};
 use std::fmt;
-use std::ops::{Add, Div, Mul, Sub};
 use std::ops::{AddAssign, DivAssign, MulAssign, RemAssign, SubAssign};
 
 /// Elements that support linear algebra operations.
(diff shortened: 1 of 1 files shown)


bluss:
Thanks.

I'm sorry that we don't have a plan for this trait, but at the moment I don't think we should simplify it

PR #742, opened 2019-10-09
Add `sampling_axis_inplace` and `sampling_axis_inplace_using`

LukeMathWalker:
Other two simple functions that I think might useful when working with arrays.

Should we add also `sampling_axis` and `sampling_axis_using` as convenient shims for
```rust
let a = self.clone();
a.shuffle_axis_inplace(axis);
a
```
?

### DIFF
--- a/ndarray-rand/src/lib.rs
+++ b/ndarray-rand/src/lib.rs
@@ -34,8 +34,7 @@ use crate::rand::rngs::SmallRng;
 use crate::rand::seq::index;
 use crate::rand::{thread_rng, Rng, SeedableRng};
 
-use ndarray::{Array, Axis, RemoveAxis, ShapeBuilder};
-use ndarray::{ArrayBase, DataOwned, Dimension};
+use ndarray::{Array, Axis, Data, DataMut, RemoveAxis, ShapeBuilder, ArrayBase, DataOwned, Dimension};
 #[cfg(feature = "quickcheck")]
 use quickcheck::{Arbitrary, Gen};
 
@@ -64,7 +63,7 @@ pub mod rand_distr {
 /// [`.random_using()`](#tymethod.random_using).
 pub trait RandomExt<S, A, D>
 where
-    S: DataOwned<Elem = A>,
+    S: Data<Elem = A>,
     D: Dimension,
 {
     /// Create an array with shape `dim` with elements drawn from
@@ -87,6 +86,7 @@ where
     /// # }
     fn random<Sh, IdS>(shape: Sh, distribution: IdS) -> ArrayBase<S, D>
     where
+        S: DataOwned<Elem=A>,
         IdS: Distribution<S::Elem>,
         Sh: ShapeBuilder<Dim = D>;
 
@@ -116,6 +116,7 @@ where
     /// # }
     fn random_using<Sh, IdS, R>(shape: Sh, distribution: IdS, rng: &mut R) -> ArrayBase<S, D>
     where
+        S: DataOwned<Elem=A>,
         IdS: Distribution<S::Elem>,
         R: Rng + ?Sized,
         Sh: ShapeBuilder<Dim = D>;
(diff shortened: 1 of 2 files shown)


ndarray-rand/src/lib.rs
```diff
@@ -280,6 +358,41 @@ where
...
+        for i in (1..self.len_of(axis)).rev() {
+            // Invariant: elements with index > i have been locked in place.
+            let j = rng.gen_range(0, i + 1);
+
+            if i != j {
+                // Swap the two slices along `axis`
+                let slice1 = self.index_axis(axis, i);
+                let slice2 = self.index_axis(axis, j);
+
+                for (x, y) in slice1.iter().zip(slice2.iter()) {
```
bluss: We are modifying through read-only/shared array views and this is invalid, unless we have interior mutability (i.e using Cell, Mutex or similar).

And we'd like to avoid `iter().zip()` for ndarray, because it is much slower than it needs to be, we have Zip.

ndarray-rand/src/lib.rs
```diff
@@ -225,17 +226,93 @@ where
         R: Rng + ?Sized,
         A: Copy,
         D: RemoveAxis;
+
+    /// Shuffle `self`'s slices along `axis`.
```
bluss: Shuffle the array's xyz along `axis`.

We need a name for it, not sure if it should be slices. For the `.lanes(axis)` method we call them lanes - but those are perpendicular to `axis` in that case.

ndarray-rand/src/lib.rs
bluss: I'm working on making `ArrayView<Cell<T>, D>` a reality - it allows us to have.. more fun like this. :)

LukeMathWalker: > We are modifying through read-only/shared array views and this is invalid, unless we have interior mutability (i.e using Cell, Mutex or similar).

On the other hand the method is taking a mutable reference to `self`, hence we are guaranteed to be the only ones operating on the array. Or am I interpreting this incorrectly?

bluss: Modifying through an ArrayView (without the Cell) exception is full stop invalid, because we treat it as we would treat a `&[T]`. (We could go into details and exceptions at another time, maybe there's some argument for a more lenient view?)

bluss: In the short range it is easy to point to what is invalid: `x` is a `&T` but we are modifying through it. That's not permitted.

Exceptions would come in to place if you somehow got a raw pointer out of an array view, *maybe* you'd then be able to make an argument. But since we go through explicit shared references here, we are at invalid full stop.

LukeMathWalker: Given the `S: DataMut` trait bound on the method, aren't we dealing with an `ArrayViewMut`, which is equivalent to a `&mut [T]`? 
Just trying to *truly* understand, I am not trying to argue :sweat_smile:

bluss: Sure, no problem. The types matter and there are shared refs used on the path to the swap. If we read the reference this is UB, we can't just cast from &T and mutate.

bluss:
This is a good starting point:
In Rust, a shared reference is more than a pointer.

http://smallcultfollowing.com/babysteps/blog/2017/02/01/unsafe-code-and-shared-references/

ndarray-rand/src/lib.rs
bluss: I'd be tempted to call them *planes*. We shuffle among the items in an AxisIter, right? Each such item is *n - 1* dimensional in an ndarray.

LukeMathWalker: An alternative name could be _projections_, which has the plus of not carrying an intuitive dimensionality along with it (I think 2D when I read planes).

bluss:
`ArrayView<Cell<..>>` would actually be useful here - and is not so far off - because `Cell::<T>::swap` is an allowed method for all `T` (unlike `.get()` which is restricted to `T: Copy`). But there are other tools that should make it possible to write the equivalent code with ArrayViewMut, too.

Not sure about how to solve this. Another way would be to make one permutation as a vector of `len_of(axis)` and then apply that permutation identically to every lane in `lanes_mut(axis)`. But that hinges on applying permutations quickly, and I don't know about that.

bluss:
How to make an `ArrayView<Cell<T>`

See lines 14-15 https://github.com/rust-ndarray/ndarray/blob/6b388101d05e124217b46d04ec0a6332fa028adf/tests/raw_views.rs#L14-L15

and mind the lifetime of the conversion. NB we go through `RawArrayViewMut<T, D>` into `ArrayView<Cell<T>, D>`.

LukeMathWalker:
Wouldn't #717 be enough to get this method written in safe Rust? 
Both slices would be mutable, hence I could get mutable reference to their elements and swap them using `std::mem::swap`.

bluss:
Maybe, and split at or axis iter mut could also make it possible.

With slicing I think we'd struggle too much to construct the slice argument, not sure?

LukeMathWalker:
I tried to get it done with slicing and I am indeed struggling to construct the slice argument :sweat_smile:

Issue #743, opened 2019-10-16
Sub (and Add, Div, etc) not implemented for ArrayView/ViewRepr?


jamestwebber:
First off, very impressed with this crate so far. I'm a long-time numpy user and I appreciate the task you took on on behalf of the Rust ecosystem.

I only recently started using Rust, so I might just be trying to do the wrong thing here, but I'm curious why an `ArrayView` can not be used on the LHS of a binary operation like `+`, `-`, etc. That seems like a natural use-case, since you need a read-only view of the data in the array but won't be modifying it.

In my case I'd like to compare an array of values to a particular query by broadcasting and then operate on the matching rows later–this means I can't consume the array in this operation*.

Is the appropriate thing to `Zip` over the rows and compare them that way? I can see how to do that but it seemed verbose, so I thought I'd check. And if this is an oversight I can try to help implement a PR.

* a broadcasting comparison operator would be even better for this but the existence of #35 suggests that might be difficult.

jturner314:
For arithmetic operations, adding `&` in front of the view should work. See #238 and [the docs for arithmetic operations](https://docs.rs/ndarray/0.13.0/ndarray/struct.ArrayBase.html#arithmetic-operations). I've created #744 to implement arithmetic operations on `ArrayView` directly (so the `&` wouldn't be required); I'll be interested to get feedback on this idea.

For elementwise comparison of values, we can't provide direct support for the normal comparison operators (`<`, `>=`, etc.) since those operators require the result to be a `bool` (instead of e.g. an array of `bool`s). For the time being, the best way to do something like this is to use `Zip` or [`azip`](https://docs.rs/ndarray/0.13.0/ndarray/macro.azip.html).

I wouldn't mind adding elementwise comparison support (using methods, since we can't do it using the normal operators) so that `Zip`/`azip` wouldn't be necessary for elementwise comparison, although I'm not sure what others think about this.

jamestwebber:
Oh, right you are! I read that section multiple times and I thought I had tried that before making an issue, but I guess I was wrong. Thanks for the tip!

I think element-wise comparisons would be pretty handy as convenience functions. Or if that's deemed unnecessary it would be a worthy addition to the "ndarray_for_numpy_users" doc.

xldenis:
[...]

Issue #237, opened 2016-11-15
Binary operation for ArrayView


termoshtt:
I'm confusing errors arising around `+`.

```rust
let dist = Range::new(0, 1);
let a = Array::<f64, _>::random((5, 5));
let x0 = Array::<f64, _>::random(5);
let xs: Vec<_> = a.axis_iter(Axis(0)).map(|x| x + &x0);
```

I expect that `x + &x0` produces new array since [document](http://bluss.github.io/rust-ndarray/master/ndarray/struct.ArrayBase.html#arithmetic-operations) says
> &A @ &A which produces a new Array

But it produces an error:

```
error[E0369]: binary operation `+` cannot be applied to type `ndarray::ArrayBase<ndarray::ViewRepr<&f64>, usize>` 
```

After struggling, I find it works when I rewrite the last line as following:

```rust
let xs: Vec<_> = a.axis_iter(Axis(0)).map(|x| x.to_owned() + &x0);
```

I hope the first my code works. Is there any technical problem to implement `Add` for ArrayView?

bluss:
`x` in `x + &x0` is an array view, which is an "A", not an "&A". So `&x + &x0` should work.

termoshtt:
Oh...
I now understand my mistake and what document says.
The above error message means `ArrayView` cannot be consumed. It is hard to read :<

Thanks for your help!

bluss:
It's a bit unfortunate that one needs to pass `&ArrayView` in many places where just passing a view by value should be enough. It's partly due to missing support for custom view types in Rust. Feel free to report any issues you find with ndarray.

Issue #699, opened 2019-09-08
Improved arithmetic operations for arrays


bluss:
(This is a collaborative issue, please edit and add your thoughts, if applicable)

For context, please read the [ArrayBase documentation on Arithmetic Operations](https://docs.rs/ndarray/0.12.1/ndarray/struct.ArrayBase.html#arithmetic-operations) first.

### Goals

- Ease of use: Arithmetic operations should be available when the user expects it
- Transparency: It should be possible to understand how much copying is involved in an operation

### Prioritization

- I think we should continue to focus on primitive numeric types as operands, these are always the most important, while improving the situation for generic array elements. This focus means that we can accept solutions that are perfect for primitives but not yet perfect for other elements.

### Known Problems in Current Implementation

- Excessive copying of the whole array.
  - Example: In &A @ &A we use `self.to_owned().add(rhs)` to implement it, while it could be implemented without copying the first operand's data
- Excessive copying of elements: We use `elt1.clone() + elt2.clone()` in some places.
  - The alternative would be: Use by-reference operators or other general interface
  - Copying primitive numeric types (integers, floats) like this is not a problem, and for this reason, the current implementation only has a problem when concerning non-primitive array elements.
- Right hand side and left hand side scalar operands are implemented in completely different ways, due to Rust limitations in how this is expressed.
  - We should continue to strive for symmetry in which operands are supported on the lhs and rhs. But as a compromise, in generic code, we can document that users sometimes must prefer array-scalar operations where the scalar is the right hand side operand(?)

&A @ &A

### Related issues:

- Additional arithmetic operations/variants #697
- Previously decided: Don't mutate ArrayViewMut in place: Non-assignment binary operations shouldn't mutate input #83
- (Tangentially related): and_broadcast_mut for Zip #478

bluss:
About compile time

Remember: When working on the crate, compile time of the `ndarray` crate itself can seem “pretty good”. It's mostly generic code, we don't generate the machine code here, and we avoid the slowest part of compiling a rust program — feeding all the code to LLVM! The compile time when all the generic functionality is instantiated in the user's project might be quite something different. (This also happens in our tests).

There are some strategies to keep in mind

1. Split code so that the main parts of it depend on as few generic parameters as possible
2. Reduce the space of possible type parameter values up front.
   a. Rely on coercions to support multiple types while implementing for a lesser number of them
[...]

PR #744, opened 2019-10-16
Implement arithmetic ops on more combinations of types

jturner314:
In particular, this adds implementations for `ArrayView` where`&ArrayBase` was allowed previously, and it adds support for in-place operations where the RHS is any array (instead of requiring the RHS to be a reference).

This functionality seems useful. It doesn't allow anything that wasn't previously possible by adding a `&` in front of a parameter, but removing the requirement of a `&` reduces verbosity and may be more friendly to new users.

Thoughts?

See also #237 and #743.

### DIFF
--- a/src/impl_ops.rs
+++ b/src/impl_ops.rs
@@ -129,6 +129,74 @@ where
 
 /// Perform elementwise
 #[doc=$doc]
+/// between references `self` and `rhs`,
+/// and return the result as a new `Array`.
+///
+/// If their shapes disagree, `rhs` is broadcast to the shape of `self`.
+///
+/// **Panics** if broadcasting isn’t possible.
+impl<'a, 'b, A, B, S2, D, E> $trt<&'b ArrayBase<S2, E>> for ArrayView<'a, A, D>
+where
+    A: Clone + $trt<B, Output=A>,
+    B: Clone,
+    S2: Data<Elem=B>,
+    D: Dimension,
+    E: Dimension,
+{
+    type Output = Array<A, D>;
+    fn $mth(self, rhs: &'b ArrayBase<S2, E>) -> Array<A, D> {
+        // FIXME: Can we co-broadcast arrays here? And how?
+        self.to_owned().$mth(rhs)
+    }
+}
+
+/// Perform elementwise
+#[doc=$doc]
+/// between references `self` and `rhs`,
+/// and return the result as a new `Array`.
+///
+/// If their shapes disagree, `rhs` is broadcast to the shape of `self`.
+///
+/// **Panics** if broadcasting isn’t possible.
+impl<'a, 'b, A, B, S, D, E> $trt<ArrayView<'b, B, E>> for &'a ArrayBase<S, D>
+where
+    A: Clone + $trt<B, Output=A>,
+    B: Clone,
+    S: Data<Elem=A>,
+    D: Dimension,
+    E: Dimension,
+{
+    type Output = Array<A, D>;
+    fn $mth(self, rhs: ArrayView<'b, B, E>) -> Array<A, D> {
+        // FIXME: Can we co-broadcast arrays here? And how?
+        self.to_owned().$mth(rhs)
+    }
+}
+
+/// Perform elementwise
+#[doc=$doc]
+/// between references `self` and `rhs`,
+/// and return the result as a new `Array`.
+///
+/// If their shapes disagree, `rhs` is broadcast to the shape of `self`.
+///
+/// **Panics** if broadcasting isn’t possible.
+impl<'a, 'b, A, B, D, E> $trt<ArrayView<'b, B, E>> for ArrayView<'a, A, D>
+where
+    A: Clone + $trt<B, Output=A>,
+    B: Clone,
+    D: Dimension,
+    E: Dimension,
+{
+    type Output = Array<A, D>;
+    fn $mth(self, rhs: ArrayView<'b, B, E>) -> Array<A, D> {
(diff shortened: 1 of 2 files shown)


LukeMathWalker:
I agree that it reduces verbosity and accidental complexity: you are not left puzzled trying to understand why only certain combinations are allowed.
On the other hand, I think I remember @bluss being concerned this could lead to code bloat/sensibly longer compile times.

jamestwebber:
As the initiator of #743 I will say that I just straight-up missed the relevant documentation (and/or didn't properly grok it) and while adding this feature would have saved me a little time I wouldn't want you to trade nice code or fast compiles just to have it.

Looking at the documentation now I can't figure what went wrong. Maybe I was thrown off by the use of `@` to mean "any binary operator". Making that section more verbose might be a simpler and sufficient solution.

bluss:
The summary issue is #699 where we coordinate work on this.

I'd investigate what's the total count of arith ops impl methods before and after this change? What's the clean build time of the ndarray crate before and after?

One suggested strategy for improving compilation time given in #699 is to have fewer impl blocks that cover more cases. We can use marker traits if needed. (We can use https://doc.rust-lang.org/nightly/std/array/trait.LengthAtMost32.html as inspiration in a way - it's a marker that allows each other array trait impl to be written as one `impl` instead of 33 pieces.)

I think this should be seen in context of #699. Here we add more impls that do the same "clone whole array and then do the op", which is something we want to fix.

bluss:
Treating ArrayView and &ArrayBase same seems sensible. So I'd say it would be great if those could be treated in the same impl.

src/impl_ops.rs
```diff
@@ -218,6 +301,23 @@ impl<'a, S, D> $trt<&'a ArrayBase<S, D>> for $scalar
         })
     }
 }
+
+/// Perform elementwise
+/// between the scalar `self` and array `rhs`,
+/// and return the result as a new `Array`.
```
bluss: Not important, but just something I saw. Allegedly this doc is not visible in rustdoc, but that comment may be outdated now(!). then we want that `#[doc=$doc]` back to make a complete sentence.

bluss: I would especially discourage these impl blocks. We have to spend a lot of impl blocks on the arithmetic ops with scalars on the *left hand side*. Part of me want to just remove everything with left hand side scalars, because it can never(?) be properly general because of how we need to express it as an `impl .. for $scalar` (for a specific single type).

src/lib.rs
```diff
@@ -646,18 +646,19 @@ pub type Ixs = isize;
...
-/// for an arbitrary unary operator denoted by `@` (it can be `-` or `!`).
+/// Let `A` be an array or view of any kind. Let `O` be an array with owned
+/// storage (either `Array` or `ArcArray`). Let `V` be an `ArrayView`. The
+/// following operands are supported for an arbitrary unary operator denoted by
+/// `@` (it can be `-` or `!`).
 ///
-/// - `@&A` which produces a new `Array`
-/// - `@B` which consumes `B`, updates it with the result, and returns it
+/// - `@&A` or `@V` which produces a new `Array`
+/// - `@O` which consumes `O`, updates it with the result, and returns it
```
bluss: Thanks for updating this, even if it's a draft PR or whatever, it makes it easy to discuss changes :)

jturner314:
@bluss's comments made me realize that we don't actually need separate implementations for `ArrayView`; we can write implementations for arbitrary `ArrayBase`, using `.into_owned()` as necessary. This is no more expensive than the approach in `ndarray` 0.13 (which requires `DataOwned` in many cases) because:

* For owned arrays, `.into_owned()` is a no-op.
* For `ArcArray`s, `.into_owned()` calls `ensure_unique`, which was called in the old implementation anyway because it was mutating the array.
* For `ArrayView`s (which weren't supported in the old implementation), the cost is equivalent to using `&array`. This is similarly true for `ArrayViewMut` and `CowArray`.

This is really nice. The only issue is that it's a breaking change because the return type is always `Array` now. (Before, some implementations would return `ArrayBase<S, D>`.)

src/impl_ops.rs
jturner314: I've removed this impl block in the most recent commit because it's not necessary.

> Part of me want to just remove everything with left hand side scalars, because it can never(?) be properly general because of how we need to express it as an `impl .. for $scalar` (for a specific single type).

I understand the objection, and AFAIK it will never be possible to write these impls for generic scalars even after [re_rebalance_coherence](https://doc.rust-lang.org/unstable-book/language-features/re-rebalance-coherence.html). However, IMO supporting left-hand-side scalars is important, even if it's just `f32`/`f64`. I use these impls all the time in my code. They are especially important for non-commutative ops, where it's not straightforward to just move the scalar to the other side.

bluss:
It's nice to express, but it risks cementing one of the things we want to fix - excessive cloning?

I don't think it's right to change the return type of `ArcArray<T> + T` or similar. That should still be `ArcArray`. This is still the spirit of predictiveness. We want to preserve the kind of the array if we can, and we even want to preserve the layout of the array if we can (if the user uses F arrays, we return those where reasonable).

bluss:
(Fwiw, there is a benchmark to compare the "excessive cloning vs not" and there isn't so much difference. The benchmarks are called add_2d_alloc_* and I'm not sure it gains so much. The "x.to_owned() + y" approach has one benefit of having simpler loops - a "memcpy" to copy, one binary operand loop to add in place; when we lose the extra cloning we have a ternary operand loop instead, to add x and y and store into z)

PR #746, opened 2019-11-04
Require std::hash::Hash trait on Dimension::Pattern

HactarCE:
I was trying to create a generic sparse N-dimensional array type, using `Dimension::Pattern` as the keys in a `std::collections::HashMap`, but `std::hash::Hash` is not required to be implemented by `Dimension::Pattern`. All of `Dimension::Pattern`'s implementations do have `std::hash::Hash`, and `Dimension` is never implemented outside of ndarray, so this shouldn't cause any problems in calling code.

I'm still somewhat new to Rust, so please do tell if there's something I'm missing.

### DIFF
--- a/src/dimension/dimension_trait.rs
+++ b/src/dimension/dimension_trait.rs
@@ -7,6 +7,7 @@
 // except according to those terms.
 
 use std::fmt::Debug;
+use std::hash::Hash;
 use std::ops::{Add, AddAssign, Mul, MulAssign, Sub, SubAssign};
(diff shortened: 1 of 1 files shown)


HactarCE:
After some more experimenting, I don't think this will help me in my sparse array adventures. (It looks like I want `NdArray<D> + std::cmp::Eq + std::hash::Hash`, not `D::Pattern where D::Pattern: std::hash::Hash`.)

PR #750, opened 2019-11-10
Remove unnecessary parens around types

viniciusd:
rust-lang/rust#64169 adds a compiler warning for parenthesis around simple types. It has already made its way into the beta channel.

Also, thoughts on adding `-D warnings` to the nightly channel build as well? Gotta confess it was weird when I saw it failing because of this reason on beta while it didn't on nightly


viniciusd:
It is now failing because of unrelated issues

LukeMathWalker:
I think we can start addressing these linting issues one by one. 
I'll merge this :+1:

PR #752, opened 2019-11-11
Remove main function from doctests

viniciusd:
This pull requests addresses the [needless_doctest_main](https://rust-lang.github.io/rust-clippy/master/index.html#needless_doctest_main) linting issues.

Most of the remaining ones are with regards missing `Safety` section in the docs, using clippy words, [missing_safety_doc](https://rust-lang.github.io/rust-clippy/master/index.html#missing_safety_doc), which was added by rust-lang/rust-clippy#4608. Those aren't as straightforward for fixing as they require a per-case analysis in order to verify the safety pre-requisites.

There is also one occurrency of a [redundant clone](https://rust-lang.github.io/rust-clippy/master/index.html#redundant_clone)

### DIFF
--- a/src/impl_constructors.rs
+++ b/src/impl_constructors.rs
@@ -132,7 +132,6 @@ where
     /// use approx::assert_abs_diff_eq;
     /// use ndarray::{Array, arr1};
     ///
-    /// # fn example() -> Option<()> {
     /// # #[cfg(feature = "approx")] {
     /// let array = Array::geomspace(1e0, 1e3, 4)?;
     /// assert_abs_diff_eq!(array, arr1(&[1e0, 1e1, 1e2, 1e3]), epsilon = 1e-12);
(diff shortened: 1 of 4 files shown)


src/impl_constructors.rs
```diff
@@ -132,18 +132,13 @@ where
     /// use approx::assert_abs_diff_eq;
     /// use ndarray::{Array, arr1};
     ///
-    /// # fn example() -> Option<()> {
```
LukeMathWalker: Does this cause a `clippy` failure? Given that the return type is `Option<()>` I wouldn't an expect this to be deemed a `needless_doctest_main`.

viniciusd: Lemme double-check why I've done so. It might be for the sake of simplicity for the example, but I agree it wouldn't be the exact scope of this pull request.
Given I couldn't find an explanation in the commit message (I usually leave explanatory messages), going to re-run it.

viniciusd: Yep, running fine with the example function. So it was, indeed, a matter of simplicity

Issue #655, opened 2019-07-04
Standard deviation and variance


LukeMathWalker:
For completeness, we should add a `std` and a `var` method to `ndarray`. See https://github.com/rust-ndarray/ndarray-stats/pull/43 for more details.

allmywatts:
I started working on this. Hopefully someone else is not doing the same in parallel :smile:

bluss:
Fixed by #790 in 0.15.0

PR #753, opened 2019-11-12
Issue/655: Standard deviation and variance

allmywatts:
This PR implements `std` and `var` methods (see: #655) using the same method as used for `var_axis`, `std_axis`.
The variance is computed by flattening the array of N dimensions into a 1D array of the length `.len()` of the original array.

The return type is scalar A instead of Array<A, D:Smaller>

An attempt was made to refactor `var_axis` to use `var`, but this resulted in a performance regression so original implementation is kept and we accept some code duplication. :stuck_out_tongue: 

```rust
pub fn var_axis(&self, axis: Axis, ddof: A) -> Array<A, D::Smaller>
where
    A: Float + FromPrimitive,
    D: RemoveAxis,
{
    let mut sum_sq = Array::<A, D::Smaller>::zeros(self.dim.remove_axis(axis));

    for (lane, sum) in self.lanes(axis).into_iter().zip(sum_sq.iter_mut()) {
        *sum = lane.var(ddof);
    }
    sum_sq
}
```

Thanks @LukeMathWalker for the mentoring. :grinning:


LukeMathWalker:
The main issue we had here was that `Zip` does not allow to iterator over two producers that yield items with different dimensionality (`0` for the output array and `1` for the lanes that we want to reduce).

bluss:
I don't quite see the issue with using Zip here

LukeMathWalker:
Sorry, it took me a while to get back to this - @bluss is indeed right, I was able to put it together using `Zip`. If you still have the attempts we made together @allmywatts at RustFest I'd be curious to understand what I was doing wrong there :thinking: 

I opened a PR against your branch with the `Zip` version @allmywatts - https://github.com/allmywatts/ndarray/pull/1
We might want to have a `var_unchecked`, as we did at the beginning, that we use from both `var` and `var_axis` to avoid redundant input checks in `var_axis`.

hameerabbasi:
@allmywatts If you don't mind can I take over this PR and open a new one?

allmywatts:
@hameerabbasi hey, sorry for the late reply, yes please go ahead.

bluss:
This looks ready to merge, right?

src/numeric/impl_numeric.rs
```diff
@@ -306,6 +294,67 @@ where
...
+        let n = A::from_usize(self.len()).expect("Converting length to `A` must not fail.");
+        assert!(
+            !(ddof < zero || ddof > n),
+            "`ddof` must not be less than zero or greater than the length of \
+             the axis",
+        );
+        let dof = n - ddof;
+        let mut mean = A::from_usize(0).expect("Converting 0 to `A` must not fail.");
+        let mut sum_sq = A::from_usize(0).expect("Converting 0 to `A` must not fail.");
+        for (count, x) in self.iter().enumerate() {
```
bluss, line 327: Just to note, iter is not an efficient way to iterate a "flattened" array, so this should potentially use unary Zip instead, for example, or `ArrayBase::fold`.

bluss:
Super minor thing, but as long as there is a division and a checked type conversion inside the actual loop, I don't think `var_unchecked` with the purpose of saving input checks saves that much.

src/numeric/impl_numeric.rs
```diff
@@ -306,6 +294,67 @@ where
...
+    /// # Example
+    ///
+    /// ```
+    /// use ndarray::{arr2, Axis};
+    ///
+    /// let a = arr2(&[[1., 2.],
+    ///                [3., 4.],
+    ///                [5., 6.]]);
+    ///
+    /// let a_flat = a.view().into_shape(6).expect("This must not fail.");
```
bluss, line 348: "This must not fail" does not add information, just use `.unwrap()`. I'm just afraid that the user will conclude some kinds of `into_shape()` (to 1D) always succeed, and they don't.

bluss:
Superseded by #790

Issue #749, opened 2019-11-05
Iteration performance for 'f' same-order arrays


nilgoyette:
We were wondering if mixing 'c' and 'f' order was as bad as the documentation says, so I created a series of benchmarks, comparing different/same order and loop/iter.zip/Zip::from/Zip::indexed/etc. Please read my next comment for more information. (Maybe this issue should be in #469?)

nilgoyette:
I also tested `zip_mut_with` in my benchmarks and the results are also surprising.

    zip_mut_with_c ... bench:   8,235,780 ns/iter (+/- 1,323,708)
    zip_mut_with_f ... bench:   7,551,085 ns/iter (+/- 266,037)

- `zip_mut_with_c` is as fast as `zip_c`. This makes sense to me.
- `zip_f` is around 7x faster than `zip_mut_with_f `. This doesn't seem normal.

bluss:
Please make this more specific with some example code, at least the stanzas that call Zip

nilgoyette:
Haha, wow, you're right, this is missing a lot of details! Here's my benchmark

```
#![feature(test)]
extern crate test;
use test::{black_box, Bencher};
use ndarray::{Array3, ShapeBuilder, Zip};

pub fn zip_copy(data: &Array3<f32>, out: &mut Array3<f32>) {
    Zip::from(data).and(out).apply(|&i, o| {
        *o = i;
    });
}

pub fn zip_indexed(data: &Array3<f32>, out: &mut Array3<f32>) {
    Zip::indexed(data).and(out).apply(|idx, &i, o| {
        *o = i;
    });
}

pub fn zip_mut_with(data: &Array3<f32>, out: &mut Array3<f32>) {
    out.zip_mut_with(&data, |o, &i| {
        *o = i;
    });
}

#[bench]
fn zip_cf(b: &mut Bencher) {
    let data: Array3<f32> = Array3::zeros((137, 171, 151));
    let mut out = Array3::zeros(data.dim().f());
    b.iter(|| black_box(zip_copy(&data, &mut out)));
}

#[bench]
fn zip_cc(b: &mut Bencher) {
    let data: Array3<f32> = Array3::zeros((137, 171, 151));
    let mut out = Array3::zeros(data.dim());
    b.iter(|| black_box(zip_copy(&data, &mut out)));
}

#[bench]
fn zip_fc(b: &mut Bencher) {
    let data: Array3<f32> = Array3::zeros((137, 171, 151).f());
[...]

PR #754, opened 2019-11-18
zip_mut_with 'f' order

nilgoyette:
As adviced by @jturner314 in #749, I now use `as_slice_memory_order` in `zip_mut_with_same_shape` for same-order arrays. (I also wanted to optimize `zip_indexed` but it's more complex so I'll probably do it later)

I modified `zip_mut_with_same_shape` more than I intended at first because there were some useless statements. At that point, we have same-shape arrays, so they have the same length, which make the slicing useless.

The bench I added show that 'cc' is not slower than it was and 'ff' is as fast as 'cc'. On windows and WSL.

### DIFF
--- a/src/impl_methods.rs
+++ b/src/impl_methods.rs
@@ -1917,8 +1917,7 @@ where
     {
         debug_assert_eq!(self.shape(), rhs.shape());
 
-        // Same shape and order should have same strides
-        if self.strides() == rhs.strides() {
+        if self.dim.strides_equivalent(&self.strides, &rhs.strides) {
             if let Some(self_s) = self.as_slice_memory_order_mut() {
                 if let Some(rhs_s) = rhs.as_slice_memory_order() {
                     for (s, r) in self_s.iter_mut().zip(rhs_s) {
--- a/src/impl_methods.rs
+++ b/src/impl_methods.rs
@@ -6,7 +6,6 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
-use std::cmp;
 use std::ptr as std_ptr;
 use std::slice;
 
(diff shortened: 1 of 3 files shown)


src/impl_methods.rs
```diff
@@ -1917,18 +1916,20 @@ where
...
-        if let Some(self_s) = self.as_slice_mut() {
-            if let Some(rhs_s) = rhs.as_slice() {
-                let len = cmp::min(self_s.len(), rhs_s.len());
-                let s = &mut self_s[..len];
-                let r = &rhs_s[..len];
-                for i in 0..len {
-                    f(&mut s[i], &r[i]);
+
+        // Same shape and order should have same strides
+        if self.strides() == rhs.strides() {
```
jturner314: This is a little more restrictive than necessary. For axes with length <= 1, the stride doesn't matter. For example, if we have two arrays like this:

* shape: [1, 5], strides: [42, 1]
* shape: [1, 5], strides: [3, 1]

their memory layout is the same even though the strides for axis 0 are not equal (since the length of axis 0 is <= 1).

We can handle this case by adding a function like this to the `Dimension` trait (in `src/dimension/dimension_trait.rs`):

```rust
    /// Returns `true` iff `strides1` and `strides2` are equivalent for the
    /// shape `self`.
    ///
    /// The strides are equivalent if, for each axis with length > 1, the
    /// strides are equal.
    ///
    /// Note: Returns `false` if any of the ndims don't match.
    #[doc(hidden)]
    fn strides_equivalent(&self, strides1: &Self, strides2: &Self) -> bool {
        let shape_ndim = self.ndim();
        shape_ndim == strides1.ndim()
            && shape_ndim == strides2.ndim()
            && izip!(self.slice(), strides1.slice(), strides2.slice())
                .all(|(&d, &s1, &s2)| d <= 1 || s1 as isize == s2 as isize)
    }
```

and changing the check to

```rust
        if self.dim.strides_equivalent(&self.strides, &rhs.strides) {
```

I wouldn't expect this additional complication to have a significant performance impact (since I would expect most of the time to be in the iteration over the two arrays), but it's worth double-checking.

jturner314:
I re-ran CI and it passed now that #762 resolved the clippy warnings. Thanks for this PR @nilgoyette! It's always nice to boost performance with internal improvements like this.

PR #757, opened 2019-11-23
fixup build error on linux

baloo:
When compiling with `features=["blas"]` on linux, I would hit the
following error:
```
src/linalg/impl_linalg.rs:115: undefined reference to `cblas_ddot'
```

I believe cblas_ddot is included on macos, but not on any other OS, and
would therefor need some linker argument.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -8,4 +8,7 @@ fn main() {
     if cfg!(feature = "test-blas-openblas-sys") {
         println!("cargo:rustc-link-lib={}=openblas", "dylib");
     }
+    if cfg!(all(not(target_os = "macos"), feature = "blas")) {
+        println!("cargo:rustc-link-lib={}=blas", "dylib");
+    }
 }


baloo:
thinking about it again, and maybe this fix should live in https://github.com/blas-lapack-rs/cblas-sys instead. Please advise, I'm not too sure.

jturner314:
Will you please post your `Cargo.toml`? The error message seems to indicate that you need to specify a BLAS provider. See the [README](https://github.com/rust-ndarray/ndarray#how-to-use-with-cargo) for more information.

On Arch Linux, I can build and run the following project without issues:

```toml
# Cargo.toml

[package]
name = "example"
version = "0.1.0"
edition = "2018"

[dependencies]
ndarray = { version = "0.13", features = ["blas"] }
blas-src = { version = "0.2.0", default-features = false, features = ["openblas"] }
openblas-src = { version = "0.6.0", default-features = false, features = ["cblas"] }
```

```rust
// src/main.rs

use ndarray::prelude::*;

fn main() {
    let a = Array2::<f64>::eye(2);
    let b = Array2::<f64>::ones((2, 2));
    assert_eq!(b, a.dot(&b));
}
```

Note that the README suggests using `features = ["cblas", "system"]` for `openblas-src`. However, I've never gotten that to work on my machine. I haven't investigated why, since building from source (i.e. not including the `"system"` feature, as shown above) has always worked for me.

baloo:
I'm actually using it as a dependency of rsrl:
```
[package]
name = "smarthvac"
version = "0.1.0"
authors = ["Arthur Gautier <baloo@superbaloo.net>"]
edition = "2018"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[dependencies]
rsrl = "0.7.0"

slog = "*"
[patch.crates-io]
ndarray = { path = '../ndarray' }
```

baloo:
Sounds like you\re suggesting this is an error in the dependencies of rsrl and rsrl should have a feature to switch between cblas and openblas. Is this correct?

baloo:
well: 

```
[dev-dependencies]
blas-src = { version = "0.3", default-features = false, features = ["openblas"] }
openblas-src = { version = "0.6", default-features = false, features = ["cblas", "system"] }
```
in my Cargo.toml works, sorry for the noise, thanks for the help!

jturner314:
Okay, great! I'm glad it's working now. I think `rsrl` should mention in its documentation that the user needs to select a BLAS backend. I've created a PR (tspooner/rsrl#66).

Just a clarification that may be useful, since the naming is confusing -- `cblas` is C-language bindings to BLAS, which are enabled via the `cblas` feature on the BLAS implementation crate. (Standard BLAS is a Fortran library.) OpenBLAS is an implementation of BLAS. Other implementations include Accelerate, Intel MKL, and Netlib. See the `blas-src` crate for more information.

PR #777, opened 2020-01-21
Clarify versions of BLAS dependencies in README

jturner314:
This adds a table of the BLAS versions corresponding to old `ndarray` versions. It also changes the versions of the dependencies in the example to be consistent with the version of `ndarray` in the example.


bluss:
Nice thanks

PR #790, opened 2020-03-21
Added implementation of `var` and `std` methods for ArrayBase

kdubovikov:
I have found that ArrayBase is missing single-dimensional implementations for `var` and `std`. While you can use `var_axis(Axis(0), 1.)`, it can become verbose and unnecessary in code with lots of single-dimensional arrays. I think `sum_axis` and `mean_axis` have `sum` and `mean` for this reason too.

Please consider merging this pull request.

### DIFF
--- a/src/numeric/impl_numeric.rs
+++ b/src/numeric/impl_numeric.rs
@@ -213,7 +213,7 @@ where
     ///
     /// let a = array![1., -4.32, 1.14, 0.32];
     /// let stddev = a.std(1.);
-    /// assert_abs_diff_eq!(stddev, 6.7331, epsilon = 1e-4);
+    /// assert_abs_diff_eq!(stddev, 2.59483, epsilon = 1e-4);
     /// ```
     pub fn std(&self, ddof: A) -> A
     where
--- a/src/numeric/impl_numeric.rs
+++ b/src/numeric/impl_numeric.rs
@@ -14,7 +14,6 @@ use crate::itertools::enumerate;
 use crate::numeric_util;
 
 use crate::{FoldWhile, Zip};
-use num_traits::real::Real;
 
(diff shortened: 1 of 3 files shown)


kdubovikov:
I see that the build fails at `cargo clippy` check un an unrelated file. Unsure if I need to fix that.

bluss:
Thanks! Looks similar to #753 I don't mind whichever gets merged first

ZuseZ4:
@jturner314 and @bluss I think it makes sense to merge this?
If we merge #848 first and rerun Travis here it shouldn't fail anymore. Just to be sure.

kdubovikov:
@jturner314 , @bluss , @ZuseZ4 anything else I should do to help with the merge?

bluss:
Looks great. I've rebased to current master so that tests will run properly with new CI. Given the commit history, I'd merge this with squash to clean it up a bit

bluss:
Thanks!

bluss:
cc @xd009642 Your comment would actually be interesting - w.r.t overlapping functionality with ndarray-stats. I don't think ndarray needs to include more of this, also unsure how we ended up here. Moving this method to ndarray-stats wouldn't be wrong IMO.

xd009642:
Yeah I wouldn't be opposed to moving this functionality to ndarray-stats and then deprecating here

bluss:
Oh so we ended up with var and std here - because we already had them in the axis variants here since long ago. It's hard to draw the line what's so basic that it could stay here - but from the summary functions I guess only `.sum()` is the so basic one, the rest could really move to stats.

bluss:
Unfortunately this .var() implementation should be rather slow - using the iterator for array traversal unfortunately. Should use Zip like var_axis does.

PR #792, opened 2020-03-29
Dimension::Larger should implement RemoveAxis

TheLortex:
Hi, this is something that has been useful to me while doing generic programming on array dimensions. All the implementors of `Dimension` already verify `T::Larger: RemoveAxis`, so this doesn't break anything internally.


bluss:
This looks good but we need CI to run for the PR

bluss:
CI ran in build https://travis-ci.org/github/rust-ndarray/ndarray/builds/668523130  ok except for rustfmt which is crashing on a macro

Issue #685, opened 2019-08-15
Change uninitialized to return an array of MaybeUninit


jturner314:
The `.uninitialized()` method on `ArrayBase` has some issues. For example, `Array1::<bool>::uninitialized(2)` is undefined behavior.

Rust 1.36 added `std::mem::MaybeUninit` for safer handling of uninitialized data. We should replace the existing `.uninitialized()` method on `ArrayBase` with one that returns an array of `MaybeUninit` instances. We then need to add the necessary methods to cleanly work with arrays like this (e.g. an array-level equivalent of `assume_init`).

bluss:
Internally I'd kind of just like a `Zip` with ndproducers that would give out raw pointers to the elements (so that we avoid references and iteration is sound). That way we can use uninitialized or similar correctly using some raw pointer code, at least.

bluss:
I'm not sure the recipe in the title of this issue is what we should do - there might be better ways to do it, for example to provide raw pointer traversals

bluss:
Can we explain why `Array1::<bool>::uninitialized(2)` is UB?  What's the difference to `Vec::<bool>::with_capacity(2)`?

bluss:
Internal implementation of this is added in #797, that's a start.

PR #797, opened 2020-04-13
Add methods to collect Zip into an array

### LINKED ISSUE #755: Is it possible to collect a Zip into Array?
Currently if we need to do elementwise operation on array, we can create a result array, and wrap it together with parameters arrays in `Zip` and use `apply` to assign the result. However, since `Zip` have the shape information, can we add a helper function to allocate a result array with correct shape, so that we don't need to create the result array manually?

bluss:
Add

- Zip::apply_collect
- Zip::apply_assign_into
- Zip::par_apply_collect
- Zip::par_apply_assign_into

Add internal (for now) constructors for `MaybeUninit<T>`-filled arrays.
Restrict to Copy elements to use collect (because we don't implement dropping a
halfway filled array yet).

This is added to all but the last arity of Zip (because the last item is needed for the result that we collect into.

The assignment functions are generalized so that we can assign to `&mut T`, `&mut MaybeUninit<T>` and `&Cell<T>` when we have elements of `T`. This seems powerful, and could be a concept that's explored further (methods .assign() and .fill() come to mind).
A very minor question is what to name and in which module to put the new trait for this - `AssignElem<T>`.

Fixes #755

### DIFF
--- a/src/zip/mod.rs
+++ b/src/zip/mod.rs
@@ -18,7 +18,6 @@ use crate::Layout;
 use crate::NdIndex;
 
 use crate::indexes::{indices, Indices};
-use crate::layout::LayoutPriv;
 use crate::layout::{CORDER, FORDER};
 
 /// Return if the expression is a break value.
--- a/src/zip/mod.rs
+++ b/src/zip/mod.rs
@@ -10,6 +10,7 @@
 mod zipmacro;
 
 use crate::imp_prelude::*;
+use crate::AssignElem;
 use crate::IntoDimension;
 use crate::Layout;
 use crate::NdIndex;
@@ -579,6 +580,7 @@ pub struct Zip<Parts, D> {
     layout: Layout,
 }
 
+
 impl<P, D> Zip<(P,), D>
 where
     D: Dimension,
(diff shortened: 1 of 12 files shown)


src/zip/mod.rs
```diff
@@ -982,6 +982,29 @@ macro_rules! map_impl {
...
+            ///
+            /// Restricted to functions that produce copyable results for technical reasons; other
+            /// cases are not yet implemented.
+            pub fn apply_collect<R>(self, mut f: impl FnMut($($p::Item,)* ) -> R) -> Array<R, D>
+                where R: Copy,
+            {
+                unsafe {
+                    let is_c = self.layout.is(CORDER);
+                    let is_f = !is_c && self.layout.is(FORDER);
+                    let mut output = Array::uninitialized(self.dimension.clone().set_f(is_f));
```
bluss, line 999: If this is considered UB - possibly, see comments in #685 - then a separate buffer type needs to be implemented.

bluss:
In the grand scheme of things, it would be more exciting if

1. any NdProducer could be collected
2. Zip can be an NdProducer of tuples by default.
3. One could with a method - `.map(f)` - create a different NdProducer from a `Zip` value itself.

but it's not certain that the required changes are worth it. Or if @jturner314 hasn't implemented something similar already :)

src/zip/mod.rs
bluss: It looks like the problematic part remaining with this is `Vec::drop` which uses effectively `&mut self[..]`, and this code needs to be revisited.

bluss:
Now using MaybeUninit, because `Vec::drop` is what makes it unsound to toss uninit elements around. This is a lot closer to what @jturner314 said all along.

bluss:
For non-Copy elements (more specifically, elements with drop, but we don't have a trait bound for that); there are hints solutions - I have an unbeautiful prototype for counting elements and dropping the right number of them on panic; but this doesn't transfer at all to the parallel case (in the parallel case, we have multiple different "write heads" into the array, and they all need to "roll back" if any thread panics), and that's a sign a better solution is needed of that problem. So it's not coming here.

I'm also saying that I think it's important to handle drop correctly - on panic, the elements that were created should drop. It would be easy to ignore the problem by just saying that there is no drop handling, but I would judge that's not robust.

That said, if `MaybeUninit` arrays are just exposed, they are very useful and the user can decide themselves if they want to skip handling drops on panic.

PR #798, opened 2020-04-13
Support uninitialized elements in .assign() and .fill()

bluss:
The trick is simply to traverse the target array using a raw view / raw
pointers; this means we don't create any references to
potentially-uninitialized values.

Assignment `*x = y.clone()` is used where x is a *mut A, y is a &A;
this will work as intended when A: Copy; assignment should not need that
the value being overwritten, when Copy, is initialized.

Unfortunately we use Zip here instead of the older methods
(zip_mut_with, unordered_foreach_mut); because Zip implements equivalent
functionality but in a generic way. The access pattern and performance
is not guaranteed to be identical.

Related to issue #685

### DIFF
--- a/src/impl_constructors.rs
+++ b/src/impl_constructors.rs
@@ -471,8 +471,9 @@ where
     /// ### Safety
     ///
     /// Accessing uninitalized values is undefined behaviour. You must
-    /// overwrite *all* the elements in the array after it is created; for
-    /// example using the methods `.fill()` or `.assign()`.
+    /// overwrite *all* the elements in the array after it is created; this must
+    /// be done carefully, for example using the methods `.fill()` or `.assign()`,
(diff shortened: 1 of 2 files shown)


bluss:
Benchmarks suggest .fill() is unchanged. Looks like we don't have an assign benchmark.

bluss:
Source to back us up on that assignment through `*x` on raw pointer will have the intended effect https://old.reddit.com/r/rust/comments/cefgec/howto_replace_memuninitialized_with_memmaybeuninit/eu87689/

bluss:
Thoughts for next time: Rethink - we introduced `maybe_uninit` instead. But is this still worth doing? Maybe, but need to include new trait AssignElem in the design. Additional trait for overwriting?

PR #810, opened 2020-04-25
Bump all dependencies

### LINKED ISSUE #765: Bump blas to 0.4.0
ndarray-linalg uses 0.4.0 too

Eijebong:
* Update blas-src to 0.6
* Update itertools to 0.9
* Update openblas-src to 0.9

Fixes #765

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,7 +38,7 @@ approx = { version = "0.3.2", optional = true }
 
 # Use via the `blas` crate feature!
 cblas-sys = { version = "0.1.4", optional = true, default-features = false }
-blas-src = { version = "0.2.0", optional = true, default-features = false }
+blas-src = { version = "0.6.1", optional = true, default-features = false }
 
 matrixmultiply = { version = "0.2.0" }
 serde = { version = "1.0", optional = true }
@@ -48,7 +48,7 @@ rawpointer = { version = "0.2" }
 defmac = "0.2"
 quickcheck = { version = "0.9", default-features = false }
 approx = "0.3.2"
-itertools = { version = "0.8.0", default-features = false, features = ["use_std"] }
+itertools = { version = "0.9.0", default-features = false, features = ["use_std"] }
 
 [features]
 # Enable blas usage
--- a/blas-tests/Cargo.toml
+++ b/blas-tests/Cargo.toml
@@ -10,7 +10,7 @@ test = false
 [dev-dependencies]
 approx = "0.3.2"
 ndarray = { path = "../", features = ["approx", "blas"] }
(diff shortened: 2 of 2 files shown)


termoshtt:
num-complex 0.3.0 has also released: https://github.com/rust-num/num-complex/releases/tag/num-complex-0.3.0

nilgoyette:
With rayon, num-complex, approx + the three above, this PR is obsolete. What's the procedure? Replace this PR with another or nicely ask @Eijebong to update?

cjordan:
Can I chime in naively - I forked `ndarray` just so I could bump its `num-complex` dependency, as another dependency was already using `num-complex` 0.3.0 and the different version apparently means that cargo refused to compile my project. So, is there any huge reason why `ndarray` should be restricted to 0.2.* or 0.3.*?  Could it just be 0.* instead?  I've noticed a similar problem for `rayon`.  In both cases, I did not see an API change so everything just worked fine.

ZuseZ4:
@cjordan 
https://semver.org/

From what I know, specifying a dependency like done here (a.b.*) assures, that the project will always compile correctly.
By only specifying the major version a.*.* you may end up in a situation where one of your dependencies adds a pub function (or anything else) that you already implement in your own library. Cargo will pull the newest version that you allow. This might lead to nasty bugs where the runtime behavior silently changes, since the compiler adds the new implementation from your dependency, instead of your own one which has been used before.
Of course that's an unlucky situation and mostly won't happen, but the idea is to generally avoid such pitfalls by specifying the minor version too.
Please feel free to correct what I wrote in case that I got something wrong.

mulimoen:
@ZuseZ4 You would only run into this problem when using glob imports, and this would be a compile time error at worst. You would not avoid this by specifying the minor version, this would simply mean all versions `>=major.minor` would be compatible with this crate, and minor bumps are allowed to add new functions.

@cjordan One has to specify `major` (or `0.minor`) to publish a crate

`ndarray` could use the bound `>=0.2.0, <0.4.0` for `num-complex` to allow either major version of `num-complex` if we require backwards compatability.

bluss:
I would encourage a release to be made with this - I know I'm late to do it, but I will, if nobody else has time.

xd009642:
@bluss yeah I'd like a release with what we currently have, a dependency update and also these small PRs to go in :smile: 

* https://github.com/rust-ndarray/ndarray/pull/848 
* https://github.com/rust-ndarray/ndarray/pull/847
* https://github.com/rust-ndarray/ndarray/pull/834

If you have time to look at ndarray that would be grand :grin:

bluss:
Excellent @xd009642

PR #811, opened 2020-04-25
Fix clippy lints

bluss:
The clippy.toml setting is a workaround for the currently broken single
char variable lint (a fix is pending in clippy).


bluss:
Adding the workaround since we use clippy in the beta channel, and it might not be fixed in a while(?)

PR #815, opened 2020-04-29
Implement IndexedParallelIterator for one-dimensional arrays

bluss:
The rayon IndexedParallelIterator trait maps well to any one-dimensional indexed
sequence, and we can implement this trait as a special case (of the general
unindexed parallel iterator of an arbitrary array).

Implementing this is logical, it fits, but we still will recommend ndarray's `Zip`
for lock step iteration. The reason is performance. `Zip` will do it better.

### DIFF
--- a/src/parallel/par.rs
+++ b/src/parallel/par.rs
@@ -184,7 +184,14 @@ macro_rules! par_iter_view_wrapper {
         fn fold_with<F>(self, folder: F) -> F
             where F: Folder<Self::Item>,
         {
-            self.into_iter().fold(folder, move |f, elt| f.consume(elt))
+            Zip::from(self.0).fold_while(folder, |mut folder, elt| {
+                folder = folder.consume(elt);
+                if folder.full() {
+                    FoldWhile::Done(folder)
+                } else {
+                    FoldWhile::Continue(folder)
+                }
+            }).into_inner()
         }
     }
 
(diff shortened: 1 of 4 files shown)


src/parallel/par.rs
```diff
@@ -174,12 +188,49 @@ macro_rules! par_iter_view_wrapper {
...
 
+    impl<'a, A> Producer for ParallelProducer<$view_name<'a, A, Ix1>>
+        where A: $($thread_bounds)*,
+    {
+        type Item = <$view_name<'a, A, Ix1> as IntoIterator>::Item;
+        type IntoIter = <$view_name<'a, A, Ix1> as IntoIterator>::IntoIter;
+
+        fn into_iter(self) -> Self::IntoIter {
+            self.0.into_iter()
+        }
```
bluss, line 199: It actually reduces to the default iterator as a base case, that could be a regression in performance from currently (default iterator uses logical order)

bluss: Visit order doesn't matter for 1D, only implementation quality (unless we count preferring forward over reverse, that is positive vs negative stride, and we don't have evidence for that).

Before we had the view's `self.into_iter().fold(..)`, so the fold case for the reduction. In the new indexed case, we don't always access the iterator using fold, and that's the possible quality problem.

bluss:
These implementations logically fit, but the performance of the enabled methods is not good - so not having these is a way to push users towards using `ndarray::Zip` for zipping and unordered parallel iterator methods anyway, which perform many times better.

Issue #726, opened 2019-09-23
Decide which auto traits the raw views should implement


bluss:
Decide which (if any at all) of the auto traits the `RawArrayView/Mut` should implement.

PR #817, opened 2020-05-11
Implement parallel collect to array for non-Copy elements

bluss:
Allow non-copy elements by implementing dropping partial results from
collect (needed if there is a panic with unwinding during the
apply-collect process).

It is implemented by:

   1. allocate an uninit output array of the right size and layout
   2. use parallelsplits to split the Zip into chunks processed in parallel
   3. for each chunk keep track of the slice of written elements
   4. each output chunk is contiguous due to the layout being picked to
   match the Zip's preferred layout
   5. Use reduce to merge adjacent partial results; this ensures
   we drop all the rests correctly, if there is a panic in any thread


bluss:
Depends on #815

bluss:
The SendProducer wrapper here is needed before #726 is decided. I'd be in favor of letting them have Send/Sync impls now.

PR #826, opened 2020-06-12
Add WindowsMut to complement Windows

hanmertens:
It's the only iterator for which I missed a mutable counterpart.


mulimoen:
Would this not allow one to get two mutable slices to the same elements? The windows will overlap like:
win0: xXX----
win1: -XXx---
unless the iterator disallows having only one active item (which it can't)?

hanmertens:
Well, I guess I didn't really think this through enough :sweat_smile:, you're completely right. I only considered using it in a for loop, but in general there is no guarantee the previous window is dropped before you generate a new one.

bluss:
Maybe with ArrayView of Cells, we can get there. It's on the way (and the cast method for raw views is already there, to enable others to experiment)

PR #839, opened 2020-08-17
Numpy like cumsum

npapapietro:
Newish to rust and ndarray library. I'm binding rust to python and implemented the cumsum functions that mimic the numpy API and thought I'd share them back the to main repo. Not sure if I added something or just reinvented the wheel here. Any input appreciated!


npapapietro:
already an accumulate function, maybe i'll add a documentation edit instead to the ndarray for python users

PR #847, opened 2020-11-11
Update docs

xd009642:


xd009642:
@bluss @jturner314 I think this is a better solution for now than supporting conversions in ndarray.

xd009642:
Just a note it was mentioned on #554 but the creator of nshare updates it whenever tagged by someone that they need a new version, but I also have write access to the repo and he's willing to transfer it to the ndarray org if we want it :smile:

bluss:
Thanks. Eventually I guess we will need a website, something more dynamic for this info. But I like using the docs page as well

PR #850, opened 2020-11-24
Make stack and concatenate compliant with numpy naming.

andrei-papou:
This PR changes the naming used for stack/concatenate to be compliant with the one from numpy. In particular:

- `stack` function and macro are now aliases to `stack_new_axis`. They join the arrays along a new axis.
- the implementation of former `stack` function is now moved to the `concatenate` function. It is used to join the arrays along the existing axis.
- `stack` function and macro are no longer deprecated.

This change is **not** backward-compatible and should be released within 0.14.


bluss:
Thanks, let's go for this in 0.14

PR #854, opened 2020-12-03
Add cross testing of 32-bit architectures in CI

bluss:
Test on 32-bit (we even have an open bug on 32-bit).
We avoid native libraries (no BLAS). Rayon code seems to test fine, though.


bluss:
Merge with 32-bit tests failing, then we can test the bugfix of that in its PR (coming).

Issue #708, opened 2019-09-12
no_std for ndarray


bluss:
**Using std** gives us the following features:

- `std::arch` and target feature detection for seamless simd
- .

Having optional `no_std` support would give us these tangible benefits:

- .

I have created this issue so that the pros and cons of `no_std` can be explained. As of this writing, using `no_std` does not seem to be attractive.

As usual, all new features are to be used because of what gain they give us, not to score points. Formulate goal & gain before implementing.

LukeMathWalker:
I guess the `no_std` game is more about enabling the usage of `ndarray` in `no_std` context than any tangible benefit to the library itself in terms of functionality.
Being quite ignorant when it comes to `no_std` environments (e.g. embedded), I'd be curious to understand:
- what is the current situation in terms of math libraries?
- is there any interest in using something similar to `ndarray`?
- would stack-allocated arrays be more desirable, given the device constraints?

jturner314:
Fwiw, adding (optional) `no_std` support to `ndarray` is pretty much just a matter of changing `use` statements from `std` to `core`/`alloc`, with the following exceptions:

* `matrixmultiply` (a required dependency of `ndarray`) currently doesn't support `no_std`. bluss/matrixmultiply#46 is a work-in-progress PR to resolve this.

* The `Float` trait from `num-traits` isn't currently available in `no_std` environments. Hopefully, that'll be resolved soon with rust-num/num-traits#99.

So once those things are resolved, adding `no_std` support should be straightforward. There'll be only marginal additional complexity if we start using explicit SIMD. (In `no_std`, we have to rely on compile-time feature detection; the `is_x86_feature_detected` macro isn't available.)

I do think `ndarray` could be useful in embedded contexts, even without an allocator. With an allocator, it would be possible to use the full functionality of `ndarray`.

vadixidav:
@bluss It looks like all the road blocks are now out of the way to make this possible. If nobody else volunteers to do this then I can do it.

bluss:
Nobody has answered my question of what the benefit of doing it is 😅

vadixidav:
[...]

PR #861, opened 2020-12-13
Place NdFloat and num-traits/std behind a new "std" feature

xd009642:
This PR adds a new default feature "std" and when present this activates the "std" feature in num-traits and includes the `NdFloat` trait defined in ndarray. This is part of the tasks required in order to provide a `no_std + alloc` version of ndarray for issue #708 

This is (an unavoidable) breaking change for any users that set `default-features = false`, other ndarray users will be unaffected.

### DIFF
--- a/src/linalg_traits.rs
+++ b/src/linalg_traits.rs
@@ -6,7 +6,9 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 use crate::ScalarOperand;
-use num_traits::{Float, One, Zero};
+#[cfg(feature = "std")]
+use num_traits::Float;
+use num_traits::{One, Zero};
 use std::fmt;
 use std::ops::{Add, Div, Mul, Sub};
 use std::ops::{AddAssign, DivAssign, MulAssign, RemAssign, SubAssign};
@@ -47,6 +49,7 @@ impl<T> LinalgScalar for T where
 /// operations (`ScalarOperand`).
 ///
 /// This trait can only be implemented by `f32` and `f64`.
+#[cfg(feature = "std")]
 pub trait NdFloat:
     Float
     + AddAssign
@@ -65,5 +68,8 @@ pub trait NdFloat:
 {
 }
 
+#[cfg(feature = "std")]
 impl NdFloat for f32 {}
+#[cfg(feature = "std")]
 impl NdFloat for f64 {}
+
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ test = true
 
 [dependencies]
 num-integer = "0.1.39"
-num-traits = "0.2"
+num-traits = { version = "0.2", default-features = false }
 num-complex = { version = "0.3", default-features = false }
 
 rayon = { version = "1.0.3", optional = true }
@@ -51,6 +51,7 @@ approx = "0.4"
 itertools = { version = "0.9.0", default-features = false, features = ["use_std"] }
 
 [features]
+default = ["std"]
 # Enable blas usage
 # See README for more instructions
 blas = ["cblas-sys", "blas-src"]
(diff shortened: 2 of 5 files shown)


xd009642:
So this is designed as a PoC on a potential `NdFloatCore` with more minimal floating point requirements, as well as cfg'ing `NdFloat` behind a new `std` feature. This is part of the work/discussion to make ndarray no_std from here https://github.com/rust-ndarray/ndarray/issues/708

vadixidav:
I don't see where the new std feature is added here. Make sure you add that too.

xd009642:
Oops forgot to add to the commit. I ran a cargo fmt and it changed pretty much everything so missed the cargo.toml when adding just non-fmt changes 😬

xd009642:
Well tests pass so I guess I'll mark this as ready to review. As a summary of the motivation behind this PR. 

It is some initial work for #708 

This adds an std feature and makes the std feature of num-traits optional. It adds std to the default features so no breaking changes for crates that don't specify `default-features = false`, however for crates that do that and use `NdFloat` there will be a breaking change.

For the tests that use `NdFloat` I just changed them to use `NdFloatCore` for no_std tests. I'm aware this means the tests are using different traits for the std or no_std tests, but I think this is alright in the circumstance but if anyone disagrees it's an easy change for me to make :smile:

bluss:
> This trait is only available with the std feature, or with the libm feature otherwise.

Num-traits Float says it can be used with no-std in this way. So we have exactly this problem: https://stackoverflow.com/questions/61769086/  (using std or libm but not both at the same time), with corresponding cargo feature request https://github.com/rust-lang/cargo/issues/1839

vadixidav:
@bluss Now that you mention it, we should probably just require the libm feature. I think that the crate doesn't use libm if std is present, but I haven't double checked yet. This would take care of this float problem.

bluss:
We don't want to have the extra build time of libm if we don't use it.

src/linalg_traits.rs
```diff
@@ -65,5 +68,36 @@ pub trait NdFloat:
...
+
+/// A no_std floating-point element type represented by `f32` and `f64`.
+///
+/// Trait `NdFloatCore` is only implemented for `f32` and `f64` and encompasses
+/// much float-relevant ndarray functionality as possible, while working for no_std
+/// code, including the traits needed for linear algebra and for *right hand side* scalar
+/// operations (`ScalarOperand`).
+///
+/// This trait can only be implemented by `f32` and `f64`.
+pub trait NdFloatCore:
```
bluss, line 84: I'd propose that we don't add a new NdFloatCore trait (The story behind NdFloat itself is fuzzy as it is) - we don't need it.

vadixidav: So just remove NdFloat then?

xd009642: ndarray-linalg relies on NdFloat and maybe some other ndarray using crates.. I'm not sure about just removing it. But I'll remove `NdFloatCore`

xd009642:
So @bluss I've removed `NdFloatCore`, and also made the oper tests f32 specifically so they'll still run without the std feature. I don't think it's worth using cfg to only run them for std given it's functionality that could exist on no_std as well

bluss:
Thanks, makes sense to me to do this piecemeal, even though I made a suggestion make most of the PR disappear :) Could you update PR title and description before merge? I'd also rebase/squash it all together in this case I think (as you want). Lgtm.

xd009642:
@bluss renamed, redescribed, and squashed :smile:

bluss:
Thanks!

bluss (approved):

bluss:
You can merge it, since you can

xd009642:
my first merge to ndarray :tada:

PR #862, opened 2020-12-14
Fix Zip::indexed for the 0-dimensional case

jturner314:
This commit fixes a panic for 0-dimensional, indexed `Zip` instances
which results from an out-of-bounds index in a call to
`IndexPtr::stride_offset` in `Zip::inner`. Basically, the "stride" for
`IndexPtr` is the axis to update, but for the 0-dimensional case,
there are no axes, so `IndexPtr::stride_offset` cannot be called
without panicking due to the `self.index[stride]` access.

The chosen solution is to add a special check to `Zip::apply_core` for
the 0-dimensional case. Another possible solution would be to modify
the loop of `Zip::inner` such that an offset would not be performed
for an index of zero.

### DIFF
--- a/src/zip/mod.rs
+++ b/src/zip/mod.rs
@@ -691,12 +691,14 @@ impl<P, D> Zip<P, D>
 where
     D: Dimension,
 {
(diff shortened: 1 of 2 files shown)


bluss:
Thanks!

PR #864, opened 2020-12-17
No_std support for ndarray

### LINKED ISSUE #708: no_std for ndarray

SparrowLii:
This pr enables ndarray to be used in the no_std environment, and maintains most of the functions. Fixes #708.
The `geomspace` `linspace` `logspace` `range` `var_axis` and `std_axis` methods are only available when `std` is enabled at current. And 'blas''rayon''serde' features are not available without std too.


SparrowLii:
libcore cannot be used in blas-tests/oper.rs, which is a bit strange

bluss:
Hm, do we need to start it this way? I'd suggest using the extern crate core as std solution, as I mentioned in the issue.

We need to figure out which *tests* to run in the no-std configuration. Crate test in itself depends on std, so we can keep using std in the test files.

SparrowLii:
In that case, it means that all dependencies on std in ndarray must be handled at once. Well, I will try that method.

src/iterators/mod.rs
```diff
@@ -26,7 +27,7 @@ pub use self::chunks::{ExactChunks, ExactChunksIter, ExactChunksIterMut, ExactCh
 pub use self::lanes::{Lanes, LanesMut};
 pub use self::windows::Windows;
 
-use std::slice::{self, Iter as SliceIter, IterMut as SliceIterMut};
+use alloc::slice::{self, Iter as SliceIter, IterMut as SliceIterMut};
```
bluss, line 30: This could still be std (it's in core)

SparrowLii: Thanks for reviewing! I will modify it in the next commit

SparrowLii:
@bluss It seems that we need to resolve the dependencies of external packages next. Do you think we should do it together here or in a new pr?

Cargo.toml
```diff
@@ -28,19 +28,19 @@ bench = false
...
 
-approx = { version = "0.4", optional = true }
+approx = { version = "0.4", optional = true , default-features = false }
 
 # Use via the `blas` crate feature!
 cblas-sys = { version = "0.1.4", optional = true, default-features = false }
 blas-src = { version = "0.6.1", optional = true, default-features = false }
 
-matrixmultiply = { version = "0.2.0" }
+matrixmultiply = { version = "0.2.0", default-features = false}
```
bluss, line 43: ndarray/std must enable matrixmultiply/std; runtime target feature detection is one of the most important features of matrixmultiply

Cargo.toml
```diff
@@ -28,19 +28,19 @@ bench = false
 test = true
 
 [dependencies]
-num-integer = "0.1.39"
-num-traits = { version = "0.2", default-features = false }
+num-integer = { version = "0.1.39", default-features = false }
+num-traits = { version = "0.2", default-features = false , features = ["libm"] }
```
bluss, line 32: This adds a dependency to the default configuration, will have to look at that

Cargo.toml
SparrowLii: Is there any solution? like first detect features and then introduce modules

Cargo.toml
SparrowLii: We do need some methods in Float( ln() exp() etc), I think the best way is to use libm features. Or is there any way to detect [default-features] first and then introduce crates?

Cargo.toml
bluss: just write std = ["matrixmultiply/std"] under features and it's what we need

Cargo.toml
bluss: The exclusion of the float trait (already merged) assumed we do it without libm, so I think we can continue on that path. What would break if we don't have libm? We need num-traits with std, when ndarray/std is enabled.

Cargo.toml
SparrowLii: Ok, got it.

Cargo.toml
SparrowLii: Ok, i will try this way. Thanks for reviwing!

bluss: Is it usable for ndarray's purposes? Users can add features to num that they need for their purposes. :)

SparrowLii: I'm afraid it is. Will be used in src/geomspace.rs:36, in Iterator for Geomspace.

bluss: FloatCore is public. It's in num-trait, it's not our trait.

Then we just exclude geomspace when std is not enabled, for now

SparrowLii: sry it's my mistake. It works :) I will do this way. Thanks a lot!

SparrowLii:
@bluss now it use FloatCore as Float in no_std environment. I think it will not affect the function under the std environment and can theoretically be used under no_std

bluss:
@SparrowLii That kind of solution (FloatCore as Float) is not someting we can use in our public interface, it will create incompatibilities between our dependencies (it's a non-additive change to our API depending on features, which is not allowed)

src/lib.rs
```diff
@@ -109,6 +110,19 @@
...
+
+#[cfg(feature = "std")]
+extern crate std;
+#[cfg(not(feature = "std"))]
+extern crate core as std;
+
+#[cfg(feature = "std")]
+pub use num_traits::Float;
+#[cfg(not(feature = "std"))]
+pub use num_traits::float::FloatCore as Float;
```
bluss, line 124: This line must be deleted. Then we follow the consequence of that for `not(feature="std")`.

bluss: If we need a creative solution, then we can "redirect" to support only f32, f64 on no-std (not general FloatCore!). Then we can feature-safely expand our methods bounded on `A: Float` to support the full `A: Float` range with std.

SparrowLii: Apologies for my lack of experience. I will change it. 
I think it is cool to expand for FloatCore under no_std, but if we don’t use libm, we might have to implement exp() and ln() functions ourselves. I doubt about that.

SparrowLii:
> If we need a creative solution, then we can "redirect" to support only f32, f64 on no-std (not general FloatCore!). Then we can feature-safely expand our methods bounded on A: Float to support the full A: Float range with std.

Apologies for my lack of experience. Have changed it.
I think it is cool to expand for FloatCore under no_std, but if we don’t use libm, we might have to implement exp() and ln() functions ourselves. I doubt about that.

bluss:
Replacing the libm feature is not a goal no, don't worry about that.

SparrowLii:
> Replacing the libm feature is not a goal no, don't worry about that.

Sorry for a little not understanding. So it means adding libm features in ndarray?

SparrowLii:
Added libm features, so we can use --no-defualt-features --features "libm" to use all Float functions under no_std

SparrowLii:
@bluss any more advice? I think it can already be used under no_std. Maybe we should find a good way to verify?

src/impl_constructors.rs
```diff
@@ -11,9 +11,9 @@
 //!
 
 #![allow(clippy::match_wild_err_arm)]
-#[cfg(feature = "std")]
+#[cfg(any(feature = "std", feature = "libm"))]
 use num_traits::Float;
-#[cfg(not(feature = "std"))]
+#[cfg(not(any(feature = "std", feature = "libm")))]
 use num_traits::float::FloatCore as Float;
```
bluss: These two lines still need to be removed - we can't have a public interface (for example Array::linspace) that changes API in a breaking way (from Float to FloatCore) depending on feature flags

bluss:
I think this is a hard task if we try to do the impossible (making FloatCore/Float fit together). Let's follow a simple recipe which means that std is an additive feature. With std disabled, some features are not there.

Anything else than that can be improved later.

src/lib.rs
```diff
@@ -12,6 +12,7 @@
     clippy::deref_addrof,
     clippy::unreadable_literal
 )]
+#![no_std]
```
bluss, line 15: I would make this line conditional with cfg-attr

SparrowLii:
> I think this is a hard task if we try to do the impossible (making FloatCore/Float fit together). Let's follow a simple recipe which means that std is an additive feature. When std is disabled, some API is going to be missing because of that.
> 
> Anything else than that can be improved later.

Okay then, it should be better.

SparrowLii:
@bluss Now we avoid using the Float feature in the no_std environment. It can run under no_std. WDYT?

bluss:
Great work, looks like it works. That's all we need for now. Features can be added back later.

1. Do we have enough tests? There should be a test that ensures we build properly in no-std, not using std, and it should run in CI.
2. Look at feature documentation. Features are documented in the module doc and in the README like this, and we need to add something for std. See link https://docs.rs/ndarray/0.14.0/ndarray/#crate-feature-flags
3. We need to document on the features themselves. See in our docs you can find we say `Requires crate feature "approx"` on items that require that feature and so on.  (There is an experimental Rust feature for nicer looking labels for this - but we don't need to use that feature yet if you don't want to - we can keep it simple and migrate everything to that later)
4. Update the PR title, PR description and make them ready for merge. Update the description so that it properly says that it fixes the issue that it fixes, not just talks about it.
5. Use interactive rebase to squash and clean up history (optional) - the PR commits are a bit messy - if you want to skip this, this change can be "squash merged", as one commit instead.

Cargo.toml
```diff
@@ -66,15 +67,15 @@ test = ["test-blas-openblas-sys"]
...
+std = ["num-traits/std", "matrixmultiply/std"]
 
 [profile.release]
 [profile.bench]
 debug = true
 
 [workspace]
 members = ["ndarray-rand", "serialization-tests", "blas-tests"]
-exclude = ["numeric-tests"]
+exclude = ["numeric-tests","ensure_no_std"]
```
bluss, line 78: What's this `ensure_no_std`, it sounds like a good idea but I don't think it exists in this PR

SparrowLii:
@bluss I updated the description of std feature and updated the pr description. But there are still two problems:
I did build test on linux-armv7(cargo build --target=thumbv7m-none-eabi --no-default-features)and it built successfully. But because I don’t know much about CI, I don’t know how to add it to the CI test script. I try to refer to the no_std of matrixmultiply, but there is no .travis.yml file in the ndarray. 
I am happy to help optimize the document, but apart from adding a description of the std feature, I can’t find specific places to do it. 
In addition, I tried to rebase, but encountered a lot of conflicts that I could not handle, so I thought it would be better to execute "squash merged".

Cargo.toml
SparrowLii: I tried to refer to the no_std of matrixmultiply, But because I don’t know much about CI,  this idea cannot be put into practice at currect :(

README.rst
```diff
@@ -48,6 +48,17 @@ Crate Feature Flags
 The following crate feature flags are available. They are configured in
 your `Cargo.toml`.
 
+- ``std``
+
+  - This crate can be used without the standard library by disabling the
+    default `std` feature. Use this in `Cargo.toml`:
+
+    [dependencies]
+    ndarray = { version = "0.14.0", default-features = false }
```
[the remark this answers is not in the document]
SparrowLii: OK, has updated

src/lib.rs
```diff
@@ -68,6 +69,14 @@
 //! The following crate feature flags are available. They are configured in your
 //! `Cargo.toml`.
 //!
+//! - `std`
+//!  - Rust Standard Library
+//!  - This crate can be used without the standard library by disabling the
+//!    default `std` feature. To do so, use `default-features = false` in
+//!    your `Cargo.toml`.
+//!  - The `geomspace` `linspace` `logspace` `range` `std` `var` `var_axis`
+//!    `std_axis` and `all_close` methods are only available when `std` is
```
bluss, line 78:
```suggestion
//!    and `std_axis` methods are only available when `std` is
```

README.rst
```diff
@@ -48,6 +48,19 @@ Crate Feature Flags
...
+
+  - Rust Standard Library
+
+  - This crate can be used without the standard library by disabling the
+    default `std` feature. To do so, use this in your `Cargo.toml`:
+
+    [dependencies]
+    ndarray = { version = "0.x.y", default-features = false }
+
+  - The `geomspace` `linspace` `logspace` `range` `std` `var` `var_axis` `std_axis`
```
bluss, line 61:
```suggestion
  - The `geomspace` `linspace` `logspace` `range` `std` `var` `var_axis` and `std_axis`
```

README.rst
```diff
@@ -48,6 +48,19 @@ Crate Feature Flags
...
+    and `all_close` methods are only available when `std` is enabled.
```
bluss, line 62:
```suggestion
    methods are only available when `std` is enabled.
```

src/iterators/macros.rs
```diff
@@ -1,7 +1,7 @@
 // Send and Sync
 // All the iterators are thread safe the same way the slice's iterator are
 
-// read-only iterators use Sync => Send rules, same as `std::slice::Iter`.
+// read-only iterators use Sync => Send rules, same as `alloc::slice::Iter`.
```
bluss, line 4:
```suggestion
// read-only iterators use Sync => Send rules, same as `std::slice::Iter`.
```

src/iterators/macros.rs
```diff
@@ -19,7 +19,7 @@ macro_rules! send_sync_read_only {
     };
 }
 
-// read-write iterators use Send => Send rules, same as `std::slice::IterMut`.
+// read-write iterators use Send => Send rules, same as `alloc::slice::IterMut`.
```
bluss, line 22:
```suggestion
// read-write iterators use Send => Send rules, same as `std::slice::IterMut`.
```

bluss:
Thanks! Finally done

PR #874, opened 2020-12-24
Remove deprecated items `.all_close()` and `DataClone`, `ArrayView::into_slice`

### LINKED ISSUE #863: Remove deprecated items
Because next release is a breaking version (0.15), we should remove deprecated items before then.

Note: a few (very few) items are perma-deprecated and should have a comment to that effect.

In a few cases, we might want to undeprecate and keep an API.

Related #698

bluss:
- Remove deprecated `.all_close()` - use approx feature and methods like  `.abs_diff_eq` instead
- Mark `.scalar_sum()` as deprecated - use `.sum()` instead
- Remove deprecated `DataClone` - use `Data + RawDataClone` instead
- Remove deprecated `ArrayView::into_slice` - use `to_slice()` instead.

Fixes #863 

### `ArrayView::into_slice` vs `to_slice`

Looking closely at this one, I'd want to use the name `into_slice`.
The "into" semantic is for preserving the lifetime from the input
to the output - and for setting it more apart from the general `ArrayBase::as_slice`.

However, both names `into-slice` and `to-slice` have pros and cons
and in the interest of reducing churn, the already deprecated method
loses in the tiebreaker IMO, and so `into_slice` is removed because
clippy said so.

Previously in #646

### DIFF
--- a/src/aliases.rs
+++ b/src/aliases.rs
@@ -2,7 +2,6 @@
 //!
 
 use crate::dimension::Dim;
-#[allow(deprecated)]
 use crate::{ArcArray, Array, ArrayView, ArrayViewMut, Ix, IxDynImpl};
 
 /// Create a zero-dimensional index
--- a/src/data_traits.rs
+++ b/src/data_traits.rs
@@ -129,18 +129,6 @@ pub unsafe trait DataMut: Data + RawDataMut {
     }
 }
 
-/// Array representation trait.
-///
-/// An array representation that can be cloned and allows elements to be
-/// accessed with safe code.
-///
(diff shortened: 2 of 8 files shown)


bluss:
The ArrayView::to_slice/into_slice situation: it's not clear cut, it could swing both ways, but to me the `into_slice` name feels truest to what the method does.

bluss:
I changed my mind about into_slice. It's really a toss-up which is best between into_slice and to_slice. I'd prefer into_slice, but it's too churny to reverse the deprecation, so I also prefer to not do that.

Issue #804, opened 2020-04-18
Remove Array::uninitialized()


bluss:
It now has a replacement. See previous issue #685 

If 0.14.0 comes quickly after 0.13.1, probably start with just deprecating the method

bluss:
Possibly put maybe_uninit in its place (rename it)

bluss:
uninitialized is hard to use correctly. I haven't removed it yet because I think I have managed to put it on sound ground - it *is technically possible* to use it correctly. See issue #804  and issue #876 (we have just completed removing all internal usage of this method).

Unfortunately `ArrayBase::uninitialized`:

1. Has some usage by actual users - it's a basically popular and easy to use (github code search confirms)
2. This usage is from casual inspection almost always technically incorrect


The conclusion is that deprecation will have impact on users, but we need to do it, and need to make `maybe_uninit` easy to use.

xd009642:
Now I'm wondering how incorrect my usage in ndarray-vision is :sweat_smile:

bluss:
`uninitialized` was added before stronger rules crystallized. Many had the thought that `let x = uninitialized(); *(&mut x) = 1.0;` was correct when the type was copy - however that as also been judged as incorrect, so there are not many easy ways to use it left - short of using the raw views (partly why they were introduced). Yes, your code breaks this rule (as did ours before the next version).

bluss:
See also #796 (A lot of progress has been made on that issue).

bluss:
Deprecated by #902, so this issue is now fixed. But it will be finally removed in the next release.

Issue #796, opened 2020-04-13
Known issues and concerns w.r.t soundness and best practices


bluss:
I'll try to write down the concerns about ndarray's `unsafe` code that I know about. In many cases this is about aligning better with the developing unsafe code guidelines for Rust, which have been in progress for a long time. It is possible to adopt best practices and fixes, before these guidelines are finished.

***ndarray*** is written carefully and has had many incremental improvements to its internals, but  like much `unsafe` Rust code in the wild, it's unavoidable that our authors's model of unsafe Rust might not correspond to the consensus and best practices of Rust right now. So we try to list concerns and improve piece by piece.

If you know a clear-cut soundness bug with a scope-limited fix, it can be reported directly as a separate bug and does not need to be listed here. Specific concerns in this issue should be researched more deeply before they are fleshed out and fixed in pull requests.

## Uninitialized data 1

Possibly fixed by #798. Examples fixed by #805

***Concern description***

Some examples and some internal code suppose we can use `&mut T` to overwrite in unitialized data, but it is unclear if this is correct.

***Possible remedy***

Use raw views and traversal with raw pointers - now possible with `Zip` to overwrite unitialized elements

## Uninitialized data 2

See issue #685 
Partly mitigated by #802 

***Concern description***

Is the uninitialized data constructor for Arrays the right API? Does it make sense to expose it, and how do we do it in a safe way?

As food for thought, consider that `Vec::with_capacity(128)` is a well-known and safe abstraction for dealing with unitialized data - in this case 128 elements of that - can we use this thinking for our arrays.

## Ownership and aliasing through Raw Pointers

Fixed by #799

***Concern description***

The unique and shared ownership arrays have two access ways to the data - through `self.ptr` and through the owned data's pointer accessor. The Vec assumes unique ownership of its data (like the Box), so having an aliasing extra pointer is possibly problematic.

***Possible remedy***

When creating an Array from a Vec, "take apart" the Vec into data pointer, len, capacity and don't store it as a Vec but in a representation that shows the compiler that we no longer assert unique ownership through the Vec's data pointer.

Details and possible remedy for `Arc<Vec<_>>` are not known.

***Further info***

In principle, this problem is described in issue https://github.com/Kimundi/owning-ref-rs/issues/49

bluss:
[...]

PR #876, opened 2020-12-29
Improve .assume_init() and use Array::maybe_uninit instead of Array::uninitialized

bluss:
Improve `.assume_init()` so that we can see that it is on sound ground (and not transmuting wildly).

Then the remaining internal uses of `Array::uninitialized` are removed and replaced with `Array::maybe_uninit`.

Fixes part of #796
Prerequisite of #804

This pull request demonstrates two different ways of working with `maybe_uninit`. In concatenate and stack,
we can see how we're using the more "Rust-level" approach of using regular array traversal and interacting with
elements of type `MaybeUninit<A>` (by just assigning to them, but still).

In the implementation of `dot` and the benchmark, we instead use raw array views - `RawArrayViewMut`,
cast the raw view from element type `MaybeUninit<A>` to `A`, and handle the resulting raw view and
raw pointers carefully. We have some power in the ndarray abstractions because we can use `Zip`
to traverse raw views just like regular arrays.

### DIFF
--- a/src/data_traits.rs
+++ b/src/data_traits.rs
@@ -526,28 +526,60 @@ unsafe impl<'a, A> DataMut for CowRepr<'a, A> where A: Clone {}
 pub trait RawDataSubst<A>: RawData {
     /// The resulting array storage of the same kind but substituted element type
     type Output: RawData<Elem = A>;
+
+    /// Unsafely translate the data representation from one element
+    /// representation to another.
+    ///
+    /// ## Safety
+    ///
+    /// Caller must ensure the two types have the same representation.
+    unsafe fn data_subst(self) -> Self::Output;
 }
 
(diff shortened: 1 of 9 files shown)


src/data_traits.rs
```diff
@@ -526,28 +526,60 @@ unsafe impl<'a, A> DataMut for CowRepr<'a, A> where A: Clone {}
...
     type Output: RawData<Elem = A>;
+
+    /// Unsafely translate the data representation from one element
+    /// representation to another.
+    ///
+    /// ## Safety
+    ///
+    /// Caller must ensure the two types have the same representation.
+    unsafe fn data_subst(self) -> Self::Output;
 }
```
bluss, line 537: Note that the data storage types are never publically available so the user of ndarray can never reach this method and call it.

PR #877, opened 2020-12-29
Add methods .cell_view() and .into_cell_view()

### LINKED ISSUE #478: and_broadcast_mut for Zip
Currently Zip has no mutable equivalent for the `and_broadcast` method, which is presumably to allow it to safely be `split()` and parallelized. However this makes some calculations difficult. The examples below are for backprop through a broadcast-able element-wise multiplication, but the problem appears as a pattern often.

Currently I'm looking at using this:
```
unsafe{
	let input2_grad = data.get_mut(&self.input2_id.gradient_id())?;

	// do not split/parallelise this Zip!
	Zip::from(&input1)
		.and(&output_grad)
		.and_broadcast(&input2_grad)
		.apply(|input1, out_grad, input2_grad| {
			let input2_grad = input2_grad as *const f32 as *mut f32;
			*input2_grad += input1 * out_grad;
		});
}
```
which works correctly, but also has UB written all over it depending on how Zip is implemented internally (I realize Zip owes me nothing here). 

The safe alternative I've found is:
```
let iter = input1.exact_chunks(input2.shape()).into_iter()
	.zip(output_grad.exact_chunks(input2.shape()));

for (input1_chunk, out_grad_chunk) in iter {
	Zip::from(&mut input2_grad)
		.and(&input1_chunk)
		.and(&out_grad_chunk)
		.apply(|input2_grad, input1, out_grad| {
			*input2_grad += input1 * out_grad;
		});
}
```
which is catastrophically (over 30x) slower for some cases.

So my questions are:
 * is there a safe high performance workaround that I've missed?
 * would it be possible to add a `and_broadcast_mut()` method which would set a flag or something which makes `split()` panic?

bluss:
Add `ArrayBase::cell_view` (similar to methods `.view()` and `.view_mut()`) and the conversion method `ArrayViewMut::into_cell_view()`

We've wanted to add these methods since we added `.cast::<T>()` to the raw views.

The translation of `ArrayViewMut<'a, T, D>` -> `ArrayView<'a, MathCell<T>, D>` works under the same principle
as the method `Cell:from_mut` in the standard library.

These methods should give a lot of flexibility - suddenly the same array view can be traversed multiple times
in the same `Zip` and mutated while being traversed in a different way than before. The price to pay for this
is using the Cell methods to manipulate the elements.

We use a new cell type - `MathCell` - which is just a transparent wrapper around `Cell`. This is so that
we can implement arithmetic ops for cells. However, finding the right implementations requires more design
work so that will be in a later PR.

This fixes #478 - the use case is fixed by using `Zip::from(a).and_broadcast(b.cell_view())` - i.e composing `cell_view`
with the existing broadcasting method.

### DIFF
--- a/src/argument_traits.rs
+++ b/src/argument_traits.rs
@@ -1,6 +1,7 @@
 use std::cell::Cell;
 use std::mem::MaybeUninit;
 
+use crate::math_cell::MathCell;
 
 /// A producer element that can be assigned to once
 pub trait AssignElem<T> {
@@ -17,6 +18,13 @@ impl<'a, T> AssignElem<T> for &'a mut T {
 
 /// Assignable element, simply `self.set(input)`.
 impl<'a, T> AssignElem<T> for &'a Cell<T> {
+    fn assign_elem(self, input: T) {
+        self.set(input);
+    }
+}
+
+/// Assignable element, simply `self.set(input)`.
+impl<'a, T> AssignElem<T> for &'a MathCell<T> {
     fn assign_elem(self, input: T) {
         self.set(input);
     }
--- a/src/impl_methods.rs
+++ b/src/impl_methods.rs
@@ -6,6 +6,7 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
+use std::cell::Cell;
 use std::ptr as std_ptr;
 use std::slice;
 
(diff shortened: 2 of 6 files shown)


bluss:
Oh, now I remember why I didn't add this before. I wanted to wait for adding some kind of "ArithmeticCell<A>" which would be easier to use transparently in arithmetic ops.

bluss:
Any preference between the names `MathCell<T>` and `ArithCell<T>` for the arithmetic ops supporting (transparent) `Cell<T>` wrapper?  I think `MathCell<T>` would be a good name.  Or should it just be `Cell<T>` in the ndarray crate, too confusing?

jturner314:
IMO, reusing the name `Cell` for a wrapper would be too confusing. Either `MathCell` or `ArithCell` would be fine with me. Other options are `OpsCell` and `AutoCell`.

PR #881, opened 2021-01-04
extended SliceInfo by DerefMut

johanneshoetzer:
To play in a more flexible maner with SliceInfo a added the DerefMut trait, so i can change the SliceInfo afterwards.


bluss:
If it was mutable, then it breaks the type checking (dimensions) that is done other places in the code, so this change can not be permitted

bluss:
I'll close this, it can be reopened when back with a different kind of solution.

Issue #842, opened 2020-09-19
Construction from negative strides?


kngwyu:
Hi, thank you for this great library.

I encountered a problem when creating an `ArrayView` with a pointer and negative strides (see https://github.com/PyO3/rust-numpy/issues/151 for the full context).
We can do this by converting `&[isize]` to `&[usize]`, but this causes panic by [a debug assertion](https://github.com/rust-ndarray/ndarray/blob/f69248e977f42b7afb01f5c280baba1da09738c9/src/impl_raw_views.rs#L75) when non-release build.
So,
- How about adding some unsafe API to do this? It would be like a public `new_`. I'm happy to create a PR if you're willing to.
- **Why strides are internally usize?** It can be negative by slicing and I don't think it's reasonable to use `usize`.

jturner314:
I started working on support for negative strides in the various constructors a while ago ([branch with work-in-progress changes](https://github.com/rust-ndarray/ndarray/compare/master...jturner314:negative-stride-constructors?expand=1)) but didn't get a chance to finish. Most of the necessary changes are in the internal functions which check the validity of the inputs. For the public API, the primary change will be modifying [`ShapeBuilder`](https://docs.rs/ndarray/0.13.1/ndarray/trait.ShapeBuilder.html) to allow `isize` strides.

You could finish up the necessary changes to `ndarray` to support negative strides in the constructors (or, I suppose, the subset of changes needed for just the unsafe view constructors). Or, as a temporary workaround, you could use a function which does the following:

1. For each axis with a negative stride, offset the pointer by `(axis_len - 1) * stride`.
2. Construct a view using the absolute values of the strides.
3. For each axis that had a negative stride, call [`.invert_axis()`](https://docs.rs/ndarray/0.13.1/ndarray/struct.ArrayBase.html#method.invert_axis).

> Why are strides internally usize?

They are represented as `usize` in the `ArrayBase` struct, but everywhere they're used, they're cast to `isize`. Representing them as `usize` is a hack to have a relatively simple design while dealing with limitations of the type system. With some of the upcoming Rust features (generic associated types in particular), it'll be possible to resolve this in a much cleaner way.

kngwyu:
Thank you for the detailed explanation!
I'll use a temporary workaround for now, but am willing to tackle the `ShapeBuilder` change if I could make time.

> Representing them as usize is a hack to have a relatively simple design

[...]

PR #885, opened 2021-01-06
Fix memory continuity judgment when stride is negative

SparrowLii:
It comes from FIXME in is_cotiguous. Whether the stride is negative will not affect the arrangement of the array in the memory, therefore, sorting and continuity detection can be performed according to the absolute value of the step size. That's what this pr dose in dimension_trait.rs.

Then we need to deal with the as_slice_memory_order function in impl_methods.rs. When stride is negative, although we can know that the array is continuous, because the ptr of the array does not point to the head of the memory data at this time, the ptr value cannot be used directly.

This is why I added the head_ptr_offset function in dimension/mod.rs. It simply calculates the offset of the head address of the array data relative to the current ptr based on the dim and stride of the array. The offset is always less than or equal to 0. After calculating the offset, as_slice_memory_order can get continuous data. as_slice_memory_order_mut is the same.

The last thing to be solved is the to_owned function in impl_method.rs. After calling as_slice_memory_order, it will call from_shape_vec_unchecked function( in impl_constructors.rs ) to generate a new ArrayBase based on the obtained array head pointer. This is incorrect when stride is negative, because the head address of the data array is not the ptr value that ArrayBase should have. Fortunately, we have the head_ptr_offset method, which can simply calculate the offset based on dim and strides, and then get the ptr we need.

### DIFF
--- a/src/dimension/dimension_trait.rs
+++ b/src/dimension/dimension_trait.rs
@@ -314,7 +314,8 @@ pub trait Dimension:
             *elt = i;
         }
         let strides = self.slice();
-        indices.slice_mut()
+        indices
+            .slice_mut()
             .sort_by_key(|&i| (strides[i] as isize).abs());
         indices
     }
--- a/src/dimension/dimension_trait.rs
+++ b/src/dimension/dimension_trait.rs
@@ -286,17 +286,16 @@ pub trait Dimension:
             return true;
         }
         if dim.ndim() == 1 {
-            return false;
+            return strides[0] as isize == -1;
         }
         let order = strides._fastest_varying_stride_order();
         let strides = strides.slice();
 
-        // FIXME: Negative strides
         let dim_slice = dim.slice();
         let mut cstride = 1;
         for &i in order.slice() {
             // a dimension of length 1 can have unequal strides
-            if dim_slice[i] != 1 && strides[i] != cstride {
+            if dim_slice[i] != 1 && (strides[i] as isize).abs() as usize != cstride {
                 return false;
             }
             cstride *= dim_slice[i];
@@ -307,8 +306,7 @@ pub trait Dimension:
     /// Return the axis ordering corresponding to the fastest variation
     /// (in ascending order).
     ///
-    /// Assumes that no stride value appears twice. This cannot yield the correct
-    /// result the strides are not positive.
+    /// Assumes that no stride value appears twice.
     #[doc(hidden)]
     fn _fastest_varying_stride_order(&self) -> Self {
         let mut indices = self.clone();
@@ -316,7 +314,8 @@ pub trait Dimension:
             *elt = i;
         }
         let strides = self.slice();
-        indices.slice_mut().sort_by_key(|&i| strides[i]);
+        indices.slice_mut()
+            .sort_by_key(|&i| (strides[i] as isize).abs());
         indices
     }
 
@@ -645,7 +644,7 @@ impl Dimension for Dim<[Ix; 2]> {
 
     #[inline]
     fn _fastest_varying_stride_order(&self) -> Self {
-        if get!(self, 0) as Ixs <= get!(self, 1) as Ixs {
+        if (get!(self, 0) as Ixs).abs() <= (get!(self, 1) as Ixs).abs() {
             Ix2(0, 1)
         } else {
             Ix2(1, 0)
@@ -805,7 +804,7 @@ impl Dimension for Dim<[Ix; 3]> {
         let mut order = Ix3(0, 1, 2);
         macro_rules! swap {
             ($stride:expr, $order:expr, $x:expr, $y:expr) => {
-                if $stride[$x] > $stride[$y] {
+                if ($stride[$x] as isize).abs() > ($stride[$y] as isize).abs() {
                     $stride.swap($x, $y);
                     $order.ixm().swap($x, $y);
                 }
--- a/src/dimension/mod.rs
+++ b/src/dimension/mod.rs
@@ -280,6 +280,19 @@ pub fn stride_offset_checked(dim: &[Ix], strides: &[Ix], index: &[Ix]) -> Option
     Some(offset)
 }
 
+/// Checks if strides are non-negative.
+pub fn strides_non_negative<D>(strides: &D) -> Result<(), ShapeError>
+where
+    D: Dimension,
+{
+    for &stride in strides.slice() {
+        if (stride as isize) < 0 {
+            return Err(from_kind(ErrorKind::Unsupported));
+        }
+    }
+    Ok(())
+}
+
 /// Implementation-specific extensions to `Dimension`
 pub trait DimensionExt {
     // note: many extensions go in the main trait if they need to be special-
--- a/src/dimension/mod.rs
+++ b/src/dimension/mod.rs
@@ -207,10 +207,7 @@ where
 ///
 /// 2. The product of non-zero axis lengths must not exceed `isize::MAX`.
 ///
-/// 3. For axes with length > 1, the pointer cannot move outside the
-///    slice. For axes with length ≤ 1, the stride can be anything.
-///
-/// 4. If the array will be empty (any axes are zero-length), the difference
+/// 3. If the array will be empty (any axes are zero-length), the difference
 ///    between the least address and greatest address accessible by moving
 ///    along all axes must be ≤ `data.len()`. (It's fine in this case to move
 ///    one byte past the end of the slice since the pointers will be offset but
@@ -221,13 +218,19 @@ where
 ///    `data.len()`. This and #3 ensure that all dereferenceable pointers point
 ///    to elements within the slice.
 ///
-/// 5. The strides must not allow any element to be referenced by two different
+/// 4. The strides must not allow any element to be referenced by two different
 ///    indices.
 ///
 /// Note that since slices cannot contain more than `isize::MAX` bytes,
 /// condition 4 is sufficient to guarantee that the absolute difference in
 /// units of `A` and in units of bytes between the least address and greatest
 /// address accessible by moving along all axes does not exceed `isize::MAX`.
+///
+/// Warning: This function is sufficient to check the invariants of ArrayBase only
+/// if the pointer to the first element of the array is chosen such that the element
+/// with the smallest memory address is at the start of data. (In other words, the
+/// pointer to the first element of the array must be computed using offset_from_ptr_to_memory
+/// so that negative strides are correctly handled.)
 pub(crate) fn can_index_slice<A, D: Dimension>(
     data: &[A],
     dim: &D,
@@ -244,7 +247,7 @@ fn can_index_slice_impl<D: Dimension>(
     dim: &D,
     strides: &D,
 ) -> Result<(), ShapeError> {
-    // Check condition 4.
+    // Check condition 3.
     let is_empty = dim.slice().iter().any(|&d| d == 0);
     if is_empty && max_offset > data_len {
         return Err(from_kind(ErrorKind::OutOfBounds));
@@ -253,7 +256,7 @@ fn can_index_slice_impl<D: Dimension>(
         return Err(from_kind(ErrorKind::OutOfBounds));
     }
 
-    // Check condition 5.
+    // Check condition 4.
     if !is_empty && dim_stride_overlap(dim, strides) {
         return Err(from_kind(ErrorKind::Unsupported));
     }
@@ -384,8 +387,8 @@ fn to_abs_slice(axis_len: usize, slice: Slice) -> (usize, usize, isize) {
 
 /// This function computes the offset from the logically first element to the first element in
 /// memory of the array. The result is always <= 0.
-pub fn offset_from_ptr_to_memory(dim: &[Ix], strides: &[Ix]) -> isize {
-    let offset = izip!(dim, strides).fold(0, |_offset, (d, s)| {
+pub fn offset_from_ptr_to_memory<D: Dimension>(dim: &D, strides: &D) -> isize {
+    let offset = izip!(dim.slice(), strides.slice()).fold(0, |_offset, (d, s)| {
         if (*s as isize) < 0 {
             _offset + *s as isize * (*d as isize - 1)
         } else {
--- a/src/dimension/mod.rs
+++ b/src/dimension/mod.rs
@@ -46,15 +46,12 @@ pub fn stride_offset(n: Ix, stride: Ix) -> isize {
 /// There is overlap if, when iterating through the dimensions in order of
 /// increasing stride, the current stride is less than or equal to the maximum
 /// possible offset along the preceding axes. (Axes of length ≤1 are ignored.)
-///
-/// The current implementation assumes that strides of axes with length > 1 are
-/// nonnegative. Additionally, it does not check for overflow.
 pub fn dim_stride_overlap<D: Dimension>(dim: &D, strides: &D) -> bool {
     let order = strides._fastest_varying_stride_order();
     let mut sum_prev_offsets = 0;
     for &index in order.slice() {
         let d = dim[index];
-        let s = strides[index] as isize;
+        let s = (strides[index] as isize).abs();
         match d {
             0 => return false,
             1 => {}
@@ -210,8 +207,7 @@ where
 ///
 /// 2. The product of non-zero axis lengths must not exceed `isize::MAX`.
 ///
-/// 3. For axes with length > 1, the stride must be nonnegative. This is
-///    necessary to make sure the pointer cannot move backwards outside the
+/// 3. For axes with length > 1, the pointer cannot move outside the
 ///    slice. For axes with length ≤ 1, the stride can be anything.
 ///
 /// 4. If the array will be empty (any axes are zero-length), the difference
@@ -255,14 +251,6 @@ fn can_index_slice_impl<D: Dimension>(
     }
     if !is_empty && max_offset >= data_len {
         return Err(from_kind(ErrorKind::OutOfBounds));
-    }
-
-    // Check condition 3.
-    for (&d, &s) in izip!(dim.slice(), strides.slice()) {
-        let s = s as isize;
-        if d > 1 && s < 0 {
-            return Err(from_kind(ErrorKind::Unsupported));
-        }
     }
 
     // Check condition 5.
@@ -392,6 +380,19 @@ fn to_abs_slice(axis_len: usize, slice: Slice) -> (usize, usize, isize) {
     );
     ndassert!(step != 0, "Slice stride must not be zero");
     (start, end, step)
+}
+
+/// This function computes the offset from the logically first element to the first element in
+/// memory of the array. The result is always <= 0.
+pub fn offset_from_ptr_to_memory(dim: &[Ix], strides: &[Ix]) -> isize {
+    let offset = izip!(dim, strides).fold(0, |_offset, (d, s)| {
+        if (*s as isize) < 0 {
+            _offset + *s as isize * (*d as isize - 1)
+        } else {
+            _offset
+        }
+    });
+    offset
 }
 
 /// Modify dimension, stride and return data pointer offset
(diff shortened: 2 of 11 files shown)


src/impl_methods.rs
```diff
@@ -1281,8 +1281,7 @@ where
 
     /// Return true if the array is known to be contiguous.
     ///
-    /// Will detect c- and f-contig arrays correctly, but otherwise
-    /// There are some false negatives.
+    /// Will detect c- and f-contig arrays correctly.
```
bluss, line 1284: Just checking, isn't it supposed to detect all contiguous arrays now correctly? Could there be some test cases for that?

I mean like this:

Start with C- or F- contig array.

Swap any axes. Invert any axes.

The result should always be contiguous, right? If there are cases we still don't detect, it would be great to clarify what that is (and keep the "disclaimer" in the doc).

src/dimension/mod.rs
```diff
@@ -394,6 +382,19 @@ fn to_abs_slice(axis_len: usize, slice: Slice) -> (usize, usize, isize) {
     (start, end, step)
 }
 
+/// The offset value of the data pointer relative to the actual pointer of the given
+/// dim and strides. The result is always <= 0.
+pub fn head_ptr_offset(dim: &[Ix], strides: &[Ix]) -> isize{
+    let offset=izip!(dim, strides)
+        .fold(0,|_offset, (d, s)|{
```
bluss, line 389: Missing spaces
```suggestion
        .fold(0, |_offset, (d, s)| {
```

src/dimension/mod.rs
```diff
@@ -394,6 +382,19 @@ fn to_abs_slice(axis_len: usize, slice: Slice) -> (usize, usize, isize) {
...
+            if (*s as isize) < 0{
```
bluss, line 390:
```suggestion
            if (*s as isize) < 0 {
```

src/dimension/mod.rs
```diff
@@ -394,6 +382,19 @@ fn to_abs_slice(axis_len: usize, slice: Slice) -> (usize, usize, isize) {
...
+                _offset + *s as isize * (*d as isize -1)
+            }else{
```
bluss, line 392:
```suggestion
            } else {
```

I won't mention all missing spaces in the PR I think :)
 
Still, please *don't* reformat changes outside the diffs that are relevant to this PR.

src/dimension/mod.rs
```diff
@@ -394,6 +382,19 @@ fn to_abs_slice(axis_len: usize, slice: Slice) -> (usize, usize, isize) {
     (start, end, step)
 }
 
+/// The offset value of the data pointer relative to the actual pointer of the given
+/// dim and strides. The result is always <= 0.
+pub fn head_ptr_offset(dim: &[Ix], strides: &[Ix]) -> isize{
+    let offset=izip!(dim, strides)
```
bluss, line 388:
```suggestion
    let offset = izip!(dim, strides)
```

src/impl_constructors.rs
```diff
@@ -457,7 +460,7 @@ where
         // debug check for issues that indicates wrong use of this constructor
         debug_assert!(dimension::can_index_slice(&v, &dim, &strides).is_ok());
         ArrayBase {
-            ptr: nonnull_from_vec_data(&mut v),
+            ptr: nonnull_from_vec_data(&mut v).offset(head_ptr_offset(dim.slice(),strides.slice()).abs()),
```
bluss, line 463: Why is there an abs here?

src/dimension/mod.rs
```diff
@@ -394,6 +382,19 @@ fn to_abs_slice(axis_len: usize, slice: Slice) -> (usize, usize, isize) {
     (start, end, step)
 }
 
+/// The offset value of the data pointer relative to the actual pointer of the given
+/// dim and strides. The result is always <= 0.
+pub fn head_ptr_offset(dim: &[Ix], strides: &[Ix]) -> isize{
```
bluss, line 387: Let's formulate this using a wording of "from.. to", then the sign of the pointer difference should be clear. For example, "this computes the offset **from** the first element in memory **to** the logically first element of the array".

I'd like the function name to be based on the same thinking.

Should this function not use a type parameter `D` so that it is special cased automatically for the most important dimensions (1 and 2)?

And missing space:

```suggestion
pub fn head_ptr_offset(dim: &[Ix], strides: &[Ix]) -> isize {
```

bluss (changes requested):
Thanks a lot for working on this, it would be awesome to get this fixed!

src/dimension/dimension_trait.rs
```diff
@@ -291,12 +291,11 @@ pub trait Dimension:
         let order = strides._fastest_varying_stride_order();
```
bluss, line 291: Reading this function, a few lines above - it seems there is still the case of an 1D array with stride -1 and it will still say it's not contig :(

SparrowLii:
@bluss  This line is a bit strange( tests/oper.rs:775 )
It runs &lhs.to_owned().into_shape(), but  to_owned() does not guarantee that the memory arrangement of the return value is standard C or fortran( which into_shape() needs), regardless of whether the stride is negative.
This is the reason why my test failed, I will try to modify it, but I hope you can provide some suggestions.

src/impl_methods.rs
SparrowLii: I did the following test, no matter the layout is C, Fortran or custom, it got the correct result. I tried to put these tests into pr, but because is_contiguous is not a public API (it is pub(crate) ), it is not possible to add tests directly.
    use ndarray::ArrayView;
    use ndarray::Axis;
    use ndarray::arr3;
    use ndarray::arr2;
    use ndarray::ShapeBuilder;
    use ndarray::s;
    fn main() {
        test3()
    }
    fn test2(){
        let mut a = ArrayView::from_shape((2, 3, 2).strides((1, 4, 2)),
                                      &s).unwrap();
        assert_eq!(a, arr3(&[[[0, 2], [4, 6], [8, 10]], [[1, 3], [5, 7], [9, 11]]]));
        assert!(a.is_contiguous());

        let mut b=a.slice(s![..;1, ..;-1, ..;-1]);
        assert_eq!(b, arr3(&[[[10, 8], [6, 4], [2, 0]], [[11, 9], [7, 5], [3, 1]]]));
        assert!(b.is_contiguous());

        b.swap_axes(1,2);
        assert_eq!(b, arr3(&[[[10, 6, 2], [8, 4, 0]], [[11, 7, 3], [9, 5, 1]]]));
        assert!(b.is_contiguous());

        b.invert_axis(Axis(0));
        assert_eq!(b, arr3(&[[[11, 7, 3], [9, 5, 1]], [[10, 6, 2], [8, 4, 0]]]));
        assert!(b.is_contiguous());

        let mut c=b.reversed_axes();
        assert_eq!(c, arr3(&[[[11, 10],[9, 8]],[[7, 6],[5, 4]],[[3, 2],[1, 0]]]));
        assert!(c.is_contiguous());

        c.merge_axes(Axis(1),Axis(2));
        assert_eq!(c, arr3(&[[[11, 10, 9, 8]], [[7, 6, 5, 4]], [[3, 2, 1, 0]]]));
        assert!(c.is_contiguous());

        let d=b.remove_axis(Axis(1));
        assert_eq!(d, arr2(&[[11, 7, 3], [10, 6, 2]]));
        assert!(!d.is_contiguous());

        let e=b.remove_axis(Axis(2));
        assert_eq!(e, arr2(&[[11, 9],[10, 8]]));
        assert!(e.is_contiguous());

        let f=e.insert_axis(Axis(2));
        assert_eq!(f, arr3(&[[[11], [9]], [[10], [8]]]));
        assert!(f.is_contiguous());

        let mut g=b.clone();
        g.collapse_axis(Axis(1),0);
        assert_eq!(g, arr3(&[[[11, 7, 3]],[[10, 6, 2]]]));
        assert!(!g.is_contiguous());

        b.collapse_axis(Axis(2),0);
        assert_eq!(b, arr3(&[[[11], [9]], [[10], [8]]]));
        assert!(b.is_contiguous());

        let h = ArrayView::from_shape((1, 3, 2).strides((2, 4, 2)),
                                      &s).unwrap();
        assert_eq!(h, arr3(&[[[0, 2], [4, 6], [8, 10]]]));
        assert!(!h.is_contiguous());
    }

src/dimension/mod.rs
SparrowLii: Sorry for that. I will pay attention to it.

src/impl_constructors.rs
SparrowLii: When calculating the offset, the result is always less than or equal to zero (this can avoid using abs() when calculating the offset). Therefore, we need to use abs() to obtain a positive offset when getting the logical head pointer according to the actual head pointer in the memory.

src/dimension/mod.rs
SparrowLii: I'm sorry I lacked some skills. Why change the parameter can provide a specific implementation for a 1d or 2d array? I hope there are examples for me to refer to.

src/dimension/dimension_trait.rs
SparrowLii: That's right. I checked the 1d array in dimension_trait.rs:289 and it now gets the correct result.

src/impl_constructors.rs
```diff
@@ -457,7 +460,8 @@ where
         // debug check for issues that indicates wrong use of this constructor
         debug_assert!(dimension::can_index_slice(&v, &dim, &strides).is_ok());
         ArrayBase {
-            ptr: nonnull_from_vec_data(&mut v),
+            ptr: nonnull_from_vec_data(&mut v)
+                .offset(offset_from_ptr_to_memory(dim.slice(), strides.slice()).abs()),
```
bluss, line 464: Style wise I would use a minus here instead of `.abs()`. Abs would signal that this can return both -3 and 3 and we're fine with either.

tests/oper.rs
```diff
@@ -707,7 +707,7 @@ fn gen_mat_vec_mul() {
         S2: Data<Elem = A>,
     {
         let ((m, _), k) = (lhs.dim(), rhs.dim());
-        reference_mat_mul(lhs, &rhs.to_owned().into_shape((k, 1)).unwrap())
+        reference_mat_mul(lhs, &rhs.as_standard_layout().into_shape((k, 1)).unwrap())
```
bluss, line 710: Thanks, this change makes sense - it is a sign of `.to_owned()` improving and becoming more capable which is great. But we'll have to make a compatibility note or something.

bluss (commented):
Looks good. With a few minor changes - and tests as mentioned in other comments - I think this could be ready for merge.

@jturner314 Any comments on the approach here?

bluss (approved):

jturner314 (changes requested):
I'm glad this will finally be addressed.

This PR makes it possible to directly construct arrays with negative strides using `from_shape_vec`, but it doesn't address the awkwardness of doing this using the existing API (since the existing API requires casting `isize` to `usize` in a confusing way), but that's okay. We could document how to construct an array with negative strides using `from_shape_vec`, but due to the awkwardness, IMO it's fine to leave it undocumented until we have a better API for this.

I haven't reviewed the PR in detail, but the general approach looks fine. I would like to see a few tests for the constructors with negative strides.

I commented on changes that need to be made to the `can_index_slice` docs (see below). Personally, I'd prefer for `can_index_slice` to take as a parameter either the offset to the first element or the pointer to the first element, instead of just adding a warning to the docs of how the pointer needs to be computed. (I think adding it as an explicit parameter would make accidentally misusing the function less likely.) That said, the current implementation is fine as long as the warning is added.

src/dimension/dimension_trait.rs
SparrowLii: I just found a good way. Refer to the practice in test_contiguous, use as_slice_memory_order().is_some()

SparrowLii:
@jturner314 I added the corresponding warning. I think it is a matter to construct an array based on the negative step size, and there are related [issues](https://github.com/rust-ndarray/ndarray/issues/842) in ndarray. I think I can fix them in the next pr. WDYT?

tests/oper.rs
SparrowLii: I added relevant instructions in the comment of to_owned. Maybe we should clarify the role of the return value of to_owned and improve it?

src/impl_constructors.rs
SparrowLii: That makas sense. We should replace abs with a minus sign.

src/dimension/mod.rs
```diff
@@ -394,6 +382,19 @@ fn to_abs_slice(axis_len: usize, slice: Slice) -> (usize, usize, isize) {
     (start, end, step)
 }
 
+/// This function computes the offset from the logically first element to the first element in
+/// memory of the array. The result is always <= 0.
+pub fn offset_from_ptr_to_memory(dim: &[Ix], strides: &[Ix]) -> isize {
```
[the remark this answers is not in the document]
SparrowLii: Thanks for the explanation. I have made the corresponding changes.

src/dimension/mod.rs
```diff
@@ -210,8 +207,7 @@ where
 ///
 /// 2. The product of non-zero axis lengths must not exceed `isize::MAX`.
 ///
-/// 3. For axes with length > 1, the stride must be nonnegative. This is
-///    necessary to make sure the pointer cannot move backwards outside the
+/// 3. For axes with length > 1, the pointer cannot move outside the
```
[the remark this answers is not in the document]
SparrowLii: I added the corresponding test in tests/array.rs, I think it should work well now.

SparrowLii:
I added the corresponding test in tests/array.rs, I think it should work well now.

tests/oper.rs
bluss: to_owned's docs are fine - it's `into_shape` that's a bit tricky to use, and `to_owned` doesn't need to account for that 🙂

src/impl_methods.rs
```diff
@@ -154,12 +154,12 @@ where
 
     /// Return an uniquely owned copy of the array.
     ///
-    /// If the input array is contiguous and its strides are positive, then the
-    /// output array will have the same memory layout. Otherwise, the layout of
-    /// the output array is unspecified. If you need a particular layout, you
-    /// can allocate a new array with the desired memory layout and
-    /// [`.assign()`](#method.assign) the data. Alternatively, you can collect
-    /// an iterator, like this for a result in standard layout:
+    /// If the input array is contiguous, then theoutput array will have the same
```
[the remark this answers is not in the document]
jturner314: Also, I don't want to guarantee that: "Otherwise, the layout of the output array is standard C." I'd much rather leave the sentence as-is: "Otherwise, the layout of the output array is unspecified." In the future, we may change `.to_owned()` to more intelligently choose the layout of the output array for the non-contiguous-input cases.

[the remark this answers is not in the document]
SparrowLii: Ok, got it

src/impl_methods.rs
```diff
@@ -182,6 +182,10 @@ where
     /// # assert!(owned.t().is_standard_layout());
     /// # assert_eq!(arr, owned);
     /// ```
+    ///
+    /// Warning: Using to_owned().into_shape() may cause shape mismatch errors.
+    /// Because to_owned() doesnot guarantee that the memory arrangement of
+    /// the return value isstandard C orfortran( which into_shape() needs ).
```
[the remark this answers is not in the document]
SparrowLii: Got it.

src/impl_methods.rs
```diff
@@ -154,8 +154,8 @@ where
 
     /// Return an uniquely owned copy of the array.
     ///
-    /// If the input array is contiguous, then theoutput array will have the same
-    /// memory layout. Otherwise, the layout ofthe output array is standard C.
+    /// If the input array is contiguous, then the output array will have the same
+    /// memory layout. Otherwise, the layout ofthe output array is unspecified.
```
bluss, line 158:
```suggestion
    /// memory layout. Otherwise, the layout of the output array is unspecified.
```
And so on, there's a missing space on every line

SparrowLii: So sorry for that missing

bluss: No problem. It's a good PR if both the maintainer and contributor have time and patience for reviewing in several cycles. Even though it of course feels awkward to request more changes etc. For an open source project I think on balance it makes sense to require high quality - contributors are in general not known if they will stay around and tie up loose ends and we avoid merging half-baked features. And merging PRs with too many loose ends that the maintainer has to fix - that burns out the maintainer. That's the sum of my experience.

bluss:
This ready to merge IMO.

You have fixed `is_contiguous` and the many methods that rely on it, and that's great, I think that's why the maintainers are excited about this PR.

We have left the interface unspecified in the current state of the PR - w.r.t creating array views and raw views with negative strides. For the user, it's not clear if it's allowed and how to specify the pointer - should it be the first-in-memory or the first-in-array pointer that they should pass?

For this reason, I think all the public from-raw-pointer constructors for array views and raw views must have a debug assertion that denies negative strides, and a method doc that states that strides cannot be negative - the reason is that this must be a closed possibility until the interface is decided. This change should not affect our own functionality - we internally use private constructors for array views (that always use the first-in-the-array pointer).

I can do that change that closes down the negative strides in those constructors - if you agree.

jturner314:
> For this reason, I think all the public from-raw-pointer constructors for array views and raw views must have a debug assertion that denies negative strides, and a method doc that states that strides cannot be negative - the reason is that this must be a closed possibility until the interface is decided.

I agree.

We could also do this for the public owned constructors such as `from_shape_vec_unchecked`, if there's a possibility that we'll add a way for the user to specify the offset to the first element instead of just assuming that the lowest-memory-address element should be at the start of the `Vec`. (We'd need make sure to change `to_owned`, `map`, and `map_mut` to use `from_vec_dim_stride_unchecked` instead of `from_shape_vec_unchecked` so that they'd still work for the "contiguous with negative strides" case.)

SparrowLii:
@bluss I very much agree. Many APIs currently using ptr values do not specify the meaning of ptr. ( from_raw_parts, etc.) A single pointer and the number of elements can only represent a contiguous memory, but cannot represent a data structure that can be directly used by an array. I think that the names and documents of these APIs (and perhaps implementations) should be modified to clarify the definition and use of parameters, like @jturner314 pointed out. But I think it may not be easy to find out all APIs that should be modified. 

I think this pr can be merged, and thank you very much for your reviewing and help.

src/impl_methods.rs
SparrowLii: Thank you very much, it means a lot to me.

bluss:
@jturner314 That's an interesting and complicating point. I didn't think of that - with the constructors that start from a Vec, it is already clear where the start of the memory is, I was thinking, so it's not strictly necessary to allow the user to specify a different offset. And `to_owned` and `map` already handle that correctly with this PR if I read it correctly.

jturner314:
> with the constructors that start from a Vec, it is already clear where the start of the memory is, I was thinking, so it's not strictly necessary to allow the user to specify a different offset.

I agree that the most obvious use-case is when the lowest-address element corresponds to the start of the owned data. However, once we allow negative strides in the constructors, it'll be possible for users to directly construct everything `ArrayBase` can represent except for the case where the lowest-address element is not at the start of the owned data. While I don't immediately see an obvious use-case for directly constructing an owned array like this, it would be nice to support it in the constructors because the alternative ways to construct an array like this would be painful. (You'd have to create an `Array1` and then perform a complicated series of slicing/reshaping operations. Alternatively, you could drop the elements from the start of the `Vec` that you don't want so that the lowest-address element would be at the start of the `Vec`, but AFAIK, there's not a cheap way to do that.)

> `to_owned` and `map` already handle every applicable case right with this PR if I read it correctly.

That's correct. We'd only need to change them if we added debug assertions for negative strides to `from_shape_vec_unchecked` (since they use `from_shape_vec_unchecked` with strides that can be negative).

bluss:
Without a use case, we can decline to implement it 🙂

jturner314:
That would be fine with me; I'd just like to keep this possible functionality in mind while updating the constructor API for negative strides, so that we don't accidentally make it too difficult to add in the future (in case this functionality is desirable for someone). If we do add this functionality in the future, I'd make it optional, similar to how `.strides()` is currently handled; the default would be to place the lowest-address element at the start of the owned data.

jturner314:
I just thought of one example where it could be useful -- if you wanted an alignment greater than the alignment of the element type, e.g. for SIMD. (A common way to do this is to allocate a larger `Vec` than needed and ignore the first few elements.)

SparrowLii:
I think we should update the description of these constructors, but there is no need to add an offset parameter. As long as the meaning of the ptr parameter in the API is clear (is it a memory head pointer or an array head pointer), we can create it correctly.
Even if we add functions constructed array with negative strides in the future.

For the support of smid, I think we should implement a new data structure for it. Then we can add the offset parameter to it.

bluss:
@jturner314 That sounds intriguing, since it can be supported without any extra flags or carrying extra alignment info inside `Array<A, D>`. But it can have its own constructor I think

jturner314:
> @jturner314 That sounds intriguing, since it can be supported without any extra flags or carrying extra alignment info inside `Array<A, D>`. But it can have its own constructor I think

I agree, it's not a strong justification for an offset parameter.

One other reason it could be useful for owned constructors to support an offset parameter is for testing. (Ideally, it would be easy to write tests where the lowest-address element doesn't match the start of the owned data for arrays with arbitrary strides. While most user code shouldn't care about how the array maps to the underlying storage, it may be relevant for code which converts back-and-forth from `Vec` using `into_raw_vec`, or code involving `unsafe`.)

bluss:
Rebased to squash together to more concise history

bluss:
Only the raw pointer constructors deny constructing with negative strides. The vec ones simply leave it unspecified, and it's not intended to be supported - pending design.

PR #893, opened 2021-01-13
Fix to_shared to avoid cloning when it is already ArcArray

SparrowLii:
From FIXME in to_shared. The current implementation simply calls to_owned().into_shared(), so cloning cannot be avoided when the array is already ArcArray. 
So use to_owned().shared() in a method of Data trait (just like into_owned()) and add a default implementation to it, which will not affect the existing structure. Then modify its implementation in OwnedArcRepr, which implements the Data trait, to avoid cloning when the array is already ArcArray.

### DIFF
--- a/src/data_traits.rs
+++ b/src/data_traits.rs
@@ -96,6 +96,15 @@ pub unsafe trait Data: RawData {
     where
         Self::Elem: Clone,
         D: Dimension;
+
+    fn to_share<D>(self_: &ArrayBase<Self, D>) -> ArrayBase<OwnedArcRepr<Self::Elem>, D>
+    where
+        Self::Elem: Clone,
+        D: Dimension,
+    {
+        // clone to shared
+        self_.to_owned().into_shared()
+    }
 }
 
 /// Array representation trait.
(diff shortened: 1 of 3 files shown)


src/data_traits.rs
```diff
@@ -96,6 +96,15 @@ pub unsafe trait Data: RawData {
     where
         Self::Elem: Clone,
         D: Dimension;
+
+    fn to_share<D>(self_: &ArrayBase<Self, D>) -> ArrayBase<OwnedArcRepr<Self::Elem>, D>
```
bluss, line 100: Very nice, let's name this method `to_shared` IMO

src/data_traits.rs
```diff
@@ -238,6 +247,20 @@ unsafe impl<A> Data for OwnedArcRepr<A> {
...
         }
     }
+
+    fn to_share<D>(self_: &ArrayBase<Self, D>) -> ArrayBase<OwnedArcRepr<Self::Elem>, D>
+    where
+        Self::Elem: Clone,
+        D: Dimension,
+    {
+        // to shared using clone of OwnedArcRepr without clone of raw data.
+        ArrayBase {
```
bluss, line 257: Can't this method body just be `self_.clone()`. We should avoid writing out the ArrayBase constructor just to keep maintainability.

src/data_traits.rs
bluss, line 100:
```suggestion
    #[doc(hidden)]
    fn to_share<D>(self_: &ArrayBase<Self, D>) -> ArrayBase<OwnedArcRepr<Self::Elem>, D>
```

Like the rest of the methods. And it should have a doc comment here still - as documentation.

SparrowLii: Okay, they should be consistent

src/data_traits.rs
SparrowLii: That's right. Thanks for pointing that out.

src/data_traits.rs
SparrowLii: Yes, the corresponding modifications have been added.

bluss:
I prefer merging with a merge commit, makes it clear what every unit of contribution has been. This PR contains a commit called "back consist" which seems messy. I'll prefer to squash merge so that that one is not visible.

bluss:
Thanks for this fix!

SparrowLii:
> This PR contains a commit called "back consist" which seems messy. I'll prefer to squash merge so that that one is not visible.
That is my mistake. I made some unnecessary changes on my master branch, so I used reset to eliminate it. Squash merge is all right.

Issue #858, opened 2020-12-08
Rename Zip::apply to Zip::for_each?


bluss:
Rename Zip::apply to Zip::for_each (and par_apply to par_for_each).

This seems like a suggestion that can cause some churn, (unnecessarily?), but I think the code is easier to read and write if we use the more established name `.for_each()`.

For example, new code after this change:

```rust
Zip::from(result.index_axis_mut(axis, perm_i))
    .and(self.index_axis(axis, i))
    .for_each(|to, from| {
        copy_nonoverlapping(from, to.as_mut_ptr(), 1);
        moved_elements += 1;
    });
```

The old name would be deprecated and removed a release later. At the moment, there is no way for us to integrate with "cargo fix" (only rustc_deprecated attribute can do that), so we can't have this change be applied automatically.

PR #894, opened 2021-01-15
Rename Zip::apply to for_each including other apply methods

bluss:
As mentioned in #858. Use the name `Zip::for_each` which is more idiomatic than `apply`.

Renames the following and deprecates the old name:

- apply to for_each
- apply_collect to map_collect
- apply_collect_into to map_collect_into
- par_apply to par_for_each
- par_apply_collect to par_map_collect
- par_apply_collect_into to par_map_collect_into


SparrowLii:
@bluss The method in benches does not seem to be changed. Is there any hindrance?

bluss:
I just forgot to update those - and then I noticed and forgot again, so it just needs to be fixed to use the new name.

PR #898, opened 2021-01-20
Implement co-broadcasting in operator overloading

SparrowLii:
This PR did the following to implement co_broadcasting in operator overloading
1. Implement DimMax trait, which uses the same broadcast mechanism as Numpy, for all Dimension so as to get the returned array type in operator overloading. 
2. Use map_collect() and map_collect_owned() in operator overloading to avoid redundant array traversal.

### DIFF
--- a/src/dimension/broadcast.rs
+++ b/src/dimension/broadcast.rs
@@ -23,15 +23,13 @@ where
     // (https://docs.scipy.org/doc/numpy/user/basics.broadcasting.html#general-broadcasting-rules).
     //
     // Zero dimension element is not in the original rules of broadcasting.
-    // We currently treat it as the same as 1. Especially, when one side is
-    // zero with one side is empty, or both sides are zero, the result will
-    // remain zero.
+    // We currently treat it like any other number greater than 1. As numpy does.
     for i in 0..shape1.ndim() {
         out_slice[i] = s1[i];
     }
     for i in 0..shape2.ndim() {
-        if out_slice[i + k] != s2[i] && s2[i] != 0 {
-            if out_slice[i + k] <= 1 {
+        if out_slice[i + k] != s2[i] {
+            if out_slice[i + k] == 1 {
                 out_slice[i + k] = s2[i]
             } else if s2[i] != 1 {
                 return Err(from_kind(ErrorKind::IncompatibleShape));
--- a/src/dimension/broadcast.rs
+++ b/src/dimension/broadcast.rs
@@ -1,12 +1,12 @@
 use crate::error::*;
 use crate::{Dimension, Ix0, Ix1, Ix2, Ix3, Ix4, Ix5, Ix6, IxDyn};
 
-/// Calculate the common shape for a pair of array shapes, which can be broadcasted
-/// to each other. Return an error if shapes are not compatible.
+/// Calculate the common shape for a pair of array shapes, that they can be broadcasted
+/// to. Return an error if the shapes are not compatible.
 ///
 /// Uses the [NumPy broadcasting rules]
 //  (https://docs.scipy.org/doc/numpy/user/basics.broadcasting.html#general-broadcasting-rules).
-fn co_broadcasting<D1, D2, Output>(shape1: &D1, shape2: &D2) -> Result<Output, ShapeError>
+fn co_broadcast<D1, D2, Output>(shape1: &D1, shape2: &D2) -> Result<Output, ShapeError>
     where
         D1: Dimension,
         D2: Dimension,
@@ -15,7 +15,7 @@ fn co_broadcasting<D1, D2, Output>(shape1: &D1, shape2: &D2) -> Result<Output, S
     let (k, overflow) = shape1.ndim().overflowing_sub(shape2.ndim());
     // Swap the order if d2 is longer.
     if overflow {
-        return co_broadcasting::<D2, D1, Output>(shape2, shape1);
+        return co_broadcast::<D2, D1, Output>(shape2, shape1);
     }
     // The output should be the same length as shape1.
     let mut out = Output::zeros(shape1.ndim());
@@ -38,7 +38,7 @@ pub trait BroadcastShape<Other: Dimension> {
     /// The resulting dimension type after broadcasting.
     type Output: Dimension;
 
-    /// Determines the shape after broadcasting the dimensions together.
+    /// Determines the shape after broadcasting the shapes together.
     ///
     /// If the shapes are not compatible, returns `Err`.
     fn broadcast_shape(&self, other: &Other) -> Result<Self::Output, ShapeError>;
@@ -51,7 +51,7 @@ impl<D: Dimension> BroadcastShape<D> for D {
     type Output = D;
 
     fn broadcast_shape(&self, other: &D) -> Result<Self::Output, ShapeError> {
-        co_broadcasting::<D, D, Self::Output>(self, other)
+        co_broadcast::<D, D, Self::Output>(self, other)
     }
 }
 
@@ -61,7 +61,7 @@ macro_rules! impl_broadcast_distinct_fixed {
             type Output = $larger;
 
             fn broadcast_shape(&self, other: &$larger) -> Result<Self::Output, ShapeError> {
-                co_broadcasting::<Self, $larger, Self::Output>(self, other)
+                co_broadcast::<Self, $larger, Self::Output>(self, other)
             }
         }
 
@@ -69,7 +69,7 @@ macro_rules! impl_broadcast_distinct_fixed {
             type Output = $larger;
 
             fn broadcast_shape(&self, other: &$smaller) -> Result<Self::Output, ShapeError> {
-                co_broadcasting::<Self, $smaller, Self::Output>(self, other)
+                co_broadcast::<Self, $smaller, Self::Output>(self, other)
             }
         }
     };
--- a/src/dimension/broadcast.rs
+++ b/src/dimension/broadcast.rs
@@ -1,0 +1,105 @@
+use crate::error::*;
+use crate::{Dimension, Ix0, Ix1, Ix2, Ix3, Ix4, Ix5, Ix6, IxDyn};
+
+/// Calculate the co_broadcast shape of two dimensions. Return error if shapes are
+/// not compatible.
+fn broadcast_shape<D1, D2, Output>(shape1: &D1, shape2: &D2) -> Result<Output, ShapeError>
+where
+    D1: Dimension,
+    D2: Dimension,
+    Output: Dimension,
+{
+    let (k, overflow) = shape1.ndim().overflowing_sub(shape2.ndim());
+    // Swap the order if d2 is longer.
+    if overflow {
+        return broadcast_shape::<D2, D1, Output>(shape2, shape1);
+    }
+    // The output should be the same length as shape1.
+    let mut out = Output::zeros(shape1.ndim());
+    let out_slice = out.slice_mut();
+    let s1 = shape1.slice();
+    let s2 = shape2.slice();
+    // Uses the [NumPy broadcasting rules]
+    // (https://docs.scipy.org/doc/numpy/user/basics.broadcasting.html#general-broadcasting-rules).
+    //
+    // Zero dimension element is not in the original rules of broadcasting.
+    // We currently treat it as the same as 1. Especially, when one side is
+    // zero with one side is empty, or both sides are zero, the result will
+    // remain zero.
+    for i in 0..shape1.ndim() {
+        out_slice[i] = s1[i];
+    }
+    for i in 0..shape2.ndim() {
+        if out_slice[i + k] != s2[i] && s2[i] != 0 {
+            if out_slice[i + k] <= 1 {
+                out_slice[i + k] = s2[i]
+            } else if s2[i] != 1 {
+                return Err(from_kind(ErrorKind::IncompatibleShape));
+            }
+        }
+    }
+    Ok(out)
+}
+
+pub trait BroadcastShape<Other: Dimension>: Dimension {
+    /// The resulting dimension type after broadcasting.
+    type BroadcastOutput: Dimension;
+
+    /// Determines the shape after broadcasting the dimensions together.
+    ///
+    /// If the dimensions are not compatible, returns `Err`.
+    ///
+    /// Uses the [NumPy broadcasting rules]
+    /// (https://docs.scipy.org/doc/numpy/user/basics.broadcasting.html#general-broadcasting-rules).
+    fn broadcast_shape(&self, other: &Other) -> Result<Self::BroadcastOutput, ShapeError> {
+        broadcast_shape::<Self, Other, Self::BroadcastOutput>(self, other)
+    }
+}
+
+/// Dimensions of the same type remain unchanged when co_broadcast.
+/// So you can directly use D as the resulting type.
+/// (Instead of <D as BroadcastShape<D>>::BroadcastOutput)
+impl<D: Dimension> BroadcastShape<D> for D {
+    type BroadcastOutput = D;
+}
+
+macro_rules! impl_broadcast_distinct_fixed {
+    ($smaller:ty, $larger:ty) => {
+        impl BroadcastShape<$larger> for $smaller {
+            type BroadcastOutput = $larger;
+        }
+
+        impl BroadcastShape<$smaller> for $larger {
+            type BroadcastOutput = $larger;
+        }
+    };
+}
+
+impl_broadcast_distinct_fixed!(Ix0, Ix1);
+impl_broadcast_distinct_fixed!(Ix0, Ix2);
+impl_broadcast_distinct_fixed!(Ix0, Ix3);
+impl_broadcast_distinct_fixed!(Ix0, Ix4);
+impl_broadcast_distinct_fixed!(Ix0, Ix5);
+impl_broadcast_distinct_fixed!(Ix0, Ix6);
+impl_broadcast_distinct_fixed!(Ix1, Ix2);
+impl_broadcast_distinct_fixed!(Ix1, Ix3);
+impl_broadcast_distinct_fixed!(Ix1, Ix4);
+impl_broadcast_distinct_fixed!(Ix1, Ix5);
+impl_broadcast_distinct_fixed!(Ix1, Ix6);
+impl_broadcast_distinct_fixed!(Ix2, Ix3);
+impl_broadcast_distinct_fixed!(Ix2, Ix4);
+impl_broadcast_distinct_fixed!(Ix2, Ix5);
+impl_broadcast_distinct_fixed!(Ix2, Ix6);
+impl_broadcast_distinct_fixed!(Ix3, Ix4);
+impl_broadcast_distinct_fixed!(Ix3, Ix5);
+impl_broadcast_distinct_fixed!(Ix3, Ix6);
+impl_broadcast_distinct_fixed!(Ix4, Ix5);
+impl_broadcast_distinct_fixed!(Ix4, Ix6);
+impl_broadcast_distinct_fixed!(Ix5, Ix6);
+impl_broadcast_distinct_fixed!(Ix0, IxDyn);
+impl_broadcast_distinct_fixed!(Ix1, IxDyn);
+impl_broadcast_distinct_fixed!(Ix2, IxDyn);
+impl_broadcast_distinct_fixed!(Ix3, IxDyn);
+impl_broadcast_distinct_fixed!(Ix4, IxDyn);
+impl_broadcast_distinct_fixed!(Ix5, IxDyn);
+impl_broadcast_distinct_fixed!(Ix6, IxDyn);
--- a/src/dimension/broadcast.rs
+++ b/src/dimension/broadcast.rs
@@ -41,7 +41,7 @@ where
     Ok(out)
 }
 
-pub trait BroadcastShape<Other: Dimension>: Dimension {
+pub trait BroadcastShape<Other: Dimension> {
     /// The resulting dimension type after broadcasting.
     type BroadcastOutput: Dimension;
 
@@ -51,9 +51,7 @@ pub trait BroadcastShape<Other: Dimension>: Dimension {
     ///
     /// Uses the [NumPy broadcasting rules]
     /// (https://docs.scipy.org/doc/numpy/user/basics.broadcasting.html#general-broadcasting-rules).
-    fn broadcast_shape(&self, other: &Other) -> Result<Self::BroadcastOutput, ShapeError> {
-        broadcast_shape::<Self, Other, Self::BroadcastOutput>(self, other)
-    }
+    fn broadcast_shape(&self, other: &Other) -> Result<Self::BroadcastOutput, ShapeError>;
 }
 
 /// Dimensions of the same type remain unchanged when co_broadcast.
@@ -61,16 +59,28 @@ pub trait BroadcastShape<Other: Dimension>: Dimension {
 /// (Instead of <D as BroadcastShape<D>>::BroadcastOutput)
 impl<D: Dimension> BroadcastShape<D> for D {
     type BroadcastOutput = D;
+
+    fn broadcast_shape(&self, other: &D) -> Result<Self::BroadcastOutput, ShapeError> {
+        broadcast_shape::<D, D, Self::BroadcastOutput>(self, other)
+    }
 }
 
 macro_rules! impl_broadcast_distinct_fixed {
     ($smaller:ty, $larger:ty) => {
         impl BroadcastShape<$larger> for $smaller {
             type BroadcastOutput = $larger;
+
+            fn broadcast_shape(&self, other: &$larger) -> Result<Self::BroadcastOutput, ShapeError> {
+                broadcast_shape::<Self, $larger, Self::BroadcastOutput>(self, other)
+            }
         }
 
         impl BroadcastShape<$smaller> for $larger {
             type BroadcastOutput = $larger;
+
+            fn broadcast_shape(&self, other: &$smaller) -> Result<Self::BroadcastOutput, ShapeError> {
+                broadcast_shape::<Self, $smaller, Self::BroadcastOutput>(self, other)
+            }
         }
     };
 }
--- a/src/dimension/broadcast.rs
+++ b/src/dimension/broadcast.rs
@@ -1,26 +1,24 @@
 use crate::error::*;
 use crate::{Dimension, Ix0, Ix1, Ix2, Ix3, Ix4, Ix5, Ix6, IxDyn};
 
-/// Calculate the co_broadcast shape of two dimensions. Return error if shapes are
-/// not compatible.
-fn broadcast_shape<D1, D2, Output>(shape1: &D1, shape2: &D2) -> Result<Output, ShapeError>
-where
-    D1: Dimension,
-    D2: Dimension,
-    Output: Dimension,
+/// Calculate the common shape for a pair of array shapes, which can be broadcasted
+/// to each other. Return an error if shapes are not compatible.
+///
+/// Uses the [NumPy broadcasting rules]
+//  (https://docs.scipy.org/doc/numpy/user/basics.broadcasting.html#general-broadcasting-rules).
+fn co_broadcasting<D1, D2, Output>(shape1: &D1, shape2: &D2) -> Result<Output, ShapeError>
+    where
+        D1: Dimension,
+        D2: Dimension,
+        Output: Dimension,
 {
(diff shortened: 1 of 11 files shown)


bluss:
Co-broadcasting is cool and interesting! I can see that you are attacking our fixme's in code one by one and that's gratifying, but it doesn't always tell the whole story of what might be the most important issues to solve. (*)

I think the type system is still limited in what can be expressed, and we seem to have a typenum-like problem here (not really the fault of typenum, just the limits of the type system). Expressing one generic arithmetic operation is ok.. but what about two? will it be usable for the user?  Imagine this:

```rust
type Mat<D> = Array<f64, D>;
fn foo<D>(x: Mat<D>, y: Mat<D>, z: Mat<D>) -> Mat<D> {
   x + &y * &z
}
```

How would we write the actual return type of the function? Maybe in some cases this makes it harder for the user of ndarray, or can we find a great solution?

It could be that co-broadcasting should be a feature that ndarray implements, without using it by default in the arithmetic ops.

(*) Specifically see the issue about #699 - there are lots of things we want to spend the - implementation, complexity, etc - budget on in arithmetic ops.

src/impl_ops.rs
```diff
@@ -79,27 +104,34 @@ where
...
+/// If their shapes disagree, `self` is broadcast to their broadcast shape,
+/// cloning the data if needed.
 ///
 /// **Panics** if broadcasting isn’t possible.
 impl<'a, A, B, S, S2, D, E> $trt<&'a ArrayBase<S2, E>> for ArrayBase<S, D>
 where
     A: Clone + $trt<B, Output=A>,
     B: Clone,
-    S: DataOwned<Elem=A> + DataMut,
+    S: Data<Elem=A>,
```
jturner314, line 117: This change is unfortunate. It would be nice to continue to reuse owned allocations when possible.

src/numeric/impl_numeric.rs
```diff
@@ -283,10 +283,11 @@ where
     ///     a.mean_axis(Axis(0)).unwrap().mean_axis(Axis(0)).unwrap() == aview0(&3.5)
     /// );
     /// ```
-    pub fn mean_axis(&self, axis: Axis) -> Option<Array<A, D::Smaller>>
+    pub fn mean_axis(&self, axis: Axis) -> Option<Array<A, <D::Smaller as BroadcastShape<Ix0>>::BroadcastOutput>>
     where
         A: Clone + Zero + FromPrimitive + Add<Output = A> + Div<Output = A>,
         D: RemoveAxis,
+        D::Smaller: BroadcastShape<Ix0>,
```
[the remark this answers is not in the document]
jturner314: This is a good point and something I hadn't considered before. Unfortunately, I don't think there's a way to avoid the extra bound without improvements to the language (either specialization or enum-variants-as-types), since the compiler needs to know that `BroadcastShape` is implemented for all possible combinations of dimension types.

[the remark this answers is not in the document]
bluss: Let's see how far we get with the blanket implementation for all `Dimension` with output `Self` and then maybe it can be added that `Ix0` can broadcast to any other dimension.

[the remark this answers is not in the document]
SparrowLii: I can't think of a good way :( If we add a statement like this:

        impl<D: Dimension> BroadcastShape<Ix0> for D {
            type BroadcastOutput = D;
        }
Then it will conflict with the statement on broadcast.rs:62( Although they don’t actually conflict ).

SparrowLii:
> How would we write the actual return type of the function? Maybe in some cases this change makes it harder for the user of ndarray, or can we find a great solution?
If the dimension types of the parameters are all D, the result can also be written directly as D. But if the dimensions are not the same, it will be really troublesome. That's right, if the cost of using operators is that additional restrictions must be added, it maybe better to risk panic. I'm sorry I can't think of a better way for the time being. Thanks for pointing out.

src/impl_ops.rs
SparrowLii: Thanks for pointing out

bluss:
Changed the title from "fix" since it's a new feature, not a fix. There is more feedback and thoughts to be given, when I have time. 🙂

bluss:
Here's a trick that actually works to resolve the `mean` situation. Just to show you a neat trick to use. Ignore the `broadcast_shape` method change for now - that one can be fixed later (debatable if it should be a method at all).

What we do here is that we put the `BroadcastShape<Self>` and `BroadcastShape<Ix0>` requirements into requirements of the `Dimension` itself. That way we rearrange the situation so that instead every type that implements Dimension has to prove - for itself - that it implements these two broadcasts. It seems to sidestep the overlapping problem.

See this temporary branch for that diff: https://github.com/rust-ndarray/ndarray/pull/new/bluss/pr-898-broadcast-shape

jturner314:
> Here's a trick that actually works to resolve the `mean` situation.

That's a really nice solution. Is it possible to add `BroadcastShape<Self::Smaller, BroadcastOutput=Self>` and `BroadcastShape<Self::Larger, BroadcastOutput=Self::Larger>` bounds too? If so, I would think all four bounds would avoid a need for users to add `BroadcastShape` bounds to their functions in most cases.

SparrowLii:
> What we do here is that we put the BroadcastShape<Self> and BroadcastShape<Ix0> requirements into requirements of the Dimension itself. That way we rearrange the situation so that instead every type that implements Dimension has to prove - for itself - that it implements these two broadcasts. It seems to sidestep the overlapping problem.

That is amazing! It's so surprising it works! However, we still can’t add statements like BroadcastShape<Self::Smaller, BroadcastOutput=Self> because it will cause loops in analysis passes. But the exciting thing is that I tried this way in the definition of Dimension:

    /// Next smaller dimension (if applicable)
    type Smaller: Dimension + BroadcastShape<Self, BroadcastOutput=Self>;
    /// Next larger dimension
    type Larger: Dimension + RemoveAxis + BroadcastShape<Self, BroadcastOutput=Self::Larger>;

And it did run successfully in my nightly version (cargo 1.49.0-nightly (d5556aeb8 2020-11-04) ), I think we can achieve this amazing work as soon as the 1.49 version is stable.

bluss:
Note that `+ BroadcastShape<Self::Smaller, BroadcastOutput=Self>` does work it just has to be written like `+ BroadcastShape<<Self as Dimension>::Smaller, BroadcastOutput=Self>` to help the compiler.

jturner314:
We could also add this bound to `RemoveAxis`: `BroadcastShape<<<Self as Dimension>::Smaller as Dimension>::Larger, BroadcastOutput = Self>`.

src/impl_ops.rs
```diff
@@ -53,24 +54,48 @@ macro_rules! impl_binary_op(
...
     S2: Data<Elem=B>,
-    D: Dimension,
+    D: Dimension + BroadcastShape<E>,
     E: Dimension,
 {
-    type Output = ArrayBase<S, D>;
-    fn $mth(self, rhs: ArrayBase<S2, E>) -> ArrayBase<S, D>
+    type Output = Array<A, <D as BroadcastShape<E>>::BroadcastOutput>;
```
bluss, line 72: A feature is removed here - the ability to have arithmetic that preserves ArcArray (ArcArray in and out).

src/impl_ops.rs
```diff
@@ -53,24 +54,48 @@ macro_rules! impl_binary_op(
...
+where
+    A: Clone + $trt<B, Output=A>,
+    B: Clone,
+    S: Data<Elem=A>,
+    S2: Data<Elem=B>,
+    D: Dimension + BroadcastShape<E>,
+    E: Dimension,
+{
+    type Output = Array<A, <D as BroadcastShape<E>>::BroadcastOutput>;
+    fn $mth(self, rhs: ArrayBase<S2, E>) -> Self::Output
```
bluss, line 98: If this impl this method should aim to use rhs' allocation

src/impl_ops.rs
bluss: They are reused still in some cases

src/impl_ops.rs
```diff
@@ -79,27 +104,34 @@ where
...
-    S: DataOwned<Elem=A> + DataMut,
+    S: Data<Elem=A>,
     S2: Data<Elem=B>,
-    D: Dimension,
+    D: Dimension + BroadcastShape<E>,
     E: Dimension,
 {
-    type Output = ArrayBase<S, D>;
-    fn $mth(mut self, rhs: &ArrayBase<S2, E>) -> ArrayBase<S, D>
+    type Output = Array<A, <D as BroadcastShape<E>>::BroadcastOutput>;
```
bluss, line 122: We'd like to preserve the type of owned array here

src/impl_ops.rs
```diff
@@ -79,27 +104,34 @@ where
...
+    fn $mth(self, rhs: &ArrayBase<S2, E>) -> Self::Output
     {
-        self.zip_mut_with(rhs, |x, y| {
+        let shape = self.dim.broadcast_shape(&rhs.dim).unwrap();
+        let mut self_ = if shape.slice() == self.dim.slice() {
+            self.into_owned().into_dimensionality::<<D as BroadcastShape<E>>::BroadcastOutput>().unwrap()
```
bluss, line 127: The method body is quite messy but it's not the highest priority to solve first.

src/impl_ops.rs
```diff
@@ -117,13 +150,21 @@ where
...
     S: Data<Elem=A>,
     S2: Data<Elem=B>,
-    D: Dimension,
+    D: Dimension + BroadcastShape<E>,
     E: Dimension,
 {
-    type Output = Array<A, D>;
-    fn $mth(self, rhs: &'a ArrayBase<S2, E>) -> Array<A, D> {
-        // FIXME: Can we co-broadcast arrays here? And how?
-        self.to_owned().$mth(rhs)
```
bluss, line 126: It needs to be mentioned that the number one thing I'd want to fix with this impl is not co-broadcasting, but removing the `.to_owned()` call - that is we don't want to traverse the lhs array twice (once to copy and once to do the arithmetic).

src/impl_ops.rs
```diff
@@ -117,13 +150,21 @@ where
...
+    type Output = Array<A, <D as BroadcastShape<E>>::BroadcastOutput>;
+    fn $mth(self, rhs: &'a ArrayBase<S2, E>) -> Self::Output {
+        let shape = self.dim.broadcast_shape(&rhs.dim).unwrap();
+        let mut self_ = if shape.slice() == self.dim.slice() {
+            self.to_owned().into_dimensionality::<<D as BroadcastShape<E>>::BroadcastOutput>().unwrap()
+        } else {
+            self.broadcast(shape).unwrap().to_owned()
```
bluss, line 162: This line of code uses `to_owned` on a broadcasted array - so we copy a large array (separately from the arithmetic) - ideally we should not do this - allocate the destination array uninitialized (`maybe_uninit`) and broadcast both inputs as views and do the operation.

tests/array.rs
```diff
@@ -485,7 +485,7 @@ fn test_add() {
     }
 
     let B = A.clone();
-    A = A + &B;
+    let A = A + &B;
```
bluss, line 488: It's a goal that we can revert this change 🙂

tests/array.rs
```diff
@@ -1558,6 +1558,53 @@ fn insert_axis_view() {
...
+        &Dim(vec![1, 1, 3, 1, 5, 1, 7]),
+        &Dim([2, 1, 4, 1, 6, 1]),
+        Ok(Dim(IxDynImpl::from(v.as_slice()))),
+    );
+    let d = Dim([1, 2, 1, 3]);
+    test_co(&d, &d, Ok(d));
+    test_co(
+        &Dim([2, 1, 2]).into_dyn(),
+        &Dim(0),
+        Ok(Dim([2, 1, 2]).into_dyn()),
```
bluss, line 1592: Should probably be a shape error when one the rhs is this zero dimension?

```python
>>> import numpy as np
>>> np.arange(4.)
array([0., 1., 2., 3.])
>>> a = np.arange(4.).reshape((2, 1, 2))
>>> a
array([[[0., 1.]],

       [[2., 3.]]])
>>> b = a[:0, :, :0]
>>> b = b.reshape((0, ))
>>> b
array([], dtype=float64)
>>> b.shape
(0,)
>>> a + b
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: operands could not be broadcast together with shapes (2,1,2) (0,)
>>>
```

bluss (commented):
I'm sorry I haven't at this time provided a high-level plan

SparrowLii:
I used uninitialized() to create an uninitialized array, and added the zip_mut_from_pair method to impl_methods.rs to get the result. Now no matter whether we need co_broadcast or not, we only need to traverse self and rhs once each. And ArcArray will not be turned into Array. But there is still a problem: both uninitialized() and maybe_uninit() methods require elements to implement the copy trait (I don’t know the reason behind this). So it is still unavoidable to change the signature of the method in impl_numeric.rs.

src/impl_ops.rs
SparrowLii: Thanks for correcting. I made a separate implementation of it now.

src/impl_ops.rs
SparrowLii: That's right. We should not abandon the limitation of DataOwned. For non-DataOwned arrays, you can use &.

tests/array.rs
SparrowLii: That's right. We can achieve it by keeping the limit of DataOwned.

tests/array.rs
SparrowLii: Yes. I did not test the zero dimension in numpy. I will test them and be consistent with numpy.

bluss:
uninitialized is hard to use correctly. I haven't removed it yet because I think I have managed to put it on sound ground - it *is possible* to use it soundly.

Unfortunately this method

1. Has some usage by actual users (github code search confirms)
2. This usage is from casual inspection almost always technically incorrect

Please use `maybe_uninit`. 🙂 If you do, then I don't have to go through and point out where your usage of `uninitialized` breaks the soundness rules 🙂

src/dimension/broadcast.rs
```diff
@@ -0,0 +1,115 @@
...
+                return Err(from_kind(ErrorKind::IncompatibleShape));
+            }
+        }
+    }
+    Ok(out)
+}
+
+pub trait BroadcastShape<Other: Dimension> {
+    /// The resulting dimension type after broadcasting.
+    type BroadcastOutput: Dimension;
```
bluss, line 46: A trait with an "output" type (See RawDataSubst and FnOnce as inspiration) should almost always prefer the associated type name "Output" as a matter of convention. Think of it as a type-level function's return value.

src/impl_methods.rs
```diff
@@ -1968,6 +1968,46 @@ where
...
+    /// Traverse two arrays in unspecified order, in lock step,
+    /// calling the closure `f` on each element pair, and put
+    /// the result into the corresponding element of self.
+    pub fn zip_mut_from_pair<B, C, S1, S2, F>(&mut self, lhs: &ArrayBase<S1, D>, rhs: &ArrayBase<S2, D>, f: F, )
+    where
+        S: DataMut,
+        S1: Data<Elem = B>,
+        S2: Data<Elem = C>,
+        F: Fn(&B, &C) -> A,
+    {
```
bluss, line 1980: Why not just use `Zip` in this method? `Zip::map_assign_into` should do it (assign into self)

src/impl_ops.rs
```diff
@@ -108,22 +175,33 @@ where
...
-        // FIXME: Can we co-broadcast arrays here? And how?
-        self.to_owned().$mth(rhs)
+    type Output = Array<A, <D as BroadcastShape<E>>::BroadcastOutput>;
+    fn $mth(self, rhs: &'a ArrayBase<S2, E>) -> Self::Output {
+        let shape = self.dim.broadcast_shape(&rhs.dim).unwrap();
+        // SAFETY: Overwrite all the elements in the array after
+        // it is created via `zip_mut_from_pair`.
+        let mut out = unsafe {
+            Self::Output::uninitialized(shape.clone().into_pattern())
+        };
```
bluss, line 198: On closer thought, it would be better to just use plainly `map_collect` here - it allocates the output array for you, in an ok memory layout and you don't have to write the uninit logic correctly yourself.

SparrowLii:
> Please use maybe_uninit that I mentioned. 🙂 If you do, then I don't have to go through and point out where your usage of uninitialized breaks the soundness rules 🙂

I tried to use maybe_uninit(), but I can’t think of a simple way to generate an ArrayBase with the data type MaybeUninit<A>. So I added a feature called MaybeUninitSubst to get the mapping from ArrayBase with element A to MaybeUnint<A>. In addition used Zip to get the results.

src/dimension/broadcast.rs
SparrowLii: Thanks much for teaching.  I changed BroadCastOutput to Output

src/impl_methods.rs
SparrowLii: Right, we can use Zip. This function does not need to be added.

bluss:
Good fighting. The type related problems with the maybe_uninit constructor are not so fun, we're really feeling that we miss the "GAT" feature here (still experimental).

tests/array.rs
SparrowLii: As for zero dimension in numpy, if we do like this:

    import numpy as np
    a = np.arange(4.).reshape((2, 2, 1))
    print(a.shape)
    
    b = a[:2,:2,:0]
    b = b.reshape((0))
    print(b.shape)
    
    print(a+b)

It's just fine:

    (2, 2, 1)
    (0,)
    []

In 1 and 0, numpy will also choose 0：

    import numpy as np
    a = np.arange(2.).reshape((1, 2, 1))
    print(a.shape)
    
    b = a[:0,:2,:1]
    b = b.reshape((0,2,1))
    print(b.shape)
    c=a+b
    print(c.shape)

Output:

    (1, 2, 1)
    (0, 2, 1)
    (0, 2, 1)

Then we can look PyArray_BroadcastToShape method in numpy/numpy/core/src/multiarray/iterators:204:

        for (i = 0; i < PyArray_NDIM(ao); i++, j++) {
            if (PyArray_DIMS(ao)[i] == 1) {
                continue;
            }
            if (PyArray_DIMS(ao)[i] != dims[j]) {
                compat = 0;
                break;
            }
        }
        if (!compat) {
            goto err;
        }

numpy treats 0 values just like other values greater than 1.

bluss:
I'm working on fixing the maybe uninit functionality so that it can do what we need.

SparrowLii:
> I'm working on fixing the maybe uninit functionality so that it can do what we need.

Cool! It will be very exciting if it and co_broadcast both land.

SparrowLii:
@bluss Thank you so much for the amazing work! I have rebased the pr and used map_collect_owned according to your guide to avoid unsafe code.

src/dimension/broadcast.rs
```diff
@@ -0,0 +1,113 @@
...
+    let mut out = Output::zeros(shape1.ndim());
+    let out_slice = out.slice_mut();
+    let s1 = shape1.slice();
+    let s2 = shape2.slice();
+    // Uses the [NumPy broadcasting rules]
+    // (https://docs.scipy.org/doc/numpy/user/basics.broadcasting.html#general-broadcasting-rules).
+    //
+    // Zero dimension element is not in the original rules of broadcasting.
+    // We currently treat it like any other number greater than 1. As numpy does.
+    for i in 0..shape1.ndim() {
```
bluss, line 27: index loop that should be a `zip(&mut out_slice, &s1)` loop.

src/dimension/broadcast.rs
```diff
@@ -0,0 +1,113 @@
...
+        out_slice[i] = s1[i];
+    }
+    for i in 0..shape2.ndim() {
```
bluss, line 30: index loop that should be a `zip(&mut out_slice[k..], &s2)` loop

SparrowLii: OK. Thanks a lot for taking the time to review.

src/dimension/broadcast.rs
```diff
@@ -0,0 +1,110 @@
...
+    // Swap the order if d2 is longer.
+    if overflow {
+        return broadcast_shape::<D2, D1, Output>(shape2, shape1);
+    }
+    // The output should be the same length as shape1.
+    let mut out = Output::zeros(shape1.ndim());
+    // Uses the [NumPy broadcasting rules]
+    // (https://docs.scipy.org/doc/numpy/user/basics.broadcasting.html#general-broadcasting-rules).
+    //
+    // Zero dimension element is not in the original rules of broadcasting.
```
bluss, line 22: Which original rules? `ArrayBase::broadcast` seems concsistent with these rules already. Also, we need to ensure here that we do exactly the same thing as in `ArrayBase::broadcast`.

bluss (changes requested):
It's starting to getting in shape, we've managed to reduce the arithmetic ops trait requirements to a minimum (optimal would be to have all dimensions implement `BroadcastShape<E>` but it seems impossible and this is acceptable for now).

Arithmetic ops and broadcasting docs (on ArrayBase) look like they need an update

tests/array.rs
```diff
@@ -1566,6 +1613,49 @@ fn arithmetic_broadcast() {
...
+
+    // shared array
+    let sa = a.to_shared();
+    let sa2 = sa.to_shared();
+    let sb = b.to_shared();
+    let sb2 = sb.to_shared();
+    let sc = c.to_shared();
+    let sc2 = sc.into_shared();
+    assert_eq!(
+        sa2 + &sb2 + sc2.into_owned(),
```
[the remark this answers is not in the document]
SparrowLii: Yes, just for mixed operations

src/impl_ops.rs
```diff
@@ -108,22 +160,25 @@ where
...
 {
-    type Output = Array<A, D>;
-    fn $mth(self, rhs: &'a ArrayBase<S2, E>) -> Array<A, D> {
-        // FIXME: Can we co-broadcast arrays here? And how?
-        self.to_owned().$mth(rhs)
+    type Output = Array<A, <D as BroadcastShape<E>>::Output>;
+    fn $mth(self, rhs: &'a ArrayBase<S2, E>) -> Self::Output {
+        let shape = self.dim.broadcast_shape(&rhs.dim).unwrap();
+        let lhs = self.broadcast(shape.clone()).unwrap();
+        let rhs = rhs.broadcast(shape).unwrap();
```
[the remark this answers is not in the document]
SparrowLii: That's right. But the method to convert the dim of ArrayBase seems to have only into_dimensionality and broadcast. However, the former cannot be used for reference types. I think we can add the judgment whether the shape is the same in the broadcast.

src/dimension/broadcast.rs
SparrowLii: I did not find a description of the zero value in the Numpy broadcast rules cited above. However, if there is no difference between a zero value and a value greater than 1, I think this explanation can be omitted. 
`ArrayBase::broadcast` only detects the shape mismatch, and does not make any changes to dim.

src/impl_ops.rs
```diff
@@ -90,16 +96,59 @@ where
...
-    type Output = ArrayBase<S, D>;
-    fn $mth(mut self, rhs: &ArrayBase<S2, E>) -> ArrayBase<S, D>
+    type Output = ArrayBase<S, <D as BroadcastShape<E>>::Output>;
+    fn $mth(self, rhs: &ArrayBase<S2, E>) -> Self::Output
     {
-        self.zip_mut_with(rhs, |x, y| {
-            *x = x.clone() $operator y.clone();
-        });
-        self
+        if self.ndim() == rhs.ndim() && self.shape() == rhs.shape() {
```
jturner314, line 105: I haven't reviewed the PR in detail, but I just wanted to make a small comment here. There's another case that's worth considering: when it's possible to broadcast the RHS to match the existing shape of the LHS. In this case, it's possible to reuse the existing LHS allocation (like in the current release of `ndarray` and the `if` branch here), instead of collecting into a new array (like the `else` branch here).

This observation also applies analogously to the other ops impl with `DataOwned`.

src/dimension/broadcast.rs
```diff
@@ -0,0 +1,113 @@
...
+                out_slice[i + k] = s2[i]
+            } else if s2[i] != 1 {
+                return Err(from_kind(ErrorKind::IncompatibleShape));
+            }
+        }
+    }
+    Ok(out)
+}
+
+pub trait BroadcastShape<Other: Dimension> {
```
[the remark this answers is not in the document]
jturner314: I don't know the answer to the tuple question, but I had another thought about this trait -- we may want to name it `DimMax`. We're going to add `DimAdd` in #570, and we'll want `DimSub` for slicing with ellipses. The name `DimMax` would be more consistent with `DimAdd`/`DimSub`.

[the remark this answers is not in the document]
bluss: ok, that's good. Can be fixed after this PR is merged, so not blocking, but we'll use that name.

src/dimension/broadcast.rs
```diff
@@ -0,0 +1,113 @@
...
+    /// The resulting dimension type after broadcasting.
+    type Output: Dimension;
+
+    /// Determines the shape after broadcasting the dimensions together.
```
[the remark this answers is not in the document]
bluss:
```suggestion
    /// Determines the shape after broadcasting the shapes together.
```

bluss (commented):
The PR is basically ready, a few minor things we want to change and then we merge. A few of the other issues mentioned can be fixed after merging.

src/impl_ops.rs
bluss: This is room for future improvement.

bluss (approved):

src/impl_ops.rs
```diff
@@ -79,27 +83,75 @@ where
...
+            let mut out = self.into_dimensionality::<<D as BroadcastShape<E>>::Output>().unwrap();
+            out.zip_mut_with_same_shape(rhs, clone_iopf(A::$mth));
+            out
+        } else {
+            let shape = self.dim.broadcast_shape(&rhs.dim).unwrap();
+            let lhs = self.broadcast(shape.clone()).unwrap();
+            let rhs = rhs.broadcast(shape).unwrap();
```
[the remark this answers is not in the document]
SparrowLii: @bluss It does make sense. We can't directly use `(&self).$mth(rhs)`, which will limit the return type to Array. I think it is a good choice to write it separately as a function.

SparrowLii:
> The PR is basically ready, a few minor things we want to change and then we merge. A few of the other issues mentioned can be fixed after merging.

Thank you very much for reviewing. It does have points that need improvement. I added a function called broadcast_with, but if this is not the focus at the moment, maybe we can consider it in the future.

bluss:
Ok that's good, please make broadcast_with `pub(crate)` for now - otherwise we have to design name, signature and doc comment more. Let's get to the merge.

bluss:
I'll just make a list of "known issues" that we can fix **after** the PR review and merge

- Rename BroadcastShape to `DimMax` - we also move the broadcast method out of the trait (it can just be a free function)
- It is a wish that the `D: Dimension + BroadcastShape<E>` bound in arith ops becomes just `D: Dimension`, i.e. the DimMax with associated type becomes implied by `D: Dimension`. At this point it is not certain how this can be done.
- Preserve the allocation of the input array in arithmetic ops in more cases (when only the left or right side needs broadcast)
- Improve performance of "no-broadcasting-needed" scenario in &array + &array operation
- Potentially provide public function or method similar to `broadcast_with` - that does the broadcasting of two arrays together.

src/impl_methods.rs
```diff
@@ -1707,6 +1707,36 @@ where
         unsafe { Some(ArrayView::new(self.ptr, dim, broadcast_strides)) }
     }
 
+    /// Calculate the views of two ArrayBases after broadcasting each other, if possible.
```
bluss, line 1710: It will be an internal method so it doesn't matter so much, but we don't use this terminology ("ArrayBases") in the docs.
```suggestion
    /// For two arrays or views, find their common shape if possible and broadcast them as array views into that shape
```

SparrowLii:
@bluss It has been modified accordingly.  I very much look forward to our ability to solve the above problems in the future. Especially removing the restriction like BroadCast<E>, it may represent a new language feature IMO, which will be so cool and exciting.

bluss:
I've rebased this to fix the conflict (sorry about that). Could we update the PR description for the current state of the PR before merge?

SparrowLii:
> I've rebased this to fix the conflict (sorry about that). Could we update the PR description for the current state of the PR before merge?

OK, it has been modified accordingly

bluss:
Thanks :)

PR #902, opened 2021-01-30
A better `Array::uninit` and deprecate `Array::uninitialized`

bluss:
Rename maybe_uninit to Array::uninit and move to base type

Old type situation: `Array<MaybeUninit<i32>, D>::maybe_uninit()` returns `Array<MaybeUninit<i32>, D>`

New type situation: `Array<i32, D>::uninit()` returns `Array<MaybeUninit<i32>, D>`

This is easier to use in generic code. There's also a new (internal) constructor `build_uninit` that makes
it possible to support any owned array for uninit initialization - for example in variants of `map_collect`.

Old `Array::uninitialized` is also deprecated. This is the first part of #804.

MSRV increases to 1.49 to be able to use the new definition of the `DataOwned` trait.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,7 +20,7 @@ jobs:
           - stable
           - beta
           - nightly
-          - 1.42.0  # MSRV
+          - 1.49.0  # MSRV
 
     steps:
       - uses: actions/checkout@v2
--- a/benches/bench1.rs
+++ b/benches/bench1.rs
@@ -271,7 +271,7 @@ fn add_2d_alloc_zip_uninit(bench: &mut test::Bencher) {
     let a = Array::<i32, _>::zeros((ADD2DSZ, ADD2DSZ));
     let b = Array::<i32, _>::zeros((ADD2DSZ, ADD2DSZ));
     bench.iter(|| unsafe {
-        let mut c = Array::<MaybeUninit<i32>, _>::maybe_uninit(a.dim());
+        let mut c = Array::<i32, _>::uninit(a.dim());
         azip!((&a in &a, &b in &b, c in c.raw_view_mut().cast::<i32>())
             c.write(a + b)
         );
--- a/examples/sort-axis.rs
+++ b/examples/sort-axis.rs
@@ -102,7 +102,7 @@ where
         assert_eq!(axis_len, perm.indices.len());
         debug_assert!(perm.correct());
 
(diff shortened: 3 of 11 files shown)


bluss:
@SparrowLii I have a commit here that shows my idea of how you'd use this to adapt your arithmetic ops changes for this  https://github.com/rust-ndarray/ndarray/pull/new/bluss/pr-898-use-collect  now that's a very messy commit, only one of the arithmetic ops impls is fully implemented, but it's meant to get the idea across without finishing it. Please don't merge master or this branch into your work (at least not until this PR is merged - I'd prefer rebasing after that, not merge).

bluss:
Rust 1.42 doesn't support the associated type and trait bound situation, trying to find which Rust fix or feature that we now depend on. Suspecting Rust 1.49 https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1490-2020-12-31

bluss:
Rust 1.49 confirmed by CI. It looks like this change https://github.com/rust-lang/rust/pull/73905 was needed for DataOwned's new definition to compile.

bluss:
Breaking just by updating MSRV to 1.49

bluss:
cc #796 for deprecating uninitialized

SparrowLii:
@bluss Thank you very much for submitting this pr and it is amazing! I am trying to rebase my work to be able to use it

jturner314:
This new API is very nice. It's great to have a convenient, sound way for users to create uninitialized arrays and initialize them, while working cleanly with generic code.

bluss:
So far `build_uninit` is internal - it could be made public (probably then using ArrayViewMut instead of a raw view) - I wrote we want som experience before making it public.

SparrowLii:
> So far `build_uninit` is internal - it could be made public (probably then using ArrayViewMut instead of a raw view) - I wrote we want som experience before making it public. But it's not necessary for users that work with `Array` only generic S: DataOwned

I think it’s better not to let users distinguish between A and MaybeUinit<A>. If build_uninit becomes public, users will inevitably use unsafe code to call assume_init(). It feels a little strange for me.

bluss:
Users can use our unsafe APIs, we need to enable that if nothing else for the expert use - personal projects and related crates that use ndarray. Uninitialized is deprecated because it's very hard to use, and we make a new try with the much improved MaybeUninit interface, and it does indeed require using unsafe code blocks, but not as much, and it should be easier to reason about.

PR #903, opened 2021-01-30
Use raw pointers for .swap(i, j)

### LINKED ISSUE #796: Known issues and concerns w.r.t soundness and best practices

bluss:
This fixes a minor issue where we want to avoid overlapping &mut from
the same array.

Closes #796

### DIFF
--- a/src/impl_methods.rs
+++ b/src/impl_methods.rs
@@ -6,7 +6,6 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
-use std::ptr as std_ptr;
 use alloc::slice;
 use alloc::vec;
 use alloc::vec::Vec;
@@ -630,11 +629,18 @@ where
         S: DataMut,
         I: NdIndex<D>,
     {
-        let ptr1: *mut _ = &mut self[index1];
-        let ptr2: *mut _ = &mut self[index2];
-        unsafe {
-            std_ptr::swap(ptr1, ptr2);
+        let ptr = self.as_mut_ptr();
+        let offset1 = index1.index_checked(&self.dim, &self.strides);
+        let offset2 = index2.index_checked(&self.dim, &self.strides);
+        if let Some(offset1) = offset1 {
+            if let Some(offset2) = offset2 {
+                unsafe {
+                    std::ptr::swap(ptr.offset(offset1), ptr.offset(offset2));
+                }
+                return;
+            }
         }
+        panic!("swap: index out of bounds for indices {:?} {:?}", index1, index2);
     }
 
     /// Swap elements *unchecked* at indices `index1` and `index2`.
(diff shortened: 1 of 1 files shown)


src/impl_methods.rs
```diff
@@ -630,11 +629,18 @@ where
...
         I: NdIndex<D>,
     {
-        let ptr1: *mut _ = &mut self[index1];
-        let ptr2: *mut _ = &mut self[index2];
-        unsafe {
-            std_ptr::swap(ptr1, ptr2);
+        let ptr = self.as_mut_ptr();
+        let offset1 = index1.index_checked(&self.dim, &self.strides);
+        let offset2 = index2.index_checked(&self.dim, &self.strides);
+        if let Some(offset1) = offset1 {
```
jturner314, line 635: Fwiw, as of Rust 1.46, things like this can be written a little more nicely using [`Option::zip`](https://doc.rust-lang.org/std/option/enum.Option.html#method.zip):

```rust
if let Some((offset1, offset2)) = offset1.zip(offset2) {
    unsafe { ... }
} else {
    panic!(...);
}
```

(or `if let Some((offset1, offset2)) = Option::zip(offset1, offset2)` if that looks better)

bluss: if let (Some(_), Some(_)) = (offset1, offset2) also works IMO. Now which option gives better code :) I wouldn't bother to investigate though.

PR #904, opened 2021-02-01
replace Zip::apply with Zip::for_each in benches and docs

SparrowLii:
Work as a supplement to #894:
·Replace Zip::apply with Zip::for_each including other methods in benches so that they can function normally
·Update related calls and descriptions in the document


bluss:
Thanks!

PR #907, opened 2021-02-03
Rename .visit() to .for_each()

jturner314:
This is more consistent with `std::iter::Iterator` and `ndarray::Zip`.

See also #894, which is similar.

By the way, what do you think about renaming the existing (private) `.unordered_foreach_mut()` method to `.for_each_mut()` and making it public? (I'd also add an explicit lifetime parameter, like the one in `.visit()`, to make it more versatile.)


bluss:
Both ideas sound good

PR #908, opened 2021-02-03
Add internal constructors from_data_ptr(...).with_strides_dim(...) and use everywhere

bluss:
Use array builder from_data_ptr(...).with_strides_dim(...) where possible

This avoids using the naked ArrayBase { .. } and similar struct
constructors, which should make it easier for us to change
representation in the future if needed.

The constructor methods are unsafe since they rely on the caller to
assert certain invariants; this means that this change increases the
number of `unsafe` code blocks in ndarray, but that's only correct since
every ArrayBase construction is safety-critical w.r.t the validity of
the pointer, dimensions and strides taken together.

The resulting code is often easier to read - especially when we are only
updating the strides and dimensions of an existing array or view.

The .with_strides_dim() method requires the Dimension trait so that
appropriate debug assertions can be added in this method if desired.
This precludes us from using this method in the `Clone` impl, but that's
just a minor flaw and the only exception.

### DIFF
--- a/src/impl_internal_constructors.rs
+++ b/src/impl_internal_constructors.rs
@@ -55,6 +55,7 @@ where
     where
         E: Dimension
     {
+        debug_assert_eq!(strides.ndim(), dim.ndim());
         ArrayBase {
             data: self.data,
             ptr: self.ptr,
--- a/src/impl_internal_constructors.rs
+++ b/src/impl_internal_constructors.rs
@@ -1,0 +1,65 @@
+// Copyright 2021 bluss and ndarray developers.
+//
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+use std::ptr::NonNull;
+
+use crate::imp_prelude::*;
+
+// internal "builder-like" methods
+impl<A, S> ArrayBase<S, Ix1>
+where
+    S: RawData<Elem = A>,
+{
+    /// Create an (initially) empty one-dimensional array from the given data and array head
+    /// pointer
+    ///
+    /// ## Safety
+    ///
+    /// The caller must ensure that the data storage and pointer is valid.
+    /// 
+    /// See ArrayView::from_shape_ptr for general pointer validity documentation.
+    pub(crate) unsafe fn from_data_ptr(data: S, ptr: NonNull<A>) -> Self {
+        let array = ArrayBase {
+            data: data,
+            ptr: ptr,
+            dim: Ix1(0),
+            strides: Ix1(1),
+        };
+        debug_assert!(array.pointer_is_inbounds());
+        array
+    }
+}
+
+// internal "builder-like" methods
+impl<A, S, D> ArrayBase<S, D>
+where
+    S: RawData<Elem = A>,
+    D: Dimension,
+{
+
+    /// Set strides and dimension of the array to the new values
+    ///
+    /// The argument order with strides before dimensions is used because strides are often
+    /// computed as derived from the dimension.
+    ///
+    /// ## Safety
+    ///
+    /// The caller needs to ensure that the new strides and dimensions are correct
+    /// for the array data.
+    pub(crate) unsafe fn with_strides_dim<E>(self, strides: E, dim: E) -> ArrayBase<S, E>
+    where
+        E: Dimension
+    {
+        ArrayBase {
+            data: self.data,
+            ptr: self.ptr,
+            dim,
+            strides,
+        }
+    }
+}
--- a/src/impl_methods.rs
+++ b/src/impl_methods.rs
@@ -1942,7 +1942,7 @@ where
         self.index_axis_move(axis, 0)
     }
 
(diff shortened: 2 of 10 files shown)


bluss:
It is now using an empty one-dimensional array for the return value from `ArrayBase::from_data_ptr`. In the first draft it was a zero-dim array which is a smaller struct, but unfortunately those always have length 1, and thus there would be some arrays that were invalid). It wouldn't matter so much since it's temporary and nothing's accessing the array, but it's best to avoid anyway.

src/impl_internal_constructors.rs
```diff
@@ -0,0 +1,65 @@
...
+    /// computed as derived from the dimension.
+    ///
+    /// ## Safety
+    ///
+    /// The caller needs to ensure that the new strides and dimensions are correct
+    /// for the array data.
+    pub(crate) unsafe fn with_strides_dim<E>(self, strides: E, dim: E) -> ArrayBase<S, E>
+    where
+        E: Dimension
+    {
```
jturner314, line 57: It would be nice to add a debug check:

```suggestion
    {
        debug_assert_eq!(strides.ndim(), dim.ndim());
```

src/impl_methods.rs
```diff
@@ -757,11 +753,10 @@ where
...
         let dim = self.dim.remove_axis(axis);
         let strides = self.strides.remove_axis(axis);
-        ArrayBase {
-            ptr: self.ptr,
-            data: self.data,
-            dim,
-            strides,
+        // safe because new dimension, strides allow access to a subset of old data
+        unsafe {
+            ArrayBase::from_data_ptr(self.data, self.ptr)
```
jturner314, line 758: The `from_data_ptr` call here shouldn't be necessary.

```suggestion
            self
```

src/impl_internal_constructors.rs
```diff
@@ -0,0 +1,65 @@
...
+    /// Set strides and dimension of the array to the new values
+    ///
+    /// The argument order with strides before dimensions is used because strides are often
+    /// computed as derived from the dimension.
+    ///
+    /// ## Safety
+    ///
+    /// The caller needs to ensure that the new strides and dimensions are correct
+    /// for the array data.
+    pub(crate) unsafe fn with_strides_dim<E>(self, strides: E, dim: E) -> ArrayBase<S, E>
```
jturner314, line 54: I think the following would be a little more consistent with other internal functions:

```suggestion
    pub(crate) unsafe fn with_dim_strides<E>(self, dim: E, strides: E) -> ArrayBase<S, E>
```

src/impl_methods.rs
```diff
@@ -1576,11 +1560,10 @@ where
...
     /// ```
     pub fn into_dyn(self) -> ArrayBase<S, IxDyn> {
-        ArrayBase {
-            data: self.data,
-            ptr: self.ptr,
-            dim: self.dim.into_dyn(),
-            strides: self.strides.into_dyn(),
+        // safe because new dims equivalent
+        unsafe {
+            ArrayBase::from_data_ptr(self.data, self.ptr)
```
jturner314, line 1565:
```suggestion
            self
```

src/impl_methods.rs
```diff
@@ -1604,27 +1587,19 @@ where
...
-        } else if D::NDIM == None || D2::NDIM == None { // one is dynamic dim
-            if let Some(dim) = D2::from_dimension(&self.dim) {
-                if let Some(strides) = D2::from_dimension(&self.strides) {
-                    return Ok(ArrayBase {
-                        data: self.data,
-                        ptr: self.ptr,
-                        dim,
-                        strides,
-                    });
+                return Ok(ArrayBase::from_data_ptr(self.data, self.ptr)
```
jturner314, line 1595:
```suggestion
                return Ok(self
```

jturner314 (commented):
I like the new `with_strides_dim` method. I made a few minor comments.

src/impl_methods.rs
bluss: oh right ok, remove_axis takes &self, thanks

src/impl_methods.rs
bluss: Here it's a partial move - `into_dyn()` moves the D values.

src/impl_methods.rs
bluss: Same, partial move

src/impl_internal_constructors.rs
bluss: Yes but I think the strides, dim order works best for us - see usage in `into_shape` and the motivation written in this method, The argument order with strides before dimensions is used because strides are often computed as derived from the dimension.

src/impl_methods.rs
jturner314: Oh, right. Sorry about that.

src/impl_internal_constructors.rs
jturner314: Okay, that makes sense.

src/impl_methods.rs
bluss: fixed

src/impl_methods.rs
```diff
@@ -1498,22 +1490,15 @@ where
             return Err(error::incompatible_shapes(&self.dim, &shape));
         }
         // Check if contiguous, if not => copy all, else just adapt strides
```
bluss, line 1492: I see this comment is entirely incorrect (copied from reshape?) but the coming change that will fix into_shape should take care of that..

PR #909, opened 2021-02-03
implement creating SliceArg from arbitrary Dimension

### LINKED ISSUE #892: I would like a simpler way to create a SliceArg that matches an arbitrary Dimension
The code below is my solution for Advent Of Code 2020 Day 17. The problem involves running Conway's Game of Life in both 3 and 4 dimensions. I want to use slice_mut to create a mutable slice for Dimension D. The only way I could figure out how to do it was to convert the array to IxDyn using into_dimensionality, slice the array using a Vec I created, and then convert the array back to dimension D.

To solve a similar problem, I was able to call raw_dim on my existing array, and then iterate over it and modify it. My initial proposal would be to have a similar method (raw_slice?) that would return a SliceArg that I could then iterate over and modify in place.

https://play.rust-lang.org/?version=stable&mode=release&edition=2018&gist=4a028d474bec27d8f32f938ece81b0e4

SparrowLii:
Fix #892 
I added a field named in_dim to SliceInfo to indicate the entire length of the incides, which is the length of the SliceArg to be generated. Then the following implementation is done in slice.rs:

    impl<T, D, Do> AsRef<SliceInfo<D::SliceArg, Do, D>> for SliceInfo<T, Do, D>
        where
            T: AsRef<[SliceOrIndex]>,
            Do: Dimension,
            D: Dimension,
    {
        fn as_ref(&self) -> &SliceInfo<D::SliceArg, Do, D> {
            let index = self.indices.as_ref();
            if let Some(dim) = D::NDIM {
                debug_assert!(index.len() == dim);
            }
            let arg_ref = D::slice_arg_from(index);
            unsafe {
                // This is okay because the only non-zero-sized member of
                // `SliceInfo` is `indices`, so `&SliceInfo<[SliceOrIndex], D>`
                // should have the same bitwise representation as
                // `&[SliceOrIndex]`.
                &*(arg_ref as *const D::SliceArg
                    as *const SliceInfo<D::SliceArg, Do, D>)
            }
        }
    }
In this way, not only IxDyn, but SliceArg of other dimensions can also be generated from AsRef<[SliceOrIndex]>. 
Then `Impl<T, Do> AsRef<SliceInfo<[SliceOrIndex], Do, IxDyn>> for SliceInfo<T, Do, $dim>` is to achieve the conversion of other dimensions to IxDyn

### DIFF
--- a/src/dimension/dimension_trait.rs
+++ b/src/dimension/dimension_trait.rs
@@ -16,7 +16,7 @@ use super::axes_of;
 use super::conversion::Convert;
 use super::{stride_offset, stride_offset_checked};
 use crate::itertools::{enumerate, zip};
-use crate::Axis;
+use crate::{Axis, SliceInfo};
 use crate::IntoDimension;
 use crate::RemoveAxis;
 use crate::{ArrayView1, ArrayViewMut1};
@@ -64,7 +64,7 @@ pub trait Dimension:
     /// - and so on..
     /// - For `IxDyn`: `[SliceOrIndex]`
     ///
-    /// The easiest way to create a `&SliceInfo<SliceArg, Do, D>` is using the
+    /// The easiest way to create a `&SliceInfo<SliceArg, Do>` is using the
     /// [`s![]`](macro.s!.html) macro.
     type SliceArg: ?Sized + AsRef<[SliceOrIndex]>;
     /// Pattern matching friendly form of the dimension value.
@@ -78,9 +78,35 @@ pub trait Dimension:
     type Smaller: Dimension;
     /// Next larger dimension
     type Larger: Dimension + RemoveAxis;
+
     /// Convert index to &Self::SliceArg. Make sure that length of index
     /// consists with Self::NDIM(if it exists).
+    ///
+    /// Panics if conversion failed.
+    #[doc(hidden)]
     fn slice_arg_from(index: &[SliceOrIndex]) -> &Self::SliceArg ;
+
+    /// Convert &SliceInfo<AsRef<[SliceOrIndex]>, Do> to &SliceInfo<D::SliceArg, Do>.
+    /// Generate SliceArg of any dimension via this method.
+    ///
+    /// Panics if conversion failed.
+    #[doc(hidden)]
+    fn slice_info_from<T, Do>(indices: &T) -> &SliceInfo<Self::SliceArg, Do>
+    where
+        T: AsRef<[SliceOrIndex]>,
+        Do: Dimension,
+    {
+        let arg_ref = Self::slice_arg_from(indices.as_ref());
+        unsafe {
+            // This is okay because the only non-zero-sized member of
+            // `SliceInfo` is `indices`, so `&SliceInfo<[SliceOrIndex], D>`
+            // should have the same bitwise representation as
+            // `&[SliceOrIndex]`.
+            &*(arg_ref as *const Self::SliceArg
+                as *const SliceInfo<Self::SliceArg, Do>)
+        }
+    }
+
     /// Returns the number of dimensions (number of axes).
     fn ndim(&self) -> usize;
 
--- a/src/dimension/dimension_trait.rs
+++ b/src/dimension/dimension_trait.rs
@@ -971,10 +971,16 @@ impl Dimension for IxDyn {
     fn from_dimension<D2: Dimension>(d: &D2) -> Option<Self> {
         Some(IxDyn(d.slice()))
     }
+
     #[inline]
     fn slice_arg_from<T: AsRef<[SliceOrIndex]>>(index: &T) -> Result<&Self::SliceArg, ShapeError> {
         Ok(index.as_ref())
     }
+
+    fn into_dyn(self) -> IxDyn {
+        self
+    }
+
     private_impl! {}
 }
 
--- a/src/dimension/dimension_trait.rs
+++ b/src/dimension/dimension_trait.rs
@@ -16,7 +16,7 @@ use super::axes_of;
 use super::conversion::Convert;
 use super::{stride_offset, stride_offset_checked};
 use crate::itertools::{enumerate, zip};
-use crate::{Axis, SliceInfo};
+use crate::{Axis, ShapeError, ErrorKind};
 use crate::IntoDimension;
 use crate::RemoveAxis;
 use crate::{ArrayView1, ArrayViewMut1};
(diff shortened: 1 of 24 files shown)


bluss:
I don't really know the design of the existing code enough to review. Why is there a new type parameter on SliceInfo? 🙂

jturner314:
I can review this when I have time. I am also curious about the new type parameter.

Fwiw, I wrote up some of the reasoning behind the current design here: [Implementing Improved Slicing in `ndarray` 0.11](https://jim.turner.link/pages/impl-slicing-subviews-ndarray/).

SparrowLii:
@jturner314 Thank you very much for reviewing my code! I must say that I just found a much better way. And it will not make any changes to the existing code @bluss —— Why don't we directly create the `&SliceInfo` we need from `AsRef<[SliceOrIndex]>`? Just like the implementation of slice.rs:371, `&[SliceOrIndex]` and `&SliceInfo` can be converted to each other without cost. If the length does not match, just let it simply report an error in slice_arg_from. All we need to do is specify the type of `Do`. This point may be improved (maybe refer to the realization of `s![]`). But for the particular case of  #892, it is completely enough.

src/dimension/dimension_trait.rs
```diff
@@ -78,6 +79,34 @@ pub trait Dimension:
     /// Next larger dimension
     type Larger: Dimension + RemoveAxis;
 
+    /// Convert index to &Self::SliceArg. Make sure that length of index
+    /// consists with Self::NDIM(if it exists).
+    ///
+    /// Panics if conversion failed.
+    #[doc(hidden)]
+    fn slice_arg_from(index: &[SliceOrIndex]) -> &Self::SliceArg ;
```
bluss, line 87: I'm unsure if this should be a method on Dimension. It could be, it's a bit wonky? This is a hidden method so it is then not intended to be public user-facing API.

bluss: This method should not be listed as the first method in the Dimension trait. It is not the most important method in the trait.

jturner314:
I took some time to think about this, and came to the following conclusions:

1. I really don't think the example in #892 should be implemented using `SliceInfo`, even if we added the necessary conversions. Using `SliceInfo` for this case is awkward, and creating it from a `Vec` requires an unnecessary allocation (for the `Vec`).  IMO, the best way to implement the example is like this:

   ```rust
   let mut v = x.view_mut();
   field
       .axes()
       .for_each(|ax| v.slice_axis_inplace(ax.axis(), Slice::from(2..2 + ax.len())));
   v.assign(field);
   ```

   I've created PR #912 to hopefully document this better.

2. Long-term, to support "ellipsis" and "new axis" functionality in slicing, the cleanest way to express the API will be something like this:

   ```rust
   pub fn slice<I>(&self, info: &I) -> ArrayView<'_, A, I::OutDim>
   where
       I: CanSlice<D>,
       S: Data,
   ```

   In other words, we'll no longer need the `Dimension::SliceArg` associated type.

   If we wanted, we could allow creating the slicing information from a `Vec` and then panic in `.slice()` if the dimensions don't match. This avoids the need for the user to convert the slicing information to a particular type.

   Huh, I just discovered that PR #570 exists is still open. I had forgotten about that.

src/dimension/dimension_trait.rs
jturner314: We could do something like this:

```rust
use std::convert::{TryFrom, TryInto};

pub trait TryFromRef<T: ?Sized> {
    type Error;
    fn try_from_ref(orig: &T) -> Result<&Self, Self::Error>;
}

impl<T: ?Sized, U: ?Sized> TryFromRef<T> for U
where
    for<'a> &'a U: TryFrom<&'a T>,
{
    // Once Rust has const generics, we can implement `TryFromRef` separately
    // for `[SliceOrIndex]` and `[SliceOrIndex; N]`, and use
    // `TryFromSliceError` for the `[SliceOrIndex; N]` case.
    type Error = ();

    fn try_from_ref(orig: &T) -> Result<&Self, Self::Error> {
        orig.try_into().map_err(|_| ())
    }
}

pub trait Dimension {
    ...
    type SliceArg: ?Sized + AsRef<[SliceOrIndex]> + TryFromRef<[SliceOrIndex]>;
    ...
}
```

but a hidden method on `Dimension` is no less awkward, IMO.

src/dimension/dimension_trait.rs
```diff
@@ -78,6 +79,34 @@ pub trait Dimension:
...
+
+    /// Convert &SliceInfo<AsRef<[SliceOrIndex]>, Do> to &SliceInfo<D::SliceArg, Do>.
+    /// Generate SliceArg of any dimension via this method.
+    ///
+    /// Panics if conversion failed.
+    #[doc(hidden)]
+    fn slice_info_from<T, Do>(indices: &T) -> &SliceInfo<Self::SliceArg, Do>
```
jturner314, line 94: This isn't correct, because the `Do` here is unconstrained and is not checked. Instead, we should add a method to `SliceInfo`:

```rust
impl<T: ?Sized, D> SliceInfo<T, D>
where
    T: AsRef<[SliceOrIndex]>,
    D: Dimension,
{
    pub fn for_dimensionality<E>(&self) -> Result<&SliceInfo<E::SliceArg, D>, ShapeError>
    where
        E: Dimension,
    {
        ...
    }
}
```

which makes sure that the `D` is correctly preserved.

jturner314 (commented):
I added a couple of comments for how I think the conversion should be implemented, if we decide to add a conversion like this.

src/dimension/dimension_trait.rs
jturner314: For what it's worth, we could do this instead to avoid the need for `Dimension::slice_arg_from`:

```rust
impl<T: ?Sized, D> SliceInfo<T, D>
where
    T: AsRef<[SliceOrIndex]>,
    D: Dimension,
{
    pub fn for_dimensionality<'a, E>(&'a self) -> &'a SliceInfo<E::SliceArg, D>
    where
        E: Dimension,
        &'a D::SliceArg: TryFrom<&'a [SliceOrIndex]>,
    {
        unimplemented!()
    }
}
```

but that would make it awkward to use in generic functions because users would have to add the `&'a D::SliceArg: TryFrom<&'a [SliceOrIndex]>` bound to their functions.

SparrowLii: That's right, function should be put in SliceInfo. But I think it seems difficult to move slice_arg_from out of Dimension.

SparrowLii:
> In other words, we'll no longer need the Dimension::SliceArg associated type, and the type of the thing containing the slice information doesn't need to be any particular type. We could allow SliceInfo to be created from a Vec and check the dimensionality in the constructor.

I think it might be difficult. If we implement CanSlice<D> for the Vec type, how do we determine its Outdim type?

jturner314:
> If we implement `CanSlice<D>` for `Vec<SliceOrIndex>` , how do we determine its Outdim type?

Oh, I meant that it would be possible to create a `SliceInfo` from a `Vec`, rather than implementing `CanSlice` for `Vec<SliceOrIndex>` directly. [The user has to specify the input and output dimensions when manually constructing a `SliceInfo` without `s![]`.](https://github.com/jturner314/ndarray/blob/slice-new-axis/src/slice.rs#L421-L440) Like in your original proposal on this PR, `SliceInfo` has a new type parameter for the dimension of the array being sliced.

I've rebased #570 off the latest `master`. I propose merging #570 instead of this PR, because #570 happens to solve the problem addressed by this PR, via the `SliceInfo::new` method (in addition to the new functionality).

SparrowLii:
> I've rebased #570 off the latest master. I propose merging #570 instead of this PR, because #570 happens to solve the problem addressed by this PR, via the SliceInfo::new method (in addition to the new functionality).
That's OK for me. I would love to see this issue can be solved.

bluss:
IIUC, #570 takes over this use case

PR #910, opened 2021-02-03
Add for_each_mut method

jturner314:
This PR does the following:

- Rename the private `unordered_foreach_mut` method to `for_each_mut`.
- Make `for_each_mut` public.
- Unify some of the implementation of `fold` and `for_each_mut`.
- Lengthen the lifetimes of the `for_each_mut` element borrows to match `for_each`. Unfortunately, the compiler rejects the previous implementation due to the longer borrows, so a compromise involving two checks for contiguity is used to avoid the compilation error.

### DIFF
--- a/src/impl_methods.rs
+++ b/src/impl_methods.rs
@@ -2253,7 +2253,7 @@ where
     /// Call `f` for each element in the array.
     ///
     /// Elements are visited in arbitrary order.
-    pub(crate) fn for_each_mut<F>(&mut self, mut f: F)
+    pub fn for_each_mut<F>(&mut self, mut f: F)
     where
         S: DataMut,
         F: FnMut(&mut A),
--- a/src/impl_methods.rs
+++ b/src/impl_methods.rs
@@ -1459,17 +1459,26 @@ where
     where
         S: DataMut,
     {
+        self.try_as_slice_memory_order_mut().ok()
+    }
+
+    /// Return the array’s data as a slice if it is contiguous, otherwise
+    /// return `self` in the `Err` variant.
+    pub(crate) fn try_as_slice_memory_order_mut(&mut self) -> Result<&mut [A], &mut Self>
+    where
+        S: DataMut,
+    {
         if self.is_contiguous() {
             self.ensure_unique();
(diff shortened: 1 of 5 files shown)


src/impl_methods.rs
```diff
@@ -2250,6 +2230,25 @@ where
...
+    /// Call `f` for each element in the array.
+    ///
+    /// Elements are visited in arbitrary order.
+    pub fn for_each_mut<'a, F>(&'a mut self, f: F)
+    where
+        F: FnMut(&'a mut A),
+        A: 'a,
+        S: DataMut,
+    {
+        if self.is_contiguous() {
```
bluss, line 2242: To avoid the double check, another version of the `try_into_slice` internal method would be needed (`try_into_memory_order_slice` or similar).

This whole method looks like it could be `self.iter_mut().for_each(f)` except that the iterator is order-preserving 🙁

jturner314: > To avoid the double check, another version of the `try_into_slice` internal method would be needed (`try_into_memory_order_slice` or similar).

Okay, I added a `try_as_slice_memory_order_mut` method to do this.

> This whole method looks like it could be `self.iter_mut().for_each(f)` except that the iterator is order-preserving

Yeah, I've been thinking about the API of the iterators, `.into_shape()`, and other methods where there's a choice of order. One approach would be to introduce a wrapper type for specifying the order of things, similar to your [builder suggestion](https://github.com/rust-ndarray/ndarray/issues/390#issuecomment-761769759), but applied before calling `.iter()`/`.into_shape()`/etc., rather than after. So, it would be used like this: `arr.order(Order::Any).iter()` for a "fast-order" iterator, or like this: `arr.order(Order::Fortran).into_shape(new_shape)` for reshaping with Fortran order.

bluss: try_as_slice_memory_order_mut  is a nice solution

bluss (approved):

bluss:
I just realized this method overlaps with the `map_inplace` method

jturner314:
> I just realized this method overlaps with the `map_inplace` method

Oh, I missed that. Thanks for catching it. I've created #911 to replace this PR.

PR #912, opened 2021-02-05
Suggest single-axis slicing methods for generic fns

jturner314:
This makes the single-axis slicing methods more easily discoverable and illustrates why they're especially useful in generic code.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -503,6 +503,16 @@ pub type Ixs = isize;
 /// [`.slice_move()`]: #method.slice_move
 /// [`.slice_collapse()`]: #method.slice_collapse
 ///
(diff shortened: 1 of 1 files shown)


bluss:
This makes sense. Would it also be good to add a helper method for this - like `slice_each_axis_inplace(f: impl FnMut(AxisDescription) -> Slice)` that makes it even more accessible?

bluss (approved):

jturner314:
> This makes sense. Would it also be good to add a helper method for this - like `slice_each_axis_inplace(f: impl FnMut(AxisDescription) -> Slice)` that makes it even more accessible?

Yeah, that seems like a good idea. I've created #913 to replace this PR.

PR #913, opened 2021-02-05
Add slice_each_axis/_mut/_inplace methods

jturner314:
This is an alternative to #912.

Closes #912.


bluss (approved):

bluss:
That's neat for both generic and dynamic dimensional, so I think we arrived at something good :)

jturner314:
Thanks for reviewing this.

PR #915, opened 2021-02-09
Make AxisDescription a non-tuple struct

### LINKED ISSUE #914: Make `AxisDescription` a non-tuple-struct
I'm curious why [`AxisDescription` is a tuple struct](https://docs.rs/ndarray/0.14.0/ndarray/struct.AxisDescription.html). Since the fields are `pub` anyway, I think it would be more convenient to change the definition to this

```rust
pub struct AxisDescription {
    pub axis: Axis,
    pub len: usize,
    pub stride: isize,
}
```

and remove the `.axis()`, `.len()`, and `.stride()` methods.

jturner314:
Closes #914.

### DIFF
--- a/src/impl_methods.rs
+++ b/src/impl_methods.rs
@@ -557,7 +557,7 @@ where
     where
         F: FnMut(AxisDescription) -> Slice,
     {
-        (0..self.ndim()).for_each(|ax| {
+        for ax in 0..self.ndim() {
             self.slice_axis_inplace(
                 Axis(ax),
                 f(AxisDescription {
@@ -566,7 +566,7 @@ where
                     stride: self.strides[ax] as isize,
                 }),
             )
-        })
+        }
     }
 
     /// Return a reference to the element at `index`, or return `None`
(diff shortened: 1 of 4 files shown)


bluss (approved):

src/impl_methods.rs
```diff
@@ -560,11 +560,11 @@ where
         (0..self.ndim()).for_each(|ax| {
```
bluss, line 560: Sorry if too frivolous, but I missed before that style wise this would be preferred to be a regular for loop. Slightly more idiomatic and avoids creating a few more functions to codegen.

jturner314: Okay, no problem. I've added that change.

I guess I'm used to using `for_each` instead of `for` loops whenever possible because, for some iterators, `for_each` can be faster. For a range iterator like this, though, that's probably a premature optimization, and you're right that a `for` loop is more idiomatic and avoids the extra closure.

The docs say:

> It's generally more idiomatic to use a `for` loop, but `for_each` may be more legible when processing items at the end of longer iterator chains. In some cases `for_each` may also be faster than a loop, because it will use internal iteration on adaptors like `Chain`.

I wonder if the compiler could be changed to automatically implement `for` loops using `for_each` instead of `next` when possible.

bluss:
Thanks

PR #916, opened 2021-02-09
Fix sort-axis example, add test to confirm original error and resolution

dam5h:
Noticed a sorting issue when using this example code in my application.  After building a couple tests and sharing with @bluss , he determined that the `perm_i` and `i` vars were swapped in the Zip comprehension.  Once fixed, a single test case was left which would have caught the original failure.

### DIFF
--- a/examples/sort-axis.rs
+++ b/examples/sort-axis.rs
@@ -102,21 +102,27 @@ where
         assert_eq!(axis_len, perm.indices.len());
         debug_assert!(perm.correct());
 
+        if self.is_empty() {
+            return self;
+        }
+
         let mut result = Array::uninit(self.dim());
 
         unsafe {
             // logically move ownership of all elements from self into result
             // the result realizes this ownership at .assume_init() further down
             let mut moved_elements = 0;
-            for i in 0..axis_len {
-                let perm_i = perm.indices[i];
-                Zip::from(result.index_axis_mut(axis, perm_i))
-                    .and(self.index_axis(axis, i))
-                    .for_each(|to, from| {
-                        copy_nonoverlapping(from, to.as_mut_ptr(), 1);
-                        moved_elements += 1;
-                    });
-            }
+            Zip::from(&perm.indices)
+                .and(result.axis_iter_mut(axis))
+                .for_each(|&perm_i, result_pane| {
+                    // possible improvement: use unchecked indexing for `index_axis`
+                    Zip::from(result_pane)
+                        .and(self.index_axis(axis, perm_i))
+                        .for_each(|to, from| {
+                            copy_nonoverlapping(from, to.as_mut_ptr(), 1);
+                            moved_elements += 1;
+                        });
+                });
             debug_assert_eq!(result.len(), moved_elements);
             // panic-critical begin: we must not panic
             // forget moved array elements but not its vec
@@ -129,6 +135,7 @@ where
         }
     }
 }
+
 #[cfg(feature = "std")]
 fn main() {
     let a = Array::linspace(0., 63., 64).into_shape((8, 8)).unwrap();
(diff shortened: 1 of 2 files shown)


examples/sort-axis.rs
```diff
@@ -145,3 +175,140 @@ fn main() {
...
+        let mut sorted_f = a.column(0).to_vec();
+        sorted_f.sort_by(|a, b| a.partial_cmp(b).unwrap());
+        assert_eq!(sorted_f[0], 74880.05859375);
+        assert_eq!(sorted_f[283103], 109439.94140625);
+
+        let perm = a.sort_axis_by(Axis(0), |i, j| a[[i, 0]] < a[[j, 0]]);
+        assert_eq!(perm.indices[0], 5897);
+        assert_eq!(perm.indices[283103], 176939);
+
+        let b = permute_axis_simple(a, Axis(0), &perm);
```
bluss, line 291: if you want to compare permute_axis_simple and permute_axis - why isn't everything else in the function equal?

dam5h: Good point, that was left over from some earlier experimentation.  Just pushed a commit so that the test setups are identical.

bluss:
I pushed, but I'd prefer if real tests were added before merge. And we'd need to fix CI so that it actually tests the examples.

examples/sort-axis.rs
```diff
@@ -102,21 +114,27 @@ where
...
-            for i in 0..axis_len {
-                let perm_i = perm.indices[i];
-                Zip::from(result.index_axis_mut(axis, perm_i))
-                    .and(self.index_axis(axis, i))
-                    .for_each(|to, from| {
-                        copy_nonoverlapping(from, to.as_mut_ptr(), 1);
-                        moved_elements += 1;
-                    });
-            }
+            Zip::from(&perm.indices)
```
bluss, line 127: tip, we can use slices just fine in `Zip`.

examples/sort-axis.rs
```diff
@@ -84,6 +87,15 @@ where
...
+    where
+        F: FnMut(usize, usize) -> Ordering,
+    {
+        let mut perm = self.identity(axis);
+        perm.indices.sort_by(move |&a, &b| less_than(a, b));
+        perm
+    }
 }
 
 impl<A, D> PermuteArray for Array<A, D>
```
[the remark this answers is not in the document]
dam5h: Ah, makes sense!  Now that this is resolved I think we can drop the "simple" method and the associated test.  Also may not need the big array file test either, or at least can trim the file up some to avoid bloat, it was primarily needed just to chase this down as a one off.

examples/sort-axis.rs
dam5h: Thanks for the tip, I haven't used `Zip` much but should start.  I noticed this change being beyond the index swap.

dam5h:
> I pushed, but I'd prefer if real tests were added before merge - we can find a shorter example that was wrong before. And we'd need to fix CI so that it actually tests the examples.

Ah, I made a comment above about the test being outdated now (before seeing this one), I will trim up the file to reproduce the original error and amend my last commit.  Also added `--examples` to the test script.

bluss:
Can we squash together the commits? In particular so that the big array file disappears from the history-to-merge. Thanks :)

examples/sort-axis.rs
```diff
@@ -143,5 +150,42 @@ fn main() {
...
+            [75600.5810546875, 0.000010122515835114419],
+            [75600.703125, -0.000024209462534106906],
+            [75600.8251953125, -0.000057581837088067716],
+            [75600.947265625, -0.00008825070702257629],
+            [75601.0693359375, -0.00009862067207474166],
+        ];
+
+        let perm = a.sort_axis_by(Axis(0), |i, j| a[[i, 0]] < a[[j, 0]]);
+        let b = a.permute_axis(Axis(0), &perm);
+        assert_eq!(b.row(0), array![75600.0927734375, 0.00011850080901105835]);
```
bluss, line 185: sorry for the nitpick but sorting is best by asserting that every element is <= its predecessor (and some way of checking the other elements in the same row are moved correctly).

dam5h: Agreed!  I'll do this and then squash them up.

bluss (approved):

bluss:
Don't forget to update pr description & title before merge for the new state of the PR. Thanks for this

bluss:
Thanks for this!

dam5h:
Sure, thanks for your help and the library!  Used `ndarray` for my first non-toy rust app, took a little understanding at first but its a great library!

PR #918, opened 2021-02-15
Document NdProducer::raw_dim

jturner314:
This is useful because, otherwise, users have to manually calculate the shape of the producer to e.g. allocate an array of the correct shape to zip with it.

### DIFF
--- a/src/zip/ndproducer.rs
+++ b/src/zip/ndproducer.rs
@@ -72,7 +72,7 @@ pub trait NdProducer {
 
     #[doc(hidden)]
     fn layout(&self) -> Layout;
-    #[doc(hidden)]
+    /// Return the shape of the producer.
     fn raw_dim(&self) -> Self::Dim;
     #[doc(hidden)]
     fn equal_dim(&self, dim: &Self::Dim) -> bool {


bluss (approved):

bluss:
Users can use `.map_collect` to have an array of the right shape and layout. I'm also drafting something that makes it easy to do `Array::zeros(a.shape_builder()` to get an array of the same shape and layout (coarsely defined - probably only selecting C/F).

jturner314:
> Users can use `.map_collect` to have an array of the right shape and layout.

Right; my earlier comment wasn't clear. I'm thinking specifically of a discussion with a user on #rust-sci who wanted to copy windows of an `ArrayView1` into columns of a new `Array2`. Calling `Zip::from(arr_1d.windows()).map_collect(...)` would result in `Array1<...>`. If `NdProducer::raw_dim` is exposed, then, the following works:

```rust
fn windows_to_cols<A>(arr: ArrayView1<'_, A>, window_len: usize) -> Array2<A>
where
    A: Clone + Zero,
{
    let windows = arr.windows(window_len);
    let mut out = Array2::zeros([window_len, windows.raw_dim()[0]].f());
    azip!(
        (mut col in out.axis_iter_mut(Axis(1)), window in windows)
        col.assign(&window)
    );
    out
}
```

PR #922, opened 2021-02-18
Add rustfmt check to CI

jturner314:
`rustfmt` is very convenient when writing PRs, but when it's not enforced by CI, the formatting in the project inevitably diverges from the `rustfmt` style. So, when I run `rustfmt` to format a PR, lines unrelated to the changes in the PR are reformatted (since those lines don't follow the `rustfmt` style), and then I have to manually stage only the relevant lines and discard the irrelevant formatting changes. If `rustfmt` style is enforced by CI, then I can just run `rustfmt` before submitting a PR and not have to worry about discarding unrelated formatting changes (since there shouldn't be any).

This PR should fail CI right now. Once we've merged the large open PRs, we can run `cargo fmt` on the whole crate, rerun CI for this PR to make sure it passes, and then merge this PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -76,3 +76,18 @@ jobs:
           components: clippy
       - run: cargo clippy
 
+  format:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        rust:
+          - stable
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: ${{ matrix.rust }}
+          override: true
+          components: rustfmt
+      - run: cargo fmt -- --check


bluss:
While we certainly need some formatting help sometimes, I think rustfmt makes the wrong decisions too often.

jturner314:
That makes sense. I prioritize the usefulness of automatic formatting over the quality of the formatting, but that's just my personal preference.

[The various rustfmt configuration options are listed here.](https://github.com/rust-lang/rustfmt/blob/master/Configurations.md) I don't see any way to configure rustfmt to preserve the formatting in the first example. For the second example, there are options to preserve this indentation style (called "Visual" in the configuration) for some things (imports, expressions, function args, etc.), but I don't see an existing option like this for attributes. The third example could work by increasing the relevant width option. The fourth example may work with [`overflow_delimited_expr`](https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#overflow_delimited_expr) set to `true`, but I'm not sure.

dvc94ch:
My two cents are that the first three examples aren't that bad, but might take some getting used to. The last one isn't very good, but the way to fix it would be to split it into two lines like this:
```
let base = ArrayBase::from_data_ptr(self.data, self.ptr);
Ok(base.with_strides_dim(strides, dim))
```

bluss:
I think the way to go is to use unstable format settings, as soon as rustfmt 2 is out it is also possible on stable.

danieleades:
I strongly believe that the value of consistent formatting *massively* outweighs the pain of small disagreements about the formatting of specific cases.

That said, if you can find a configuration that does what you want, everyone wins

PR #923, opened 2021-02-18
Add clippy lints as annotations to PRs

jturner314:
This makes CI annotate PRs with clippy lints. See https://github.com/actions-rs/clippy-check for details.

Huh, it doesn't look like it worked. When I run `clippy` locally, there are at least a few warnings.


jturner314:
Oh, never mind. This doesn't work for PRs from forked repositories, due to token permissions. [See the message at the bottom of the README.](https://github.com/actions-rs/clippy-check#limitations)

PR #924, opened 2021-02-18
Treat warnings as errors in CI

jturner314:
This helps avoid accidentally missing a warning.

### DIFF
--- a/tests/iterators.rs
+++ b/tests/iterators.rs
@@ -6,12 +6,9 @@
 )]
 
 use ndarray::prelude::*;
-use ndarray::Ix;
-use ndarray::{arr2, arr3, aview1, indices, s, Axis, Data, Dimension, Slice, Zip};
+use ndarray::{arr3, aview1, indices, s, Axis, Slice, Zip};
 
-use itertools::assert_equal;
-use itertools::{enumerate, rev};
-use std::iter::FromIterator;
+use itertools::{assert_equal, enumerate};
 
 macro_rules! assert_panics {
     ($body:expr) => {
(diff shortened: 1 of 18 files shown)


bluss:
I'm thinking we maybe want to have the continue on error flag set on the clippy job (so that other tests also run) (?)

tests/iterators.rs
```diff
@@ -544,9 +543,9 @@ fn axis_chunks_iter_corner_cases() {
     assert_equal(
         it,
         vec![
-            arr2(&[[7.], [6.], [5.]]),
-            arr2(&[[4.], [3.], [2.]]),
-            arr2(&[[1.], [0.]]),
```
bluss, line 549: what was the warning here?

jturner314: Unused import of `arr2` when `std` feature is not enabled.

jturner314:
> I'm thinking we maybe want to have the continue on error flag set on the clippy job (so that other tests also run) (?)

I think they were running. (I was just force-pushing to fix warnings before they got a chance to finish.) I have seen items canceled when other items fail, but from what I've seen, it's always been within a job (i.e. within "tests" or within "cross_test"), so the tests should still run if clippy fails. I'll check to make sure.

bluss:
I think you'd want continue on error, but yeah I wonder why they are running anyway.

https://github.com/petgraph/petgraph/blob/master/.github/workflows/ci.yml#L45

jturner314:
I came across these pages regarding continue-on-error:

- https://stackoverflow.com/q/62045967
- https://github.community/t/continue-on-error-allow-failure-ui-indication/16773

which seem to indicate that continue-on-error makes the job always "pass", i.e. get a green check. They suggest to use `if: succeeded() || failed()` in the `run` step if you want a job to continue running after a step fails or `if: always()` if you want it to also continue running after a step is canceled.

I'll try adding `continue-on-error` to the (currently failing) clippy job to see what happens.

bluss:
Thanks

PR #930, opened 2021-03-03
Speed up applying the permutation in sort-axis example

bluss:
Speed it up by avoiding bounds checking when looking up the pane to move
in the source array.

This works because for any given element pointer in the array we have the
relationship:

.index_axis(axis, 0) + .stride_of(axis) * j == .index_axis(axis, j)

where + is pointer arithmetic on the element pointers.

### DIFF
--- a/examples/sort-axis.rs
+++ b/examples/sort-axis.rs
@@ -1,6 +1,8 @@
 use ndarray::prelude::*;
 use ndarray::{Data, RemoveAxis, Zip};
 
+use rawpointer::PointerExt;
+
 use std::cmp::Ordering;
 use std::ptr::copy_nonoverlapping;
 
@@ -97,8 +99,8 @@ where
     where
         D: RemoveAxis,
     {
-        let axis = axis;
         let axis_len = self.len_of(axis);
+        let axis_stride = self.stride_of(axis);
         assert_eq!(axis_len, perm.indices.len());
         debug_assert!(perm.correct());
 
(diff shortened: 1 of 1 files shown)


examples/sort-axis.rs
```diff
@@ -112,26 +114,48 @@ where
...
             Zip::from(&perm.indices)
                 .and(result.axis_iter_mut(axis))
                 .for_each(|&perm_i, result_pane| {
-                    // possible improvement: use unchecked indexing for `index_axis`
+                    // Use a shortcut to avoid bounds checking in `index_axis` for the source.
+                    //
+                    // It works because for any given element pointer in the array we have the
+                    // relationship:
+                    //
+                    // .index_axis(axis, 0) + .stride_of(axis) * j == .index_axis(axis, j)
```
bluss, line 139: coding style wise this is a bit ugh, since it depends on manual pointer offsetting instead of using an abstraction. Probably this would be ok as an implementation but not great as an *example* and that's where it is now.

jturner314: Yeah, I've never really thought this was a great example to showcase the library, although I'm not sure where else to put it. I do think it would be worth considering adding some of this functionality to `ndarray` itself.

bluss: yes, of course, it's just that it needs API design to be added to ndarray

bluss:
From the discussion we can say that this is not an example but more like a "staging area" for future features, and I'll merge this so that we have this improvement in store for when we move this over as a real feature in ndarray. It's non-obvious what the API should be IMO, and that's the reason it hasn't been included yet.

PR #932, opened 2021-03-05
append, append_row, append_column: methods for appending an array or single rows and columns

### LINKED ISSUE #269: Support `Clone` elements in `stack` and `select`
Related issue to #249

bluss:
For owned arrays specifically (Array), allow appending rows and/or columns or 
whole arrays (along a given axis).

**New methods:**

- `.append_row(ArrayView1<A>) -> Result`
- `.append_column(ArrayView1<A>) -> Result`
- `.append(axis, ArrayView<A, D>) -> Result`
- `.move_into(impl Into<ArrayViewMut>)`

**New features:**

- `stack`, `concatenate` and `.select(...)` now support Clone elements
  (previously only Copy).

The axis we append along should be a "growing axis", i.e the axis with greatest,
and positive, stride. However if that axis is of length 0 or 1, we can always
convert it to the new growing axis.

These methods automatically move the whole array to a compatible memory
layout for appending, if needed.

The examples show that empty arrays are valid both ways (both memory
layouts) - and that might be the easiest way to use these methods for the users.

There were quite a few corner cases in the implementation in this PR, but hopefully
they should all be dealt with in a moderately clean way and with little duplication.

Fixes #269

### DIFF
--- a/src/data_repr.rs
+++ b/src/data_repr.rs
@@ -6,6 +6,8 @@ use alloc::borrow::ToOwned;
 use alloc::vec::Vec;
 use crate::extension::nonnull;
 
+use rawpointer::PointerExt;
+
 /// Array's representation.
 ///
 /// *Don’t use this type directly—use the type alias
@@ -55,6 +57,37 @@ impl<A> OwnedRepr<A> {
         self.ptr
     }
 
+    /// Return end pointer
+    pub(crate) fn as_end_nonnull(&self) -> NonNull<A> {
+        unsafe {
+            self.ptr.add(self.len)
+        }
+    }
+
+    /// Reserve `additional` elements; return the new pointer
+    /// 
+    /// ## Safety
+    ///
+    /// Note that existing pointers into the data are invalidated
+    #[must_use = "must use new pointer to update existing pointers"]
+    pub(crate) fn reserve(&mut self, additional: usize) -> NonNull<A> {
+        self.modify_as_vec(|mut v| {
+            v.reserve(additional);
+            v
+        });
+        self.as_nonnull_mut()
+    }
+
+    /// Set the valid length of the data
+    ///
+    /// ## Safety
+    ///
+    /// The first `new_len` elements of the data should be valid.
+    pub(crate) unsafe fn set_len(&mut self, new_len: usize) {
+        debug_assert!(new_len <= self.capacity);
+        self.len = new_len;
+    }
+
     /// Cast self into equivalent repr of other element type
     ///
     /// ## Safety
@@ -70,6 +103,11 @@ impl<A> OwnedRepr<A> {
             len: self_.len,
             capacity: self_.capacity,
         }
+    }
+
+    fn modify_as_vec(&mut self, f: impl FnOnce(Vec<A>) -> Vec<A>) {
+        let v = self.take_as_vec();
+        *self = Self::from(f(v));
     }
 
     fn take_as_vec(&mut self) -> Vec<A> {
--- a/src/impl_owned_array.rs
+++ b/src/impl_owned_array.rs
@@ -480,9 +480,11 @@ impl<A, D> Array<A, D>
         } else if current_axis_len == 1 {
             // This is the outermost/longest stride axis; so we find the max across the other axes
             let new_stride = self.axes().fold(1, |acc, ax| {
-                if ax.axis == axis { acc } else {
-                    let this_ax = ax.len as isize * ax.stride;
-                    if this_ax.abs() > acc { this_ax } else { acc }
+                if ax.axis == axis || ax.len <= 1 {
+                    acc
+                } else {
+                    let this_ax = ax.len as isize * ax.stride.abs();
+                    if this_ax > acc { this_ax } else { acc }
                 }
             });
             let mut strides = self.strides.clone();
--- a/src/impl_owned_array.rs
+++ b/src/impl_owned_array.rs
@@ -201,14 +201,20 @@ impl<A, D> Array<A, D>
         }
 
         let len_to_append = array.len();
-        if len_to_append == 0 {
-            return Ok(());
-        }
 
         let array_shape = array.raw_dim();
         let mut res_dim = self.raw_dim();
         res_dim[axis.index()] += array_shape[axis.index()];
         let new_len = dimension::size_of_shape_checked(&res_dim)?;
+
+        if len_to_append == 0 {
+            // There are no elements to append and shapes are compatible:
+            // either the dimension increment is zero, or there is an existing
+            // zero in another axis in self.
+            debug_assert_eq!(self.len(), new_len);
+            self.dim = res_dim;
+            return Ok(());
+        }
 
         let self_is_empty = self.is_empty();
 
--- a/src/impl_owned_array.rs
+++ b/src/impl_owned_array.rs
@@ -2,6 +2,8 @@
 use alloc::vec::Vec;
 use std::mem::MaybeUninit;
 
+use rawpointer::PointerExt;
+
 use crate::imp_prelude::*;
 
 use crate::dimension;
@@ -332,15 +334,18 @@ impl<A, D> Array<A, D>
     /// - If the array is empty (the axis or any other has length 0) or if `axis`
     ///   has length 1, then the array can always be appended.
     ///
-    /// ***Errors*** with a layout error if the array is not in standard order or
-    /// if it has holes, even exterior holes (from slicing). <br>
-    /// ***Errors*** with shape error if the length of the input row does not match
-    /// the length of the rows in the array. <br>
+    /// ***Errors*** with shape error if the shape of self does not match the array-to-append;
+    /// all axes *except* the axis along which it being appended matter for this check.
     ///
-    /// The memory layout of the `self` array matters, since it determines in which direction the
-    /// array can easily grow. Notice that an empty array is compatible both ways. The amortized
-    /// average complexity of the append is O(m) where *m* is the number of elements in the
-    /// array-to-append (equivalent to how `Vec::extend` works).
+    /// The memory layout of the `self` array matters for ensuring that the append is efficient.
+    /// Appending automatically changes memory layout of the array so that it is appended to
+    /// along the "growing axis".
+    ///
+    /// Ensure appending is efficient by for example starting from an empty array and/or always
+    /// appending to an array along the same axis.
+    ///
+    /// The amortized average complexity of the append, when appending along its growing axis, is
+    /// O(*m*) where *m* is the length of the row.
     ///
     /// The memory layout of the argument `array` does not matter.
     ///
@@ -404,8 +409,11 @@ impl<A, D> Array<A, D>
         // array must be empty or have `axis` as the outermost (longest stride) axis
         if !self_is_empty && current_axis_len > 1 {
             // `axis` must be max stride axis or equal to its stride
-            let max_stride_axis = self.axes().max_by_key(|ax| ax.stride).unwrap();
-            if max_stride_axis.axis != axis && max_stride_axis.stride > self.stride_of(axis) {
+            let max_axis = self.axes().max_by_key(|ax| ax.stride.abs()).unwrap();
+            if max_axis.axis != axis && max_axis.stride.abs() > self.stride_of(axis) {
+                incompatible_layout = true;
+            }
+            if self.stride_of(axis) < 0 {
                 incompatible_layout = true;
             }
         }
@@ -442,7 +450,8 @@ impl<A, D> Array<A, D>
             // This is the outermost/longest stride axis; so we find the max across the other axes
             let new_stride = self.axes().fold(1, |acc, ax| {
                 if ax.axis == axis { acc } else {
-                    Ord::max(acc, ax.len as isize * ax.stride)
+                    let this_ax = ax.len as isize * ax.stride;
+                    if this_ax.abs() > acc { this_ax } else { acc }
                 }
             });
             let mut strides = self.strides.clone();
@@ -454,37 +463,18 @@ impl<A, D> Array<A, D>
 
         unsafe {
             // grow backing storage and update head ptr
-            debug_assert_eq!(self.data.as_ptr(), self.as_ptr());
-            self.ptr = self.data.reserve(len_to_append); // because we are standard order
-
-            // copy elements from view to the array now
-            //
-            // make a raw view with the new row
-            // safe because the data was "full"
-            let tail_ptr = self.data.as_end_nonnull();
-            let mut tail_view = RawArrayViewMut::new(tail_ptr, array_shape, strides.clone());
-
-            struct SetLenOnDrop<'a, A: 'a> {
-                len: usize,
-                data: &'a mut OwnedRepr<A>,
-            }
-
-            let mut length_guard = SetLenOnDrop {
-                len: self.data.len(),
-                data: &mut self.data,
+            let data_to_array_offset = if std::mem::size_of::<A>() != 0 {
+                self.as_ptr().offset_from(self.data.as_ptr())
+            } else {
+                0
             };
+            debug_assert!(data_to_array_offset >= 0);
+            self.ptr = self.data.reserve(len_to_append).offset(data_to_array_offset);
 
-            impl<A> Drop for SetLenOnDrop<'_, A> {
-                fn drop(&mut self) {
-                    unsafe {
-                        self.data.set_len(self.len);
-                    }
-                }
-            }
-
+            // clone elements from view to the array now
+            //
             // To be robust for panics and drop the right elements, we want
-            // to fill the tail in-order, so that we can drop the right elements on
-            // panic.
+            // to fill the tail in memory order, so that we can drop the right elements on panic.
             //
             // We have: Zip::from(tail_view).and(array)
             // Transform tail_view into standard order by inverting and moving its axes.
@@ -495,23 +485,56 @@ impl<A, D> Array<A, D>
             // doesn't have drop. However, in the interest of code coverage, all elements
             // use this code initially.
 
-            if tail_view.ndim() > 1 {
-                for i in 0..tail_view.ndim() {
-                    if tail_view.stride_of(Axis(i)) < 0 {
-                        tail_view.invert_axis(Axis(i));
+            // Invert axes in tail_view by inverting strides
+            let mut tail_strides = strides.clone();
+            if tail_strides.ndim() > 1 {
+                for i in 0..tail_strides.ndim() {
+                    let s = tail_strides[i] as isize;
+                    if s < 0 {
+                        tail_strides.set_axis(Axis(i), -s as usize);
                         array.invert_axis(Axis(i));
                     }
                 }
-                sort_axes_to_standard_order_tandem(&mut tail_view, &mut array);
+            }
+
+            // With > 0 strides, the current end of data is the correct base pointer for tail_view
+            let tail_ptr = self.data.as_end_nonnull();
+            let mut tail_view = RawArrayViewMut::new(tail_ptr, array_shape, tail_strides);
+
+            if tail_view.ndim() > 1 {
+                sort_axes_in_default_order_tandem(&mut tail_view, &mut array);
+                debug_assert!(tail_view.is_standard_layout(),
+                              "not std layout dim: {:?}, strides: {:?}",
+                              tail_view.shape(), tail_view.strides());
             } 
+
+            // Keep track of currently filled lenght of `self.data` and update it
+            // on scope exit (panic or loop finish).
+            struct SetLenOnDrop<'a, A: 'a> {
+                len: usize,
+                data: &'a mut OwnedRepr<A>,
+            }
+
+            impl<A> Drop for SetLenOnDrop<'_, A> {
+                fn drop(&mut self) {
+                    unsafe {
+                        self.data.set_len(self.len);
+                    }
+                }
+            }
+
+            let mut data_length_guard = SetLenOnDrop {
+                len: self.data.len(),
+                data: &mut self.data,
+            };
+
             Zip::from(tail_view).and(array)
                 .debug_assert_c_order()
                 .for_each(|to, from| {
                     to.write(from.clone());
-                    length_guard.len += 1;
+                    data_length_guard.len += 1;
                 });
-
-            drop(length_guard);
+            drop(data_length_guard);
 
             // update array dimension
             self.strides = strides;
@@ -520,6 +543,7 @@ impl<A, D> Array<A, D>
         // multiple assertions after pointer & dimension update
         debug_assert_eq!(self.data.len(), self.len());
         debug_assert_eq!(self.len(), new_len);
+        debug_assert!(self.pointer_is_inbounds());
 
         Ok(())
     }
@@ -537,20 +561,6 @@ where
         return;
     }
     sort_axes1_impl(&mut a.dim, &mut a.strides);
-}
-
-fn sort_axes_to_standard_order_tandem<S, S2, D>(a: &mut ArrayBase<S, D>, b: &mut ArrayBase<S2, D>)
-where
-    S: RawData,
-    S2: RawData,
-    D: Dimension,
-{
-    if a.ndim() <= 1 {
-        return;
-    }
-    sort_axes_impl(&mut a.dim, &mut a.strides, &mut b.dim, &mut b.strides);
-    debug_assert!(a.is_standard_layout(), "not std layout dim: {:?}, strides: {:?}",
-                  a.shape(), a.strides());
 }
 
 fn sort_axes1_impl<D>(adim: &mut D, astrides: &mut D)
@@ -579,7 +589,23 @@ where
 }
 
 
-fn sort_axes_impl<D>(adim: &mut D, astrides: &mut D, bdim: &mut D, bstrides: &mut D)
+/// Sort axes to standard order, i.e Axis(0) has biggest stride and Axis(n - 1) least stride
+///
+/// Axes in a and b are sorted by the strides of `a`, and `a`'s axes should have stride >= 0 before
+/// calling this method.
+fn sort_axes_in_default_order_tandem<S, S2, D>(a: &mut ArrayBase<S, D>, b: &mut ArrayBase<S2, D>)
+where
+    S: RawData,
+    S2: RawData,
+    D: Dimension,
+{
+    if a.ndim() <= 1 {
+        return;
+    }
+    sort_axes2_impl(&mut a.dim, &mut a.strides, &mut b.dim, &mut b.strides);
+}
+
+fn sort_axes2_impl<D>(adim: &mut D, astrides: &mut D, bdim: &mut D, bstrides: &mut D)
 where
     D: Dimension,
 {
--- a/src/impl_owned_array.rs
+++ b/src/impl_owned_array.rs
@@ -203,3 +203,143 @@ impl<A> Array<A, Ix2> {
     }
 }
 
+impl<A, D> Array<A, D>
+    where D: Dimension
+{
+    /// Append a row to an array with row major memory layout.
+    ///
+    /// ***Errors*** with a layout error if the array is not in standard order or
+    /// if it has holes, even exterior holes (from slicing). <br>
+    /// ***Errors*** with shape error if the length of the input row does not match
+    /// the length of the rows in the array. <br>
+    ///
+    /// The memory layout matters, since it determines in which direction the array can easily
+    /// grow. Notice that an empty array is compatible both ways. The amortized average
+    /// complexity of the append is O(m) where *m* is the length of the row.
+    ///
+    /// ```rust
+    /// use ndarray::{Array, ArrayView, array};
+    ///
+    /// // create an empty array and append
+    /// let mut a = Array::zeros((0, 4));
+    /// a.try_append_row(ArrayView::from(&[1., 2., 3., 4.])).unwrap();
+    /// a.try_append_row(ArrayView::from(&[0., -2., -3., -4.])).unwrap();
+    ///
+    /// assert_eq!(
+    ///     a,
+    ///     array![[1., 2., 3., 4.],
+    ///            [0., -2., -3., -4.]]);
+    /// ```
+    pub fn try_append_array(&mut self, axis: Axis, array: ArrayView<A, D>)
+        -> Result<(), ShapeError>
+    where
+        A: Clone,
+        D: RemoveAxis,
+    {
+        if self.ndim() == 0 {
+            return Err(ShapeError::from_kind(ErrorKind::IncompatibleShape));
+        }
+
+        let remaining_shape = self.raw_dim().remove_axis(axis);
+        let array_rem_shape = array.raw_dim().remove_axis(axis);
+
+        if remaining_shape != array_rem_shape {
+            return Err(ShapeError::from_kind(ErrorKind::IncompatibleShape));
+        }
+
+        let len_to_append = array.len();
+        if len_to_append == 0 {
+            return Ok(());
+        }
+
+        let array_shape = array.raw_dim();
+        let mut res_dim = self.raw_dim();
+        res_dim[axis.index()] += array_shape[axis.index()];
+        let new_len = dimension::size_of_shape_checked(&res_dim)?;
+
+        let self_is_empty = self.is_empty();
+
+        // array must be empty or have `axis` as the outermost (longest stride)
+        // axis
+        if !(self_is_empty ||
+             self.axes().max_by_key(|ax| ax.stride).map(|ax| ax.axis) == Some(axis))
+        {
+            return Err(ShapeError::from_kind(ErrorKind::IncompatibleLayout));
+        }
+
+        // array must be be "full" (have no exterior holes)
+        if self.len() != self.data.len() {
+            return Err(ShapeError::from_kind(ErrorKind::IncompatibleLayout));
+        }
+        let strides = if self_is_empty {
+            // recompute strides - if the array was previously empty, it could have
+            // zeros in strides.
+            res_dim.default_strides()
+        } else {
+            self.strides.clone()
+        };
+
+        unsafe {
+            // grow backing storage and update head ptr
+            debug_assert_eq!(self.data.as_ptr(), self.as_ptr());
+            self.ptr = self.data.reserve(len_to_append); // because we are standard order
+
+            // copy elements from view to the array now
+            //
+            // make a raw view with the new row
+            // safe because the data was "full"
+            let tail_ptr = self.data.as_end_nonnull();
+            let tail_view = RawArrayViewMut::new(tail_ptr, array_shape, strides.clone());
+
+            struct SetLenOnDrop<'a, A: 'a> {
+                len: usize,
+                data: &'a mut OwnedRepr<A>,
+            }
+
+            let mut length_guard = SetLenOnDrop {
+                len: self.data.len(),
+                data: &mut self.data,
+            };
+
+            impl<A> Drop for SetLenOnDrop<'_, A> {
+                fn drop(&mut self) {
+                    unsafe {
+                        self.data.set_len(self.len);
+                    }
+                }
+            }
+
+            // we have a problem here XXX
+            //
+            // To be robust for panics and drop the right elements, we want
+            // to fill the tail in-order, so that we can drop the right elements on
+            // panic. Don't know how to achieve that.
+            //
+            // It might be easier to retrace our steps in a scope guard to drop the right
+            // elements.. (PartialArray style).
+            //
+            // assign the new elements
+            Zip::from(tail_view).and(array)
+                .for_each(|to, from| {
+                    to.write(from.clone());
+                    length_guard.len += 1;
+                });
+
+            //length_guard.len += len_to_append;
+            dbg!(len_to_append);
+            drop(length_guard);
+
+            // update array dimension
+            self.strides = strides;
+            self.dim = res_dim;
+            dbg!(&self.dim);
+
+        }
+        // multiple assertions after pointer & dimension update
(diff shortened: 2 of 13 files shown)


bluss:
Note to others: there already exists other ways to build arrays row-by-row or column-by-column: the stack and concat functions.

src/impl_owned_array.rs
```diff
@@ -59,3 +64,143 @@ where
...
+    /// let mut a = Array::zeros((0, 4));
+    /// a.try_append_row(ArrayView::from(&[1., 2., 3., 4.])).unwrap();
+    /// a.try_append_row(ArrayView::from(&[0., -2., -3., -4.])).unwrap();
+    ///
+    /// assert_eq!(
+    ///     a,
+    ///     array![[1., 2., 3., 4.],
+    ///            [0., -2., -3., -4.]]);
+    /// ```
+    pub fn try_append_row(&mut self, row: ArrayView<A, Ix1>) -> Result<(), ShapeError>
```
bluss, line 98: Consider using the name `append_row`?

bluss: The input could be `&[A]` or ArrayView, but it could also be a `Vec<A>` or `Array1<A>` for ownership semantics. Doesn't look possible to cover both in the same method in a simple way.

jturner314:
Functionality like this would be really nice. I think a slightly different API would be more generally useful:

```rust
impl<A, D> Array<A, D>
where
    D: Dimension,
{
    /// Appends the subview to the specified axis.
    ///
    /// The memory layout of the array will be changed if necessary (by copying the data)
    /// so that subsequent calls to `.append_to_axis()` with the same axis will be fast.
    /// This occurs in at least the following cases:
    ///
    /// - If the outermost (i.e. largest absolute stride) axis of the array is not `axis`.
    /// - If the array is discontiguous.
    /// - If the stride for `axis` is negative.
    ///
    /// So, calling `.append_to_axis()` with varying values of `axis` will likely be very slow, but
    /// after one call to `.append_to_axis()`, subsequent calls to `.append_to_axis()` with the
    /// *same* axis will be relatively fast.
    ///
    /// **Errors** if the axis lengths of `subview` do not match the correponding axis lengths of `self`.
    pub fn append_to_axis<S2>(&mut self, axis: Axis, subview: &ArrayBase<S2, D::Smaller>) -> Result<(), ShapeError>
    where
        A: Clone,
        S2: Data<Elem = A>,
    {}
}
```

In other words, it would be nice for the operation to work for n-dimensional arrays of arbitrary layout. The layout of the array would be changed as necessary so that the specified axis would be the outermost axis, so that repeated calls to `.append_to_axis()` with the same axis would be relatively fast. The remaining question with this design is how to handle the layout of the other axes. We could decide to (1) preserve the existing order, (2) use the order in the subview being appended, or (3) something in-between. I'd suggest option (2), because I think in most cases `.append_to_axis()` will be called repeatedly with the same axis and subviews of the same layout, and because it would simplify the handling of drops if the `.clone()` of one of the elements in `subview` panics (since the new elements would be added in order, like in this PR).

src/impl_owned_array.rs
jturner314: In principle, it would be possible to cover both the owned and non-owned cases by accepting `row: ArrayBase<S2, Ix1> where S2: Data<Elem = A>` and adding the necessary functionality to `Data` to either clone or move elements based on the ownership of the data. I haven't thought through the details of how difficult this would be to implement, though.

bluss:
Yeah that would be more useful, but the complexity is higher. Do we have an easy way to get to the implementation?

How to do the trick where we make `axis` the largest stride axis? If this is enough then maybe it's not too complex to implement:

```rust
a.swap_axis(0, axis);
let a = a.as_standard_layout().into_owned();
a.swap_axis(0, axis);
// axis is now the largest stride axis; and with that axis ignored, the rest is c-contig
```

jturner314:
I suppose I'd suggest something like the following:

```rust
impl<A, D> Array<A, D>
where
    D: Dimension,
{
    /// Rearranges the layout if necessary (by copying the data) such that:
    ///
    /// - if the specified axis has length > 1, it is the outermost axis and
    ///   has a positive stride
    /// - the other axes are in standard layout
    /// - the array is contiguous
    ///
    /// # Example
    ///
    /// ```rust
    /// use ndarray::{Array4, Axis};
    ///
    /// let mut arr = Array4::<u8>::zeros([2, 3, 4, 5]);
    /// assert_eq!(arr.shape(), &[2, 3, 4, 5]);
    /// assert_eq!(arr.strides(), &[60, 20, 5, 1]);
    ///
    /// arr.make_axis_outer_pos(Axis(2));
    /// assert_eq!(arr.shape(), &[2, 3, 4, 5]);
    /// assert_eq!(arr.strides(), &[15, 5, 30, 1]);
    /// ```
    pub(crate) fn make_axis_outer_pos(&mut self, axis: Axis)
    where
        A: Clone,
        D: RemoveAxis,
    {
        let ax = axis.index();
        if self.is_contiguous()
            && is_standard_layout(&self.dim.remove_axis(axis), &self.strides.remove_axis(axis))
            && (self.len_of(axis) <= 1
                || axis == self.dim.max_stride_axis(&self.strides) && self.stride_of(axis) > 0)
        {
            return;
        }
        self.dim.slice_mut()[..=ax].rotate_right(1);
        self.strides.slice_mut()[..=ax].rotate_right(1);
        *self = Array::from_shape_vec(self.dim.clone(), self.iter().cloned().collect()).unwrap();
        self.dim.slice_mut()[..=ax].rotate_left(1);
        self.strides.slice_mut()[..=ax].rotate_left(1);
    }
}
```

(The existing `is_standard_layout` function would need to be moved out of the `.is_standard_layout()` method and into e.g. `dimension/mod.rs`.)

In principle, we could move the elements instead of cloning them, although that would require some unsafe code.

This is basically the same approach as what you wrote above, but uses a rotate instead of a swap so that the other axes are in standard layout. This would make it easy to append the subview since the layout would match the order of `subview.iter()`. Note that if `axis` has length 1 or any of the axes has length 0, we'd need to be careful to make sure the strides are correct after the subview is appended (since the stride of a length 1 axis can be anything, and if there's a length 0 axis, then the strides of the other axes can be 0).

bluss:
I've looked at the append array case, but I'm stuck at the usual, place, handling non-copy elements. I'm stubbornly preferring to not forget/leak values in case of panic, I don't think it's the right thing to do for a Rust data structure.

jturner314:
I'd suggest something like the following:

1. Check that `subview` has the correct shape.
2. Call `self.make_axis_outer_pos(axis)` to adjust the layout of `self`.
3. (Debug) assert that `self.data.ptr == self.ptr`. (`make_axis_outer_pos` should ensure that this is the case.)
4. Call `let v = self.data.take_as_vec();` to convert the `OwnedRepr` to a `Vec`.
5. Call `v.extend(subview.iter().cloned())`.
6. Convert `v` back into an `OwnedRepr` and assign it to `self.data`.
7. Update `self.ptr` to point at the new data.
8. Update `self.shape` by adding 1 to the length of the axis.
9. If the original axis length was <= 1, then make sure the new stride of the axis is correct, since the stride of an axis with length <= 1 could be almost anything. (This step could be performed here, or it could be handled with a modification to `make_axis_outer_pos`.)

This way, the tricky stuff is handled by `Vec::extend`, which AFAIK handles panics without leaks. I'm curious how you were thinking about implementing it, and where the leak could occur.

bluss:
Hm, my implementation uses Zip to make sure shape is traversed in the right order. To use Vec::extend we need to guarantee the array is C-contig except for the axis we are extending, which is a further requirement. So I started a bit different. I'd ideally see that we don't depend on Vec at all - looking forward to future custom allocation, but that concern is just a minor thing in the details here.

jturner314:
> To use Vec::extend we need to guarantee the array is C-contig except for the axis we are extending, which is a further requirement (or, at least, whatever order `subview.iter()` or similar would be using).

The `make_axis_outer_pos` implementation from my earlier comment does this (cloning the data if necessary), although I see now that the name doesn't indicate that. Maybe a better name would be `make_axis_outer_standard`, or the body should just be placed in `append_to_axis` instead of having a separate function. An implementation based on `make_axis_outer_pos` isn't necessarily the fastest implementation, but it's fairly simple.

I guess you were thinking of preserving the original layout of the array as long as the axis being appended to is the outermost one with a positive stride? That would be possible, although as you pointed out, the implementation would be tricky due to iteration order / leaking issues.

> I'd ideally see that we don't depend on Vec at all - looking forward to future custom allocation, but that concern is just a minor thing in the details here.

I'm curious about this. I'd think we'd want to write an allocator which implements the [`std::alloc::Allocator`](https://doc.rust-lang.org/std/alloc/trait.Allocator.html) trait (which will hopefully be stabilized at some point), and `Vec` has an `A: Allocator` type parameter, so it would work with our custom allocator. Or, are you thinking about creating an `ndarray`-specific allocator before `std::alloc::Allocator` is stabilized?

bluss:
I just think I'm following my own implementation, not writing down your ideas here :) So that's the reason. I was standing still at that problem in my implementation, not using your implementation idea from your comment. Just to clarify.

jturner314:
Oh, okay, I understand.

Regarding the allocator stuff, I was thinking of something like this:

```rust
pub struct MinAlignAllocator<A: Allocator, const MIN_ALIGN: usize>(A);

impl<A: Allocator, const MIN_ALIGN: usize> Allocator for MinAlignAllocator {
    fn allocate(&self, layout: Layout) -> Result<NonNull<[u8]>, AllocError> {
        self.0.allocate(layout.align_to(MIN_ALIGN).unwrap_or(AllocError)?)
    }

    fn deallocate(&self, ptr: NonNull<u8>, layout: Layout) {
        self.0.deallocate(ptr, layout.align_to(MIN_ALIGN).unwrap())
    }
}
```

We'd add an `L: Allocator` parameter to `ArrayBase`. We could use a default of `L = Global` and always wrap the allocator in `MinAlignAllocator` internally (if we always want a minimum alignment), or we could use a default of `L = MinAlignAllocator<Global, 64>` and directly use `L` internally (if we want to allow allocations without the minimum alignment). This would be compatible with zero-cost conversions to/from `Vec<T, MinAlignAllocator<A, 64>>`.

This is the cleanest way I see to handle the minimum-alignment issue, since the `GlobalAlloc`/`Allocator` traits require using the same alignment in `alloc`/`allocate` and `dealloc`/`deallocate`. So, there currently isn't a good way to get a minimum alignment for array allocations while also supporting zero-cost conversions to `Vec` (since the `Vec` needs to know the alignment of the original allocation so that it can be correctly dropped, but `Vec<T>` just assumes that the alignment of the allocation is the alignment of `T`).

In current Rust, I suppose we could change `OwnedRepr` to use `GlobalAlloc` directly and force a minimum alignment, since we have complete control over allocation/deallocation of `OwnedRepr`, but that would mean that we couldn't have a zero-cost conversion from `OwnedRepr` to `Vec`.

bluss:
I solved my predicament with the "memory fill order" with Zip. Using a technique I think we have discussed before but I don't recall that it's in the source anywhere: (I think it's cool :slightly_smiling_face:)

To recover the desired traversal order for array views `memory` and `b`:

1. Given `Zip::from(memory).and(b)`
2. Apply axis inversions to `memory` to give it all-positive stride axes, and invert the same axes in `b`
3. Apply axis order swaps on `memory` to give it standard memory layout and swap the same axes in `b`

Thanks @jturner314 for the reminder to fix up strides. I didn't quite realize how involved it might be.

src/impl_owned_array.rs
bluss: No action on this IMO

bluss:
Current state of the PR does not implement adjusting the array layout to fit automatically. But this is not far away, so we might get there in this PR (or a later one).

src/impl_owned_array.rs
```diff
@@ -59,3 +67,538 @@ where
...
+            }
+            self.data
+        } else {
+            self.drop_unreachable_elements_slow()
+        }
+    }
+
+    #[inline(never)]
+    #[cold]
+    fn drop_unreachable_elements_slow(mut self) -> OwnedRepr<A> {
```
bluss, line 201: This method basically contains radioactive amounts of `unsafe`.

bluss:
Benchmarks show a large regression in .select() performance (which is based on concatenate, which is now based on `.append()`). The feature we gain in this PR w.r.t. select is that `Clone` elements are supported.

src/impl_owned_array.rs
```diff
@@ -59,3 +67,541 @@ where
...
+
+/// Methods specific to `Array2`.
+///
+/// ***See also all methods for [`ArrayBase`]***
+///
+/// [`ArrayBase`]: struct.ArrayBase.html
+impl<A> Array<A, Ix2> {
+    /// Append a row to an array
+    ///
+    /// ***Errors*** with shape error if the length of the input row does not match
```
bluss, line 79:
```suggestion
    /// ***Errors*** with a shape error if the length of the input row does not match
```

src/impl_owned_array.rs
```diff
@@ -59,3 +67,541 @@ where
...
+    /// memory order:
+    ///
+    /// `axis` must be the growing axis of the current array, an axis with length 0 or 1.
+    ///
+    /// - This is the 0th axis for standard layout arrays
+    /// - This is the *n*-1 th axis for fortran layout arrays
+    /// - If the array is empty (the axis or any other has length 0) or if `axis`
+    ///   has length 1, then the array can always be appended.
+    ///
+    /// ***Errors*** with shape error if the shape of self does not match the array-to-append;
```
bluss, line 326:
```suggestion
    /// ***Errors*** with a shape error if the shape of self does not match the array-to-append;
```

bluss:
Benchmark:

before:

```text
test select_1d    ... bench:       1,339 ns/iter (+/- 20)
test select_axis0 ... bench:       1,204 ns/iter (+/- 21)
test select_axis1 ... bench:      11,136 ns/iter (+/- 35)
```

after

```text
test select_1d    ... bench:         110 ns/iter (+/- 2)
test select_axis0 ... bench:       1,169 ns/iter (+/- 20)
test select_axis1 ... bench:      21,087 ns/iter (+/- 176)
```

`.select()` used `concatenate` before, and select uses `.append()` after this PR.

jturner314:
Does "before" refer to the current release of `ndarray`? That's a nice improvement!

If you let me know once the PR is ready for review, I'll take a look when I get a chance.

bluss:
Before is a commit before this PR from master. Equivalent to 0.15.1

bluss:
I think it's ready now. It could use some more reworking to make the code neater, but it has received some rounds of that already.

tests/stacking.rs
```diff
@@ -15,6 +15,13 @@ fn concatenating() {
     let d = concatenate![Axis(0), a.row(0), &[9., 9.]];
     assert_eq!(d, aview1(&[2., 2., 9., 9.]));
 
+    let d = concatenate![Axis(1), a.row(0).insert_axis(Axis(1)), aview1(&[9., 9.]).insert_axis(Axis(1))];
+    assert_eq!(d, aview2(&[[2., 9.],
+                           [2., 9.]]));
+
+    let d = concatenate![Axis(0), a.row(0).insert_axis(Axis(1)), aview1(&[9., 9.]).insert_axis(Axis(1))];
+    assert_eq!(d.t(), aview2(&[[2., 2., 9., 9.]]));
```
jturner314, line 23: It took me a while to understand this because I didn't notice the `.t()` at first. This is easier for me to read:

```suggestion
    assert_eq!(d, aview2(&[[2.], [2.], [9.], [9.]]));
```

src/zip/mod.rs
```diff
@@ -427,6 +430,25 @@ where
     }
 }
 
+impl<D, P1, P2> Zip<(P1, P2), D>
+    where D: Dimension,
```
jturner314, line 434: The formatting of this `where` block doesn't match how the `where` blocks for the other `impl`s are formatted.

tests/array.rs
```diff
@@ -709,6 +709,19 @@ fn test_select() {
     assert_abs_diff_eq!(c, c_target);
 }
 
+#[test]
+fn test_select_1d() {
+    let x = arr1(&[0, 1, 2, 3, 4, 5, 6]);
+    let r1 = x.select(Axis(0), &[1, 3, 5]);
```
jturner314, line 715: It would be nice to have a repeated index and an index which is not in increasing order.

jturner314 (commented):
So far, I've reviewed everything except for `src/impl_owned_array.rs`, `tests/append.rs`, and `tests/assign.rs`. I'll try to get to those soon. I'm submitting the comments I have so far, even though I'm not done with the review, because GitHub doesn't reliably preserve pending comments when the code changes.

src/data_repr.rs
```diff
@@ -55,6 +57,35 @@ impl<A> OwnedRepr<A> {
...
+            self.ptr.add(self.len)
+        }
+    }
+
+    /// Reserve `additional` elements
+    /// 
+    /// ## Safety
+    ///
+    /// Note that existing pointers into the data are invalidated
+    pub(crate) unsafe fn reserve(&mut self, additional: usize) {
```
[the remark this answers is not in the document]
bluss: You're right, that sounds good.

src/impl_owned_array.rs
```diff
@@ -59,3 +69,558 @@ where
...
+            if tail_view.ndim() > 1 {
+                sort_axes_in_default_order_tandem(&mut tail_view, &mut array);
+                debug_assert!(tail_view.is_standard_layout(),
+                              "not std layout dim: {:?}, strides: {:?}",
+                              tail_view.shape(), tail_view.strides());
+            } 
+
+            // Keep track of currently filled lenght of `self.data` and update it
+            // on scope exit (panic or loop finish).
+            struct SetLenOnDrop<'a, A: 'a> {
```
jturner314, line 503: What is the purpose of using a guard? (Why not directly update the length of `self.data`?)

src/impl_owned_array.rs
```diff
@@ -59,3 +69,558 @@ where
...
+            let mut data_length_guard = SetLenOnDrop {
+                len: self.data.len(),
+                data: &mut self.data,
+            };
+
+
+            // Safety: tail_view is constructed to have the same shape as array
+            Zip::from(tail_view)
+                .and_unchecked(array)
+                .debug_assert_c_order()
```
jturner314, line 525: This seems risky because it's dependent on the details of `Zip`'s heuristic for choosing iteration order, which could change in the future. I suppose as long as we have this debug assertion and adequate testing it's okay.

src/impl_owned_array.rs
```diff
@@ -59,3 +69,558 @@ where
...
+    sort_axes_impl(&mut a.dim, &mut a.strides, &mut b.dim, &mut b.strides);
+}
+
+fn sort_axes1_impl<D>(adim: &mut D, astrides: &mut D)
+where
+    D: Dimension,
+{
+    debug_assert!(adim.ndim() > 1);
+    debug_assert_eq!(adim.ndim(), astrides.ndim());
+    // bubble sort axes
```
jturner314, line 580: Personally, I'd prefer an indirect sort with the standard library's sorting functions over a custom bubble sort.

src/impl_owned_array.rs
```diff
@@ -59,3 +69,558 @@ where
...
+}
+
+
+fn sort_axes_impl<D>(adim: &mut D, astrides: &mut D, bdim: &mut D, bstrides: &mut D)
+where
+    D: Dimension,
+{
+    debug_assert!(adim.ndim() > 1);
+    debug_assert_eq!(adim.ndim(), bdim.ndim());
+    // bubble sort axes
```
jturner314, line 606: Same here.

src/impl_owned_array.rs
```diff
@@ -59,3 +69,558 @@ where
...
+    debug_assert_eq!(adim.ndim(), astrides.ndim());
+    // bubble sort axes
+    let mut changed = true;
+    while changed {
+        changed = false;
+        for i in 0..adim.ndim() - 1 {
+            let axis_i = i;
+            let next_axis = i + 1;
+
+            // make sure higher stride axes sort before.
```
jturner314, line 588: Fwiw, I'd consider sorting length 1 axes to be the outermost loops, although I doubt it would have much effect either way.

jturner314 (commented):
I'm submitting my comments so far. At this point, I've reviewed everything except for `tests/append.rs` and `tests/assign.rs`.

I think, taken together, the names `append_row`, `append_column`, and `append` are somewhat confusing because `append_row/column` take an array with one less axis than `self`, while `append` takes an array with the same number of axes as `self`. What do you think about changing the names of `append_row/column` to `push_row/column`?

The implementation mostly looks good. I found only a couple of small bugs, and most of the other comments are style/docs. There's a lot of `unsafe` and safety-critical code, but I think it's necessary in order for the implementation to be so flexible with respect preserving the existing array layout.

src/impl_owned_array.rs
```diff
@@ -59,3 +69,558 @@ where
...
+        self.append(Axis(1), column.insert_axis(Axis(1)))
+    }
+}
+
+impl<A, D> Array<A, D>
+    where D: Dimension
+{
+    /// Move all elements from self into `new_array`, which must be of the same shape but
+    /// can have a different memory layout. The destination is overwritten completely.
+    ///
```
jturner314, line 159: We should also mention in the docs that if an element panics on drop, some elements may be leaked. (This can occur in `drop_unreachable_elements_slow`.)

src/impl_owned_array.rs
```diff
@@ -59,3 +69,558 @@ where
...
+    /// ***Panics*** if the shapes don't agree.
+    pub fn move_into(mut self, new_array: ArrayViewMut<MaybeUninit<A>, D>) {
```
jturner314, line 161: I just realized that we could change this signature to

```rust
    pub fn move_into<S2>(mut self, new_array: &mut ArrayBase<S2, D>) {
    where
        S2: AssignElem<A>,
```

which would make it more useful. The primary issue is dealing with panics, especially since `AssignElem` is a safe trait. It's tricky to deal with panics in the implementation without leaking elements on panic or aborting on panic. Even if we used a new `unsafe` trait instead of `AssignElem`, the overwritten elements can panic when they're dropped. I suppose an easy solution would be to move the elements being overwritten into a temporary `Vec`, and then drop the temporary `Vec`, but that seems pretty inefficient.

bluss: Re ArrayViewMut vs &mut ArrayBase: view mut is closer to what we actually need (we convert to view mut unconditionally in the Zip); so "zero-cost" interface would be something like `Into<ArrayViewMut<_>>`. It's only an ergonomic difference and not functional, I believe.

It's right that we only use MaybeUninit here, so we only overwrite, so that we can never have panics in that critical section. I would put supporting anything more as something to be added later.  It is intended that we add a method that makes it easy to convert a view mut of T into a view mut of MaybeUninit<T>, and use this method broadly that way.

bluss: It can feel like we're not making the 100% solution here and maybe that's ok. We grow features bit by bit.

src/impl_owned_array.rs
bluss: Hm right. Maybe that a difference to what drop_in_place on [T] does (I think it attempts to continue dropping?) Good call, I think we can keep doing that, but document it.

src/impl_owned_array.rs
bluss: That's interesting, and also another uncommon but low hanging fruit would be removing dim 1 axes for dyn dim.

bluss: I actually have another WIP on this topic in Baseiter - low hanging fruit is to swap axis `n - 2` and `n - 1`  if `Axis(n - 1)` is length 1.

src/impl_owned_array.rs
bluss: I have the feeling that the bubble sort is a good optimization for code size/perf for the common cases - 1-3 dimensional - but no evidence so far.

src/impl_owned_array.rs
```diff
@@ -59,3 +69,558 @@ where
...
+            let array_memory_head_ptr = self_.ptr.as_ptr();
+            let data_end_ptr = data_ptr.add(data_len);
+            debug_assert!(data_ptr <= array_memory_head_ptr);
+            debug_assert!(array_memory_head_ptr <= data_end_ptr);
+
+            // iter is a raw pointer iterator traversing self_ in its standard order
+            let mut iter = Baseiter::new(self_.ptr.as_ptr(), self_.dim, self_.strides);
+            let mut dropped_elements = 0;
+
+            // The idea is simply this: the iterator will yield the elements of self_ in
```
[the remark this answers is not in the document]
bluss: I want to do that but I didn't find the clean way to do it. I suppose it's about treating every pointer from the iterator as `&[T; N]` where `N` is the length of the last axis, so you jump that whole distance.

src/impl_owned_array.rs
```diff
@@ -59,3 +69,558 @@ where
...
+                self_.invert_axis(Axis(i));
+            }
+        }
+
+        // Sort axes to standard order, Axis(0) has biggest stride and Axis(n - 1) least stride
+        // Note that self_ has holes, so self_ is not C-contiguous
+        sort_axes_in_default_order(&mut self_);
+
+        unsafe {
+            let array_memory_head_ptr = self_.ptr.as_ptr();
```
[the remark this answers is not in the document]
bluss: I'll think of a name. `self.ptr` could also be called the logical head or "head", so this is "memory head", i.e the start of the memory that the array is actually using.  So yeah, we have three heads we need to name, in order:

- start of allocation
- first used address of the array
- address of the logically first array element (at index 0, .. 0)

src/impl_owned_array.rs
bluss: We should try that, as long as benchmarks are still ok. I think this pattern became a habit (in Vec, arrayvec etc). From working around aliasing bugs that should now be gone, now that we have noalias enabled for `&mut` (recent rustc + llvm 12 fix).

src/impl_owned_array.rs
```diff
@@ -59,3 +69,558 @@ where
...
+        unsafe {
+            *self = new_array.assume_init();
+        }
+    }
+
+
+    /// Append an array to the array
+    ///
+    /// The axis-to-append-to `axis` must be the array's "growing axis" for this operation
+    /// to succeed. The growing axis is the outermost or last-visited when elements are visited in
```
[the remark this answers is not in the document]
bluss: oh, thanks

src/impl_owned_array.rs
```diff
@@ -59,3 +69,558 @@ where
...
+    /// Create new_array with the right layout for appending to `growing_axis`
+    #[inline(never)]
+    fn change_to_contig_append_layout(&mut self, growing_axis: Axis) {
+        let ndim = self.ndim();
+        let mut dim = self.raw_dim();
+
+        // The array will be created with 0 (C) or ndim-1 (F) as the biggest stride
+        // axis. Rearrange the shape so that `growing_axis` is the biggest stride axis
+        // afterwards.
+        let prefer_f_layout = growing_axis == Axis(ndim - 1);
```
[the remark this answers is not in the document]
bluss: Is there a reason for preferring the standard layout instead of being based on the current layout?

src/impl_owned_array.rs
bluss: I've already riced this to death in Dimension::_fastest_varying_stride_order, so this method should simply use that :)

src/impl_owned_array.rs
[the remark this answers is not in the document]
jturner314: Right, that's what I was thinking. The "smallest stride of 1" case could be implemented by removing the stride-one axis before creating the iterator, and then using basically the same implementation as the "not smallest stride of 1" case, but change

```rust
                // Next interval will continue one past the current element
                last_ptr = elem_ptr.add(1);
```

to

```rust
                // Next interval will continue one past the current lane
                last_ptr = elem_ptr.add(removed_axis_len);
```

src/impl_owned_array.rs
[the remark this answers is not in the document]
jturner314: > Is there a reason for preferring the standard layout instead of being based on the current layout?

I'm not sure what you mean. With a swap instead of a rotate, the layout still isn't based on the current layout; it's a modification of the standard layout. The reason for preferring a rotate over a swap is that it's fairly likely that the array being appended to `self` will be in standard layout and have length 1 for the "growing axis".

To consider an example, let's define the "axis order" as the axis indices in order of decreasing absolute stride. For example, a 4-D standard layout array would have axis order

```
[0, 1, 2, 3]
```

Now, let's say we want to append to a 4-D array, and we're calling `change_to_contig_append_layout` to create a new owned array with axis 2 as the largest-stride axis.

- If we do a swap, then the strides of the owned array will correspond to the following axis order:

   ```
   [2, 1, 0, 3]
   ```

   so if we append a standard-layout array of shape e.g. `[5, 5, 1, 5]` to the owned array, the order of iteration of the array being appended to the owned array will not be in increasing memory order.

- If we do a rotate, then the strides of the owned array will correspond to the following axis order:

   ```
   [1, 2, 0, 3]
   ```

   so if we append a standard-layout array of shape e.g. `[5, 5, 1, 5]`, the order of iteration will be in increasing memory order for both the owned array and the array being appended.

src/impl_owned_array.rs
jturner314: Ah, yes, I see [a comment in `std`](https://github.com/rust-lang/rust/blob/1.51.0/library/alloc/src/vec/mod.rs#L1988-L1990) which explains this kind of thing:

```rust
            // Use SetLenOnDrop to work around bug where compiler
            // may not realize the store through `ptr` through self.set_len()
            // don't alias.
```

I agree that it would be worth trying without the separate guard to see if the performance is the same.

src/impl_owned_array.rs
jturner314: I don't think the performance of this function really matters much, at least for `append`, since it's called only once, and I'd expect copying the elements to dominate the execution time.

src/impl_owned_array.rs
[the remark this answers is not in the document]
bluss: done :slightly_smiling_face:

src/impl_owned_array.rs
bluss: What I can see, it's still a worthwhile optimization to do this unfortunately. My rustc version is old `rustc 1.53.0-nightly (07e0e2ec2 2021-03-24)` but still, then current stable is likely affected and the optimization can stay.

What I see - using the benches/append.rs benchmarks, is that the inner loop doesn't optimize very well without the drop guard.

src/impl_owned_array.rs
jturner314: > I actually have another WIP on this topic in Baseiter - low hanging fruit is to swap axis `n - 2` and `n - 1` if `Axis(n - 1)` is length 1.

A somewhat more general improvement is to merge axes, when possible, into axis `n-1`, with something like this:

```rust
let mut take = n-2;
loop {
    let merged = arr.merge_axes(Axis(take), Axis(n-1));
    if !merged || take == 0 {
        break;
    } else {
        take -= 1;
    }
}
```

For example, in addition to cases where axis `n-1` has length 1, this helps cases where a standard layout array has been sliced with a step larger than 1 for the outermost axis.

> With that said, the goal of the axis sorting here is iteration in memory order, not performance.

Right, it probably doesn't matter much here.

src/impl_owned_array.rs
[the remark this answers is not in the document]
jturner314: That's a nice way to handle the axis "removal" to avoid code duplication for the two cases. :)

src/impl_owned_array.rs
```diff
@@ -59,3 +69,558 @@ where
...
+        // Note that self_ has holes, so self_ is not C-contiguous
+        sort_axes_in_default_order(&mut self_);
+
+        unsafe {
+            let array_memory_head_ptr = self_.ptr.as_ptr();
+            let data_end_ptr = data_ptr.add(data_len);
+            debug_assert!(data_ptr <= array_memory_head_ptr);
+            debug_assert!(array_memory_head_ptr <= data_end_ptr);
+
+            // iter is a raw pointer iterator traversing self_ in its standard order
```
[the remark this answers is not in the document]
bluss: Thanks, will incorporate with rewording

src/impl_owned_array.rs
jturner314: Okay. That is unfortunate. I hope this will become unnecessary in the future with compiler improvements.

jturner314: It would be good to add a comment saying that the guard is a performance optimization.

src/impl_owned_array.rs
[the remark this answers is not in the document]
bluss: Thanks, now I totally see what you mean. That gives a good chance that the `Zip` will use a contiguous order.

src/impl_owned_array.rs
bluss: I think I managed to write the same as that in #979

jturner314: I like your implementation better. :)

src/impl_owned_array.rs
bluss: I'm thinking that unfortunately append could be 5 elements or 5 million depending on the user. I don't mind keeping the simplest possible sort around - anything that sorts the dimensions in-place seems superior.

src/impl_owned_array.rs
bluss: I'm incorporating that, but it seems so general and expected that maybe it doesn't need mentioning? It is nice to be so overly specific when we can, maybe.

bluss:
Thanks a lot for the comments and bugfixes :slightly_smiling_face:.

I don't know about append versus push, it's quite arbitrary to me.

I looked into std collections and they are consistent with append always being something like `Vec::append(&mut self, &mut Vec<_>)`; where append moves ownership from another `&mut Self` into the receiver.

My only thought is that it's not inconceivable to generalize our `.append(arrayview)` to take either an arrayview (and Clone) or an Array (and move elements), but it's a bit much to do as an extra project inside this PR.

src/impl_owned_array.rs
bluss: As a note the benchmark `select_axis0` copies all of 256 elements of f32 but they are contiguous - that copying is so fast that this benchmark is quite sensitive to performance improvements in all the checking and dimension munging code around the copying, for example the improvement in 7564ad29109f7d8c77f875bb65aaac8ff7e7ffc2 was significant.

src/data_repr.rs
```diff
@@ -55,6 +57,35 @@ impl<A> OwnedRepr<A> {
         self.ptr
     }
 
+    /// Return end pointer
+    pub(crate) fn as_end_nonnull(&self) -> NonNull<A> {
```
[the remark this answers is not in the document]
bluss: I'll leave this for now, I forgot about it, that's mostly why. And that the as_ convention is weirdly preferring "as_" for pointer accessors, so that's what a couple of our methods have. I don't even like `.as_ptr()` - it should normally be just `.ptr()` etc.

bluss:
Merging to finish this work. Renaming can still be done in follow-ups.

gussmith23:
this is awesome! really helpful to have clone supported on these methods

bluss:
I'm glad you share my enthusiasm. :slightly_smiling_face: I'm happy to finally have arrays that can be appended to.

jturner314:
I'm really looking forward to this functionality too. :tada: There are a number of places where it'll simplify my code.

To clarify my comment about the naming question -- I think that if we name the row/column methods and the general methods similarly, then the general method should also take `D::Smaller` for consistency:

```rust
// On 2-D arrays
fn append_row(&mut self, row: ArrayView<A, D::Smaller>) {}
fn append_column(&mut self, column: ArrayView<A, D::Smaller>) {}

// On n-D arrays
fn the_general_method(&mut self, axis: Axis, array: ArrayView<A, D::Smaller>) {}
```

I agree that it's helpful to name the methods similarly to `Vec`'s methods. IMO, it makes sense for the axis being appended to be thought of analogously to `Vec`. Conceptually, you could think of appending to axis 1 of an `Array3` of shape [2, 3, 4] as being analogous to appending to a `Vec<Array2>` where the `Vec` has length 3 and the `Array2`s have shape [2, 4]. With this analogy, adding a single subview (of dimension `D::Smaller`) to the array is like adding a single element to the `Vec` with `push`, while adding a view of dimension `D` (which is similar to `&[ArrayView<A, D::Smaller>]`) is like adding a sequence of elements to a `Vec` with `append`/`extend_from_slice`.

So, I'd actually suggest four methods:

```rust
// On 2-D arrays
fn push_row(&mut self, row: ArrayView<A, D::Smaller>) {}
fn push_column(&mut self, column: ArrayView<A, D::Smaller>) {}

// On n-D arrays
fn push(&mut self, axis: Axis, array: ArrayView<A, D::Smaller>) {}
fn append(&mut self, axis: Axis, array: ArrayView<A, D>) {} // or "extend_from_view"
```

The two separate methods on n-D arrays are both useful. `push` is useful for building arrays by stacking subviews one-by-one, while `append` is useful for building arrays from larger chunks. For my personal use, `push` is actually more useful than `append`, because I'm usually performing some calculations in a loop and want to stack the results into an array along a new dimension corresponding to the loop.

The primary way in which this analogy isn't quite right is that `ndarray`'s methods take views, while `Vec`'s `push` and `append` methods take owned values, but I think that's less important with respect to naming than the number of dimensions. If we add support for moving owned values in the future, we could either modify the existing methods to take an argument with `S2: Data, A: Clone` (like `into_owned` does), or we could rename the existing methods which take views to something like `push_view`/`push_cloned`/`push_from_view`/etc.

bluss:
That sounds good. I think you have convinced me.

PR #940, opened 2021-03-15
Simplifications for slicing-related types

jturner314:
I realized that now that we no longer have a slicing API based on the `Dimension::SliceArg` associated type, the `s![]` macro can return an owned type, the tricky conversions involving pointer casts no longer need to exist, and `SliceInfo` no longer needs to be `?Sized`.

This PR does the following:

- Implement `SliceArg<D> for &T`.
- Change the slicing methods to take `I` rather than `&I`.
- Make the `s![]` macro return an owned `SliceInfo` rather than a reference.
- Replace the conversions of `SliceInfo` involving tricky pointer casts with similar conversions not involving pointer casts.
- Make `SliceInfo` be `Sized` and remove the `#[repr(transparent)]` attribute. These changes aren't necessary, but seem like a nice simplification.


bluss:
Nice!

PR #939, opened 2021-03-13
Update release notes in preparation for 0.15.0

bluss:
The list of new features and improvements is pretty large for this release, this is looking good.

[Rendered preview](https://github.com/rust-ndarray/ndarray/blob/doc-0.15/RELEASES.md)

PR #943, opened 2021-03-15
Check expr type for expr;step in s![] macro at compile time, and remove AxisSliceInfo::step_by

jturner314:
Before, if the `expr;step` notation was used with an index as the `expr`, e.g. `s![4;2]`, the macro would panic at runtime. Now, it errors at compile time. This is accomplished by converting the `expr` to a `Slice` before converting it into the final `AxisSliceInfo`.

The `AxisSliceInfo::step_by` method is awkward because it panics for all cases except the `Slice` variant. This method is no longer necessary because the `s![]` macro now calls `Slice::step_by()` instead.

### DIFF
--- a/blas-tests/tests/oper.rs
+++ b/blas-tests/tests/oper.rs
@@ -6,7 +6,7 @@ extern crate num_traits;
 use ndarray::linalg::general_mat_mul;
 use ndarray::linalg::general_mat_vec_mul;
 use ndarray::prelude::*;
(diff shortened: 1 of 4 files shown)


bluss (approved):

bluss:
Thanks. Would you like to describe your recent PRs in the change log?

bluss:
We can both push to https://github.com/rust-ndarray/ndarray/pull/939 as long as it's open for example. Or just open up for PRs to add to the changelog by themselves.

PR #946, opened 2021-03-17
Add "forwards" of the approx methods .abs_diff_eq and .relative_eq

### LINKED ISSUE #763: Should `ArrayBase::all_close` be un-deprecated?
Apparently ArrayBase::all_close() has been deprecated in favor of approx crate and feature. 
I think this new method brings unnecessary complications to the workflow i.e. extra crate must be included and feature must be enabled by hand in Cargo.toml. So my question is: Is the deprecation really worth the benefits it brings? Or why is it deprecated at all?

bluss:
This is a nod to the all_close method that was removed. The replacement
methods still require the "approx" feature gate, but we forward them as
inherent methods on the array, to make them easier to call (no trait
import needed).

Closes #763

### DIFF
--- a/src/array_approx.rs
+++ b/src/array_approx.rs
@@ -21,9 +21,10 @@ where
         if self.shape() != other.shape() {
             return false;
         }
+
         Zip::from(self)
             .and(other)
-            .all(|a, b| A::abs_diff_eq(a, b, epsilon.clone()))
+            .all(move |a, b| A::abs_diff_eq(a, b, epsilon.clone()))
     }
 }
 
@@ -49,9 +50,10 @@ where
         if self.shape() != other.shape() {
             return false;
         }
+
         Zip::from(self)
             .and(other)
-            .all(|a, b| A::relative_eq(a, b, epsilon.clone(), max_relative.clone()))
+            .all(move |a, b| A::relative_eq(a, b, epsilon.clone(), max_relative.clone()))
     }
 }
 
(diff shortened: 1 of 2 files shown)


src/array_approx.rs
```diff
@@ -72,12 +74,51 @@ where
...
+impl<A, S, D> ArrayBase<S, D>
+where
+    S: Data<Elem = A>,
+    D: Dimension,
+{
+    /// A test for equality that uses the elementwise absolute difference to compute the
+    /// approximate equality of two arrays.
+    ///
+    /// **Requires crate feature `"approx"`**
+    pub fn abs_diff_eq<B, S2>(&self, other: &ArrayBase<S2, D>, epsilon: A::Epsilon) -> bool
```
jturner314, line 93: IMO, this would be better with a single type parameter for simplicity:

```rust
pub fn abs_diff_eq<S2>(&self, other: &ArrayBase<S2, D>, epsilon: A::Epsilon) -> bool
where
    A: AbsDiffEq<S2::Elem>,
    A::Epsilon: Clone,
    S2: Data,
{}
```

src/array_approx.rs
```diff
@@ -72,12 +74,51 @@ where
...
+        S2: Data<Elem = B>,
+    {
+        <Self as AbsDiffEq<_>>::abs_diff_eq(self, other, epsilon)
+    }
+
+    /// A test for equality that uses an elementwise relative comparison if the values are far
+    /// apart; and the absolute difference otherwise.
+    ///
+    /// **Requires crate feature `"approx"`**
+    pub fn relative_eq<B, S2>(
```
jturner314, line 106: Same here.

jturner314 (commented):
This is a good idea. I just had a small comment regarding the type parameters of the methods.

There are a few places in doc tests (e.g. for `logspace` and `mapv_inplace`) which `use approx`, which could optionally be replaced with these methods, if desired.

bluss:
Thanks for the input :slightly_smiling_face:

PR #947, opened 2021-03-17
Add new method .assign_to()

bluss:
This method does the same job as Zip::from(a).map_assign_into(f, b) with
cloning as the function f.

Comparison with existing assign:

- Reverse argument order: source, destination
- The new method doesn't broadcast.
- The new method is generic over AssignElem (assign into &Cell, &mut
ManuallyUninit etc).


bluss:
Unsure if these methods pull their weight (they add nothing new, but are expressive because they are easy to use). Their properties contrast with the existing `.assign()`.

Best we can do for existing .assign is to generalize it so that it can assign to both "T" and "ManuallyUninit<T>" arrays.  The assign_into solution additionally handles "Cell<T>" arrays.

src/impl_methods.rs
```diff
@@ -2049,6 +2050,41 @@ where
...
+    }
+
+    /// Perform an elementwise assigment of values mapped with `f` from `self` into array or
+    /// producer `to`.
+    ///
+    /// `to` can be another array or a producer of assignable elements.
+    /// [`AssignElem`] determines how elements are assigned.
+    ///
+    /// **Panics** if shapes disagree.
+    pub fn map_assign_into<'a, F, B, P>(&'a self, f: F, to: P)
```
bluss, line 2076: Maybe skip exposing this as a public method, it's not as used.

jturner314: Yeah, I'm not sure this method is necessary. The parameters are also a little awkward, because while I think the order `f: F, to: P` better fits the name of the method, it's more common for Rust methods to put the closure as the last parameter since closures are often formatted as multiple lines. (Consider e.g. `Result::map_or`, which IMO would better fit the name as `map_or<U, F>(self, f: F, default: U)` but is actually `map_or<U, F>(self, default: U, f: F)` in the standard library.)

jturner314 (commented):
I do think it's worth having a simple way to assign to arrays of `MaybeUninit`, although as you pointed out, this could be handled with the existing `assign`. I'm not opposed to an `assign_to` method, though, because it's a simple addition, and the swapped parameter order could be useful in some cases. The fact that it works with arrays of `Cell`/`MathCell` is a bonus, although I'm unclear on the typical use cases for arrays of `Cell`/`MathCell`.

src/impl_methods.rs
```diff
@@ -2049,6 +2050,41 @@ where
         self.zip_mut_with(rhs, |x, y| *x = y.clone());
     }
 
+    /// Perform an elementwise assigment of values from `self` into array or producer `to`.
+    ///
+    /// `to` can be another array or a producer of assignable elements.
+    /// [`AssignElem`] determines how elements are assigned.
+    ///
+    /// **Panics** if shapes disagree.
+    pub fn assign_into<P>(&self, to: P)
```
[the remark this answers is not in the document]
bluss: Yes, I guess it's better to stick to that convention, even though it's not always clear cut.

PR #948, opened 2021-03-22
Fix a few inconsistencies around negative strides in constructors

### LINKED ISSUE #942: Fix inconsistencies around negative strides in constructors
For 0.15 we need to fix some of the bugs that SparrowLii found in #901

bluss:
The ArrayView from_shape constructors that allow and have a natural interpretation for the array head pointer, now support negative strides properly.

Note that the raw view and raw pointer constructors for regular views don't support negative strides yet. There are assertions to check this. The goal of this PR is to have consistency for the next release, not to add new features for negative strides.

The first commit is by @SparrowLii (some changes dropped from that commit) from PR #901.

Closes #942


bluss:
This is sort of a follow-up to #885

PR #951, opened 2021-03-25
Drop direct blas-src dependency, update docs for blas integration

### LINKED ISSUE #950: Versioning confusion in Readme
Dear all,

Tl;dr: In the Readme.md dependencies-example, there is ndarray 0.14 together with blas-src 0.7, while according to the table below it should actually be 0.14 with 0.6.1 (or 0.15 with 0.7).

I hoped it is ok that I opened an issue for such a (seemingly) minor problem. But here's the thing: I was trying to get a simple program running, basically only matrix multiplication:

```rust
extern crate ndarray;
use ndarray::prelude::*;

fn main() {
	let	a_id_mat: Array2<f64> = 2. * Array::eye(3);

	let x = array![[1.,2.,3.]];
	let x = x.t();
	let y = a_id_mat.dot(&x);
	
	println!("x={}",x);
	println!("y={}",y);
}
```
with a simple Cargo-file, as suggested in the Readme:
```
[package]
name = "nda_mm"
version = "0.1.0"
authors = ["M. M."]
edition = "2018"

[dependencies]
ndarray = { version = "0.14.0", features = ["blas"] }
blas-src = { version = "0.7", default-features = false, features = ["openblas"] }
openblas-src = { version = "0.9", default-features = false, features = ["cblas", "system"] }
```
As warned in the Readme, this won't link correctly (basically, I get `undefined reference to cblas_dgemm`), at least on Debian stable.

(Actually, I would like to ask how the colliding versions are handled with ndarray-linalg, but that's maybe not the right place).

Kind regards and thanks for creating ndarray :)

bluss:
We can in theory unchain ourselves from the blas-src version and let the
user be responsible for linking in blas in the final product.

Update xtest-blas to test this configuration.

Update instructions in readme for how to use it.

Remove ndarray build.rs file since it was only ever used for testing (and now not used anymore!)
Update xtest-numeric so that it actually tests with/without blas. Now there's a bit of
overlap between xtest-numeric and xtest-blas.

It happens that this fixes #950


bluss:
This came up because I was waiting for a blas-src release before pulling the trigger on ndarray 0.15. But we're doing it wrong already, we should just depend on the interface (cblas-sys), and let the end user crate supply the implementation.

We also should port ourselves to blas-sys when possible so that we are compatible all backend configurations.

src/lib.rs
```diff
@@ -124,8 +124,6 @@ extern crate std;
 #[cfg(not(feature = "std"))]
 extern crate core as std;
 
-#[cfg(feature = "blas")]
-extern crate blas_src;
```
bluss, line 128: This was the main feature in the old solution - when we don't link this, the end user has to.

PR #956, opened 2021-03-27
Elaborate doc for Axis and fix an example for Axes

### LINKED ISSUE #564: Document reason for strongly typed `Axis`?
While using `ndarray`, I was wondering why certain methods use the `Axis` type instead of just `usize`, especially given that `Axis` is just a simple wrapper around `usize`.  I tried looking into the documentation, and could not find any explanation; and searching through past issues/pull requests, I found just #91, #96 and #97 (though none of them given a clear reason why this is the case).

Would it be worth documenting this choice?  

And is there a reason why `From<usize>`/`Into<Axis>` isn't implemented for `Axis`/`usize`?

bluss:
Follow-up to deprecation of AxisDescription methods #915.
Also includes a code style change to `is_square`.

Fixes #564

### DIFF
--- a/src/dimension/axes.rs
+++ b/src/dimension/axes.rs
@@ -1,7 +1,7 @@
 use crate::{Axis, Dimension, Ix, Ixs};
 
 /// Create a new Axes iterator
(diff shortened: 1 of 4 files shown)


bluss:
Failing clippy was fixed in a different PR.

PR #962, opened 2021-03-29
Shape info in error

bluss:
Description forthcoming


bluss:
I guess I needed to experiment with this no-alloc approach. This is a good starting point for also experimenting with an error in a Box, which also does fine on microbenchmarks. Maybe it can be optional.

bluss:
After exploring this, I'm ready to try adding something like `ShapeError(Box<ErrorDetails>)` to pack in more information. Thoughts?  What I don't like is the code bloat that results from cleanup pads (code to drop the extra box in various places). It could even be made optional, to have such error details.

Issue #936, opened 2021-03-12
Fixes and follow-ups to two-sided broadcasting in arithmethic ops


bluss:
In #898 we left the following "known issues" open that we can fix **after** that PR's review and merge:

- [ ] **Rename** BroadcastShape to `DimMax` - we also move the broadcast method out of the trait (it can just be a free function)
- [ ] It is a wish that the `D: Dimension + BroadcastShape<E>` bound in arith ops becomes just `D: Dimension`, i.e. the DimMax with associated type becomes implied by `D: Dimension`. At this point it is not certain how this can be done.
- [ ] Preserve the allocation of the input array in arithmetic ops in more cases (when only the left or right side needs broadcast)
- [ ] Improve performance of "no-broadcasting-needed" scenario in &array + &array operation
- [ ] Potentially provide public function or method similar to `broadcast_with` - that does the broadcasting of two arrays together.

The issues marked in bold should be solved before the next release.

SparrowLii:
[Specialization](https://github.com/rust-lang/rust/issues/31844) can solve DimMax's problems:

    impl<D1: Dimension, D2: Dimension> DimMax<D2> for D1 {
        default type Output = D1;
    }

But it seems that there are still many problems to be solved before stability.

PR #963, opened 2021-03-30
Preserve the allocation of the input array  in arithmetic ops

SparrowLii:
Update #936 
Preserve the allocation of the input array whose data is DataOwned in arithmetic ops, if its shape stays the same.


bluss:
Nice, thanks!

PR #965, opened 2021-04-01
Improve performance of "no-broadcasting-needed" scenario in &array +&array operation

SparrowLii:
This PR does the following:
1. Add a method `to_dimensionality` which creats an array view from an array with the same shape, but different dimensionality type. 
2. Use `to_dimensionality` to avoid unnecessary  calls of `broadcast` in &array + &array operation.

Updates #936

### DIFF
--- a/src/impl_methods.rs
+++ b/src/impl_methods.rs
@@ -1805,35 +1805,20 @@ where
     {
         let shape = co_broadcast::<D, E, <D as DimMax<E>>::Output>(&self.dim, &other.dim)?;
         let view1 = if shape.slice() == self.dim.slice() {
-            self.to_dimensionality::<<D as DimMax<E>>::Output>()
+            self.view().into_dimensionality::<<D as DimMax<E>>::Output>().unwrap()
+        } else if let Some(view1) = self.broadcast(shape.clone()) {
+            view1
         } else {
-            self.broadcast(shape.clone())
+            return Err(from_kind(ErrorKind::IncompatibleShape))
         };
         let view2 = if shape.slice() == other.dim.slice() {
-            other.to_dimensionality::<<D as DimMax<E>>::Output>()
+            other.view().into_dimensionality::<<D as DimMax<E>>::Output>().unwrap()
+        } else if let Some(view2) = other.broadcast(shape) {
+            view2
         } else {
-            other.broadcast(shape)
+            return Err(from_kind(ErrorKind::IncompatibleShape))
         };
-        if let Some(view1) = view1 {
-            if let Some(view2) = view2 {
-                return Ok((view1, view2));
-            }
-        }
-        Err(from_kind(ErrorKind::IncompatibleShape))
-    }
-
-    /// Creat an array view from an array with the same shape, but different dimensionality
-    /// type. Return None if the numbers of axes mismatch.
-    #[inline]
-    pub(crate) fn to_dimensionality<D2>(&self) -> Option<ArrayView<'_, A, D2>>
-    where
-        D2: Dimension,
-        S: Data,
-    {
-        let dim = <D2>::from_dimension(&self.dim)?;
-        let strides = <D2>::from_dimension(&self.strides)?;
-
-        unsafe { Some(ArrayView::new(self.ptr, dim, strides)) }
+        Ok((view1, view2))
     }
 
     /// Swap axes `ax` and `bx`.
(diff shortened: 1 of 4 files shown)


src/impl_ops.rs
```diff
@@ -179,7 +179,13 @@ where
 {
     type Output = Array<A, <D as DimMax<E>>::Output>;
     fn $mth(self, rhs: &'a ArrayBase<S2, E>) -> Self::Output {
-        let (lhs, rhs) = self.broadcast_with(rhs).unwrap();
+        let (lhs, rhs) = if self.ndim() == rhs.ndim() && self.shape() == rhs.shape() {
+            let lhs = self.to_dimensionality::<<D as DimMax<E>>::Output>().unwrap();
```
bluss, line 183: These to_dimensionality calls seem redundant with the ones already in `broadcast_with`. Is there a reason this PR shouldn't just do one or the other, not both?

If it was just the calls in this method, here it looks like `self.view().into_dimensionality::<...>()` is enough, i.e. using the existing method.

src/impl_methods.rs
```diff
@@ -1804,14 +1804,38 @@ where
...
                 return Ok((view1, view2));
             }
         }
         Err(from_kind(ErrorKind::IncompatibleShape))
     }
 
+    /// Creat an array view from an array with the same shape, but different dimensionality
+    /// type. Return None if the numbers of axes mismatch.
+    #[inline]
+    pub(crate) fn to_dimensionality<D2>(&self) -> Option<ArrayView<'_, A, D2>>
```
bluss, line 1828: If this method is added, it should be just next to (after) `into_dimensionality` in this file.

bluss:
Is there a benchmark for this case, and does it show an improvement? Benchmarks for non-dyn dimensions would be the most interesting. :slightly_smiling_face:

src/impl_ops.rs
SparrowLii: That's right. We should use view().into_dimension() here. It is faster in the benchmark test (though I don’t know why)

SparrowLii: Adding the same shape detection in impl_ops can avoid executing co_broadcast. I think it has reserved value.

SparrowLii:
I did benchmark tests in two cases: 1.no need broadcasting 2. only one side need broadcasting. The result shows that view().into_dimensionality() is faster than to_dimensionality(). And judging whether broadcasting is not needed before broadcast_with can significantly increase the speed.
The following is the result of the benchmark：

    no need broadcast
    (ns/iter)
    origin:
    441 430 434 436 453
    broadcast_with using view().into_dimensionality:
    406 411 410 410 404
    broadcast_with using to_dimensionality:
    423 404 420 410 402
    no broadcast_with and use view().into_dimensionality():
    384 388 392 380 386
    no broadcast_with and use to_dimensionality():
    389 394 399 386 390
    
    one side broadcast
    (ns/iter)
    origin:
    477 473 468 491 479
    broadcast_with using view().into_dimensionality:
    457 456 452 458 454
    broadcast_with using to_dimensionality:
    463 463 456 458 473

bluss:
Is the "origin" benchmark from before this PR, or with changes in this PR? Only the changes from before this PR to after it are interesting :)

SparrowLii:
> Is the "origin" benchmark from before this PR, or with changes in this PR? It's mostly the changes from before this PR to after it that are interesting :)

It is from before this PR

src/impl_ops.rs
```diff
@@ -179,7 +179,13 @@ where
...
     fn $mth(self, rhs: &'a ArrayBase<S2, E>) -> Self::Output {
-        let (lhs, rhs) = self.broadcast_with(rhs).unwrap();
+        let (lhs, rhs) = if self.ndim() == rhs.ndim() && self.shape() == rhs.shape() {
+            let lhs = self.view().into_dimensionality::<<D as DimMax<E>>::Output>().unwrap();
+            let rhs = rhs.view().into_dimensionality::<<D as DimMax<E>>::Output>().unwrap();
+            (lhs, rhs)
+        } else {
+            self.broadcast_with(rhs).unwrap()
+        };
         Zip::from(&lhs).and(&rhs).map_collect(clone_opf(A::$mth))
```
bluss, line 189:
```suggestion
        Zip::from(lhs).and(rhs).map_collect(clone_opf(A::$mth))
```

I think here we should just consume the views, saves more redundant view creations in Zip (won't be a very noticeable change, maybe the compiler can remove the difference).

SparrowLii: All right. It has been correct.

bluss: *improved. Cool.

bluss:
Thanks!

PR #967, opened 2021-04-04
Add method .remove_index(axis, index)

### LINKED ISSUE #931: Method to remove index along axis
A user on #rust-sci asked the best way to remove an element from an `Array1`. With current `ndarray`, if it's known to be contiguous (and with no extra elements in the `Vec` before/after the array data), probably the best way to do this is to convert to a `Vec` with `.into_raw_vec()`, remove the element with `.remove()`, and then convert the `Vec` back to an `Array1`. If the array is possibly discontiguous, it's simpler to allocate a new `Array1` and copy the data into it with slicing and `.assign()`.

It would be nice to provide a `remove_index_axis` method to provide this functionality for n-dimensional arrays.

Ideally, the implementation would reuse the existing allocation and just shift the existing elements.

- For the "remove index along outermost axis for a contiguous array" special case, it would probably be best to use `std::ptr::copy` to shift the elements (i.e. implement it like `Vec::remove`). For `Array`, we'd need to be careful about how to drop the removed data and adjust the length of the underlying `Vec`. If we want to implement `remove_index_axis` for `ArrayViewMut`, we'd have to consider what to place in the last subview along the axis (before reducing the length of the axis in the view) so that we aren't illegally duplicating the elements in the last subview. If we want to move the removed data into that location, we'd need to store it in a temporary allocation before shifting the other elements with `std::ptr::copy`. We'd also need to be careful to avoid panics while in an invalid state.

- For the general case, I'd suggest using something like `accumulate_axis_inplace` with `Clone` to copy the elements. In principle, we could implement the general case without `Clone`, but this would have the same safety challenges as using `std::ptr::copy` for the special case, while being more complex to implement.

For a first implementation, just implementing the general case would be useful, even if it isn't the most efficient approach in all cases.

bluss:
Add method to remove the `index`th entry along all lanes of `axis`.
The method doesn't change the number of initialized elements, just like slicing in place works
(this could be confusing, w.r.t remove semantics).

Fixes #931

### DIFF
--- a/src/impl_methods.rs
+++ b/src/impl_methods.rs
@@ -26,6 +26,7 @@ use crate::dimension::broadcast::co_broadcast;
 use crate::error::{self, ErrorKind, ShapeError, from_kind};
 use crate::math_cell::MathCell;
 use crate::itertools::zip;
+use crate::low_level_util::AbortIfPanic;
 use crate::zip::{IntoNdProducer, Zip};
 use crate::AxisDescription;
 
@@ -2476,6 +2477,7 @@ where
             //      dst = elt;
             //  }
             //
+            let guard = AbortIfPanic(&"remove_index: temporarily moving out of owned value");
             let mut slot = MaybeUninit::<A>::uninit();
             unsafe {
                 slot.as_mut_ptr().copy_from_nonoverlapping(dst, 1);
@@ -2485,6 +2487,7 @@ where
                 }
                 (dst as *mut A).copy_from_nonoverlapping(slot.as_ptr(), 1);
             }
+            guard.defuse();
         });
         // then slice the axis in place to cut out the removed final element
         self.slice_axis_inplace(axis, Slice::new(0, Some(-1), 1));
--- a/src/impl_methods.rs
+++ b/src/impl_methods.rs
@@ -6,7 +6,7 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
-use std::mem::{size_of, ManuallyDrop};
+use std::mem::{size_of, ManuallyDrop, MaybeUninit};
 use alloc::slice;
 use alloc::vec;
 use alloc::vec::Vec;
@@ -2460,9 +2460,22 @@ where
             let mut lane_iter = lane.iter_mut();
             let mut dst = if let Some(dst) = lane_iter.next() { dst } else { return };
 
-            for elt in lane_iter {
-                std::mem::swap(dst, elt);
-                dst = elt;
+            // Logically we do a circular swap here, all elements in a chain
+            // Using MaybeUninit to avoid unecessary writes in the safe swap solution
+            //
+            //  for elt in lane_iter {
+            //      std::mem::swap(dst, elt);
+            //      dst = elt;
+            //  }
+            //
+            let mut slot = MaybeUninit::<A>::uninit();
+            unsafe {
+                slot.as_mut_ptr().copy_from_nonoverlapping(dst, 1);
+                for elt in lane_iter {
+                    (dst as *mut A).copy_from_nonoverlapping(elt, 1);
+                    dst = elt;
+                }
+                (dst as *mut A).copy_from_nonoverlapping(slot.as_ptr(), 1);
             }
         });
         // then slice the axis in place to cut out the removed final element
--- a/src/impl_methods.rs
+++ b/src/impl_methods.rs
@@ -2449,13 +2449,13 @@ where
     /// Decreases the length of `axis` by one.
     ///
     /// ***Panics** if `axis` or `index` is out of bounds.
-    pub fn shift_remove_index(&mut self, axis: Axis, index: usize)
+    pub fn remove_index(&mut self, axis: Axis, index: usize)
     where
         S: DataOwned + DataMut,
     {
         let (_, mut tail) = self.view_mut().split_at(axis, index);
-        // shift elements to the back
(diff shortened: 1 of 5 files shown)


bluss:
This is my approach. Safe version (commit 1) and optimized (commit 2) are not so far apart, and they should be pretty good and remarkably simple as a first approach. They don't change the number of initialized elements, just like slicing in place works.

src/impl_methods.rs
```diff
@@ -2444,6 +2444,46 @@ where
...
+            //
+            //  for elt in rest.iter_mut() {
+            //      std::mem::swap(dst, elt);
+            //      dst = elt;
+            //  }
+            //
+            let mut slot = MaybeUninit::<A>::uninit();
+            unsafe {
+                slot.as_mut_ptr().copy_from_nonoverlapping(dst, 1);
+                for elt in lane_iter {
```
jturner314, line 2476: If `lane_iter.next()` panics, then an element will be dropped twice, and, less importantly, the element in `slot` won't be dropped. I don't think the current implementation can panic, but this still makes me uncomfortable. I'd feel more comfortable with getting a raw pointer to the start of the lane, the stride for the axis, and then `.offset()`ting the pointer by the stride on each iteration, because this would be easier to audit for potential panics.

src/impl_methods.rs
```diff
@@ -2444,6 +2444,46 @@ where
...
+    ///
+    /// ***Panics** if `axis` or `index` is out of bounds.
+    pub fn shift_remove_index(&mut self, axis: Axis, index: usize)
+    where
+        S: DataOwned + DataMut,
+    {
+        // TODO: It's possible to choose to shift the elment to the front or the back here,
+        // whichever is closer.
+        let (_, mut tail) = self.view_mut().split_at(axis, index);
+        // shift elements to the back
```
jturner314, line 2459:
```suggestion
        // shift elements to the front
```

src/impl_methods.rs
```diff
@@ -2444,6 +2444,46 @@ where
...
+        // use swap to keep all elements initialized (as required by owned storage)
```
jturner314, line 2460:
```suggestion
        // use swapping to keep all elements initialized (as required by owned storage)
```

src/impl_methods.rs
jturner314, line 2476: IMO, `src` would be a clearer name than `elt`.

src/impl_methods.rs
```diff
@@ -2444,6 +2444,46 @@ where
...
+    /// Decreases the length of `axis` by one.
+    ///
+    /// ***Panics** if `axis` or `index` is out of bounds.
+    pub fn shift_remove_index(&mut self, axis: Axis, index: usize)
+    where
+        S: DataOwned + DataMut,
+    {
+        // TODO: It's possible to choose to shift the elment to the front or the back here,
+        // whichever is closer.
+        let (_, mut tail) = self.view_mut().split_at(axis, index);
```
jturner314, line 2458: I think we should also check that `index < self.len_of(axis)`. The `split_at` method allows an `index` equal to the length of the axis, in which case the upper portion of the split has zero length. I don't think it makes sense to allow this case, because the index wouldn't really be *removed* -- it wasn't a valid index in the first place.

src/impl_methods.rs
```diff
@@ -2444,6 +2444,46 @@ where
...
+        // shift elements to the back
+        // use swap to keep all elements initialized (as required by owned storage)
+        Zip::from(tail.lanes_mut(axis)).for_each(|mut lane| {
+            let mut lane_iter = lane.iter_mut();
+            let mut dst = if let Some(dst) = lane_iter.next() { dst } else { return };
```
jturner314, line 2463: If we panic on the `index == self.len_of(axis)` case as I suggested above, the `None` case cannot occur.

src/impl_methods.rs
```diff
@@ -2444,6 +2444,46 @@ where
         }
     }
 
+    /// Remove the `index`th elements along `axis` and shift down elements from higher indexes.
+    ///
+    /// Decreases the length of `axis` by one.
+    ///
+    /// ***Panics** if `axis` or `index` is out of bounds.
+    pub fn shift_remove_index(&mut self, axis: Axis, index: usize)
```
jturner314, line 2452:
```suggestion
    pub fn remove_index(&mut self, axis: Axis, index: usize)
```

IMO, `shift_` doesn't need to be part of the name.

jturner314 (commented):
Swapping along lanes like this is a nice solution, and the temporary variable to avoid unnecessary writes is nice, too.

src/impl_methods.rs
bluss: Yes I know this is a no-panic section and it should be marked. I was comfortable apparently, but I had a thought to maybe leave an "abort on panic" guard here to make it entirely visible.

As for a simpler iteration, in that case we should implement a special 1D iterator that does just that (even though Zip is already doing about the same job). I don't want to write out a raw pointer loop, ndarray probably needs a bit more use of higher level abstractions, not less 🙂

src/impl_methods.rs
bluss: Yes. Type system wise the existing code is neat enough, don't know if anything else is worthwhile.

src/impl_methods.rs
bluss: Hm matter of taste that I haven't addressed

PR #971, opened 2021-04-06
Add link to the parallel module in top-level.

### LINKED ISSUE #964: Make ArrayView(Mut)'s parallel iterators implement IndexedParallelIterator
Axis iterators can be used as IndexedParallelIterator, but it is not the case for ArrayBase, as shown below:
```rust
use rayon::prelude::*;
fn main() {
    let a = ndarray::Array1::<u32>::zeros(5);
    test_into_par_iter(&a);
    test_into_par_iter_indexed(&a); // <- Does not compile.
}
fn test_into_par_iter<T: IntoParallelIterator>(_x: T) {}
fn test_into_par_iter_indexed<T: IntoParallelIterator>(_x: T)
where
    T::Iter: IndexedParallelIterator,
{
}
```
From an outsider perspective, such an implementation appears to be possible. Is it intentionally omitted, or could it be added ?

cassiersg:
Fixes #964

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -80,7 +80,7 @@
 //! - `serde`
 //!   - Enables serialization support for serde 1.x
 //! - `rayon`
-//!   - Enables parallel iterators, parallelized methods and [`par_azip!`].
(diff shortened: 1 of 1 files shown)


bluss:
Thanks!

PR #975, opened 2021-04-07
Track-caller panics

### LINKED ISSUE #972: Add `#[track_caller]` judiciously
Add `#[track_caller]` to some panicking methods. Slicing methods seem the most prudent (Index/Mut does not apply, sorry).

Needs benchmarks to investigate impact, preferably holistic benchmarks.

xd009642:
Add track caller to any function, method, trait that has a documented panics section. I also used cargo fmt to tidy up my semi-auto track_caller labelling so there may be other changes but they'll just be formatting.

Still need to assess the performance impact of all these track callers!

Fixes #972

### DIFF
--- a/src/dimension/mod.rs
+++ b/src/dimension/mod.rs
@@ -310,11 +310,13 @@ pub trait DimensionExt {
     /// Get the dimension at `axis`.
     ///
     /// *Panics* if `axis` is out of bounds.
+    #[track_caller]
     fn axis(&self, axis: Axis) -> Ix;
 
     /// Set the dimension at `axis`.
     ///
     /// *Panics* if `axis` is out of bounds.
+    #[track_caller]
     fn set_axis(&mut self, axis: Axis, value: Ix);
 }
 
@@ -349,6 +351,7 @@ impl DimensionExt for [Ix] {
 /// available.
 ///
 /// **Panics** if `index` is larger than the size of the axis
+#[track_caller]
 // FIXME: Move to Dimension trait
 pub fn do_collapse_axis<D: Dimension>(
     dims: &mut D,
@@ -385,6 +388,7 @@ pub fn abs_index(len: Ix, index: Ixs) -> Ix {
 /// The return value is (start, end, step).
 ///
 /// **Panics** if stride is 0 or if any index is out of bounds.
+#[track_caller]
 fn to_abs_slice(axis_len: usize, slice: Slice) -> (usize, usize, isize) {
     let Slice { start, end, step } = slice;
     let start = abs_index(axis_len, start);
@@ -426,6 +430,7 @@ pub fn offset_from_low_addr_ptr_to_logical_ptr<D: Dimension>(dim: &D, strides: &
 /// Modify dimension, stride and return data pointer offset
 ///
 /// **Panics** if stride is 0 or if any index is out of bounds.
+#[track_caller]
 pub fn do_slice(dim: &mut usize, stride: &mut usize, slice: Slice) -> isize {
     let (start, end, step) = to_abs_slice(*dim, slice);
 
--- a/src/impl_2d.rs
+++ b/src/impl_2d.rs
@@ -23,6 +23,7 @@ where
     /// let array = array![[1., 2.], [3., 4.]];
     /// assert_eq!(array.row(0), array![1., 2.]);
     /// ```
+    #[track_caller]
     pub fn row(&self, index: Ix) -> ArrayView1<'_, A>
     where
         S: Data,
@@ -40,6 +41,7 @@ where
     /// array.row_mut(0)[1] = 5.;
     /// assert_eq!(array, array![[1., 5.], [3., 4.]]);
     /// ```
+    #[track_caller]
     pub fn row_mut(&mut self, index: Ix) -> ArrayViewMut1<'_, A>
     where
         S: DataMut,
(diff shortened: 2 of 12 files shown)


xd009642:
So I tried the benchmarks in the repo just to see if there were any glaring issues but the results seemed very noisy and an even mix of faster than master or slower than master. I'll create some more directed benchmarks and try to figure out what the impact is

src/arraytraits.rs
```diff
@@ -47,6 +47,7 @@ where
 {
     type Output = S::Elem;
     #[inline]
+    #[track_caller]
     fn index(&self, index: I) -> &S::Elem {
```
bluss, line 51: It doesn't work on trait methods I thought?

ndarray-rand/src/lib.rs
```diff
@@ -35,7 +35,7 @@ use crate::rand::seq::index;
 use crate::rand::{thread_rng, Rng, SeedableRng};
 
 use ndarray::{Array, Axis, RemoveAxis, ShapeBuilder};
-use ndarray::{ArrayBase, DataOwned, RawData, Data, Dimension};
+use ndarray::{ArrayBase, Data, DataOwned, Dimension, RawData};
```
bluss, line 38: Formatting changes for the whole crate should be in a different PR, see #922

xd009642: Yeah I tried not to commit anything that I hadn't changed but I'll go through and revert what I can. It's just a bit faffy when cargo fmt is basically part of my workflow :sweat_smile:

src/arraytraits.rs
xd009642: https://doc.rust-lang.org/reference/attributes/codegen.html Looks like it does, but I should just put it on the trait definition not implementations

bluss:
If we can find, a non-microbenchmark like mentioned #972

src/arraytraits.rs
xd009642: Ah but Index and IndexMut are in the standard library... And already tracked there! I'll remove :+1:

src/arraytraits.rs
```diff
@@ -182,6 +183,7 @@ where
     /// let array = Array::from_iter((0..5).map(|x| x * x));
     /// assert!(array == arr1(&[0, 1, 4, 9, 16]))
     /// ```
+    #[track_caller]
```
bluss, line 186: Should not have track caller - it's a trait that doesn't support it, and the panic reason is very obscure, not worth it.

xd009642:
Okay so I created this microbenchmark of slicing:

```rust
use criterion::{black_box, criterion_group, criterion_main, Criterion};
use ndarray::{s, Array3};
use ndarray_rand::RandomExt;
use ndarray_rand::rand_distr::Uniform;

fn slice_benchmark(c: &mut Criterion) {
    let data: Array3<f64> = Array3::random((64,64,64), Uniform::new(0., 10.));

    c.bench_function("slice", |b| b.iter(|| {
        data.slice(black_box(s![.., 24..50, ..])).shape();
    }));
}

criterion_group!(benches, slice_benchmark);
criterion_main!(benches);
```

Results:

**0.15.0**
slice                   
time:   [34.281 ns 34.723 ns 35.189 ns]                   
Found 9 outliers among 100 measurements (9.00%)
  4 (4.00%) high mild
  5 (5.00%) high severe

**This PR**
slice                   
time:   [32.815 ns 33.108 ns 33.458 ns]                   
change: [-8.6198% -5.7554% -3.0651%] (p = 0.00 < 0.05)
Performance has improved.
Found 5 outliers among 100 measurements (5.00%)
  3 (3.00%) high mild
  2 (2.00%) high severe

Results folder here:

[criterion_results.zip](https://github.com/rust-ndarray/ndarray/files/6273524/criterion_results.zip)

xd009642:
Personally, I reckon the shift in the results between the two is probably down to something else scheduled by my OS taking some cycles away and these two versions can be considered equivalent in performance (just from this benchmark).

Whether this microbenchmark is good enough to check the impact of track caller is down to someone more familiar with ndarray internals :slightly_smiling_face:

xd009642:
Any thoughts about the benchmark @bluss ?

bluss:
The maintainers have asked for judicious use of the track caller attribute and applying it everywhere is funny to me :slightly_smiling_face:  because it's basically the literal opposite by the dictionary of what judicious means.

Maintainers have asked for a holistic benchmark - maybe a real program that uses ndarray for a lot of varied operations.

Just some real talk :slightly_smiling_face:. If the contributor doesn't want to be *judicious* or whatever, then the maintainer instead has to judge line by line if the additions make sense. And I don't have the investment to do that at the moment, sorry. In short this PR is stalling a bit since it seems to go a bit against every instruction that I have given to try to guide its resolution.

bluss:
A general nice github feature: please use a line like "Fixes #972" to the PR description when making a change that closes an issue. Issues should be closed automatically on merge, that's the nice thing.

xd009642:
So I didn't apply it everywhere the PR title is a bit inaccurate in that sense. I only applied it on public API functions figuring only panics that can be caused directly by user code should be tracked. But I'll go through them and aim to reduce the scope further.

Oh I misread the non-microbenchmark as microbenchmark sorry that ones on me. I was looking at some projects like linfa to see if I could find bigger example projects that used some methods that panicked but I didn't find anything that looked useful. I'll go back to that.

bluss:
Updated to master and conflicts resolved, I hope.

bluss:
Squashed to clean up history. Current merge queue config will create a merge commit at the end too.

bluss:
Thanks, sorry for the delay.

PR #977, opened 2021-04-08
Require len > 1 axis for layout preference

bluss:
This change means that we are not drawn to prefer an axis with length 1 as the "innermost/tightest" axis in Zip.

This fixes the bad performance reported in https://github.com/rust-ndarray/ndarray/pull/932#issuecomment-816082477, specifically select_axis1. The runtime for that benchmark decreases from 21 to 3.4 μs, which is quite substantial and just due to misidentification (in that case) of which axis to prefer for the loop.

### DIFF
--- a/src/layout/mod.rs
+++ b/src/layout/mod.rs
@@ -171,6 +171,34 @@ mod tests {
     }
 
     #[test]
+    fn no_layouts() {
+        let a = M::zeros((5, 5));
+        let b = M::zeros((5, 5).f());
+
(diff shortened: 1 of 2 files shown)


bluss:
This commit (on a pr branch - don't know if the commit will disappear) is relevant too https://github.com/rust-ndarray/ndarray/pull/932/commits/4878d1b148c618a973cbc90fe09ed1afd81e7941

```

PR #978, opened 2021-04-10
Add getter for dim in Shape and StrideShape

### LINKED ISSUE #820: Getter for dim in Shape
For `Shape`, there is not getter for the dimension stored inside the structure. There was one, but now it is solely a comment (see below): 
```rust
impl<D> Shape<D>
where
    D: Dimension,
{
    // Return a reference to the dimension
    //pub fn dimension(&self) -> &D { &self.dim }
    /// Return the size of the shape in number of elements
    pub fn size(&self) -> usize {
        self.dim.size()
    }
}
```
Is there a possibility to extract the exact dimension stored inside `Shape` to use it to modify the dimensions of struct definitions (minimal example below) outside of the crate? 

```rust
struct Tester<D: Dimension> {
    weight: Array<f64, D::Larger>,
}

impl Tester<Ix1> {
    fn new<Sh>(shape: Sh, first_dim: usize) -> Self
    where
        Sh: ShapeBuilder<Dim = Ix1>,
    {
        let dim_in = shape.into_shape().dim;  // This is not possible because the field `dim` is private. How can one obtain the dimension?
        let shape_new = Shape::from(Ix2(first_dim, dim[0]));
        let weight = Array::<f64, Ix2>::zeros(new);
        // ... modify `weight` ...
        Self { weight }
    }
}
```

stokhos:
Code should compile. fixes #820 

But there are 2 parts that I'm not sure.
1.  `(3,2).f() `returns an instance that has the trait `ShapeBuilder`. I'm confused that it seems only `Shape` and `StrideShape` implemented `ShapeBuilder`. Why does `(usize, usize)` also have `.f()` method? 
2. How can I test the `StrideShape`?  The 'Stride' is entirely new to me. Can someone give me an example over this?

It looks like my vim formats the code automatically. Is that ok, or I have to follow the original format?

### DIFF
--- a/src/shape_builder.rs
+++ b/src/shape_builder.rs
@@ -13,7 +13,7 @@ pub struct Shape<D> {
 }
 
 #[derive(Copy, Clone, Debug)]
-pub(crate) enum Contiguous { }
+pub(crate) enum Contiguous {}
 
 impl<D> Shape<D> {
     pub(crate) fn is_c(&self) -> bool {
@@ -21,12 +21,25 @@ impl<D> Shape<D> {
     }
 }
 
-
 /// An array shape of n dimensions in c-order, f-order or custom strides.
 #[derive(Copy, Clone, Debug)]
 pub struct StrideShape<D> {
     pub(crate) dim: D,
     pub(crate) strides: Strides<D>,
+}
+
+impl<D> StrideShape<D>
+where
+    D: Dimension,
+{
+    /// Return a reference to the dimension
+    pub fn raw_dim(&self) -> &D {
+        &self.dim
+    }
+    /// Return the size of the shape in number of elements
+    pub fn size(&self) -> usize {
+        self.dim.size()
+    }
 }
 
 /// Stride description
@@ -37,21 +50,26 @@ pub(crate) enum Strides<D> {
     /// Column-major ("F"-order)
     F,
     /// Custom strides
-    Custom(D)
+    Custom(D),
 }
 
 impl<D> Strides<D> {
     /// Return strides for `dim` (computed from dimension if c/f, else return the custom stride)
     pub(crate) fn strides_for_dim(self, dim: &D) -> D
-        where D: Dimension
+    where
+        D: Dimension,
     {
         match self {
             Strides::C => dim.default_strides(),
             Strides::F => dim.fortran_strides(),
             Strides::Custom(c) => {
-                debug_assert_eq!(c.ndim(), dim.ndim(),
+                debug_assert_eq!(
+                    c.ndim(),
+                    dim.ndim(),
                     "Custom strides given with {} dimensions, expected {}",
-                    c.ndim(), dim.ndim());
+                    c.ndim(),
+                    dim.ndim()
+                );
                 c
             }
         }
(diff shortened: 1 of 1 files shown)


src/shape_builder.rs
```diff
@@ -7,28 +7,41 @@ use crate::Dimension;
 #[derive(Copy, Clone, Debug)]
 pub struct Shape<D> {
     /// Shape (axis lengths)
-    pub(crate) dim: D,
+    pub dim: D,
```
bluss, line 10: Should not be a public field :)

bluss:
This looks good to me. Since we have some unfortunate naming around dimensions, these methods should probably be called `raw_dim`, that's the convention that the arrays use for now. :frowning_face:. `dimension` or `dim` is clearly a better name, but it's hard to decide if convention or better name is the way to go. Change looks good to me of course, as soon as we know what name to use.

bluss:
The `.f()` method comes from the IntoDimension blanket impl if I don't misremember. It's a strange feature, that we put some extra methods on tuples so that tuples can be shape builder values for us.

stokhos:
> This looks good to me. Since we have some unfortunate naming around dimensions, these methods should probably be called `raw_dim`, that's the convention that the arrays use for now. frowning_face. `dimension` or `dim` is clearly a better name, but it's hard to decide if convention or better name is the way to go. Change looks good to me of course, as soon as we know what name to use.

Do you have a plan for resolving the naming issue? Cause I feel this would be an great opportunity for me to learn the internals.

bluss:
Resolving the name for arrays in general won't come until we remake dimensions for const generics (ndarray 0.16 or later).

Right now I'm wondering what you think about the naming. I could see either .dimension() or .raw_dim() being a good choice here.

stokhos:
I have changed method name from `dimension` to `raw_dim`.

stokhos:
> Resolving the name for arrays in general won't come until we remake dimensions for const generics (ndarray 0.16 or later).
> 
> Right now I'm wondering what you think about the naming. I could see either .dimension() or .raw_dim() being a good choice here.

The reason why I used `dimension` is it consistent with  method `size` in below.

bluss (approved):
Thanks, looks good. I give up the naming, either would be fine, this is consistent (sadly).

stokhos:
> Thanks, looks good. I give up the naming, either would be fine, this is consistent (sadly).

So if we stick to convention, do we need to change `size` to `raw_size`?

jturner314:
I agree that `raw_dim` is the best name for now, for consistency with `ArrayBase`'s methods. I think the `size` method should be named `size`, because that's what the similar method is named in the `Dimension` trait.

While we're at it, I'd like an `is_f` method on `Shape` for use with [`ndarray_npy::write_zeroed_npy`](https://docs.rs/ndarray-npy/0.8.0/ndarray_npy/fn.write_zeroed_npy.html), assuming @bluss agrees. (I'd like to reuse `ndarray`'s `ShapeBuilder` stuff to allow the user to optionally specify that they'd like Fortran layout, like `ArrayBase::zeros` does.) I could create a separate issue/PR for that, but it seems closely enough related to this PR to include here.

bluss:
The is_f method sounds fine. Note that the Shape/StrideShape things are due for redesign

stokhos:
I'd like to work on the `is_f` if it is not urgent.

bluss:
It is good to avoid extra formatting changes, but it's not so much of it here. Thanks for this.

PR #979, opened 2021-04-11
Basic support for axis simplification and arbitrary order in iterators

bluss:


src/dimension/mod.rs
```diff
@@ -689,6 +701,33 @@ where
...
+    D: Dimension,
+{
+    debug_assert_eq!(dim.ndim(), strides.ndim());
+    match dim.ndim() {
+        0 | 1 => {}
+        n => {
+            let mut last = n - 1;
+            for i in (0..last).rev() {
+                if !merge_axes(dim, strides, Axis(i), Axis(last)) {
+                    last = i;
```
jturner314, line 724: I don't think that setting `last = i` will actually terminate the loop.

```suggestion
                    break;
```

bluss: It just changes the axis we're currently merging into.

jturner314: Oh, I understand now. I was thinking that the intent was to merge into only the last axis.

bluss: My bad, with the function name

jturner314: The function name is fine, IMO. It's more that I was thinking of the example code I wrote in [my comment on #932](https://github.com/rust-ndarray/ndarray/pull/932#discussion_r611237837), which did terminate the loop as soon as merging failed. I would suggest changing the variable name from `last` to `into` or `into_axis`, though.

bluss:
Benchmark

In the commit that adds the benchmark:

```rust
test iter_sum_1d_strided_fold           ... bench:       1,423 ns/iter (+/- 364)                                                                                     test iter_sum_1d_strided_rfold          ... bench:       1,440 ns/iter (+/- 119)                                                                                     test iter_sum_2d_col_matrix_for_strided ... bench:       1,134 ns/iter (+/- 51)                                                                                      test iter_sum_2d_col_matrix_sum_strided ... bench:         583 ns/iter (+/- 150)                                                                                     test iter_sum_2d_row_matrix_for_strided ... bench:       1,659 ns/iter (+/- 146)                                                                                     test iter_sum_2d_row_matrix_sum_strided ... bench:       4,828 ns/iter (+/- 598)   
```

After the iterator takes advantage of the optimizations (after this PR):

```rust
test iter_sum_1d_strided_fold           ... bench:       1,388 ns/iter (+/- 274)                                                                                     test iter_sum_1d_strided_rfold          ... bench:       1,366 ns/iter (+/- 322)                                                                                     test iter_sum_2d_col_matrix_for_strided ... bench:       1,189 ns/iter (+/- 76)                                                                                      test iter_sum_2d_col_matrix_sum_strided ... bench:         550 ns/iter (+/- 71)                                                                                      test iter_sum_2d_row_matrix_for_strided ... bench:       1,216 ns/iter (+/- 51)                                                                                      test iter_sum_2d_row_matrix_sum_strided ... bench:         561 ns/iter (+/- 52) 
```

This is for `.iter()` so only the order-preserving optimizations apply here. For example transforming those row/column matrices shapes from [10, 1] to [1, 10] etc.

PR #981, opened 2021-04-15
Add logo to docs

jturner314:
I noticed the logo in the upper left corner of the [`wgpu` docs](https://docs.rs/wgpu/0.7.1/wgpu/) and thought it was kind of fun, so I've created this PR. This is the same logo I created for the [`rust-ndarray` GitHub organization](https://github.com/rust-ndarray/). The Ferris crab inside of the brackets is from [here](https://www.rustacean.net/); it's licensed as CC0 (public domain).

Note that the image URL will not work until this is merged (since the image will be hosted by GitHub). Edit: I just saw that a PNG version of the logo is currently located in the GitHub pages repository. Would it be better to host the image from GitHub Pages instead?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -7,6 +7,7 @@
 // except according to those terms.
 #![crate_name = "ndarray"]
 #![doc(html_root_url = "https://docs.rs/ndarray/0.15/")]
+#![doc(html_logo_url = "https://rust-ndarray.github.io/images/rust-ndarray_logo.svg")]
 #![allow(
     clippy::many_single_char_names,
     clippy::deref_addrof,


bluss:
I don't know if github pages or githubusercontent is the best one, depends on what github thinks about serving files out of the latter? They probably tolerate a lot of that, but there are solutions like https://gitcdn.link/ for that too (so it must mean there is some drawback like throttling?  https://stackoverflow.com/questions/8779197/how-to-link-files-directly-from-github-raw-github-com

Use pages if you want. It should only be hosting an ancient version of the rustdoc (that's redirecting to docs.rs) (no need to change anything about the redirect I guess)

bluss (approved):
This is a good solution, but we can use pages too if you want

jturner314:
I've uploaded the logo to the [rust-ndarray/rust-ndarray.github.io](https://github.com/rust-ndarray/rust-ndarray.github.io) repository.

I did notice that the only other image in the docs (the one for [`.axis_iter()`](https://docs.rs/ndarray/0.15.1/ndarray/struct.ArrayBase.html#method.axis_iter)) is hosted from the `gh-pages` GitHub Pages branch on this repository rather than the separate repository, but there's also a similar image in the separate repository. I'm not sure what the intent is with respect to the `gh-pages` branch here vs. the separate repository.

PR #986, opened 2021-04-22
Implement by-value iterator for owned arrays

### LINKED ISSUE #196: IntoIterator on ArrayBase
Why is there no IntoIterator implementation for ArrayBase itself? I want to collect all elements into another data structure (or just map its type, but that's not available either), but all I can access are references of the elements, which is kind of bad. I could use mem::replace, but that requires me to create dummy values, which is bad for performance. So I'd like to just consume the whole thing and collect it into a new data structure instead.

bluss:
Implement `IntoIterator` for `Array`. This uses the same code for "dropping unreachable elements" that `.move_into()` does (from the #932 PR).

New features: 

- `Array::into_iter` (same for ArcArray, CowArray)

Fixes #196

### DIFF
--- a/src/data_repr.rs
+++ b/src/data_repr.rs
@@ -53,6 +53,10 @@ impl<A> OwnedRepr<A> {
         self.ptr.as_ptr()
     }
 
+    pub(crate) fn as_ptr_mut(&self) -> *mut A {
+        self.ptr.as_ptr()
+    }
+
     pub(crate) fn as_nonnull_mut(&mut self) -> NonNull<A> {
         self.ptr
     }
--- a/src/data_repr.rs
+++ b/src/data_repr.rs
@@ -92,6 +92,13 @@ impl<A> OwnedRepr<A> {
         self.len = new_len;
     }
 
+    /// Return the length (number of elements in total)
+    pub(crate) fn release_all_elements(&mut self) -> usize {
+        let ret = self.len;
+        self.len = 0;
+        ret
+    }
+
     /// Cast self into equivalent repr of other element type
     ///
     /// ## Safety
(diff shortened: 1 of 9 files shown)


nilgoyette:
Thank you @bluss . This was included in 0.15.2 and I was surprised to see that it broke our project. It's not a problem at all, it was super easy to fix! I'm more curious about which of the following is true:

1. This was an API change so it shouldn't have been included in a patch version.
2. I shouldn't have been using `into_iter` with ndarrays because it was iterating on references anyway. I should have been using `iter`.

Or none of them? Or both ^^

bluss:
Do you have an example of the code?

nilgoyette:
Here's one example

    fft_zscores // Array1 or f32 or f64
        .into_iter()
        .enumerate()
        .filter(|(_, &v)| v.abs() > threshold)
        .map(|(idx, &v)| Outlier { ... }

I had to remove both `&`.

bluss:
Thanks. Do you recall the blog poist https://blog.rust-lang.org/2021/05/11/edition-2021.html and the part about IntoIterator for arrays, it might be interesting? As it says:

> Usually we categorize this type of breakage (adding a trait implementation) 'minor' and acceptable.

So that means (2) which applies here too. I'm suprised but I could see that it can happen.

I would recommend that you convert this code by using `.iter()`, not using the by move iterator unless you need to IMO.

nilgoyette:
Oh, ok, that anwers my question. Thank you.

I don't want to abuse of your time, but may I ask you why using `.iter()` instead of removing the `&` is a better idea? I fail to see why one method is better in general than the other.

bluss:
you can see it if you read the implementation? :)

Issue #988, opened 2021-04-28
Slicing an ArrayView restricts lifetime


cassiersg:
The following doesn't work, due to the signature of ArrayBase::slice.
```
fn foo<'a>(x: ArrayView1<'a, i32>) -> ArrayView<'a, i32> {
    x.slice(s![1..])
}
```
It would be nice to have a method dedicated to this on ArrayView, similarly to `ArrayView::to_slice`, (which is a taylor-adapted version of `ArrayBase::as_slice`).

jturner314:
The [`slice_move`](https://docs.rs/ndarray/0.15.1/ndarray/struct.ArrayBase.html#method.slice_move) method on `ArrayBase` can be used for this purpose. If you don't want to lose access to the original view, note that const-ndim `ArrayView`s implement `Copy` and `Clone`, and dynamic-ndim `ArrayView`s implement `Clone`. However, I'm not opposed to adding a special method to `ArrayView` if we can come up with a good name for it, since there is a potential efficiency gain from a specialized implementation instead of `.clone().slice_move()`.

cassiersg:
Thank you, I missed `slice_move`. I'll go with that. I guess overhead of `.clone().slice_move()` will be negligible except in rather rare situations.

jturner314:
Yeah, for `ArrayView1` especially, you don't even need the `.clone()` since it implements `Copy`, and the overhead should be zero or negligible.

cassiersg:
I opened #989 which would have helped me not miss the `.clone().slice_move()` solution. Re-perf: indeed, and also for any dimensionality as long as you deal with large matrices (and actually do something with them).

Leaving up to you to close this or not.

jturner314:
Okay, I'll go ahead and close this.

To anyone coming across this issue: If `.clone().slice_move()` isn't good enough, and you need a method

```rust
impl<'a, A, D> ArrayView<'a, A, D> {
    fn foo<I>(&self, info: I) -> ArrayView<'a, A, I::OutDim> 
    where
        I: SliceArg<D>,
    { ... }
}
```

and have a good name for it, please feel free to reopen this issue or create a new one referencing it.

PR #989, opened 2021-04-28
Improve documentation of slicing methods.

cassiersg:


jturner314:
It's somewhat awkward to reference a few other slicing methods without listing all of the potentially relevant ones. However, I don't think we should list all of them, since there are a lot of slicing methods, and they're all mentioned in the *Slicing* docs already.

What I'd suggest is to address the question in #988 more directly, i.e. add something like the following to the docs for `.slice()`:

> Note that the lifetime of the resulting slice corresponds to the `&self` borrow. If you want a slice with the same lifetime as an existing view, clone/copy the view if necessary and then use `.slice_move()` or `.slice_collapse()`.

and add the following to the docs for `.slice_mut()`:

> Note that the lifetime of the resulting slice corresponds to the `&mut self` borrow. If you want a slice with the same lifetime as an existing view, use `.slice_move()` or `.slice_collapse()` on the view.

We could also add analogous comments to the docs of `slice_axis/_mut` (referencing `slice_axis_inplace`) and `slice_each_axis/_mut` (referencing `slice_each_axis_inplace`).

One other note: the markup of the links in this PR isn't quite right. A link should be formatted like this:

```
some text [the link text](url) more text
```

or like this:

```
some text [the link text] more text

[the link text]: url
```

Note also that as of Rust 1.48.0, it's possible to [use the Rust path of an in-scope item as a "URL"](https://doc.rust-lang.org/rustdoc/linking-to-items-by-name.html), and as of Rust 1.51.0, this works for associated items too, such as writing `[.slice()](ArrayBase::slice)`.

PR #994, opened 2021-05-09
Add offset_to_first_elem method for owned arrays

jturner314:
This adds a method to get the offset from the start of the allocation to the first element. Without this method, correctly reinterpreting the results of `.into_raw_vec()` as an n-D array is tricky.

I came across a bug in the `numpy` crate (PyO3/rust-numpy#182) due to not handling this properly, so the the primary purpose of this PR is to help properly fix that bug.

The zero-sized element case is a bit weird, because for many use cases, a return value of `Some(0)` would be fine. However, if we want the user to be able to compute in-bounds indices into the `Vec` using the offset, shape, and strides, we need to compute an offset that will work based on the shape and strides.

### DIFF
--- a/src/impl_owned_array.rs
+++ b/src/impl_owned_array.rs
@@ -61,6 +61,55 @@ impl<A, D> Array<A, D>
 where
     D: Dimension,
 {
+    /// Returns the offset (in units of `A`) from the start of the raw data to
+    /// the first element, or `None` if the array is empty.
+    ///
+    /// In other words, after converting the array to `Vec<A>` with
+    /// [`.into_raw_vec()`](Self::into_raw_vec), this is the offset from the
+    /// start of the `Vec` to the first element of the array.
+    ///
+    /// ```
+    /// use ndarray::{Array2, Axis};
+    ///
+    /// let mut arr: Array2<f64> = Array2::zeros([3, 4]);
+    /// arr.slice_axis_inplace(Axis(0), (1..).into());
+    /// arr[[0, 0]] = 5.;
+    ///
+    /// let offset = arr.offset_to_first_elem().unwrap();
+    /// assert_eq!(offset, 4);
+    ///
+    /// let v = arr.into_raw_vec();
+    /// assert_eq!(v[offset], 5.);
+    /// ```
+    ///
+    /// In the case of zero-sized elements, the offset is somewhat meaningless.
+    /// For convenience, an offset will be returned such that all indices
+    /// computed using the offset, shape, and strides will be in-bounds for the
+    /// `Vec<A>` returned by [`.into_raw_vec()`](Self::into_raw_vec). Note that
+    /// this offset won't necessarily be the same as the offset for an array of
+    /// nonzero-sized elements sliced in the same way.
+    ///
+    /// ```
+    /// use ndarray::{Array2, Axis};
+    ///
+    /// let mut arr: Array2<()> = Array2::from_elem([3, 4], ());
+    /// arr.slice_axis_inplace(Axis(0), (1..).into());
+    /// let offset = arr.offset_to_first_elem().unwrap();
+    /// assert_eq!(offset, 0);
+    /// ```
+    pub fn offset_to_first_elem(&self) -> Option<usize> {
+        if self.is_empty() {
+            return None;
+        }
+        let offset = if std::mem::size_of::<A>() == 0 {
+            -dimension::offset_from_ptr_to_memory(&self.dim, &self.strides)
+        } else {
+            unsafe { self.as_ptr().offset_from(self.data.as_ptr()) }
+        };
+        debug_assert!(offset >= 0);
+        Some(offset as usize)
+    }
+
     /// Return a vector of the elements in the array, in the way they are
     /// stored internally.
     ///
@@ -491,13 +540,8 @@ impl<A, D> Array<A, D>
 
         unsafe {
             // grow backing storage and update head ptr
-            let data_to_array_offset = if std::mem::size_of::<A>() != 0 {
-                self.as_ptr().offset_from(self.data.as_ptr())
-            } else {
-                0
-            };
-            debug_assert!(data_to_array_offset >= 0);
-            self.ptr = self.data.reserve(len_to_append).offset(data_to_array_offset);
+            let data_to_array_offset = self.offset_to_first_elem().unwrap_or(0);
+            self.ptr = self.data.reserve(len_to_append).add(data_to_array_offset);
 
             // clone elements from view to the array now
             //


src/impl_owned_array.rs
```diff
@@ -61,6 +61,55 @@ impl<A, D> Array<A, D>
...
+    /// arr.slice_axis_inplace(Axis(0), (1..).into());
+    /// let offset = arr.offset_to_first_elem().unwrap();
+    /// assert_eq!(offset, 0);
+    /// ```
+    pub fn offset_to_first_elem(&self) -> Option<usize> {
+        if self.is_empty() {
+            return None;
+        }
+        let offset = if std::mem::size_of::<A>() == 0 {
+            -dimension::offset_from_ptr_to_memory(&self.dim, &self.strides)
```
bluss, line 105: can't we use this dimension function in both cases?

src/impl_owned_array.rs
```diff
@@ -61,6 +61,55 @@ impl<A, D> Array<A, D>
...
+    ///
+    /// ```
+    /// use ndarray::{Array2, Axis};
+    ///
+    /// let mut arr: Array2<()> = Array2::from_elem([3, 4], ());
+    /// arr.slice_axis_inplace(Axis(0), (1..).into());
+    /// let offset = arr.offset_to_first_elem().unwrap();
+    /// assert_eq!(offset, 0);
+    /// ```
+    pub fn offset_to_first_elem(&self) -> Option<usize> {
```
bluss, line 100: Maybe we can use the name logic "offset from memory to first elem" again?

src/impl_owned_array.rs
jturner314: No, because the shape and strides are not sufficient information. Consider, for example:

```rust
let a = Array2::<f64>::zeros([3, 4]).slice_move(s![1.., ..]);
let b = Array2::<f64>::zeros([2, 4]);
assert_eq!(a.shape(), b.shape());
assert_eq!(a.strides(), b.strides());
assert_ne!(
    unsafe { a.as_ptr().offset_from(a.into_raw_vec().as_ptr()) },
    unsafe { b.as_ptr().offset_from(b.into_raw_vec().as_ptr()) },
);
```

`a` and `b` have the same shape and strides but different offsets.

The only reason for calling `dimension::offset_from_ptr_to_memory` here is to get an offset such that indices computed with `offset + index[0]*strides[0] + index[1]*strides[1] + ...` will be in-bounds for the result of `.into_raw_vec()`. We could return value return `None` or `Some(0)` for the zero-sized element case, but then it's more likely that the caller will have to explicitly deal with the zero-sized element case too.

bluss: Oh right. Thanks! That's the three "starts" of the array in memory again...

1. start of memory allocation
2. lowest address of an element in the array
3. address of logically first element in the array..

I guess if we try and find good names for these (again). I wouldn't mind allocation head, memory head, logical head or something along those lines. Maybe "memory" is too vague to be useful(?)

src/impl_owned_array.rs
jturner314: Yeah, `offset_from_memory_to_ptr`, `offset_from_memory_to_first`, and `offset_from_memory_to_first_elem` are clearer than `offset_to_first_elem`. Another, perhaps better, option is to change `.into_raw_vec()` to return `(vec, offset)` instead of adding a new method. This would make it harder to misuse `.into_raw_vec()`.

src/impl_owned_array.rs
jturner314: I like "allocation ptr" and "logical ptr/offset". "Memory head" to refer to the lowest address of any element is a bit confusing, IMO. I'd rather say "low addr ptr/offset". Fortunately, there aren't many times when that pointer/offset is important, anyway, so a concise name for it probably isn't a big concern. Pretty much the only things I could see the "low addr head" being useful for are the `drop_unreachable_elements` method or a method to "compact" an array (remove unreachable elements).

jturner314: > I mixed up (1) and (2)...

Fwiw, this isn't quite true. The mixup was assuming that (1) and (3) were the same. The `dimension::offset_from_ptr_to_memory` function actually returns the offset from the "logical pointer" (3) to the "lowest reachable address pointer" (2). This makes me realize that we should really rename `dimension::offset_from_ptr_to_memory` to `dimension::offset_from_logical_ptr_to_low_addr_ptr`. (Actually, I'd prefer to have a `dimension::offset_from_low_addr_ptr_to_logical_ptr` function instead so that the return value is always nonnegative.)

src/impl_owned_array.rs
jturner314: From the discussion in the other comment chain, I'm thinking that a good name would be `offset_from_alloc_to_logical_ptr`. How about making it an internal-only method for now and changing `.into_raw_vec()` to return `(vec, offset)` as I suggested above?

jturner314: You know, now that I think about it, I wonder if `std` will add a trait for `Vec`-like objects to minimize the proliferation of type parameters. The current `ndarray` API based on `ArrayBase` instead of traits can be inconvenient sometimes. Once custom (non-global) allocators are supported, `Vec` will be in a similar position since it'll be generic over the allocator.

bluss: Makes sense.

I've long thought that we might parameterize the Array not just by the Vec's allocator but by wholly new kinds of owned arrays - let's say some other allocation system, like always 64-byte aligned or what it could be, or maybe always using some external specific allocation function. In those cases we don't use Vec or std allocation traits at all, maybe.

bluss:
While it's a breaking change, we wait with merging this until the 0.15.x releases are done

src/impl_owned_array.rs
jturner314: Yeah, an `ArrayVec`-backed n-D array type would be useful too. Even with additional owned types/parameters, though, we could provide `.into_raw_vec()` for only the type parameters that make sense. Long-term, I'd like to move away from `ArrayBase` to a set of traits describing n-D array functionality, implemented for the various array types.

bluss: Fully agreed. It just seems to be a lot of work getting to the traitification. Maybe we can start it small and do it piece by piece, somehow?

jturner314: Yeah, I think we can do it piece-by-piece. We can introduce traits one-by-one and implement them for `ArrayBase`. For example, we can start with an unsafe trait to get the ptr, dim, and strides, and then add a trait using it for readonly indexing/slicing (with methods returning element references and `ArrayView`s). I expect it'll take multiple iterations to fully refine the set of traits we need.

src/dimension/mod.rs
```diff
@@ -404,17 +405,18 @@ fn to_abs_slice(axis_len: usize, slice: Slice) -> (usize, usize, isize) {
...
 }
 
-/// This function computes the offset from the logically first element to the first element in
-/// memory of the array. The result is always <= 0.
-pub fn offset_from_ptr_to_memory<D: Dimension>(dim: &D, strides: &D) -> isize {
+/// This function computes the offset from the lowest address element to the
+/// logically first element. The result is always >= 0.
+pub fn offset_from_low_addr_ptr_to_logical_ptr<D: Dimension>(dim: &D, strides: &D) -> usize {
     let offset = izip!(dim.slice(), strides.slice()).fold(0, |_offset, (d, s)| {
         if (*s as isize) < 0 {
```
SparrowLii, line 412: I am thinking about whether it would be better to change the condition to `if (*s as isize) < 0 && *d != 0`. In this way, we can allow the step size to be negative when the Axis value is 0 (theoretically, the step size can be any value at this time), and it can avoid panic in the following case:

    use ndarray::{Array, ShapeBuilder};
    fn main() {
        let a= Array::from_shape_vec((2, 0, 2).strides((1, -4isize as usize, 2)),s[0..12].to_vec()).unwrap();
        println!("{:?}", a);
    }

Output:

      [[[]]], shape=[2, 0, 2], strides=[1, -4, 2], layout=Cc (0x5), const ndim=3

bluss: Good point, this seems like a bug that should be fixed ASAP

src/impl_owned_array.rs
bluss: I don't think "logical ptr" is very descriptive either. Offset >= 0 sounds good.

src/dimension/mod.rs
jturner314: Yeah, good catch @SparrowLii. I've created #998 to fix this in 0.15.2.

src/impl_owned_array.rs
jturner314: I don't mind "logical ptr", but we can think of another name. Maybe "zero idx ptr" / "zero index ptr"?

PR #998, opened 2021-05-13
Fix offset_from_ptr_to_memory for axis lengths of 0, and replace it with offset_from_low_addr_ptr_to_logical_ptr

jturner314:
See [SparrowLii's comment](https://github.com/rust-ndarray/ndarray/pull/994#discussion_r630886761) regarding the bug fix. Renaming the function, inverting the sign of the return value, and changing the return type to `usize` make the function easier to understand and use, IMO.


bluss:
Thanks. I pushed in @SparrowLii's testcase + some more.

Issue #969, opened 2021-04-05
Give MathCell arithmetic ops implementations


bluss:
This is a follow-up to MathCell introduced in #877.

MathCell should have arithmetic operations so that *arrays* and *views* using `MathCell` elements can participate in some arithmetic ops. Preferably together with non-cell arrays. Detailed design might need to be decided after seeing what's permitted w.r.t impl overlap.

PR #1011, opened 2021-05-21
Give MathCell arithmetic ops implementations when MathCell is left value

SparrowLii:
Update #969 
Four functions are implemented for `ArrayView<MathCell, _>` in impl_methods.rs:
`zip_cell_with`, `zip_cell_with_elem`, `zip_cell_with_same_shape`, `zip_cell_with_by_rows`. 
Operator overloading is implemented in impl_ops.rs for the following situations:
`ArrayView<MathCell, _> binary_op &ArrayBase`
`ArrayView<MathCell, _> binary_op Scalar`
`ArrayView<MathCell, _> assign_op &ArrayBase`
`ArrayView<MathCell, _> assign_op Scalar`

The unresolved issues are:
1. Currently `MathCell` can only be left value. The main reason are:
(1) Conflicts occur when compiling `MathCell<A> op MathCell<B>`
(2) Rust's orphan rules. We can't write code like this:`implement<T> Add<MathCell<T>> for T`. This will cause [E0210 ](https://doc.rust-lang.org/error-index.html#E0210) error

2. The value in `MathCell` must be `Copy`
This is because the value in MathCell needs to be retrieved using the `get()` method. We cannot get `&T` from `&MathCell<T>` unless we create a new structure `MathRefCell(RefCell<T>)`

### DIFF
--- a/src/impl_methods.rs
+++ b/src/impl_methods.rs
@@ -2613,6 +2613,88 @@ where
     }
 }
 
+impl<AE, S, D> ArrayBase<S, D>
+    where
+        AE: Copy,
+        D: Dimension,
+        S: Data<Elem = MathCell<AE>>,
+{
+    /// Same as `zip_mut_with`, but just when element type is `MathCell`.
+    #[inline]
+    pub(crate) fn zip_cell_with<B, S2, E, F>(&self, rhs: &ArrayBase<S2, E>, f: F)
+        where
+            S2: Data<Elem = B>,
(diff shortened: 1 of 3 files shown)


bluss:
We don't implement ArrayViewMut + x either (only ArrayViewMut += x), and this should be the same way IMO. I.e we want the `+=` and similar ops. I hope we can do with without adding new copies like all the new `.zip*` methods.

PR #1013, opened 2021-05-27
Add .last() and .last_mut() methods

jturner314:
This PR adds `.last()` and `.last_mut()` methods. These are most useful with 1-D arrays, but we might as well define them for n-D arrays. This PR also adds examples to the docs for the `.first()` and `.first_mut()` methods.

### DIFF
--- a/src/impl_methods.rs
+++ b/src/impl_methods.rs
@@ -280,6 +280,66 @@ where
         }
     }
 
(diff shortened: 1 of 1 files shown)


bluss (approved):

bluss:
Neat

PR #1015, opened 2021-05-29
Add in/to_slice_memory_order for ArrayView/Mut

jturner314:
This adds a `.to_slice_memory_order()` method to `ArrayView` and an `.into_slice_memory_order()` method to `ArrayViewMut`. This has unfortunate code duplication with `.as_slice_memory_order/_mut()`, but the quantity of duplicated code isn't very large.

### DIFF
--- a/src/impl_views/conversions.rs
+++ b/src/impl_views/conversions.rs
@@ -7,12 +7,14 @@
 // except according to those terms.
 
 use alloc::slice;
+use rawpointer::PointerExt;
 use std::mem::MaybeUninit;
 
 use crate::imp_prelude::*;
 
 use crate::{Baseiter, ElementsBase, ElementsBaseMut, Iter, IterMut};
 
+use crate::dimension::offset_from_low_addr_ptr_to_logical_ptr;
 use crate::iter::{self, AxisIter, AxisIterMut};
 use crate::math_cell::MathCell;
 use crate::IndexLonger;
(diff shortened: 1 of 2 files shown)


src/impl_views/conversions.rs
```diff
@@ -221,6 +253,22 @@ where
         }
     }
 
+    /// Return the array’s data as a slice, if it is contiguous.
+    /// Otherwise return self in the Err branch of the result.
+    fn try_into_slice_memory_order(self) -> Result<&'a mut [A], Self> {
```
bluss, line 258: I guess it's even harder to make consistent, but I think this method is unfortunately generally useful. I've contemplated adding the corresponding one on ArrayBase (where the Err case would be `&mut Self`). So we should consider if this isn't the best one (or an additional one) to expose for ArrayViewMut.

jturner314: What if we change all the `.as_slice*`/`.to_slice*`/`.into_slice*` methods return a `Result`? Even if the `Err` variant isn't especially useful for all the methods, it'll just be a reference (i.e. pointer size) for all but the `.into_slice*` methods, so it won't cost much.

bluss:
Rebased and got the newest spelling from #1119 so we're up to date on that too :wink:

bluss:
Thanks I'll merge and future improvements can come later - the Result question is left here and it sounds like it could be a good idea! Nice to have the negative tests included here.

PR #1017, opened 2021-05-31
Document 'unsharing' in as_mut_ptr and raw_view_mut

jturner314:
The primary motivation for this PR is to add the warning to `.as_mut_ptr()`.

### DIFF
--- a/src/impl_methods.rs
+++ b/src/impl_methods.rs
@@ -1446,6 +1446,15 @@ where
     }
 
     /// Return a mutable pointer to the first element in the array.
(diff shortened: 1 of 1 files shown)


bluss (approved):

bluss:
Thanks

bluss:
In general it seems fraught to use this method in generic code. We could advise users of generic arrays to always take a `view_mut()` and work from that. (While ensuring users understand that the mutable array view is always uniquely referencing the array data).

PR #1019, opened 2021-05-31
Guarantee that `.as_slice_memory_order_mut()` preserves strides 

### LINKED ISSUE #1018: `.as_slice_memory_order_mut()` is very difficult to use correctly in generic code, and there are a couple of bugs as a result
I just realized that `.as_slice_memory_order_mut()` is very difficult to use correctly in code with generic `S: DataMut` for any purpose other than iterating over the elements in arbitrary order. The issue is that `.as_slice_memory_order_mut()` unshares the data, which may change its layout, so strides obtained *before* calling `.as_slice_memory_order_mut()` may not be correct for accessing elements in the returned slice. As a result, the [implementation of `.map_mut()`](https://github.com/rust-ndarray/ndarray/blob/e7600e82e629a54805a9fc33e50f34151c1faa2e/src/impl_methods.rs#L2325-L2328) and the [implementation of `.zip_mut_with_same_shape()`](https://github.com/rust-ndarray/ndarray/blob/e7600e82e629a54805a9fc33e50f34151c1faa2e/src/impl_methods.rs#L2189-L2198) produce incorrect results in some cases for `ArcArray` inputs. Here's an example:

```rust
use ndarray::prelude::*;
use ndarray::ArcArray2;

fn main() {
    // Fortran-layout `ArcArray`.
    let a: ArcArray2<u8> = {
        let orig = array![[0, 1, 2, 3, 4], [5, 6, 7, 8, 9]];
        let mut orig_f = Array2::zeros(orig.raw_dim().f());
        orig_f.assign(&orig);
        orig_f.into_shared()
    };
    assert_eq!(a, array![[0, 1, 2, 3, 4], [5, 6, 7, 8, 9]]);
    assert_eq!(
        a.as_slice_memory_order(),
        Some(&[0, 5, 1, 6, 2, 7, 3, 8, 4, 9][..])
    );

    // Shared reference of a portion of `a`.
    let mut b = a.clone().slice_move(s![.., ..2]);
    assert_eq!(b, array![[0, 1], [5, 6]]);
    assert_eq!(b.as_slice_memory_order(), Some(&[0, 5, 1, 6][..]));

    // `.map_mut()` is broken in this case, because the data is contiguous and
    // unsharing the data changes the layout. The current implementation of
    // `.map_mut()` uses the strides of `b` from *before* unsharing to construct
    // the new array, but these are no longer correct *after* the unsharing.
    //
    // As a result, this assertion fails using the current implementation.
    assert_eq!(b.map_mut(|&mut x| x + 10), array![[10, 11], [15, 16]]);
}
```

It's also worth noting that the [current implementation of `try_as_slice_memory_order_mut` is questionable](https://github.com/rust-ndarray/ndarray/blob/e7600e82e629a54805a9fc33e50f34151c1faa2e/src/impl_methods.rs#L1552-L1560), because it checks for contiguity before unsharing with `ensure_unique`, and `DataMut` doesn't guarantee that the array after calling `ensure_unique` will still be contiguous (although this is currently the case for all existing storage types).

I think the best solution is to do the following:

1. Add this constraint to `RawDataMut`: "The implementer must ensure that if the input to `try_ensure_unique` is contiguous, then the output will have the same strides as the input."
2. Change the implementation of `RawDataMut` for `OwnedArcRepr` to uphold this constraint.
3. Document that `.as_slice_memory_order_mut()` unshares the data if necessary, but the strides are preserved.

jturner314:
Fixes #1018.

### DIFF
--- a/src/data_traits.rs
+++ b/src/data_traits.rs
@@ -50,8 +50,11 @@ pub unsafe trait RawData: Sized {
 pub unsafe trait RawDataMut: RawData {
     /// If possible, ensures that the array has unique access to its data.
     ///
-    /// If `Self` provides safe mutable access to array elements, then it
-    /// **must** panic or ensure that the data is unique.
+    /// The implementer must ensure that if the input is contiguous, then the
+    /// output has the same strides as input.
+    ///
+    /// Additionally, if `Self` provides safe mutable access to array elements,
+    /// then this method **must** panic or ensure that the data is unique.
     #[doc(hidden)]
     fn try_ensure_unique<D>(_: &mut ArrayBase<Self, D>)
     where
@@ -230,14 +233,9 @@ where
             return;
         }
         if self_.dim.size() <= self_.data.0.len() / 2 {
-            // Create a new vec if the current view is less than half of
-            // backing data.
-            unsafe {
-                *self_ = ArrayBase::from_shape_vec_unchecked(
-                    self_.dim.clone(),
-                    self_.iter().cloned().collect(),
-                );
-            }
+            // Clone only the visible elements if the current view is less than
+            // half of backing data.
+            *self_ = self_.to_owned().into_shared();
             return;
         }
         let rcvec = &mut self_.data.0;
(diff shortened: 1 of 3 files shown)


bluss (approved):
Thanks a lot for working on this.

PR #1022, opened 2021-06-01
Add try_into_owned_nocopy method

jturner314:
I've been working on improvements to the layout handling in `ndarray-linalg` to make the code easier to understand, fix bugs, and avoid unnecessary cloning and by-clone transposes. One of operations I need to implement is, "Convert an `ArrayBase` into a LAPACK-compatible `Array` (i.e. `strides = [1, s]`, with `s >= nrows`), while avoiding unnecessary cloning." I don't think it's possible to implement this today without adding more methods to `ArrayBase`. To address this, I propose adding a `try_into_owned_nocopy` method which is similar to `into_owned` but returns `Err(self)` if the data would need to be cloned, instead of actually performing the cloning. This way, if the data needs to be cloned, I can clone it into an array of the correct layout, like this:

```rust
match array.try_into_owned_nocopy() {
    Ok(o) if is_lapack_compatible(&o) => o,
    Ok(o) => clone_into_new_lapack_array(&o),
    Err(a) => clone_into_new_lapack_array(&a),
}
```

There are other ways this problem could be solved, but `try_into_owned_nocopy` is the best I came up with.

A couple of notes about the PR:

1. ~If we implement `DataOwned` for `CowRepr` in another PR, then I'd change the bound on `try_into_owned_nocopy` to be `S: DataOwned`, since there wouldn't be any reason to call it on any non-`DataOwned` types.~ Edit: I was wrong; the method is actually useful for `S: Data` after all, and there isn't a strong reason to restrict it to `S: DataOwned` anyway.

2. I'm not completely satisfied with the name `try_into_owned_nocopy`. Other suggestions would be welcome.

### DIFF
--- a/src/data_traits.rs
+++ b/src/data_traits.rs
@@ -16,7 +16,9 @@ use std::ptr::NonNull;
 use alloc::sync::Arc;
 use alloc::vec::Vec;
 
-use crate::{ArrayBase, CowRepr, Dimension, OwnedArcRepr, OwnedRepr, RawViewRepr, ViewRepr};
+use crate::{
+    ArcArray, Array, ArrayBase, CowRepr, Dimension, OwnedArcRepr, OwnedRepr, RawViewRepr, ViewRepr,
+};
 
 /// Array representation trait.
 ///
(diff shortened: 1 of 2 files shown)


bluss:
We want to merge this right? I think the name will be ok

jturner314:
Yes, I still think this is a good idea, and I agree that the name is good enough.

bluss:
Great, thanks!

PR #1024, opened 2021-06-05
Update approx to >=0.4, <0.6 to support approx 0.5

jturner314:
`approx` 0.5 updates its `num-complex` dependency to 0.4, which is the same version of `num-complex` which `ndarray` depends on. So, it's very useful for `ndarray` to support `approx` 0.5. Instead of expanding the range of the `approx` versions, another option would be to add a separate feature for `approx` 0.5, but I don't think the additional complexity would be worth it. When we release the next breaking release of `ndarray`, I think we should change this version specification to "0.5".

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,7 @@ num-complex = { version = "0.4", default-features = false }
 # Use via the `rayon` crate feature!
 rayon_ = { version = "1.0.3", optional = true, package = "rayon" }
 
-approx = { version = "0.4", optional = true , default-features = false }
+approx = { version = ">=0.4, <0.6", optional = true , default-features = false }
 
 # Use via the `blas` crate feature!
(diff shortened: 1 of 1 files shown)


bluss:
Isn't the effect of this change de facto breaking, too?

jturner314:
Huh, this approach doesn't work well. It seems that if a crate depends on `ndarray` (this branch) and `approx 0.4`, then Cargo handles this by using `approx 0.5` for `ndarray` and `approx 0.4` for the crate, even though both are compatible with `approx 0.4`.

jturner314:
That's very unfortunate. It seems like the only non-breaking option is to add a separate feature for `approx` 0.5.

jturner314:
The approach in this PR doesn't work. I've created #1025 as a replacement.

Issue #1027, opened 2021-06-05
Refine conversions between array types.


jturner314:
- Move `into_shared` from `DataOwned` to `Data`, add an `A: Clone` bound, and drop the "without any copying" guarantee. (See the discussion in #1021.)

- Implement `From` for more combinations of array types.

bluss:
I'm not sure about going from DataOwned to Data. I guess there are many alternatives for how it could work. I was first thinking of - unsure if we can realize it - a conversion of owned arrays only, which would move if needed otherwise clone.

jturner314:
Okay, yeah, leaving it on `DataOwned` would be fine. We'd still add an `A: Clone` bound, and drop the "without any copying" guarantee. `OwnedRepr`, `OwnedArcRepr`, and the `Owned` variant of `CowRepr` would implement this without copying, but the `View` variant of `CowRepr` would copy.

PR #1028, opened 2021-06-05
Make `CowArray` an owned storage array, require Clone bound for `into_shared`

jturner314:
Add `A: Clone` bound to `into_shared` and implement `DataOwned` for `CowRepr`

See #1027 regarding the `A: Clone` bound for `into_shared`.

I've marked this PR as a draft because it would be good to add some tests using `CowArray` for methods with a `S: DataOwned` bound.


bluss:
rebased

bluss (approved):

bluss:
Thanks for this. Accepting without new tests :wink:

Issue #1032, opened 2021-06-14
Use intra-doc links


LeSeulArtichaut:
[Intra-doc links are now stable since 1.48.0 :tada:](https://blog.rust-lang.org/2020/11/19/Rust-1.48.html#easier-linking-in-rustdoc) Using them would simplify docs.
cc #558

LeSeulArtichaut:
(I can work on implementing this)

bluss:
Contributions are welcome. We use doc links in some places already.

PR #1033, opened 2021-06-15
Replace some links with intra-doc links

LeSeulArtichaut:
Helps with #1032.
Some links can't be fully transformed into intra-doc links because of `rustdoc` issues rust-lang/rust#80896 and rust-lang/rust#86120.


ndarray-rand/src/lib.rs
```diff
@@ -60,7 +60,7 @@ pub mod rand_distr {
 /// Note that `SmallRng` is cheap to initialize and fast, but it may generate
 /// low-quality random numbers, and reproducibility is not guaranteed. See its
 /// documentation for information. You can select a different RNG with
-/// [`.random_using()`](#tymethod.random_using).
+/// [`.random_using()`](Self::RandomUsing).
```
bluss, line 63: this seems to be a typo (should not be camelcase)

PR #1040, opened 2021-06-25
add mapv_into_any(), resolves #1031

### LINKED ISSUE #1031: Efficiently map array to a (possibly) different type
[`ArrayBase::mapv_into()`](https://docs.rs/ndarray/0.15.3/ndarray/struct.ArrayBase.html#method.mapv_into) consumes self and reuses its memory allocation to compute a new array with the same shape and type.  Avoiding unnecessary allocations is important when mapping very large arrays.

When writing generic methods, it is often necessary to assume that two arrays may be of different types even though we might expect them to be the same type most of the time.  For example, a method that is generic over complex numbers may be called with all real arguments, all complex arguments, or a mixture of real and complex arguments.  This precludes direct use of `mapv_into()` since we cannot assume the arguments are the same type.

Would it be worth adding something like this [`trans_mapv_into()`](https://gist.github.com/benkay86/b9f3eea271999eccf158bfbbcd30b70e), which maps an array with elements of one type to a (possibly) different type?  If the types are the same it delegates to `mapv_into()`, and if the types are different it falls back to using additional allocations.  Type comparison is performed at runtime pending [specialization](https://github.com/rust-lang/rust/issues/31844).

benkay86:
First attempt at pull request for #1031.

Since Rust does not yet have [specialization](https://github.com/rust-lang/rfcs/pull/1210), mapping between generic numerical types in ndarray is challenging when the types *might* be different.  For example, we might want to perform on operation on an array of real numbers to get an array that might be real or might be complex.  We can do this now with `ArrayBase::mapv()`, but if it turns out that both the input and output are the same type (e.g. real numbers) then `mapv()` will perform unnecessary allocation of a new array compared to `mapv_into()`.

This PR proposes a generic method `ArrayBase::mapv_into_any()` whose signature accepts a closure mapping between different types `FnMut(A) -> B` but which uses memory-efficient `mapv_into()` when `A` and `B` are the same type.  The type comparison is performed at runtime.  This is the most memory-efficient way to map between arrays of *possibly* different types using stable Rust.

The use of `unsafe` is discussed [on the Rust language forum](https://users.rust-lang.org/t/runtime-specialization-when-types-are-equal/60628).

### DIFF
--- a/src/impl_methods.rs
+++ b/src/impl_methods.rs
@@ -2473,6 +2473,45 @@ where
         self
     }
 
+    /// Consume the array, call `f` by **v**alue on each element, and return an
+    /// owned array with the new values. Works for **any** `F: FnMut(A)->B`.
+    ///
+    /// If `A` and `B` are the same type then the map is performed by delegating
+    /// to [`mapv_into()`] and then converting into an owned array. This avoids
+    /// unnecessary memory allocations in [`mapv()`].
+    ///
+    /// If `A` and `B` are different types then a new array is allocated and the
+    /// map is performed as in [`mapv()`].
+    ///
+    /// Elements are visited in arbitrary order.
+    pub fn mapv_into_any<B, F>(self, mut f: F) -> Array<B, D>
+    where
+        S: DataMut,
+        F: FnMut(A) -> B,
+        A: Clone + 'static,
+        B: 'static,
+    {
+        if core::any::TypeId::of::<A>() == core::any::TypeId::of::<B>() {
+            // A and B are the same type.
+            // Wrap f in a closure of type FnMut(A) -> A .
+            let f = |a| {
+                let b = f(a);
+                // Safe because A and B are the same type.
+                unsafe { unlimited_transmute::<B, A>(b) }
+            };
+            // Delegate to mapv_into() using the wrapped closure.
+            // Convert output to a uniquely owned array of type Array<A, D>.
+            let output = self.mapv_into(f).into_owned();
+            // Change the return type from Array<A, D> to Array<B, D>.
+            // Again, safe because A and B are the same type.
+            unsafe { unlimited_transmute::<Array<A, D>, Array<B, D>>(output) }
+        } else {
+            // A and B are not the same type.
+            // Fallback to mapv().
+            self.mapv(f)
+        }
+    }
+
     /// Modify the array in place by calling `f` by mutable reference on each element.
     ///
     /// Elements are visited in arbitrary order.
--- a/tests/array.rs
+++ b/tests/array.rs
@@ -990,6 +990,20 @@ fn map1() {
 }
 
 #[test]
+fn mapv_into_any_same_type() {
+    let a: Array<f64, _> = array![[1., 2., 3.], [4., 5., 6.]];
+    let a_plus_one: Array<f64, _> = array![[2., 3., 4.], [5., 6., 7.]];
+    assert_eq!(a.mapv_into_any(|a| a + 1.), a_plus_one);
+}
+
+#[test]
+fn mapv_into_any_diff_types() {
+    let a: Array<f64, _> = array![[1., 2., 3.], [4., 5., 6.]];
+    let a_even: Array<bool, _> = array![[false, true, false], [true, false, true]];
+    assert_eq!(a.mapv_into_any(|a| a.round() as i32 % 2 == 0), a_even);
+}
+
+#[test]
 fn as_slice_memory_order_mut_arcarray() {
     // Test that mutation breaks sharing for `ArcArray`.
     let a = rcarr2(&[[1., 2.], [3., 4.0f32]]);


benkay86:
Thank you for granting permission to run git workflows on this pull request.  It looks like clippy on the beta channel rejects a large portion of the ndarray codebase, but it appears that this pull request has passed all checks.

bluss:
It just looks like clippy found new lints in a new release

src/impl_methods.rs
```diff
@@ -2473,6 +2473,45 @@ where
...
+                let b = f(a);
+                // Safe because A and B are the same type.
+                unsafe { core::ptr::read(&*ManuallyDrop::new(b) as *const B as *const A) }
+            };
+            // Delegate to mapv_into() using the wrapped closure.
+            // Convert output to a uniquely owned array of type Array<A, D>.
+            let output = self.mapv_into(f).into_owned();
+            // Change the return type from Array<A, B> to Array<B, D>.
+            // Again, safe because A and B are the same type.
+            unsafe { core::ptr::read(&*ManuallyDrop::new(output) as *const Array<A, D> as *const Array<B, D>) }
```
bluss, line 2507: We happen to have a function for this in this file - called `unlimited_transmute`. :)

bluss (commented):
Looks solid to me, should be useful

src/impl_methods.rs
benkay86: Yeah, I noticed that method in the codebase used in the context of transmuting one dimensionality into another, but I wasn't sure whether or not I should rely on it.  Do you want me to refactor to use `unlimited_transmute()`?

The problem with `unlimited_transmute()` in this setting is the additional runtime check:

```rust
assert_eq!(size_of::<A>(), size_of::<B>());
```

Asserting that the size of the underlying types is equal is unnecessary because `mapv_into_any()` has already checked that the types themselves are equal (in which case their size is also equal).  The extra runtime assertion is expensive in this case because it will be called inside a tight loop on each element of the array being mapped.

bluss: I agree about ensuring no runtime overhead. `size_of` is known by the compiler at compile time, and the assertion disappears, is compiled out, if it doesn't trigger in this case.

bluss:
Please use issue closing keywords in the pr description so that the linked issue is closed when this is merged https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue

src/impl_methods.rs
benkay86: You are correct, the [runtime assertion is optimized away](https://rust.godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(fontScale:14,fontUsePx:'0',j:1,lang:rust,selection:(endColumn:12,endLineNumber:5,positionColumn:12,positionLineNumber:5,selectionStartColumn:12,selectionStartLineNumber:5,startColumn:12,startLineNumber:5),source:'use+std::mem::%7BManuallyDrop,+size_of%7D%3B%0Ause+std::ptr%3B%0A%0A//+Type+your+code+here,+or+load+an+example.%0Apub+unsafe+fn+unlimited_transmute%3CA,+B%3E(a:+A)+-%3E+B+%7B%0A++++assert_eq!!(size_of::%3CA%3E(),+size_of::%3CB%3E())%3B%0A++++ptr::read(%26*ManuallyDrop::new(a)+as+*const+A+as+*const+B)%0A%7D%0A%0A//+If+you+use+%60main()%60,+declare+it+as+%60pub%60+to+see+it+in+the+output:%0Apub+fn+main()+%7B%0A++++let+a:+i32+%3D+42%3B%0A++++let+_b:+i32+%3D+unsafe+%7B+unlimited_transmute(a)+%7D%3B%0A%7D%0A'),l:'5',n:'0',o:'Rust+source+%231',t:'0')),k:33.333333333333336,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:r1530,filters:(b:'0',binary:'1',commentOnly:'0',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1'),fontScale:14,fontUsePx:'0',j:1,lang:rust,libs:!(),options:'-C++++++++++++++++opt-level%3D1',selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'rustc+1.53.0+(Editor+%231,+Compiler+%231)+Rust',t:'0')),k:33.333333333333336,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:output,i:(compiler:1,editor:1,fontScale:14,fontUsePx:'0',wrap:'1'),l:'5',n:'0',o:'Output+of+rustc+1.53.0+(Compiler+%231)',t:'0')),k:33.33333333333333,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4) at `-C opt-level=1` or higher.  It is obviously not optimized out for debug builds with optimization turned off.

I'll refactor to use `unlimited_transmute()`.

benkay86:
Fixed commit comment.

bluss:
FWIW, we already use this kind of ad-hoc specialization in ndarray with the linalg code, so it feels familiar to us.

bluss:
(milestone - not known if we will have 0.15.4 or 0.16 atm). Thanks!

Issue #415, opened 2018-02-06
Using MKL vectorized functions for things like tanh, exp, log, ...


usamec:
I found, that using MKL library functions like vsTanh (https://software.intel.com/en-us/mkl-developer-reference-fortran-v-tanh) is quite faster than doing `vector.mapv(|x| x.tanh())`. 

Is it worth including this in ndarray crate behind feature gate?

bluss:
What's the Rust crate that provides bindings to this, and as a first step, what is the general level interface or functionality in ndarray that is needed to implement this? Access to ndarray data that can be used for this, but is still independent of for example the MKL.

usamec:
The situation around this is little bit messy.
There is https://crates.io/crates/intel-mkl-src crate, but no direct bindings for MKL.

Right now, go get working solution one has to reexport MKL functions:
```
extern "C" {
    fn vsTanh(n: c_int, a: *const c_float, y: *mut c_float);
}
```

And then instead of code mentioned above, one has to do:
```
unsafe {
    let ptr = vector.as_mut_ptr();
    vsTanh(vector_len, ptr, ptr);
}
```

What I was thinking about was to add .exp(), .log(), .tanh(), ... methods for ndarrays, which depending whether we have MKL feature would have faster implementation.

But, if you say, that this is very low priority enhacement and not worth the effort, I would understand.

termoshtt:
It sounds good.
I think it is good idea to define `tanh` and other basic math functions to allow vectorize.

```rust
pub trait VectorizedMath {
  fn tanh(self) -> Self;  // or vtanh?
  fn exp(self) -> Self;
  ...
}

impl VectorizedMath for ArrayBase<S, D> { /* like @usamec done */ }
```

It is basically independent from MKL. We will be able to implement them using SIMD extensions e.g. [vsinf] in accelerate (macOS). But this must be depends on the platform where the application works, and we must switch its backend, e.g. intel-mkl-src for Intel CPU and accelerate-src for macOS, and so on. 

One way to implement it is a feature flag "intel-mkl-vectorized" to ndarray crate, which enables `VectorizedMath` trait. We can separate this part as "ndarray-vectorized" sub-crate like ndarray-rand.

Another way is to create a binding of MKL "mkl-sys", which includes functions not included in blas-sys, lapack-sys, and fftw-sys. It should be argued on https://github.com/termoshtt/rust-intel-mkl

[...]

PR #1042, opened 2021-07-13
Provide element-wise math functions for floats

### LINKED ISSUE #992: Provide element-wise math functions for floats
This is an improvement to reduce `Array::mapv` calling.

```rust
use num_traits::Float;

// For Array<Float>
impl<A, S, D> ArrayBase<S, D>
where
    A: Float,
    S: RawData<Elem = A>,
    D: Dimension,
{
    pub fn square(&self) -> Array<A, D> {
        self.mapv(|v| v * v)
    }
    pub fn sqrt(&self) -> Array<A, D> {
        self.mapv(A::sqrt)
    }
    // abs, pow, sin, cos, round, ceil, etc.
    ...
}
```

The following is a little part of my code, a lot of mappings were used here.
(solved by a local plug-in trait)

```rust
let dt = dxy.square().sum_axis(Axis(1)).sqrt();
let phi_n = &phi * n as f64;
let cos_phi_n = (phi_n.slice(s![1..]).cos() - phi_n.slice(s![..-1]).cos()) / &dt;
let sin_phi_n = (phi_n.slice(s![1..]).sin() - phi_n.slice(s![..-1]).sin()) / &dt;
```

I hope it can be officially supported.

KmolYuan:
Closes #992.

Add a new module `impl_float_maths` under `std` feature.

If there are missing anything you like, or something should not comes here, any suggestion are welcome!

For integer things, maybe it belongs to another PR.

Three types of operators done by three local macros:

+ `boolean_op`: Returns a boolean array, join them together by `*_any` methods.
+ `unary_op`: Unary functions.
+ `binary_op`: Binary functions.

New functions I made:

+ is_nan_any: A "any" operation under `is_nan`.
+ is_infinite_any: A "any" operation under `is_infinite_any`.
+ square: A `x*x` shortcut.
+ clip: Same as [Numpy `clip`](https://numpy.org/doc/stable/reference/generated/numpy.clip.html), done by using `Float::max` and `Float::min`. This function has a doctest, copied from Numpy.

Ported functions: (here is the [`Float` trait](https://docs.rs/num-traits/0.2.14/num_traits/float/trait.Float.html))

+ is_nan
+ is_infinite
+ floor
+ ceil
+ round
+ trunc
+ fract
+ abs
+ signum
+ recip
+ powi
+ powf
+ sqrt
+ exp
+ exp2
+ ln
+ log
+ log2
+ log10
+ abs_sub
+ cbrt
+ sin
+ cos
+ tan
+ to_degrees
+ to_radians

### DIFF
--- a/src/impl_float_maths.rs
+++ b/src/impl_float_maths.rs
@@ -5,33 +5,33 @@ use num_traits::Float;
 use crate::imp_prelude::*;
 
 macro_rules! boolean_op {
-    ($(#[$meta1:meta])* fn $id1:ident $(#[$meta2:meta])* fn $id2:ident -> $func:ident) => {
-        $(#[$meta1])*
+    ($($(#[$meta1:meta])* fn $id1:ident $(#[$meta2:meta])* fn $id2:ident -> $func:ident)+) => {
+        $($(#[$meta1])*
         pub fn $id1(&self) -> Array<bool, D> {
             self.mapv(A::$func)
         }
         $(#[$meta2])*
         pub fn $id2(&self) -> bool {
             self.mapv(A::$func).iter().any(|&b|b)
-        }
+        })+
     };
 }
 
 macro_rules! unary_op {
-    ($(#[$meta:meta])* fn $id:ident) => {
-        $(#[$meta])*
+    ($($(#[$meta:meta])* fn $id:ident)+) => {
+        $($(#[$meta])*
         pub fn $id(&self) -> Array<A, D> {
             self.mapv(A::$id)
-        }
+        })+
     };
 }
 
 macro_rules! binary_op {
-    ($(#[$meta:meta])* fn $id:ident($ty:ty)) => {
-        $(#[$meta])*
+    ($($(#[$meta:meta])* fn $id:ident($ty:ty))+) => {
+        $($(#[$meta])*
         pub fn $id(&self, rhs: $ty) -> Array<A, D> {
             self.mapv(|v| A::$id(v, rhs))
-        }
+        })+
     };
 }
 
@@ -49,8 +49,7 @@ where
         fn is_nan
         /// Return `true` if any element is `NaN` (not a number).
         fn is_nan_any -> is_nan
-    }
-    boolean_op! {
+
         /// If the number is infinity, then `true` is returned for each element.
         fn is_infinite
         /// Return `true` if any element is infinity.
@@ -59,110 +58,87 @@ where
     unary_op! {
         /// The largest integer less than or equal to each element.
         fn floor
-    }
-    unary_op! {
+
         /// The smallest integer less than or equal to each element.
         fn ceil
-    }
-    unary_op! {
+
         /// The nearest integer of each element.
         fn round
-    }
-    unary_op! {
+
         /// The integer part of each element.
         fn trunc
-    }
-    unary_op! {
+
         /// The fractional part of each element.
         fn fract
-    }
-    unary_op! {
+
         /// Absolute of each element.
         fn abs
-    }
-    unary_op! {
+
         /// Sign number of each element.
         ///
         /// + `1.0` for all positive numbers.
         /// + `-1.0` for all negative numbers.
         /// + `NaN` for all `NaN` (not a number).
         fn signum
-    }
-    unary_op! {
+
         /// The reciprocal (inverse) of each element, `1/x`.
         fn recip
+
+        /// Square root of each element.
+        fn sqrt
+
+        /// `e^x` of each element. (Exponential function)
+        fn exp
+
+        /// `2^x` of each element.
+        fn exp2
+
+        /// Natural logarithm of each element.
+        fn ln
+
+        /// Base 2 logarithm of each element.
+        fn log2
+
+        /// Base 10 logarithm of each element.
+        fn log10
+
+        /// Cubic root of each element.
+        fn cbrt
+
+        /// Sine of each element. (in radians)
+        fn sin
+
+        /// Cosine of each element. (in radians)
+        fn cos
+
+        /// Tangent of each element. (in radians)
+        fn tan
+
+        /// Converts radians to degrees for each element.
+        fn to_degrees
+
+        /// Converts degrees to radians for each element.
+        fn to_radians
     }
     binary_op! {
         /// Integer power of each element.
         ///
         /// This function is generally faster than using float power.
         fn powi(i32)
-    }
-    binary_op! {
+
         /// Float power of each element.
         fn powf(A)
+
+        /// Logarithm of each element with respect to an arbitrary base.
+        fn log(A)
+
+        /// The positive difference between given number and each element.
+        fn abs_sub(A)
     }
 
     /// Square of each element.
     pub fn square(&self) -> Array<A, D> {
         self.mapv(|v| v * v)
-    }
-
-    unary_op! {
-        /// Square root of each element.
-        fn sqrt
-    }
-    unary_op! {
-        /// `e^x` of each element. (Exponential function)
-        fn exp
-    }
-    unary_op! {
-        /// `2^x` of each element.
-        fn exp2
-    }
-    unary_op! {
-        /// Natural logarithm of each element.
-        fn ln
-    }
-    binary_op! {
-        /// Logarithm of each element with respect to an arbitrary base.
-        fn log(A)
-    }
-    unary_op! {
-        /// Base 2 logarithm of each element.
-        fn log2
-    }
-    unary_op! {
-        /// Base 10 logarithm of each element.
-        fn log10
-    }
-    binary_op! {
-        /// The positive difference between given number and each element.
-        fn abs_sub(A)
-    }
-    unary_op! {
-        /// Cubic root of each element.
-        fn cbrt
-    }
-    unary_op! {
-        /// Sine of each element. (in radians)
-        fn sin
-    }
-    unary_op! {
-        /// Cosine of each element. (in radians)
-        fn cos
-    }
-    unary_op! {
-        /// Tangent of each element. (in radians)
-        fn tan
-    }
-    unary_op! {
-        /// Converts radians to degrees for each element.
-        fn to_degrees
-    }
-    unary_op! {
-        /// Converts degrees to radians for each element.
-        fn to_radians
     }
 
     /// Limit the values for each element.
--- a/src/impl_float_maths.rs
+++ b/src/impl_float_maths.rs
@@ -17,7 +17,7 @@ macro_rules! boolean_op {
     };
 }
 
-macro_rules! map_op {
+macro_rules! unary_op {
     ($(#[$meta:meta])* fn $id:ident) => {
         $(#[$meta])*
         pub fn $id(&self) -> Array<A, D> {
@@ -26,7 +26,7 @@ macro_rules! map_op {
     };
 }
 
-macro_rules! bin_op {
+macro_rules! binary_op {
     ($(#[$meta:meta])* fn $id:ident($ty:ty)) => {
         $(#[$meta])*
         pub fn $id(&self, rhs: $ty) -> Array<A, D> {
@@ -56,31 +56,31 @@ where
         /// Return `true` if any element is infinity.
         fn is_infinite_any -> is_infinite
     }
-    map_op! {
+    unary_op! {
         /// The largest integer less than or equal to each element.
         fn floor
     }
-    map_op! {
+    unary_op! {
         /// The smallest integer less than or equal to each element.
         fn ceil
     }
-    map_op! {
+    unary_op! {
         /// The nearest integer of each element.
         fn round
     }
-    map_op! {
+    unary_op! {
         /// The integer part of each element.
         fn trunc
     }
-    map_op! {
+    unary_op! {
         /// The fractional part of each element.
         fn fract
     }
-    map_op! {
+    unary_op! {
         /// Absolute of each element.
         fn abs
     }
-    map_op! {
+    unary_op! {
         /// Sign number of each element.
         ///
         /// + `1.0` for all positive numbers.
@@ -88,17 +88,17 @@ where
         /// + `NaN` for all `NaN` (not a number).
         fn signum
     }
-    map_op! {
+    unary_op! {
         /// The reciprocal (inverse) of each element, `1/x`.
         fn recip
     }
-    bin_op! {
+    binary_op! {
         /// Integer power of each element.
         ///
         /// This function is generally faster than using float power.
         fn powi(i32)
     }
-    bin_op! {
+    binary_op! {
         /// Float power of each element.
         fn powf(A)
     }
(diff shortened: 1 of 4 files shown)


bluss:
Looks to be in the right direction, that's nice - I'll be a bit unreachable during the summer, but back in a bit

src/impl_float_maths.rs
```diff
@@ -0,0 +1,157 @@
...
+        /// Base 2 logarithm of each element.
+        fn log2
+
+        /// Base 10 logarithm of each element.
+        fn log10
+
+        /// Cubic root of each element.
+        fn cbrt
+
+        /// Sine of each element. (in radians)
```
multimeric, line 108: The full stop should be at the end of the sentence here, as it is in `num_traits`

multimeric (commented):
This looks like a great PR, as the lack of these operations has been a major pain point for me trying to implement mathematical algorithms.

It's unfortunate the maintainers haven't had a chance to look at this, ideally there would be a few more people able to review PRs. In any case I'm giving some feedback in the hope that it might speed up the ultimate review process once it happens.

Some miscellaneous thoughts:

* Could you actually implement the `Float` trait for an ndarray? You seem to have provided implementations for most if not all of the required methods.
* I assume we would need (or at least, should have) some tests for all of the new methods, and not just `clip`

src/impl_float_maths.rs
```diff
@@ -0,0 +1,157 @@
+//! Element-wise methods for ndarray
+
+use num_traits::Float;
+
+use crate::imp_prelude::*;
+
+macro_rules! boolean_op {
```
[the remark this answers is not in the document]
KmolYuan: > What is your justification for this overarching macro that takes multiple function stubs and outputs multiple functions, versus the more common single-function macro like ndarray's impl_binary_op!?

I'm prefer to use keyword syntax (`fn`, `impl`) in the implementation. In my opinion, function-like calling seems to be a brief-function of the expressions or statements. Since there are many docstrings displayed, perhaps using `fn` keyword is a better syntax (docstring is a kind of attribute macros).

KmolYuan:
> Could you actually implement the Float trait for an ndarray? You seem to have provided implementations for most if not all of the required methods.

The main reason is that these methods need to import `Float` trait, otherwise they are not allowed to be used. The `Float` trait is still has some requirement, I think this is not easy to solve, such as the [`Num` trait](https://docs.rs/num-traits/0.2.14/num_traits/trait.Num.html).

> Although these are not part of the Float trait, do you think there is room for greater_than, less_than, and the rest of the family for a scalar operand? ie array![1, 2].greater_than(1) -> array![false, true]

I'm willing to add more useful methods as you mentioned, if they are compatible with the requirements.

src/impl_float_maths.rs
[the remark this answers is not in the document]
multimeric: If I'm understanding correctly, you prefer this type of macros that *don't* behave like functions? (and indeed this one does not behave like a function).

I'm pretty sure you can easily write docstrings with either approach, though. You would just:
```rust
/// The ceiling function
impl_float_op!(ceil); 
```

Anyway don't change this on my account, we'll have to see what the maintainers say.

multimeric:
> The Float trait is still has some requirement, I think this is not easy to solve, such as the Num trait.

Ah yes you're right, implementing `Float` requires implementing `Num`, which requires `from_str_radix(str: &str, radix: u32) -> Result<Self, Self::FromStrRadixErr>` which doesn't make sense for an array. It also has to implement `NumCast` which doesn't make sense either.

> I'm willing to add more useful methods as you mentioned, if they are compatible with the requirements.

What requirements are you talking about?

KmolYuan:
> What requirements are you talking about?

Oh, I mean the functions should be implemented with the `Float` trait.

multimeric:
Looks like that won't be a problem if we don't bother with implementing `Float` per se.

KmolYuan:
Additionally, I think there should be add a `#[must_use]` notation, same as the standard library.

```rust
#[must_use = "method returns a new array and does not mutate the original value"]
```

src/impl_float_maths.rs
```diff
@@ -0,0 +1,157 @@
...
+        /// Logarithm of each element with respect to an arbitrary base.
+        fn log(A)
+
+        /// The positive difference between given number and each element.
+        fn abs_sub(A)
+    }
+
+    /// Square of each element.
+    pub fn square(&self) -> Array<A, D> {
+        self.mapv(|v| v * v)
```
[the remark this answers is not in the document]
KmolYuan: I'm not sure about multiplication in Rust is faster than integer powers. Is there any reference?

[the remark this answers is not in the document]
jturner314: Fwiw, I agree that multiplication is probably the better implementation. I'd probably call it `pow2` instead of `square`, to minimize confusion with "square" arrays.

src/impl_float_maths.rs
```diff
@@ -0,0 +1,157 @@
...
+    ///
+    /// ```
+    /// use ndarray::{Array1, array};
+    ///
+    /// let a = Array1::range(0., 10., 1.);
+    /// assert_eq!(a.clip(1., 8.), array![1., 1., 2., 3., 4., 5., 6., 7., 8., 8.]);
+    /// assert_eq!(a.clip(8., 1.), array![1., 1., 1., 1., 1., 1., 1., 1., 1., 1.]);
+    /// assert_eq!(a.clip(3., 6.), array![3., 3., 3., 3., 4., 5., 6., 6., 6., 6.]);
+    /// ```
+    pub fn clip(&self, min: A, max: A) -> Array<A, D> {
```
jturner314, line 154: I'd hold off including this until `num-traits` adds a `clamp` method to `Float`. In particular, the behavior should be the same as `std`'s `clamp` methods (on `f32`, `f64`, and `Ord`) in the presence of NaNs and the `min > max` case.

src/impl_float_maths.rs
```diff
@@ -0,0 +1,157 @@
...
+where
+    A: Float,
+    S: RawData<Elem = A> + Data,
+    D: Dimension,
+{
+    boolean_op! {
+        /// If the number is `NaN` (not a number), then `true` is returned for each element.
+        fn is_nan
+        /// Return `true` if any element is `NaN` (not a number).
+        fn is_nan_any -> is_nan
```
jturner314, line 51: I'd suggest calling this `is_any_nan` instead.

src/impl_float_maths.rs
```diff
@@ -0,0 +1,157 @@
...
+
+        /// If the number is infinity, then `true` is returned for each element.
+        fn is_infinite
+        /// Return `true` if any element is infinity.
+        fn is_infinite_any -> is_infinite
```
jturner314, line 56: I'd suggest calling this `is_any_infinite` instead.

src/impl_float_maths.rs
```diff
@@ -0,0 +1,157 @@
...
+/// # Element-wise methods for Float Array
+///
+/// Element-wise math functions for any array type that contains float number.
+impl<A, S, D> ArrayBase<S, D>
+where
+    A: Float,
+    S: RawData<Elem = A> + Data,
+    D: Dimension,
+{
+    boolean_op! {
```
jturner314, line 47: If we add these methods, we should also consider adding `is_finite`, `is_any_finite`, and `is_all_finite`.

jturner314 (commented):
This mostly looks good to me. I added a few comments.

Long-term, I think we should make these methods lazy instead of eager, but that will require a decent amount of additional infrastructure. There seems to be a moderate demand for these methods today, so it's probably worth including the eager versions now and changing them to lazy in a later breaking change.

src/impl_float_maths.rs
KmolYuan: In NumPy, the boolean arrays can be converted to a single `bool` value by `all` and `any` method.

```python
if (a > b).any():
    ...
```

Instead of the case of the [`Iterator::all`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.all) and the [`Iterator::any`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.any), would it be better to add `all` and `any` methods for boolean arrays?

src/impl_float_maths.rs
bluss: Performance is also an issue and the current implementation would probably need tweaking

src/impl_float_maths.rs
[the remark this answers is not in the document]
bluss: Why should we have a method for squaring each element? I understand the motivation of this PR but this seems a bit too useless. It's easy enough to call mapv in this case? There is no .square() on f32, so no precedent there. :slightly_smiling_face:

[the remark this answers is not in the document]
multimeric: Square/sqrt is so common in scientific/statistical computing that I would like to have an implementation that does so efficiently and concisely. `powi(2)` and `powf(0.5)` are pleasantly concise, but if they are not efficient (and it looks like the consensus is that they aren't), then I would prefer an implementation that encourages users to produce faster code as is being done here.

src/impl_float_maths.rs
```diff
@@ -0,0 +1,162 @@
...
+    ($($(#[$meta:meta])* fn $id:ident($ty:ty))+) => {
+        $($(#[$meta])*
+        #[must_use = "method returns a new array and does not mutate the original value"]
+        pub fn $id(&self, rhs: $ty) -> Array<A, D> {
+            self.mapv(|v| A::$id(v, rhs))
+        })+
+    };
+}
+
+/// # Element-wise methods for Float Array
```
bluss, line 42:
```suggestion
/// # Element-wise methods for float arrays
```

src/impl_float_maths.rs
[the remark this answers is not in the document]
bluss: The current macros look smart, so whatever you like @KmolYuan. If you want you can rename them with plural, "binary_ops" since they produce multiple methods (maybe even make it "binary_methods"?)

src/impl_float_maths.rs
bluss: I'd prefer .all() or .any() to be implemented for all arrays - taking a closure, just like Iterator. Good separate PR. See comment in this PR for how to use existing `Zip::all`, i.e. the functionality already exists.

bluss:
I haven't added my general perspective, I think.

We have avoided these methods in the past for two reasons:

- Focus on general ndarray features, not specific ones. `mapv` covers all of these. But I agree that these are useful, if we have contributors that want to maintain them
- Focus on performance. These methods are not efficient, because they create new arrays - allocation and copying - for all operations. But they are convenient, and a lot of times the user doesn't need to care. But it's something to be aware of and keep in mind. We need to explain this to users as well, and we try in various places in the documentation. We provide things like `Zip` that allows the user to modify and transform elements in-place.

src/impl_float_maths.rs
```diff
@@ -0,0 +1,162 @@
...
+        fn to_radians
+    }
+    binary_op! {
+        /// Integer power of each element.
+        ///
+        /// This function is generally faster than using float power.
+        fn powi(i32)
+
+        /// Float power of each element.
+        fn powf(A)
```
bluss, line 134: Food for thought: This method is "powf(array, scalar)". We need a separate name for `powf(array, array)`. What should the name of that method be?  I don't mean that we should have that in this PR, but sometime down the line.

src/lib.rs
```diff
@@ -1535,6 +1535,8 @@ mod impl_internal_constructors;
 mod impl_constructors;
 
 mod impl_methods;
+#[cfg(feature = "std")]
+mod impl_float_maths;
```
bluss, line 1539: This line should move down. It should be below `mod numeric`. The order here determines the order of methods in the docs, and my suggestion is down there. We might actually put this inside mod numeric, below impl_numeric?

multimeric:
> These methods are not efficient, because they create new arrays - allocation and copying - for all operations. But they are convenient, and a lot of times the user doesn't need to care.

Could we not solve this by having in-place versions of all these methods? It would also force users to decide which one they need in each case instead of always picking the simpler code. Should be fairly simple to add by updating the macros.

bluss:
In-place versions are just a band aid, not a solution . Actual solutions does most the needed operations in one pass, not one by one (how do you handle chaining operations). I'd avoid in-place methods for now. 🙂

multimeric:
Is this to do with loading the elements into CPU registers or something? Because otherwise I can't see why processing the whole array in one pass and then again in a second pass is any worse than applying the same transformations to each element sequentially. I would have thought that avoiding a new array allocation would dwarf this optimisation.

In any case if that would be a significant optimisation could we do some kind of chain pattern like:
```rust
arr.chain().powi(2).exp().eval()
```

adamreichold:
> Is this to do with loading the elements into CPU registers or something?

With contemporary systems, moving the data from memory into caches and register often dominates the runtime of numerical algorithms depending on how many arithmetic operations their perform relative to each memory access. Fusing operations from multiple passes into a single pass increases this ratio and thereby the likelihood of efficient CPU utilisation.

bluss:
@multimeric Good that we got to explain processor caches. This will help a lot when working with Rust.

Your point about eval is close to what I'm thinking of too. Not in particular that we do that here, but that we lay the groundwork for a "language", where you can call `.exp()` etc on arrays.

Then a shim/layer can be inserted that has the same or sameish interface but lazy evaluation, like in your sketch! Both the in-place instead of allocating and batching operations instead of one per loop ideas are important here. There's some prior art around this. I don't know eigen (C++) very well but I think they [do something like this](https://en.wikipedia.org/wiki/Expression_templates). And the very old Rust experiment algebloat also tried this.

bluss:
It would be exciting if

1) We traitifed ndarray - not so many specific "Array", "ArrayView" in the signatures
2) A lazy expression could implement the same trait, i.e you could pass array.lazy().exp() in place of an array.

I don't know what kinds of techniques can be used in Rust to make this possible. We also don't want to make the type system situation too hard to understand, unfortunately.

multimeric:
Okay I'll move the discussion of this lazy eval into a new issue for now so we can get this PR merged.

bluss:
I wanted to say, if you can, please use rebase and not merges when updating the PR branch, we don't want merges crossing. There are of course exceptions to every rule, but in general this is what we want to do. I've force-pushed the branch with a rebase, since it was easy to do.

bluss:
In the spirit of issue #415 we might prefer to expose these methods with a bound like `A: 'static + Float`.  `'static` is for the moment the key to making ad-hoc type specialization (this is how we do it for matmul).

ethanhs:
Just to add this also solves https://github.com/rust-ndarray/ndarray/issues/1047 I think.

bluss:
fwiw, no rush, this PR is waiting on updates from the author

killme2008:
Any update of this PR?

fgsch:
This would be a very useful addition.

Is there any more work that needs to happen so this can be merged?

danjenson:
Looking forward to this...

AnotherCoolDude:
What's the status in this PR? Would be great to have all those methods implemented...

nilgoyette:
Same as last time: "this PR is waiting on updates from the author" 

Since this MR has been discussed and reviewed, I wouldn't mind merging it if @KmolYuan updates this MR.

KmolYuan:
> Same as last time: "this PR is waiting on updates from the author"
> 
> Since this MR has been discussed and reviewed, I wouldn't mind merging it if @KmolYuan updates this MR.

Sure, if the API & docs are accepted, I agree to merge this PR.

nilgoyette:
@KmolYuan What about the `clamp` function? Shouldn't we use the one from std?

clamp is stable since 1.50 and the minimum Rust version in this project is 1.51.

nilgoyette:
@adamreichold Any idea why the CI tests fail? You repaired them in the last [merged MR](https://github.com/rust-ndarray/ndarray/pull/1320), so I would have thought that the tests would still be ok. It's still using 1.51 and the errors are not related to this MR.

adamreichold:
> @adamreichold Any idea why the CI tests fail? You repaired them in the last [merged MR](https://github.com/rust-ndarray/ndarray/pull/1320), so I would have thought that the tests would still be ok. It's still using 1.51 and the errors are not related to this MR.

These errors are due to new lints introduced into rustc itself since then. It does check `pub use` in addition to `use` now and checks against collision with possibly upcoming but still unstable names. I think we have to fix the former and suppress the latter for now, c.f. #1337

nilgoyette:
Thank you @KmolYuan, and sorry for the delay.

AnotherCoolDude:
I am sorry to reopen this discussion. For some reason, my project doesn't pick up the merged changes. I would expect a version bump that tells cargo to update the package. I could probably refer to the commit specifically, but that doesn't feel right. 
Is that a mistake on my side?

AnotherCoolDude:
I did `{ version = "0.15.6", features = ["serde"] }`, which doesn't pick up the changes. When I change that to `{ git = "https://github.com/rust-ndarray/ndarray", features = ["serde"] }` and hit `cargo update`, I still don't receive the latest changes, and it creates conflicts with `ndarray_rand` in addition to that (`cargo.lock` shows two entries for `ndarray` at that point, maybe that's the reason). 

If there isn't an easy fix for that, I'll implement the changes myself. But that won't be as sophisticated, and I'd like to make use of the PR. But I also don't want to waste your time with my issues (which are not related to ndarray specifically).

adamreichold:
I think you want to use something like

```toml
[patch.crates-io]
ndarray = { git = "https://github.com/rust-ndarray/ndarray" }
```

to manage the indirect dependencies, c.f. <https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html>

PR #1053, opened 2021-08-02
Add generic pow method for arrays

### LINKED ISSUE #1047: Implement `powi` and `powf` for `ArrayBase`
Currently `a ** 3` is the only example in the "[ndarray_for_numpy_users](https://docs.rs/ndarray/0.15.3/ndarray/doc/ndarray_for_numpy_users/index.html)" docs that requires a map. Could these methods be added directly? Rust has no `^` or `**` exponentiation operator, but these two methods are implemented on numeric primitives, and I see no reason why they can't be added for arrays here for convenience.

multimeric:
Closes #1047.

Originally I was going to make separate `powi()` and `powf()` methods, but this isn't convenient because `numtraits` groups them together using one method.

One open question is how to get an integer array to take a float power and vice-versa, because `numtraits` doesn't seem to implement this use-case.

I've written two tests, which are in the documentation. I hope that's sufficient.


multimeric:
@maintainers, this seems to be failing because of something to do with `#[inline]`, which I haven't touched. Is master broken or did this PR break it in a confusing way?

jturner314:
> @ maintainers, this seems to be failing because of something to do with `#[inline]`, which I haven't touched. Is master broken or did this PR break it in a confusing way?

`master` currently fails CI due to new warnings with Rust nightly. I have a fix in #1054, which I'll merge as soon as CI passes.

multimeric:
Great thanks. Hopefully this will rebuild automatically once that gets merged?

jturner314:
> Great thanks. Hopefully this will rebuild automatically once that gets merged?

Once #1054 is merged, I think you'll need to fetch the latest `master` from GitHub, rebase the `pow` branch onto it, and then force-push `pow` to update this PR. Are you familiar with how to do that, or would you like assistance?

multimeric:
I think I can just pull master in my branch and then push my branch right?

jturner314:
> I think I can just pull master in my branch and then push my branch right?

That works too. The only disadvantage is that the commit history won't be as pretty as with a rebase.

multimeric:
By the way, can I have some advise on the "error: unused import: `Pow`" I'm getting in the build? I am using that trait in my method, but I guess when we don't have the `std` feature enabled the method is skipped and so `Pow` seems to be unused?

multimeric:
> That works too. The only disadvantage is that the commit history won't be as pretty as with a rebase.

Okay I've rebased.

jturner314:
> By the way, can I have some advice on the "error: unused import: `Pow`" I'm getting in the build? I am using that trait in my method, but I guess when we don't have the `std` feature enabled the method is skipped and so `Pow` seems to be unused?

Yes, that's the issue. The simplest fix would be move `use num_traits::Pow;` to a separate line and add `#[cfg(feature = "std")]`.

multimeric:
> Yes, that's the issue. The simplest fix would be move `use num_traits::Pow;` to a separate line and add `#[cfg(feature = "std")]`.

Done, thank you.

multimeric:
Damn, it looks like this is all duplicating a subset of #1042. I wish I knew about that!

multimeric:
Closing in favour of #1042.

PR #1061, opened 2021-09-02
Improve layout heuristic for sum_axis

### LINKED ISSUE #1058: sum_axis very slow when summing on last axis
On my computer

    let a = Array3::<f64>::zeros((500,600,700));
    let b = a.sum_axis(Axis(2));

is about 9 times slower (500ms vs 55ms) than

    let a = Array3::<f64>::zeros((500,600,700));
    let mut c = Array2::<f64>::zeros((500,600));
    Zip::from(&mut c).and(a.lanes(Axis(2))).for_each(|c,a| {
        *c = a.sum();
    });

On the other hand when summing on `Axis(0)` sum_axis is about two times faster than Zip.

The gain seems big enough to me to warrant an optimization but I'm not sure how to deal with intermediate cases (summing on some middle axis). Or can we just add the optimization for the summation on the last axis as it's a very common case ?

 (Of course it's the other way around for a column major array.)

jturner314:
This is an attempt to fix #1058. It would be nice to have some benchmarks to confirm that this is an improvement, but I haven't gotten a chance to write them.

### DIFF
--- a/src/numeric/impl_numeric.rs
+++ b/src/numeric/impl_numeric.rs
@@ -12,7 +12,6 @@ use num_traits::{self, FromPrimitive, Zero};
 use std::ops::{Add, Div, Mul};
 
 use crate::imp_prelude::*;
-use crate::itertools::enumerate;
 use crate::numeric_util;
 
 /// # Numerical Methods for Arrays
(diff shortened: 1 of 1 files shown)


bluss (commented):
Looks good, an improvement in style of implementation too (level of abstraction etc).

PR #1068, opened 2021-09-14
Add `axis_windows` method

### LINKED ISSUE #1041: Add axis_windows_iter
We should add an `axis_windows_iter` method:

```rust
impl<A, S, D> ArrayBase<S, D>
where
    S: RawData<Elem = A>,
    D: Dimension,
{
    /// Returns an iterator which traverses over all windows of a given length along an axis.
    ///
    /// The windows are all distinct, possibly-overlapping views. The shape of each window
    /// is the shape of `self`, with the length of `axis` replaced with `window_size`.
    ///
    /// **Panics** if `axis` is out-of-bounds or if `window_size` is zero.
    ///
    /// ```
    /// use ndarray::Array3;
    ///
    /// let arr = Array3::from_shape_fn([4, 5, 2], |(i, j, k)| i * 100 + j * 10 + k);
    /// let correct = vec![
    ///     arr.slice(s![.., 0..3, ..]),
    ///     arr.slice(s![.., 1..4, ..]),
    ///     arr.slice(s![.., 2..5, ..]),
    /// ];
    /// for (window, correct) in arr.axis_windows_iter(Axis(1), 3).zip(&correct) {
    ///     assert_eq!(window, correct);
    ///     assert_eq!(window.shape(), &[4, 3, 2]);
    /// }
    /// ```
    pub fn axis_windows_iter(&self, axis: Axis, window_size: usize) -> AxisWindowsIter<'_, A, D>
    where
        S: Data,
    { ... }
}
```

Optionally, we could use `NonZeroUsize` instead of `usize` as the type for `window_size`.

Existing, related methods include `axis_chunks_iter` and `windows`.

VasanthakumarV:
Closes #1041.

I have deviated a bit on the name, for consistency with [windows](https://docs.rs/ndarray/0.15.3/ndarray/struct.ArrayBase.html#method.windows) method.

I have not used `NonZeroUsize`, mainly because of the additional effort required from the User.

The doc comments have been taken verbatim from @jturner314's issue comments.

### DIFF
--- a/src/impl_methods.rs
+++ b/src/impl_methods.rs
@@ -1399,6 +1399,50 @@ where
         Windows::new(self.view(), window_size)
     }
 
+    /// Returns a producer which traverses over all windows of a given length along an axis.
+    ///
+    /// The windows are all distinct, possibly-overlapping views. The shape of each window
+    /// is the shape of `self`, with the length of `axis` replaced with `window_size`.
+    ///
+    /// **Panics** if `axis` is out-of-bounds or if `window_size` is zero.
+    ///
+    /// ```
+    /// use ndarray::{Array3, Axis, s};
+    ///
+    /// let arr = Array3::from_shape_fn([4, 5, 2], |(i, j, k)| i * 100 + j * 10 + k);
+    /// let correct = vec![
+    ///     arr.slice(s![.., 0..3, ..]),
+    ///     arr.slice(s![.., 1..4, ..]),
+    ///     arr.slice(s![.., 2..5, ..]),
+    /// ];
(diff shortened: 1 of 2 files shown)


bluss:
Thanks. We just need to drop the clippy fix and it should be fine

bluss:
I fixed the doc since it's not an iterator (but a producer and iterable). Also inserted author credit to @jturner314 for the example

bluss:
It's a producer. It's put on the future enhancements list to make it a separate producer/iterator type, something that can take advantage of the fact that in this case, it's a 1D traversal. Producers are efficient enough for that, should be no problem, but there might be some ergonomics to gain?

PR #1076, opened 2021-09-30
Add from_diag_elem method

jturner314:
This an operation I need fairly frequently and seems useful enough to include.


bluss:
thanks!

PR #1081, opened 2021-10-06
Add Parallel::with_min_len to limit splitting

adamreichold:
Splitting of parallel iterators working on `ArrayView` and `Zip` is currently not limited and continues until only one element is left which can lead to excessive overhead for algorithms formulated in terms of these iterators.

Since these iterators are also not indexed, Rayon's generic `IndexedParallelIterator::with_min_len` does not apply. However, since the number of elements is known and currently checked against one to determine if another split is possible, it appears straight-forward to replace this constant by a parameter and make it available to the user via a `Parallel::with_min_len` inherent method.

### DIFF
--- a/src/parallel/par.rs
+++ b/src/parallel/par.rs
@@ -160,8 +160,14 @@ macro_rules! par_iter_view_wrapper {
         where D: Dimension,
               A: $($thread_bounds)*,
     {
-        /// Sets the minimum number of elements desired to process in each job. This will not be split any smaller than this length, but of course a producer could already be smaller to begin with.
+        /// Sets the minimum number of elements desired to process in each job. This will not be
+        /// split any smaller than this length, but of course a producer could already be smaller
+        /// to begin with.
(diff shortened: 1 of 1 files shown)


src/parallel/par.rs
```diff
@@ -21,11 +21,23 @@ use crate::split_at::SplitPreference;
...
+impl<I> Parallel<I> {
+    pub fn with_min_len(self, min_len: usize) -> Self {
+        Self {
+            min_len,
+            ..self
+        }
+    }
+}
+
+const DEFAULT_MIN_LEN: usize = 1;
```
adamreichold, line 36: It might be useful to make this default larger than one similar to the existing [`COLLECT_MAX_SPLITS`](https://github.com/rust-ndarray/ndarray/blob/87506b8bec7cc9a265289beaa56d0b1b58ffad4f/src/parallel/impl_par_methods.rs#L50).

PR #1082, opened 2021-10-14
Mark Zip as must_use as they are lazy similar to iterators.

adamreichold:


bluss:
Thanks. The clippy failure is unrelated to the PR. :)

PR #1095, opened 2021-10-27
Add a par_fold method to Zip to improve the discoverability of Rayon's fold-reduce idiom.

### LINKED ISSUE #1094: How to do parallel array `.sum`?
Hi thanks for the lib! It seems that `array.sum()` is a bit slow. So I wonder whether there is a way to parallelly sum it? I have lookat at [parallel](https://docs.rs/ndarray/0.15.3/ndarray/parallel/index.html) but it seems only to support `map`, and does not support things like `.fold()`.

adamreichold:
Fixes #1094

### DIFF
--- a/src/parallel/impl_par_methods.rs
+++ b/src/parallel/impl_par_methods.rs
@@ -177,6 +177,56 @@ macro_rules! zip_impl {
                 self.par_map_assign_into(into, f)
             }
 
+            /// Parallel version of `fold`.
+            ///
+            /// Splits the producer in multiple tasks which each accumulate a single value
+            /// using the `fold` closure. Those tasks are executed in parallel and their results
+            /// are then combined to a single value using the `reduce` closure.
+            ///
+            /// The `identity` closure provides the initial values for each of the tasks and
+            /// for the final reduction.
+            ///
+            /// This is a shorthand for calling `self.into_par_iter().fold(...).reduce(...)`.
+            ///
+            /// Note that it is often more efficient to parallelize not per-element but rather
+            /// based on larger chunks of an array like generalized rows and operating on each chunk
+            /// using a sequential variant of the accumulation.
+            /// For example, sum each row sequentially and in parallel, taking advatange of locality
+            /// and vectorization within each task, and then reduce their sums to the sum of the matrix.
+            ///
+            /// Also note that the splitting of the producer into multiple tasks is _not_ deterministic
+            /// which needs to be considered when the accuracy of such an operation is analyzed.
+            ///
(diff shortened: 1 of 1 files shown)


src/parallel/impl_par_methods.rs
```diff
@@ -177,6 +177,48 @@ macro_rules! zip_impl {
...
+            ///
+            /// ## Examples
+            ///
+            /// ```rust
+            /// use ndarray::{Array, Zip};
+            ///
+            /// let a = Array::<usize, _>::ones((128, 1024));
+            /// let b = Array::<usize, _>::ones(128);
+            ///
+            /// let sum = Zip::from(a.rows()).and(b.view()).par_fold(
```
bluss, line 200:
```suggestion
            /// let weighted_sum = Zip::from(&a).and(&b).par_fold(
```

`weighted_sum` is just a suggestion.

bluss: You could if you want sync up even more with the existing `Zip::fold` in terms of "presentation". By no means necessary, just saying what I'm thinking - (see existing ones to see that they are not perfect). If you want, we mention something like par for each's `This is a shorthand for using .into_par_iter().for_each() on Zip.`

bluss (commented):
Seems good - par equivalent of existing Zip::fold. I don't want us to forward everything from parallel iterators to Zip, so let's be conservative with that. It is written in the docs how to use those methods already.

adamreichold:
>  I don't want us to forward everything from parallel iterators to Zip, so let's be conservative with that.

I do agree which is why I only went for the most common combination of `fold` immediately followed by `reduce`. But since my main goal here is discoverability, I think that I should definitely add 

>  If you want, we mention something like par for each's `This is a shorthand for using .into_par_iter().for_each() on Zip`.

I do have one question concerning you suggestion though, in

```rust
let weighted_sum = Zip::from(&a).and(&b).par_fold(
```

you removed the `a.rows()` call. Was this intentional? My intention for the example was to avoid showing element-wise parallelism to place idea in the readers mind that `Zip` and hence the parallel fold should often operate on larger chunks of the underlying data. Should I make that thought explicit in the doc comment?

adamreichold:
> Should I make that thought explicit in the doc comment?

Added suggestion of this could be worded. I am not to invested in that though as there are too many things to consider in each particular situation so that any advice might not apply.

bluss:
Oh sorry about removing the .rows() thing - I was reading way too fast, thought it was .view() too.

bluss (approved):

PR #1105, opened 2021-11-06
Implement Kronecker product

ethanhs:
Picking up from https://github.com/rust-ndarray/ndarray/pull/1039 and based on my original implementation in https://github.com/rust-ndarray/ndarray/issues/652 with the feedback from @bluss.

I found this implementation performs the same vs #1039 modulo using `uninit`, at least for my use case.

### DIFF
--- a/src/linalg/impl_linalg.rs
+++ b/src/linalg/impl_linalg.rs
@@ -710,7 +710,6 @@ where
     S1: Data<Elem = A>,
     S2: Data<Elem = A>,
     A: LinalgScalar,
-    A: std::ops::Mul<&'a ArrayBase<S2, Ix2>, Output = Array<A, Ix2>>,
 {
     let dimar = a.shape()[0];
     let dimac = a.shape()[1];
--- a/src/linalg/impl_linalg.rs
+++ b/src/linalg/impl_linalg.rs
@@ -711,7 +711,7 @@ where
     S1: Data<Elem = A>,
     S2: Data<Elem = A>,
     A: LinalgScalar,
-    A: std::ops::Mul<&'a ArrayBase<S2, Ix2>, Output = ArrayBase<OwnedRepr<A>, Ix2>>,
+    A: std::ops::Mul<&'a ArrayBase<S2, Ix2>, Output = Array<A, Ix2>>,
 {
     let dimar = a.shape()[0];
     let dimac = a.shape()[1];
--- a/src/linalg/impl_linalg.rs
+++ b/src/linalg/impl_linalg.rs
@@ -706,7 +706,7 @@ unsafe fn general_mat_vec_mul_impl<A, S1, S2>(
 ///
 /// The kronecker product of a LxN matrix A and a MxR matrix B is a (L*M)x(N*R)
 /// matrix K formed by the block multiplication A_ij * B.
-pub fn kron<'a, A, S1, S2>(a: &ArrayBase<S1, Ix2>, b: &'a ArrayBase<S2, Ix2>) -> ArrayBase<OwnedRepr<A>, Ix2>
+pub fn kron<'a, A, S1, S2>(a: &ArrayBase<S1, Ix2>, b: &'a ArrayBase<S2, Ix2>) -> Array<A, Ix2>
 where
     S1: Data<Elem = A>,
     S2: Data<Elem = A>,
--- a/src/linalg/impl_linalg.rs
+++ b/src/linalg/impl_linalg.rs
@@ -6,7 +6,6 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
-use crate::OwnedRepr;
 use crate::imp_prelude::*;
 use crate::numeric_util;
 #[cfg(feature = "blas")]
@@ -717,15 +716,21 @@ where
     let dimac = a.shape()[1];
     let dimbr = b.shape()[0];
     let dimbc = b.shape()[1];
-    let mut out: Array2<MaybeUninit<A>> = Array2::uninit((dimar * dimbr, dimac * dimbc));
+    let mut out: Array2<MaybeUninit<A>> = Array2::uninit((
+        dimar
+            .checked_mul(dimbr)
+            .expect("Dimensions of kronecker product output array overflows usize."),
+        dimac
+            .checked_mul(dimbc)
+            .expect("Dimensions of kronecker product output array overflows usize."),
+    ));
     Zip::from(out.exact_chunks_mut((dimbr, dimbc)))
         .and(a)
-        .for_each(|out, a| {
-            (*a * b).assign_to(out);
+        .for_each(|out, &a| {
+            (a * b).assign_to(out);
         });
     unsafe { out.assume_init() }
 }
-
 
 #[inline(always)]
 /// Return `true` if `A` and `B` are the same type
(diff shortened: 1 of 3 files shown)


ethanhs:
Also I should mention that I kept the name as `kron` because that is what numpy and matlab use and it seems familiar.

ethanhs:
Hm, I just noticed there is also https://github.com/rust-ndarray/ndarray/pull/690. Not sure what direction you want to go, but happy to work on getting an `outer` based implementation in if that's preferable.

src/linalg/impl_linalg.rs
```diff
@@ -699,6 +701,32 @@ unsafe fn general_mat_vec_mul_impl<A, S1, S2>(
...
+{
+    let dimar = a.shape()[0];
+    let dimac = a.shape()[1];
+    let dimbr = b.shape()[0];
+    let dimbc = b.shape()[1];
+    let mut out: Array2<MaybeUninit<A>> = Array2::uninit((dimar * dimbr, dimac * dimbc));
+    Zip::from(out.exact_chunks_mut((dimbr, dimbc)))
+        .and(a)
+        .for_each(|out, a| {
+            (*a * b).assign_to(out);
```
bluss, line 724: We'd prefer to not create intermediate array `a * b` here. `Zip::from(a).and(&b).and(&mut out).for_each(|&a, &b, out| *out = a * b)` is an alternative?

bluss (commented):
I don't have so much opinions about what's best here - this is outside my area.

It's easier if we maybe see it like this, some of the numerics stuff is not my area, and then it needs help from others to design and maybe even merge it, we maybe need a maintainer who likes that area. :slightly_smiling_face:  I'm mainly focusing on certain things. It doesn't mean the other things are not good, just not my focus & interest.

src/linalg/impl_linalg.rs
```diff
@@ -699,6 +701,32 @@ unsafe fn general_mat_vec_mul_impl<A, S1, S2>(
     }
 }
 
+
+/// Kronecker product of 2D matrices.
+///
+/// The kronecker product of a LxN matrix A and a MxR matrix B is a (L*M)x(N*R)
+/// matrix K formed by the block multiplication A_ij * B.
+pub fn kron<'a, A, S1, S2>(a: &ArrayBase<S1, Ix2>, b: &'a ArrayBase<S2, Ix2>) -> ArrayBase<OwnedRepr<A>, Ix2>
```
bluss, line 709:
```suggestion
pub fn kron<'a, A, S1, S2>(a: &ArrayBase<S1, Ix2>, b: &'a ArrayBase<S2, Ix2>) -> Array<A, Ix2>
```

src/linalg/impl_linalg.rs
```diff
@@ -699,6 +701,32 @@ unsafe fn general_mat_vec_mul_impl<A, S1, S2>(
...
+where
+    S1: Data<Elem = A>,
+    S2: Data<Elem = A>,
+    A: LinalgScalar,
+    A: std::ops::Mul<&'a ArrayBase<S2, Ix2>, Output = ArrayBase<OwnedRepr<A>, Ix2>>,
```
bluss, line 714:
```suggestion
    A: std::ops::Mul<&'a ArrayBase<S2, Ix2>, Output = Array<A, Ix2>>,
```

src/linalg/impl_linalg.rs
```diff
@@ -699,6 +701,32 @@ unsafe fn general_mat_vec_mul_impl<A, S1, S2>(
...
+{
+    let dimar = a.shape()[0];
+    let dimac = a.shape()[1];
+    let dimbr = b.shape()[0];
+    let dimbc = b.shape()[1];
+    let mut out: Array2<MaybeUninit<A>> = Array2::uninit((dimar * dimbr, dimac * dimbc));
```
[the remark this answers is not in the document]
ethanhs: Good point! I'll make these checked.

src/linalg/impl_linalg.rs
ethanhs: Hm, I don't think that quite works. I want to multiply every element of `a` by `b`, and assign it to the window of `out` that corresponds to that element of `a`.

What you wrote is element-wise multiplication, is it not?

ethanhs: Also I've found difficulty getting anything but `x.assign_to(y)` to work when `y` has elements that are `MaybeUninit`, is there a better way to make the types happy around that? Maybe `AssignElem` should be implemented so that one can assign an array with elements of `T` to one of `MaybeUninit<T>`?

adamreichold: > I want to multiply every element of a by the full b matrix, and assign it to the window of out that corresponds to that element of a (and is the size of the b matrix.

This should do it without the temporary, shouldn't it?

```rust
Zip::from(out.exact_chunks_mut((dimbr, dimbc)))
    .and(a)
    .for_each(|out, &a| {
        Zip::from(out).and(b).for_each(|out, &b| {
            *out = MaybeUninit::new(a * b);
        });
    });
```

> Maybe AssignElem should be implemented so that one can assign an array with elements of T to one of MaybeUninit<T>?

I think this is already the case, c.f. https://docs.rs/ndarray/0.15.3/ndarray/trait.AssignElem.html#impl-AssignElem%3CT%3E-for-%26%27a%20mut%20MaybeUninit%3CT%3E

But it isn't necessary here as we are not in generic context, i.e. we know that the left-hand side is `&mtu MaybeUninit<A>` and therefore we can just wrap the right-hand side using `MaybeUninit::new` unconditionally.

bluss: Thank you. There are a few ways to use maybeuninit, like that, that's good, the assigbelwm trait, or creating a raw array view and working with elements as *mut A

ethanhs: Good to know re using `MaybeUninit`! I was scratching my head when initially using `uninit`.

And thank you Adam for the suggestion of the nested Zips, that makes a lot of sense, I'll add that :)

src/linalg/impl_linalg.rs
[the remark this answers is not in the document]
bluss: Thanks. I missed one thing in the argument, there is an assert that happens in Zip::from(x).and(y) that x and y are the same shape. So it was already correctly constructed with that I think, but sure, it needed some kind of argument/comment. I think current version is is fine (your choice).

bluss (approved):

src/linalg/impl_linalg.rs
```diff
@@ -699,6 +700,40 @@ unsafe fn general_mat_vec_mul_impl<A, S1, S2>(
...
+/// Kronecker product of 2D matrices.
+///
+/// The kronecker product of a LxN matrix A and a MxR matrix B is a (L*M)x(N*R)
+/// matrix K formed by the block multiplication A_ij * B.
+pub fn kron<'a, A, S1, S2>(a: &ArrayBase<S1, Ix2>, b: &'a ArrayBase<S2, Ix2>) -> Array<A, Ix2>
+where
+    S1: Data<Elem = A>,
+    S2: Data<Elem = A>,
+    A: LinalgScalar,
+    A: std::ops::Mul<&'a ArrayBase<S2, Ix2>, Output = Array<A, Ix2>>,
```
bluss, line 713:
```suggestion
```

bluss: forgot about this one. this is typical of stuff we don't want to "leak" out of the implementation.

bluss:
Thanks! Please keep this comment here in mind https://github.com/rust-ndarray/ndarray/pull/1105#pullrequestreview-800600517 - we need a maintainer for "this area" (deeper numerical functions) - also collaboration with the ndarray-linalg crate.

ethanhs:
@bluss yes, I'd be happy to keep contributing and maintaining this area! I sent a message on matrix to follow up :)

PR #1106, opened 2021-11-06
Complex dot()

### LINKED ISSUE #158: Use BLAS for complex .mat_mul too

ethanhs:
This uses BLAS zgemm/cgemm to do complex matrix matrix multiplication.

I know there probably should be many more tests (the ones I added were more just a brief correctness check if nothing else) but I thought I'd open this to get feedback on whether or not something like this approach seems reasonable.

I'm mainly focusing on matrix matrix products as that is what I am using, but I can follow up with matrix vector and vector vector product implementation if there is interest.

Fixes #158

### DIFF
--- a/src/linalg/impl_linalg.rs
+++ b/src/linalg/impl_linalg.rs
@@ -416,8 +416,23 @@ fn mat_mul_impl<A>(
             rhs_trans = CblasTrans;
         }
 
+        macro_rules! cast_ty {
+            (f32, $var:ident) => {
+                cast_as(&$var)
+            };
+            (f64, $var:ident) => {
+                cast_as(&$var)
+            };
+            (c32, $var:ident) => {
+                &$var as *const A as *const _
+            };
+            (c64, $var:ident) => {
+                &$var as *const A as *const _
+            };
+        }
+
         macro_rules! gemm {
-            ($ty:ty, $gemm:ident) => {
+            ($ty:tt, $gemm:ident) => {
                 if blas_row_major_2d::<$ty, _>(&lhs_)
                     && blas_row_major_2d::<$ty, _>(&rhs_)
                     && blas_row_major_2d::<$ty, _>(&c_)
@@ -437,9 +452,9 @@ fn mat_mul_impl<A>(
                     let lhs_stride = cmp::max(lhs_.strides()[0] as blas_index, k as blas_index);
                     let rhs_stride = cmp::max(rhs_.strides()[0] as blas_index, n as blas_index);
                     let c_stride = cmp::max(c_.strides()[0] as blas_index, n as blas_index);
-
                     // gemm is C ← αA^Op B^Op + βC
                     // Where Op is notrans/trans/conjtrans
+
                     unsafe {
                         blas_sys::$gemm(
                             CblasRowMajor,
@@ -448,12 +463,12 @@ fn mat_mul_impl<A>(
                             m as blas_index,               // m, rows of Op(a)
                             n as blas_index,               // n, cols of Op(b)
                             k as blas_index,               // k, cols of Op(a)
-                            cast_as(&alpha),               // alpha
+                            cast_ty!($ty, alpha),          // alpha
                             lhs_.ptr.as_ptr() as *const _, // a
                             lhs_stride,                    // lda
                             rhs_.ptr.as_ptr() as *const _, // b
                             rhs_stride,                    // ldb
-                            cast_as(&beta),                // beta
+                            cast_ty!($ty, beta),           // beta
                             c_.ptr.as_ptr() as *mut _,     // c
                             c_stride,                      // ldc
                         );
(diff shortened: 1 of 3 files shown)


xtest-blas/tests/oper.rs
```diff
@@ -52,6 +55,20 @@ fn range_mat64(m: Ix, n: Ix) -> Array2<f64> {
         .unwrap()
 }
 
+fn range_mat_complex(m: Ix, n: Ix) -> Array2<Complex32> {
+    Array::linspace(0., (m * n) as f32 - 1., m * n)
+        .into_shape((m, n))
+        .unwrap()
+        .map_mut(|&mut f| Complex32::new(f, 0.))
```
bluss, line 62: unnecessary mutability, can just be `mapv` or `map`

src/linalg/impl_linalg.rs
```diff
@@ -454,6 +462,55 @@ fn mat_mul_impl<A>(
         }
         gemm!(f32, cblas_sgemm);
         gemm!(f64, cblas_dgemm);
+
+        macro_rules! gemm {
```
bluss, line 466: Is it possible to share this code with the previous gemm macro? It's hair enough as it is, we don't want to duplicate that code.

bluss (commented):
Good. I guess we need to document somewhere that what gemm operations we support with matrixmultiply vs blas

src/linalg/impl_linalg.rs
ethanhs: Yeah, the only difference is that with `Complex<T>` we need to cast through the pointer, whereas `cast_as` doesn't work. I can probably make a shared macro that generates the preamble to be re-used by each macro I suppose.

jturner314:
Fwiw, in the absence of BLAS, falling back to four multiplications with `matrixmultiply` and a couple of additions (`(A + B i) (C + D i) = (AC - BD) + (AD + BC) i`, where `A`, `B`, `C`, and `D` are real) may be faster than falling back to `mat_mul_general`.

ethanhs:
Good point! I can add that as well if you want.

jturner314:
If you decide to do so, PR #1029 would make it easier and is useful functionality on its own anyway. Feel free to take over work on that if you'd like; I'm very busy right now.

ethanhs:
I actually could use that functionality too (I implemented it myself for my own uses, but want to upstream things...) so I will go ahead and take over that PR and tackle it first!

src/linalg/impl_linalg.rs
```diff
@@ -6,28 +6,31 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
-use crate::imp_prelude::*;
-use crate::numeric_util;
 #[cfg(feature = "blas")]
 use crate::dimension::offset_from_low_addr_ptr_to_logical_ptr;
+use crate::imp_prelude::*;
```
bluss, line 11: (We don't need to change anything, just talking). It's just weird that tools would move down the implementation prelude import. They are wrong, that import should clearly be first. (Probably accomplished by having it with a blank line after).

adamreichold: Personally, I have always liked that Rustfmt sort use statements. Whether sorting them alphabetically is good or a bad is a reasonable discussion to have, but as usual with aesthetics, just the fact that the decision is automatic seems to bring almost all of the benefit, i.e. no more tedious discussions on how to sort here or there or writing up rules as a company-wide guideline.

Additionally, I find the ecosystem-wide readability resulting from the common application of Rustfmt's defaults highly beneficial to upstream collaboration.

(Especially since the workaround using a blank line really seems like a good idea if the import actually is special among the others next to it.)

ethanhs: Yeah I agree, it is a bit strange. Perhaps so other imports don't silently clobber items brought in from the prelude?

bluss: Sounds good @adamreichold. You'll notice that I'm not yet very fond of rustfmt, unfortunately. I'd be happier if it could accept "multiple good ways" to format the same code.

bluss:
I think I want to add fallback implementations of complex gemm into matrixmultiply. That might get a ball rolling and should perform ok-ish to start with.

Reusing dgemm seems like good way to go, but in practice I think it has a problem with the general matrix multiply and it's alpha muliplier? What's to be done if imag(alpha) != 0? Then it needs extra steps after calling dgemm, if I'm not mistaken.


For some reason I came across this, that some libraries reuse their dgemm by using a multiplication reduction algorithm. That's interesting but seems like quite the project. [(paper link)](http://citeseerx.ist.psu.edu/viewdoc/download;jsessionid=11AE11C90E32CAE32417E5AD3D68387B?doi=10.1.1.39.6887&rep=rep1&type=pdf). Just using it naively has massive memory overhead.

bild

bluss:
This PR would be fine without addressing matrixmultply or fallback much at all, of course. Just saying that anything extra can be a separate PR.

ethanhs:
Yeah I think I will leave the followup to another PR. I've fixed the small mutability issue so I think this is ready for another round of review.

src/linalg/impl_linalg.rs
```diff
@@ -670,8 +694,8 @@ unsafe fn general_mat_vec_mul_impl<A, S1, S2>(
                             a_stride,                   // lda
                             x_ptr as *const _,          // x
                             x_stride,
-                            cast_as(&beta),             // beta
-                            y_ptr as *mut _,            // y
+                            cast_as(&beta),  // beta
+                            y_ptr as *mut _, // y
```
bluss, line 698: (not important!) These comments were aligned with the others but are now not.

xtest-blas/tests/oper.rs
```diff
@@ -52,6 +55,20 @@ fn range_mat64(m: Ix, n: Ix) -> Array2<f64> {
         .unwrap()
 }
 
+fn range_mat_complex(m: Ix, n: Ix) -> Array2<Complex32> {
+    Array::linspace(0., (m * n) as f32 - 1., m * n)
```
bluss, line 59: Would it be possible to test with not purely real matrices?

bluss (commented):
looks good to me, only minor comments

src/linalg/impl_linalg.rs
```diff
@@ -406,8 +414,23 @@ fn mat_mul_impl<A>(
             rhs_trans = CblasTrans;
         }
 
+        macro_rules! cast_ty {
```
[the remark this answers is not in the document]
ethanhs: Good point, I've renamed it `gemm_scalar_cast`

xtest-blas/tests/oper.rs
ethanhs: Added a test that exercises both complex entries in the matrix and also complex alpha/beta.

bluss:
Matrixmult complex wip is at https://github.com/bluss/matrixmultiply/pull/58 performance for f32 + FMA with auto vectorization is not too bad (maybe it reaches 1/2-1/3 of the potential :). Matrixmultiply as a crate doesn't need to be so important.

bluss:
I edited the description to say it's not WIP. I think it's ready. I'd usually say we need to edit the PR description before merge, but it seems like github has gotten to the point where it doesn't even include it in the merge in git, so it's not so important anymore (did github change here?)

ethanhs:
> Matrixmult complex wip is at bluss/matrixmultiply#58 performance for f32 + FMA with auto vectorization is not too bad (maybe it reaches 1/2-1/3 of the potential :). Matrixmultiply as a crate doesn't need to be so important.

Great! I can try to take a look at that a bit later.

> I edited the description to say it's not WIP. I think it's ready. I'd usually say we need to edit the PR description before merge, but it seems like github has gotten to the point where it doesn't even include it in the merge in git, so it's not so important anymore (did github change here?)

I think(?) the message is only added if there is a single commit or depending on the kind of merge you do maybe.

bluss:
Thanks for finally closing this one :)

I think the use dgemm-for-zgemm idea still can give good benefits compared with matrixmultiply (in its current state). Handling the general case with complex alpha, beta is a bit tricky, I don't know what the best performing solution is. But alpha != 1 or beta != 0 is the uncommon case.

PR #1118, opened 2021-11-21
Use matrixmultiply cgemm and improve tests

bluss:
Refactor tests to use generics in a bit smarter way.

We test both .dot() and general_mat_mul. Tests now use relative accuracy
more explicitly (this works better with generics, instead of using
approx).

This follows up previous change #1106 by improving the default/
fallback for `Complex<f32>` and `Complex<f64>` by using
matrixmultiply.

This performance improvement is already good (2-3x on small problems),
but enabling BLAS should be much better.


bluss:
Here I wonder if complex support in general should be optional and opt-in, might be a less used feature.

PR #1123, opened 2021-11-28
Add examples for type conversion

sgasse:
- Add `type_conversion.rs` to illustrate some common conversions.
- Update the documentation for numpy users.

### DIFF
--- a/examples/type_conversion.rs
+++ b/examples/type_conversion.rs
@@ -1,5 +1,10 @@
 #[cfg(feature = "approx")]
-use {approx::assert_abs_diff_eq, ndarray::prelude::*, std::convert::TryFrom};
+use std::convert::TryFrom;
+
+#[cfg(feature = "approx")]
+use approx::assert_abs_diff_eq;
+#[cfg(feature = "approx")]
+use ndarray::prelude::*;
 
 #[cfg(feature = "approx")]
 fn main() {
--- a/examples/type_conversion.rs
+++ b/examples/type_conversion.rs
@@ -1,0 +1,114 @@
+#[cfg(feature = "approx")]
+use {approx::assert_abs_diff_eq, ndarray::prelude::*, std::convert::TryFrom};
+
+#[cfg(feature = "approx")]
+fn main() {
+    // Converting an array from one datatype to another is implemented with the
+    // `ArrayBase::mapv()` function. We pass a closure that is applied to each
+    // element independently. This allows for more control and flexiblity in
+    // converting types.
+    //
+    // Below, we illustrate four different approaches for the actual conversion
+    // in the closure.
+    // - `From` ensures lossless conversions known at compile time and is the
+    //   best default choice.
+    // - `TryFrom` either converts data losslessly or panics, ensuring that the
+    //   rest of the program does not continue with unexpected data.
+    // - `as` never panics and may silently convert in a lossy way, depending
+    //   on the source and target datatypes. More details can be found in the
+    //   reference: https://doc.rust-lang.org/reference/expressions/operator-expr.html#numeric-cast
+    // - Using custom logic in the closure, e.g. to clip values or for NaN
+    //   handling in floats.
+    //
+    // For a brush-up on casting between numeric types in Rust, refer to:
+    // https://doc.rust-lang.org/rust-by-example/types/cast.html
+
+    // Infallible, lossless conversion with `From`
+    // The trait `std::convert::From` is only implemented for conversions that
+    // can be guaranteed to be lossless at compile time. This is the safest
+    // approach.
+    let a_u8: Array<u8, _> = array![[1, 2, 3], [4, 5, 6]];
+    let a_f32 = a_u8.mapv(|element| f32::from(element));
+    assert_abs_diff_eq!(a_f32, array![[1.0, 2.0, 3.0], [4.0, 5.0, 6.0]]);
+
+    // Fallible, lossless conversion with `TryFrom`
+    // `i8` numbers can be negative, in such a case, there is no perfect
+    // conversion to `u8` defined. In this example, all numbers are positive and
+    // in bounds and can be converted at runtime. But for unknown runtime input,
+    // this would panic with the message provided in `.expect()`. Note that you
+    // can also use `.unwrap()` to be more concise.
+    let a_i8: Array<i8, _> = array![120, 8, 0];
+    let a_u8 = a_i8.mapv(|element| u8::try_from(element).expect("Could not convert i8 to u8"));
+    assert_eq!(a_u8, array![120u8, 8u8, 0u8]);
+
+    // Unsigned to signed integer conversion with `as`
+    // A real-life example of this would be coordinates on a grid.
+    // A `usize` value can be larger than what fits into a `isize`, therefore,
+    // it would be safer to use `TryFrom`. Nevertheless, `as` can be used for
+    // either simplicity or performance.
+    // The example includes `usize::MAX` to illustrate potentially undesired
+    // behavior. It will be interpreted as -1 (noop-casting + 2-complement), see
+    // https://doc.rust-lang.org/reference/expressions/operator-expr.html#numeric-cast
+    let a_usize: Array<usize, _> = array![1, 2, 3, usize::MAX];
+    let a_isize = a_usize.mapv(|element| element as isize);
+    assert_eq!(a_isize, array![1_isize, 2_isize, 3_isize, -1_isize]);
+
+    // Simple upcasting with `as`
+    // Every `u8` fits perfectly into a `u32`, therefore this is a lossless
+    // conversion.
+    // Note that it is up to the programmer to ensure the validity of the
+    // conversion over the lifetime of a program. With type inference, subtle
+    // bugs can creep in since conversions with `as` will always compile, so a
+    // programmer might not notice that a prior lossless conversion became a
+    // lossy conversion. With `From`, this would be noticed at compile-time and
+    // with `TryFrom`, it would also be either handled or make the program
+    // panic.
+    let a_u8: Array<u8, _> = array![[1, 2, 3], [4, 5, 6]];
+    let a_u32 = a_u8.mapv(|element| element as u32);
+    assert_eq!(a_u32, array![[1u32, 2u32, 3u32], [4u32, 5u32, 6u32]]);
+
+    // Saturating cast with `as`
+    // The `as` keyword performs a *saturating cast* When casting floats to
+    // ints. This means that numbers which do not fit into the target datatype
+    // will silently be clipped to the maximum/minimum numbers. Since this is
+    // not obvious, we discourage the intentional use of casting with `as` with
+    // silent saturation and recommend a custom logic instead which makes the
+    // intent clear.
+    let a_f32: Array<f32, _> = array![
+        256.0,         // saturated to 255
+        255.7,         // saturated to 255
+        255.1,         // saturated to 255
+        254.7,         // rounded down to 254 by cutting the decimal part
+        254.1,         // rounded down to 254 by cutting the decimal part
+        -1.0,          // saturated to 0 on the lower end
+        f32::INFINITY, // saturated to 255
+        f32::NAN,      // converted to zero
+    ];
+    let a_u8 = a_f32.mapv(|element| element as u8);
+    assert_eq!(a_u8, array![255, 255, 255, 254, 254, 0, 255, 0]);
+
+    // Custom mapping logic
+    // Given that we pass a closure for the conversion, we can also define
+    // custom logic to e.g. replace NaN values and clip others. This also
+    // makes the intent clear.
+    let a_f32: Array<f32, _> = array![
+        270.0,         // clipped to 200
+        -1.2,          // clipped to 0
+        4.7,           // rounded up to 5 instead of just stripping decimals
+        f32::INFINITY, // clipped to 200
+        f32::NAN,      // replaced with upper bound 200
+    ];
+    let a_u8_custom = a_f32.mapv(|element| {
+        if element == f32::INFINITY || element.is_nan() {
+            return 200;
+        }
+        if let Some(std::cmp::Ordering::Less) = element.partial_cmp(&0.0) {
+            return 0;
+        }
+        200.min(element.round() as u8)
+    });
+    assert_eq!(a_u8_custom, array![200, 0, 5, 200, 200]);
+}
+
+#[cfg(not(feature = "approx"))]
+fn main() {}
--- a/src/doc/ndarray_for_numpy_users/mod.rs
+++ b/src/doc/ndarray_for_numpy_users/mod.rs
@@ -524,6 +524,101 @@
 //! </td></tr>
 //! </table>
 //!
+//! ## Type conversions
+//!
+//! In `ndarray`, conversions between datatypes are done with `mapv()` by
+//! passing a closure to convert every element independently.
+//! For the conversion itself, we have several options:
+//! - `std::convert::From` ensures lossless, safe conversions at compile-time
+//!   and is generally recommended.
+//! - `std::convert::TryFrom` can be used for potentially unsafe conversions. It
+//!   will return a `Result` which can be handled or `unwrap()`ed to panic if
+//!   any value at runtime cannot be converted losslessly.
+//! - The `as` keyword compiles to lossless/lossy conversions depending on the
+//!   source and target datatypes. It can be useful when `TryFrom` is a
+//!   performance issue or does not apply. A notable difference to NumPy is that
+//!   `as` performs a [*saturating* cast][sat_conv] when casting
+//!   from floats to integers. Further information can be found in the
+//!   [reference on type cast expressions][as_typecast].
+//!
+//! For details, be sure to check out the type conversion examples.
+//!
+
+//! <table>
+//! <tr><th>
+//!
+//! NumPy
+//!
+//! </th><th>
+//!
+//! `ndarray`
+//!
+//! </th><th>
+//!
+//! Notes
+//!
+//! </th></tr>
+//!
+//! <tr><td>
+//!
+//! `a.astype(np.float32)`
+//!
+//! </td><td>
+//!
+//! `a.mapv(|x| f32::from(x))`
+//!
+//! </td><td>
+//!
+//! convert `u8` array infallibly to `f32` array with `std::convert::From`, generally recommended
+//!
+//! </td></tr>
+//!
+//! <tr><td>
+//!
+//! `a.astype(np.int32)`
+//!
+//! </td><td>
+//!
+//! `a.mapv(|x| i32::from(x))`
+//!
+//! </td><td>
+//!
+//! upcast `u8` array to `i32` array with `std::convert::From`, preferable over `as` because it ensures at compile-time that the conversion is lossless
+//!
+//! </td></tr>
+//!
+//! <tr><td>
+//!
+//! `a.astype(np.uint8)`
+//!
+//! </td><td>
+//!
+//! `a.mapv(|x| u8::try_from(x).unwrap())`
+//!
+//! </td><td>
+//!
+//! try to convert `i8` array to `u8` array, panic if any value cannot be converted lossless at runtime (e.g. negative value)
+//!
+//! </td></tr>
+//!
+//! <tr><td>
+//!
+//! `a.astype(np.int32)`
+//!
+//! </td><td>
+//!
+//! `a.mapv(|x| x as i32)`
+//!
+//! </td><td>
+//!
+//! convert `f32` array to `i32` array with ["saturating" conversion][sat_conv]; care needed because it can be a lossy conversion or result in non-finite values! See [the reference for information][as_typecast].
+//!
+//! </td></tr>
+//!
+//! [as_conv]: https://doc.rust-lang.org/rust-by-example/types/cast.html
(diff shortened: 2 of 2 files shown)


src/doc/ndarray_for_numpy_users/mod.rs
```diff
@@ -524,6 +524,18 @@
...
+//! ## Type conversions
+//!
+//! `ndarray` does not currently have a `astype` function as `numpy`, but
+//! conversions between datatypes can be done with `mapv()`. The behavior of
+//! Rust's datatype conversion using the `as` keyword might be surprising for
+//! some `numpy` users, be sure to check out the type conversion example.
+//!
+//! NumPy | `ndarray` | Notes
+//! ------|-----------|------
+//! `a.astype(np.uint8)` | `unsafe{ a.mapv(|x| x.to_int_unchecked::<u8>()) }` | convert float array to new datatype without checking bounds, ignoring overflow
```
adamreichold, line 536: I fear this might not be strongly worded enough, i.e. this does not just not check bounds but also imply UB if the values if out of bounds (or NaN). Personally, I would suggest strongly steering people away from these methods. Due to the above, I am also not sure whether the results will really be reliably compatible with NumPy due to unexpected compiler transformations?

examples/type_conversion.rs
```diff
@@ -0,0 +1,74 @@
...
+        254.0,         // 254, in bounds
+        -1.1,          // 255, reinterpreted 2-complement after cutting decimals
+        -2.3,          // 254, reinterpreted 2-complement after cutting decimals
+        -3.7,          // 253, reinterpreted 2-complement after cutting decimals
+        f32::INFINITY, // 0, based on representation
+        f32::NAN,      // 0, based on representation
+    ];
+    unsafe {
+        let a_u8: Array<u8, _> = a_f32.mapv(|element| element.to_int_unchecked::<u8>());
+        assert_eq!(a_u8, array![1, 0, 254, 255, 254, 253, 0, 0]);
```
adamreichold, line 51: As written below, I don't think these results can be relied on outside of `std`.

src/doc/ndarray_for_numpy_users/mod.rs
bluss: Thanks. Examples with `as` are ok. I agree with @adamreichold. Just skip mentioning `to_int_unchecked` unless you dedicate do it with care, I think I'd prefer not showing an example for it, just mention that it exists and link to its docs, if it's mentioned at all.

The perils of `to_int_unchecked` are **a lot** worse in Rust than anything you can do with `astype` in numpy.

examples/type_conversion.rs
```diff
@@ -0,0 +1,74 @@
...
+        256.0,         // 0, last 8 bits are zero
+        254.0,         // 254, in bounds
+        -1.1,          // 255, reinterpreted 2-complement after cutting decimals
+        -2.3,          // 254, reinterpreted 2-complement after cutting decimals
+        -3.7,          // 253, reinterpreted 2-complement after cutting decimals
+        f32::INFINITY, // 0, based on representation
+        f32::NAN,      // 0, based on representation
+    ];
+    unsafe {
+        let a_u8: Array<u8, _> = a_f32.mapv(|element| element.to_int_unchecked::<u8>());
```
bluss, line 50: This code is UB :boom:  so the example is meaningless :slightly_frowning_face:   Unfortunately.

sgasse: Alright thank you a lot for pointing that out - I will remove the example tomorrow :+1:

examples/type_conversion.rs
sgasse: Thank you, I'll remove it.

src/doc/ndarray_for_numpy_users/mod.rs
sgasse: OK thank you for explaining and reinforcing this @adamreichold and @bluss  - so then I will remove the `f32::to_int_unchecked` and instead put the `a.mapv()` as rough equivalent of `a.astype()` with a comment mentioning the difference with saturating casts. With only one line, it might not make sense to have this as separate section. Would it fit into `## Array Manipulation` instead?

sgasse: I added a few more examples, not it might be fine as a section on its own.

adamreichold:
I am not sure if this is too much for this MR, but similar to how we avoided the unchecked conversion, maybe we should include the infallible conversions based on [`From`](https://doc.rust-lang.org/stable/std/convert/trait.From.html) and the fallible conversions based [`TryFrom`](https://doc.rust-lang.org/stable/std/convert/trait.TryFrom.html) as well?

While `as` does not have undefined semantics in Rust any more, those two options seem preferable if applicable as they either will neither loose information guaranteed at compile time (`From`) or will make the program fail loudly at runtime if loss of information is detected (`TryFrom` and `unwrap`) which appears better suited than silently saturating for most scientific applications IMHO.

sgasse:
> I am not sure if this is too much for this MR, but similar to how we avoided the unchecked conversion, maybe we should include the infallible conversions based on [`From`](https://doc.rust-lang.org/stable/std/convert/trait.From.html) and the fallible conversions based [`TryFrom`](https://doc.rust-lang.org/stable/std/convert/trait.TryFrom.html) as well?
> 
> While `as` does not have undefined semantics in Rust any more, those two options seem preferable if applicable as they either will neither loose information guaranteed at compile time (`From`) or will make the program fail loudly at runtime if loss of information is detected (`TryFrom` and `unwrap`) which appears better suited than silently saturating for most scientific applications IMHO.

Actually, I really like the idea, so thank you for pointing this out @adamreichold ! I completely refactored the examples so that `From`/`TryFrom` are the recommended ways. This is also updated in the 'for_numpy_users' documentation. Could you take another look?

Also, I was not able to figure out why the CI fails for my patch - how is the file that I added different from the other examples? If you can point me in the right direction, I'd be happy to fix it.

examples/type_conversion.rs
```diff
@@ -0,0 +1,97 @@
...
+    // which makes the intent clear.
+    let a_f32: Array<f32, _> = array![
+        256.0,         // saturated to 255
+        255.7,         // saturated to 255
+        255.1,         // saturated to 255
+        254.7,         // rounded down to 254 by cutting the decimal part
+        254.1,         // rounded down to 254 by cutting the decimal part
+        -1.0,          // saturated to 0 on the lower end
+        f32::INFINITY, // saturated to 255
+        f32::NAN,      // converted to zero
```
sgasse, line 71: Just to be sure @adamreichold  - this is not UB, right? Otherwise, I'd at least add a comment.

adamreichold: Yes, this is well-defined, c.f. https://doc.rust-lang.org/reference/expressions/operator-expr.html#type-cast-expressions (Generally speaking, Rust code that does not use `unsafe` will not have undefined behaviour. That casts of floating point numbers used to have was a bug which was resolved by fixing the saturating semantics.)

adamreichold:
> Also, I was not able to figure out why the CI fails for my patch - how is the file that I added different from the other examples? If you can point me in the right direction, I'd be happy to fix it.

Your example has a `main` function only if `#[cfg(feature = "approx")]`, but the CI does not enable this feature. There are different ways to go about this but I think the easiest "fix" would be to add an empty main if the feature is missing, e.g.
```rust
#[cfg(not(feature = "approx"))]
fn main() {}
```

examples/type_conversion.rs
```diff
@@ -0,0 +1,97 @@
...
+#[cfg(feature = "approx")]
+fn main() {
+    // Converting an array from one datatype to another is implemented with the
+    // `ArrayBase::mapv()` function. We pass a closure that is applied to each
+    // element independently. This allows for more control and flexiblity in
+    // converting types.
+    //
+    // Below, we illustrate three different approaches for the actual conversion
+    // in the closure.
+    // - `From` ensures perfect conversions known at compile time and is the
```
adamreichold, line 14: I think the official document refers to this as "lossless" instead of "perfect".

examples/type_conversion.rs
```diff
@@ -0,0 +1,97 @@
...
+    //   best default choice.
+    // - `TryFrom` either converts data perfectly or panics, ensuring that the
```
adamreichold, line 16: `TryFrom` performs a fallible conversion, whether the program panics or not is then up to the user of the trait.

examples/type_conversion.rs
```diff
@@ -0,0 +1,97 @@
...
+    // conversion to `u8` defined. In this example, all numbers are positive and
+    // in bounds and can be converted at runtime. But for unknown runtime input,
+    // this would panic with the message provided in `.expect()`. Note that you
+    // can also use `.unwrap()` to be more concise.
+    let a_i8: Array<i8, _> = array![120, 8, 0];
+    let a_u8 = a_i8.mapv(|element| u8::try_from(element).expect("Could not convert i8 to u8"));
+    assert_eq!(a_u8, array![120u8, 8u8, 0u8]);
+
+    // Simple upcasting with `as`
+    // Every `u8` fits perfectly into a `u32`, therefore we do not need to take
```
adamreichold, line 43: In realistic programs, one often starts with this reasoning but especially with type interference the involved types can change and `as` becomes lossy even if it started out lossless. Hence, I think `From` should be recommended whenever possible. If the types changes and cannot be converted losslessly any more, compilation will fail and one can still decided to "downgrade" to a cast.

src/doc/ndarray_for_numpy_users/mod.rs
```diff
@@ -524,6 +524,30 @@
...
+//!   will panic if values are encountered at runtime that cannot be converted
+//!   perfectly.
+//! - The `as` keyword performs a *saturating* cast since Rust 1.45 and should
+//!   only be used for safe, clear conversions such as upcasting.
+//! For details, be sure to check out the type conversion examples.
+//!
+//! NumPy | `ndarray` | Notes
+//! ------|-----------|------
+//! `a.astype(np.float32)` | `a.mapv(\|x\| f32::from(x))` | convert `u8` array infallibly safe to `f32` array with `std::convert::From`, generally recommended
+//! `a.astype(np.uint8)` | `a.mapv(\|x\| u8::try_from(x))` | try to convert `i8` array to `u8` array, panic if any value cannot be converted perfectly at runtime (e.g. negative value)
```
adamreichold, line 544: I think this would result in an array of `Result`s as the intended `.unwrap()` is missing?

examples/type_conversion.rs
bluss: I think it's good to have examples with as, so we should have some

adamreichold: I do agree that there should be examples, but I would recommend against choosing one where `From` could be used instead. The `as` example should rather do a potentially lossy conversion that cannot be done using `From` probably with a comment to use this if `TryFrom` is a performance issue or does not apply.

bluss:
I don't want to pile on instructions here, but I think keep it simple. For my sake, we can write anything into the example file (but word more carefully in the documentation). Maybe this link is useful to link to (std's docs for the `as`  keyword also link it): https://doc.rust-lang.org/nightly/reference/expressions/operator-expr.html#type-cast-expressions   I don't understand all the weird caution around `as` personally, it's a well-defined and useful operation.

src/doc/ndarray_for_numpy_users/mod.rs
```diff
@@ -524,6 +524,30 @@
...
+//! In `ndarray`, conversions between datatypes are done with `mapv()` by
+//! passing a closure to convert every element independently.
+//! For the conversion itself, we have several options:
+//! - `std::convert::From` ensures perfect, safe conversions at compile-time and
+//!   is generally recommended.
+//! - `std::convert::TryFrom` can be used for potentially unsafe conversions. It
+//!   will panic if values are encountered at runtime that cannot be converted
+//!   perfectly.
+//! - The `as` keyword performs a *saturating* cast since Rust 1.45 and should
+//!   only be used for safe, clear conversions such as upcasting.
```
bluss, line 538: The note about saturating casts is for floats only, not for all `as` conversions. We need a doc link anyway to explain the full rules, I think.

```suggestion
//! - The `as` keyword performs a cast which can convert between primitive types. [Other documentation links here]
```

adamreichold:
>  I don't understand all the weird caution around as personally, it's a well-defined and useful operation.

I think that it wasn't until Rust 1.45 did not help its reputation. Leaving this aside, I see two kinds of problems when using `as`:
* Together with unsafe code, it can lead to memory safety violations, for example by small negative integers becoming large positive integers, e.g. `-1_isize as usize`.
* In numerical analyses, the fact that intermediate values saturated or wrapped might not be easily visible in the final results as they can manifest only by e.g. biasing a statistical measure.

examples/type_conversion.rs
sgasse: A great, thank you for the link!

examples/type_conversion.rs
sgasse: Updated

examples/type_conversion.rs
sgasse: Updated

examples/type_conversion.rs
sgasse: Hm the prime example for me would be `f32 as i32` - but already have an example from floats to integers to show the saturating cast. With the updated comments/documentation, do you think it is fine or which example would you like to see in particular @adamreichold ?

src/doc/ndarray_for_numpy_users/mod.rs
sgasse: Yes absolutely, thank you!

src/doc/ndarray_for_numpy_users/mod.rs
sgasse: That is a great suggestion, thank you!

sgasse:
> > Also, I was not able to figure out why the CI fails for my patch - how is the file that I added different from the other examples? If you can point me in the right direction, I'd be happy to fix it.
> 
> Your example has a `main` function only if `#[cfg(feature = "approx")]`, but the CI does not enable this feature. There are different ways to go about this but I think the easiest "fix" would be to add an empty main if the feature is missing, e.g.
> 
> ```rust
> #[cfg(not(feature = "approx"))]
> fn main() {}
> ```

Ah yes, that makes sense, thanks!

sgasse:
Thank you both @adamreichold and @bluss for the thorough review! I tried to strike a balance between warning about potential issues when using `as` while still providing some examples with it.
@bluss are the examples/docu updates in general too verbose for your taste or are they fine?

bluss:
It's good. I could send in my own edits later if I have some nitpicks.

sgasse:
> It's good. I could send in my own edits later if I have some opinions about wording.

Alright, sounds good! :slightly_smiling_face: And I hope I finally got the feature guards right for CI ^^ Sorry for the mess-up there, should have run the `all-tests.sh` script earlier on my machine.

examples/type_conversion.rs
adamreichold: > do you think it is fine or which example would you like to see in particular @adamreichold ?

I think it is fine. The other one rather common example (for example when working with integer coordinates on a grid) is between signed and unsigned types, e.g. `usize as isize` (which can fail as `usize::MAX > isize::MAX as usize`).

src/doc/ndarray_for_numpy_users/mod.rs
```diff
@@ -524,6 +524,101 @@
...
+//!
+//! `a.astype(np.int32)`
+//!
+//! </td><td>
+//!
+//! `a.mapv(|x| i32::from(x))`
+//!
+//! </td><td>
+//!
+//! upcast `u8` array to `i32` array losslessly with `std::convert::From`, preferable over `as` because it ensures a lossless conversion at compile-time
```
bluss, line 586:
```suggestion
//! upcast `u8` array to `i32` array with `std::convert::From`, preferable over `as` because it ensures at compile-time that the conversion is lossless 
```

without the ambiguity

src/doc/ndarray_for_numpy_users/mod.rs
```diff
@@ -524,6 +524,101 @@
...
+//!
+//! </td><td>
+//!
+//! convert `f32` array to `i32` array with checked, [silently saturating conversion][sat_conv] from Rust 1.45 onwards, not recommended because changing datatypes which make a lossless conversion lossy will not lead to a compile-error
+//!
+//! </td></tr>
+//!
+//! [as_conv]: https://doc.rust-lang.org/rust-by-example/types/cast.html
+//! [sat_conv]: https://blog.rust-lang.org/2020/07/16/Rust-1.45.0.html#fixing-unsoundness-in-casts
+//! [as_typecast]: https://doc.rust-lang.org/nightly/reference/expressions/operator-expr.html#type-cast-expressions
```
bluss, line 620:
```suggestion
//! [as_typecast]: https://doc.rust-lang.org/reference/expressions/operator-expr.html#type-cast-expressions
```

stable version

src/doc/ndarray_for_numpy_users/mod.rs
```diff
@@ -524,6 +524,101 @@
...
+//!
+//! `a.astype(np.int32)`
+//!
+//! </td><td>
+//!
+//! `a.mapv(|x| x as i32)`
+//!
+//! </td><td>
+//!
+//! convert `f32` array to `i32` array with checked, [silently saturating conversion][sat_conv] from Rust 1.45 onwards, not recommended because changing datatypes which make a lossless conversion lossy will not lead to a compile-error
```
bluss, line 614:
```suggestion
//! convert `f32` array to `i32` array with [“saturating” conversion][sat_conv]; care needed because it can be a lossy conversion or result in non-finite values! See [the reference for information][as_typecast].
```

We don't need to mention Rust 1.45, just confusing detail (if this version of ndarray compiles, Rust > 1.45 is already ensured, and the version info becomes less relevant the older Rust 1.45 gets - besides, it was always a bug in older Rust versions - not a feature that changed).

Just saturating conversion can make it sound like it's going to saturate to min/max finite values, but it saturates to +-inf, which is a bit more tricky.

examples/type_conversion.rs
bluss: Seems to be missing?

src/doc/ndarray_for_numpy_users/mod.rs
```diff
@@ -524,6 +524,101 @@
...
+//! For the conversion itself, we have several options:
+//! - `std::convert::From` ensures lossless, safe conversions at compile-time
+//!   and is generally recommended.
+//! - `std::convert::TryFrom` can be used for potentially unsafe conversions. It
+//!   will return a `Result` which can be handled or `unwrap()`ed to panic if
+//!   any value at runtime cannot be converted losslessly.
+//! - The `as` keyword compiles to lossless/lossy conversions depending on the
+//!   source and target datatypes. It can be useful when `TryFrom` is a
+//!   performance issue or does not apply. A notable difference to NumPy is that
+//!   `as` performs a [*saturating* cast since Rust 1.45][sat_conv] when casting
```
bluss, line 540:
```suggestion
//!   `as` performs a [*saturating* cast][sat_conv] when casting
```

examples/type_conversion.rs
sgasse: Ah thank you for the example, I included it in `type_conversion.rs` :slightly_smiling_face:

src/doc/ndarray_for_numpy_users/mod.rs
sgasse: Done

src/doc/ndarray_for_numpy_users/mod.rs
sgasse: Alright, updated :+1:

src/doc/ndarray_for_numpy_users/mod.rs
sgasse: Updated

src/doc/ndarray_for_numpy_users/mod.rs
sgasse: Updated

sgasse:
Thank you for proof-reading @bluss ! I included all your comments - what exactly do you mean seems to be missing? An example with `From` can be found [here](https://github.com/rust-ndarray/ndarray/pull/1123/files#diff-0e3b13fa5669686aafb3509ba5cd84426d59b388a08e962e4628d31a9b6e4567R25-R31). Or do I use `From` in the wrong way?

Unfortunately, I cannot reproduce the error in the `clippy` run from the last workflow - but the error messages also seem unrelated to my patch.

examples/type_conversion.rs
sgasse: Oh yes, now I understood the comment - thanks! Should be updated now.

examples/type_conversion.rs
```diff
@@ -0,0 +1,114 @@
+#[cfg(feature = "approx")]
+use {approx::assert_abs_diff_eq, ndarray::prelude::*, std::convert::TryFrom};
```
bluss, line 2:
```suggestion
#[cfg(feature = "approx")]
use std::convert::TryFrom;

#[cfg(feature = "approx")]
use approx::assert_abs_diff_eq;
#[cfg(feature = "approx")]
use ndarray::prelude::*;
```

I don't know if community standards are diverging. (I don't really like "use trees".) Seems like use trees require tool-assistance to write them correctly which I think is too much, it's not the Rust I like, smells like code golfing. Prelude imports should obviously not be joined in on the same line as anything else. Different crates should not be combined together in use trees. :slightly_smiling_face: 

The back and forth in PR review is a bit awkward so I'll just push in this fix.

bluss (approved):

bluss:
Thanks!

PR #1128, opened 2021-12-05
Add support for isize, usize scalars

### LINKED ISSUE #1090: scalar ops implementation missing for  `usize - array`
Hi thanks for the lib! I see `matrix - matrix`, `matrix - scalar` both ok, but seems that there is not `scalar - matrix`, so I would appreciate it if we could have it.

simnalamburt:
Closes #1090

###### References
- https://github.com/rust-ndarray/ndarray/issues/1090
- https://stackoverflow.com/q/70232567/13977061


bluss:
Thanks!

alper:
Thanks!

PR #1130, opened 2021-12-06
Add additional array -> array view conversions

bluss:
It is best to be "complete" with implementations when possible, since
it can always be a breaking change to add it later.

As shown in the test change in this commit, this is a minor type
inference breakage due to `ArrayView::from(&[])` being ambiguous.

Empty array views should be relatively rare - and one can use
`ArrayView1` or other means to disambiguate.

### DIFF
--- a/src/arraytraits.rs
+++ b/src/arraytraits.rs
@@ -333,7 +333,21 @@ where Slice: AsRef<[A]>
     }
 }
 
-/// Implementation of ArrayView2::from(&S) where S is a slice to a 2D array
+/// Implementation of ArrayView2::from(&[[A; N]; M])
+///
+/// **Panics** if the product of non-zero axis lengths overflows `isize` (This can only occur if A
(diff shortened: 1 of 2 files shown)


src/arraytraits.rs
```diff
@@ -323,7 +323,18 @@ where
     }
 }
 
-/// Implementation of ArrayView2::from(&S) where S is a slice to a 2D array
+/// Implementation of ArrayView2::from(&[[A; N]; M])
+///
+/// **Panics** if the product of non-zero axis lengths overflows `isize` (This can only occur if A
+/// is zero-sized because slices cannot contain more than `isize::MAX` number of bytes).
```
jturner314, line 329: This comment and the one for `ArrayViewMut` should mention that another panicking case is when `N == 0` and the number of rows is greater than `isize::MAX`, even if `A` is not zero-sized. (See #1131.)

jturner314 (commented):
This PR looks good to me, except for one comment about the docs.

PR #1132, opened 2021-12-07
Make the aview0, aview1, and aview2 free functions be const fns

jturner314:
This is something I've wanted for a long time; it allows for creation of const `ArrayView`s. With Rust 1.57, it's now possible. (This was blocked on the ability to panic in const fns.)

This PR also makes the `Ix0`, …, `Ix6` free functions into const fns.

Unfortunately, the const fn implementations of `aview0`, `aview1`, and `aview2` cannot use some existing internal functionality which would simplify the implementation, especially those which require the `Dimension` trait, since const traits / trait impls are not yet available. I think the proposed implementations are simple enough, though.

This will have a merge conflict with #1131. I'll fix it up to no longer conflict once #1131 is merged, assuming #1131 is merged first.

This is a breaking change in the sense that it requires Rust 1.57.

### DIFF
--- a/src/free_functions.rs
+++ b/src/free_functions.rs
@@ -10,6 +10,7 @@ use alloc::vec;
 #[cfg(not(feature = "std"))]
 use alloc::vec::Vec;
 use std::mem::{forget, size_of};
+use std::ptr::NonNull;
 
 use crate::imp_prelude::*;
 use crate::{dimension, ArcArray1, ArcArray2};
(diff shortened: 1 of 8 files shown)


src/free_functions.rs
```diff
@@ -64,32 +65,105 @@ pub fn rcarr1<A: Clone>(xs: &[A]) -> ArcArray1<A> {
...
-pub fn aview1<A>(xs: &[A]) -> ArrayView1<'_, A> {
-    ArrayView::from(xs)
+pub const fn aview1<A>(xs: &[A]) -> ArrayView1<'_, A> {
+    if size_of::<A>() == 0 {
+        assert!(
+            xs.len() <= isize::MAX as usize,
+            "Slice length must fit in `isize`.",
+        );
+    }
+    ArrayBase {
```
bluss, line 106: is it possible to use the internal builder methods, with_data_ptr etc? We'd prefer those instead of spreading out `ArrayBase { .. }` constructors again.

jturner314: If I try to make `from_data_ptr` into a const fn, then I get the following compilation errors:

```
error[E0658]: trait bounds other than `Sized` on const fn parameters are unstable
  --> src/impl_internal_constructors.rs:14:9
   |
14 | impl<A, S> ArrayBase<S, Ix1>
   |         ^
...
26 |     pub(crate) const unsafe fn from_data_ptr(data: S, ptr: NonNull<A>) -> Self {
   |     -------------------------------------------------------------------------- function declared as const here
   |
   = note: see issue #57563 <https://github.com/rust-lang/rust/issues/57563> for more information

error[E0015]: calls in constant functions are limited to constant functions, tuple structs and tuple variants
  --> src/impl_internal_constructors.rs:33:23
   |
33 |         debug_assert!(array.pointer_is_inbounds());
   |                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^

error[E0493]: destructors cannot be evaluated at compile-time
  --> src/impl_internal_constructors.rs:27:13
   |
27 |         let array = ArrayBase {
   |             ^^^^^ constant functions cannot evaluate destructors
...
35 |     }
   |     - value is dropped here
```

The biggest issue is the generic `S` type. I don't see a way to fix the errors while keeping a generic `S`. We could add const constructors to `impl_internal_constructors.rs` specifically for `ArrayView`, though.

jturner314:
> ArrayRef is hopefully coming in 0.16. I don't know if these conflict, but I suggest that these changes have to stand back if they can't be realized in the ArrayRef update.

I don't see any reason why they'd conflict, but if you'd feel more comfortable with waiting until later to do this, that would be fine with me. The only breaking change in this PR is the minimum Rust version, so it would be good to at least update the MSRV in the next breaking release, unless you consider Rust 1.57 to be too new.

bluss:
There are already breaking changes in the master branch (#980), so continue with this here. Thanks for this one.

Issue #427, opened 2018-03-19
Add .shrink_to_fit() method for Array


jturner314:
Some methods (e.g. `.remove_axis()`, `.into_subview()`, and `.slice_move()`) can cause some of the elements in an owned array to be come inaccessable. It would be useful to have a method that removes the inaccessable elements (to reduce memory usage without performing a new heap allocation) by shifting the remaining elements to the front of the underlying `Vec` and shrinking the `Vec`. See the discussion in #425, in particular [this comment](https://github.com/bluss/rust-ndarray/issues/425#issuecomment-373971729) for some thoughts about the implementation.

bokutotu:
Hello!
 I would like to work on this issue, but I have one question: if the `ptr` of an `Array` has an `offset`, how do I find the first element of the allocated `ptr`? 

For example, suppose you have an `Array` of `dim`-> `(3, 3)`,`stride`-> `(3, 1)` and you drop the 0th row (`dim`-> (2, 3), `stride` -> (3, 1)). Then the `ptr` is considered to be `offset` by three. Once it is `offset`, it is difficult to calculate the `offset` for these three in `shrink_to_fit()`.
```
[[0, 1, 2],           
 [3, 4, 5],      -> [[3, 4, 5],
 [6, 7, 8],].        [6, 7, 8],]

memory order
                  
ptr                           ->                   ptr(difficult to find first ptr. In this case, the pointer pointing to the zero element)
▼                                                  ▼
0, 1, 2, 3, 4, 5, 6, 7, 8                 0, 1, 2, 3, 4, 5, 6, 7, 8
```


An easy solution I can think of is to record the `offset` in an `ArrayBase`, but I don't think that's a good idea since it would require a lot of implementation and there would be few opportunities to use it. What do you think about this problem?

jturner314:
For owned arrays, you can determine this offset by the following snippet (copied from the [`append` implementation](https://github.com/rust-ndarray/ndarray/blob/6c8b82121e0ee29e5fd995900d826e28eaf947ca/src/impl_owned_array.rs#L571-L576)):

```rust
            let data_to_array_offset = if std::mem::size_of::<A>() != 0 {
                self.as_ptr().offset_from(self.data.as_ptr())
            } else {
                0
            };
            debug_assert!(data_to_array_offset >= 0);
```

[...]

PR #1141, opened 2021-12-22
Add `shrink_to_fit` to `Array`

bokutotu:
#427 

* add `shrink_to_fit` to `Array<A, D>` 
* add `to_default_strid` to `Array<A, D>` 
* add `offset_from_data` to `Array<A, D>`
* add `shrink_to_fit` to `OwnedRepr`

### DIFF
--- a/src/impl_owned_array.rs
+++ b/src/impl_owned_array.rs
@@ -452,7 +452,7 @@ where
         let mut stride_ = 1;
         for i in stride_order.iter() {
             shrinked_stride[*i] = stride_;
-            stride_ = stride_ * dim[*i];
+            stride_ *= dim[*i];
         }
 
         // Calculate which index in shrinked_stride from the pointer offset.
@@ -463,20 +463,20 @@ where
             let mut index = D::zeros(dim.ndim());
             for i in stride_order_order.iter() {
                 index[*i] = offset / shrinked_stride[*i];
-                offset = offset % shrinked_stride[*i];
+                offset %= shrinked_stride[*i];
             }
             index
         };
 
         // Change the memory order only if it needs to be changed.
         let ptr_offset = unsafe { self.offset_from_data_ptr_to_logical_ptr() };
-        self.ptr = unsafe { self.ptr.offset(ptr_offset * -1) };
+        self.ptr = unsafe { self.ptr.sub(ptr_offset as usize) };
         for offset in 0..self.len() {
             let index = offset_stride(offset);
             let old_offset = dim.stride_offset_checked(&strides, &index).unwrap();
             unsafe {
-                *self.ptr.as_ptr().offset(offset as isize) =
-                    *self.ptr.as_ptr().offset(old_offset + ptr_offset);
+                *self.ptr.as_ptr().add(offset as usize) =
+                    *self.ptr.as_ptr().add((old_offset + ptr_offset) as usize);
             }
         }
         self.strides = shrinked_stride;
--- a/tests/shrink_to_fit.rs
+++ b/tests/shrink_to_fit.rs
@@ -47,7 +47,7 @@ fn stride_negative() {
     for i in 0..4 * 5 * 6 {
         raw_vec.push(i);
     }
-    let mut a = Array::from_shape_vec((4, 5, 6), raw_vec).unwrap();
+    let a = Array::from_shape_vec((4, 5, 6), raw_vec).unwrap();
     let mut a_slice = a.slice_move(s![2..3, 0..3, 0..;-1]);
     let a_slice_clone = a_slice.view().to_owned();
     a_slice.shrink_to_fit();


src/impl_owned_array.rs
```diff
@@ -401,16 +402,88 @@ impl<A, D> Array<A, D>
...
     {
         // same-dimensionality conversion
-        self.append(axis, array.insert_axis(axis).into_dimensionality::<D>().unwrap())
+        self.append(
+            axis,
+            array.insert_axis(axis).into_dimensionality::<D>().unwrap(),
+        )
     }
 
+    /// Return the offset between the pointer to the beginning of the heap
```
bluss, line 417: prefer not to say "between" but from... to to make the sign of the result clearer.

src/impl_owned_array.rs
```diff
@@ -401,16 +402,88 @@ impl<A, D> Array<A, D>
...
+    /// allocated by Array (`self.data.ptr`) and `self.ptr`.
+    unsafe fn offset_from_data(&self) -> isize {
```
bluss, line 419: might be interesting to know there is a related function offset_from_low_addr_ptr_to_logical_ptr. Maybe it can inspire the name - offset from data ptr to logical ptr?

src/impl_owned_array.rs
```diff
@@ -401,16 +402,88 @@ impl<A, D> Array<A, D>
...
+    /// use ndarray::s;
+    ///
+    /// let a = array![[0, 1, 2], [3, 4, 5], [6, 7,8]];
+    /// let mut a = a.slice_move(s![.., 0..2]);
+    /// let b = a.clone();
+    /// a.shrink_to_fit();
+    /// assert_eq!(a, b);
+    /// ```
+    pub fn shrink_to_fit(&mut self) {
+        self.to_default_stride();
```
bluss, line 484: to default stride is an expensive operation - so we can probably avoid it on an array that is already perfectly fit.

We do prefer to not lean too strongly on a default memory order if possible. Any contiguous array should not have its memory order thrown around. I would even prefer if an array keeps its memory order the same, if it can.

src/impl_owned_array.rs
```diff
@@ -401,16 +402,88 @@ impl<A, D> Array<A, D>
...
+        if std::mem::size_of::<A>() != 0 {
+            self.as_ptr().offset_from(self.data.as_ptr())
+        } else {
+            0
+        }
+    }
+
+    /// Convert from any stride to the default stride
+    pub fn to_default_stride(&mut self) {
+        let view_mut =
```
bluss, line 429: I'm not sure about all the details of this method, but I think it's better to take a step back and see if there isn't already some functionality that can do this - maybe .to_shape()?

bluss:
shrink_to_fit is a good idea - should it modify array memory layout? When should it modify? Some more description could help us get a good design.

src/impl_owned_array.rs
```diff
@@ -774,7 +861,6 @@ where
     }
 }
 
-
 /// Sort axes to standard order, i.e Axis(0) has biggest stride and Axis(n - 1) least stride
 ///
 /// Axes in a and b are sorted by the strides of `a`, and `a`'s axes should have stride >= 0 before
```
bluss, line 866: This PR should preferably have a test for the new method :slightly_smiling_face:

src/impl_owned_array.rs
bokutotu: thank you for your review!

Let me check my understanding of `shrink_to_fit`.My understanding of `shrink_to_fit()` is that, for example, you have the following array, and you want to drop the 0,1-th row.

```
1 2 3 4    <- drop
5 6 7 8   <- drop
9 10 11 12
13 14 15 16
```
If you do `shrink_to_fit()` on this `Array`, I think it is supposed to change the memory order as follows.

before
```
1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16
```
after (Here, ○ is `free`ed element.)
```
9 10 11 12 13 14 15 16 ○ ○ ○ ○ ○ ○ ○ ○
```
My understanding of `shrink_to_fit` goes like this. Is it wrong?

As I understand it, `to_shape` and `into_shape` do not seem to change the memory order. Therefore, I created `to_default_stride()`. (Of course, you may not need to change it to default stride.The reason for changing stride to default stride is that we thought that the number of times the element would be swapped would not be significantly different if it were any other stride.)

bokutotu:
Thanks for doing the code review 😀

> shrink_to_fit is a good idea - should it modify array memory layout? When should it modify? Some more description could help us get a good design.

For example, we believe that changing the memory order is necessary in the following cases.
```rust
let a = array![[[0, 1, 2], [3, 4, 5], [6, 7, 8]], [[9, 10, 11], [12, 13, 14], [15, 16, 17]], [[18, 19, 20], [21, 22, 23], [24, 25, 26]]];
let a = a.slice_move(s![.., 1.., ..]);
// a -> 
//   [[[3, 4, 5],
//   [6, 7, 8]],
// 
// [[12, 13, 14],
//  [15, 16, 17]],
//
// [[21, 22, 23],
//  [24, 25, 26]]], shape=[3, 2, 3], strides=[9, 3, 1], layout=c (0x4), const ndim=3
``` 
If we adapt `shrink_to_fit` to this array, I think I need to change the memory order.

bluss:
I'm back next week

bluss:
Apparently not back so fast, I apologize.

bokutotu:
Don't worry about it. We all have busy times 😀

bokutotu:
What is the current status of this PR?

bluss:
Well, I want to come back to this and it would be the first thing I'd work on in ndarray

adamreichold:
@bokutotu If I you are still interested, I would like to try to pick this up again.

What would be really helpful would be to remove the formatting-related part of the diff, just so that is easier to review.

Other than that, I suspect that all of the feedback provided so far has already been addressed.

If you do reboot this, please try to ignore all the Clippy and deprecation issues for now. We will have to get the master branch into shape elsewhere.

bokutotu:
@adamreichold Thanks for your reply. I would like to make a commit to delete the diffs once they are out in cargo fmt. Please wait a bit as work is taking up a lot of my time right now.

PR #1138, opened 2021-12-20
Fix Miri failure with -Zmiri-tag-raw-pointers

jturner314:
Before this PR, running `MIRIFLAGS="-Zmiri-tag-raw-pointers" cargo miri test test_map_axis` caused Miri to report undefined behavior in the `test_map_axis` test. This PR fixes the underlying issue. Basically, Miri doesn't like us using a reference to an element to access other elements. Here's some sample code to illustrate the issue:

```rust
let a: Vec<i32> = vec![5, 6];
let first_elt: &i32 = &a[0];
let ptr: *const i32 = first_elt as *const i32;
// Okay: the data is contained in the data referenced by `first_elt`.
let a0 = unsafe { &*ptr.offset(0) };
assert_eq!(*a0, 5);
// Not okay: the data is not contained in the data referenced by `first_elt`.
let a1 = unsafe { &*ptr.offset(1) };
assert_eq!(*a1, 6);
```

Before this commit, we were using `self.index_axis(axis, 0).map(|first_elt| ...)` to create views of the lanes, from references to the first elements in the lanes. Accessing elements within those views (other than the first element) led to the Miri error, since the view's pointer was derived from a reference to a single element.

Thanks to @5225225 for reporting the issue. (This PR fixes #1137.)

Should we add Miri to CI?

PR #1142, opened 2021-12-24
Fix Miri errors for WindowsIter and ExactChunksIter/Mut

jturner314:
Before this PR, running `MIRIFLAGS="-Zmiri-tag-raw-pointers" cargo miri test` caused Miri to report undefined behavior for code using the `WindowsIter`, `ExactChunksIter`, and `ExactChunksIterMut` iterators. This PR fixes the underlying issue. Basically, Miri doesn't like code which uses a reference to an element to access other elements. Before this PR, these iterators wrapped the `ElementsBase` and `ElementsBaseMut` iterators, and they created views from the references returned by those inner iterators. Accessing elements within those views (other than the first element) led to the Miri error, since the view's pointer was derived from a reference to a single element. Now, the iterators wrap `Baseiter` instead, which produces raw pointers instead of references.

Although not necessary to satisfy Miri, this PR also changes the `Windows`, `ExactChunks`, and `ExactChunksMut` producers to wrap raw views instead of normal views. This avoids potential confusion regarding which elements are accessible through the views produced by these producers.

With this PR, #1138, and bluss/matrixmultiply#67, `MIRIFLAGS="-Zmiri-tag-raw-pointers -Zmiri-check-number-validity" RUSTFLAGS="-Zrandomize-layout" cargo +nightly miri test` runs without errors (in about 8 minutes on my machine).

### DIFF
--- a/src/impl_views/conversions.rs
+++ b/src/impl_views/conversions.rs
@@ -183,6 +183,27 @@ where
     }
 }
 
+/// Private raw array view methods
+impl<A, D> RawArrayView<A, D>
+where
+    D: Dimension,
+{
+    #[inline]
+    pub(crate) fn into_base_iter(self) -> Baseiter<A, D> {
+        unsafe { Baseiter::new(self.ptr.as_ptr(), self.dim, self.strides) }
+    }
+}
+
+impl<A, D> RawArrayViewMut<A, D>
+where
+    D: Dimension,
+{
+    #[inline]
+    pub(crate) fn into_base_iter(self) -> Baseiter<A, D> {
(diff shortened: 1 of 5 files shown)


bluss:
rebased. Safety delayed is safety denied.. yet we try to merge it now. Thanks for the nice work.

I had to do some conflict resolution on this one around where the `Window` struct is created, but I think it turned out correct this way.

bluss:
cc @adamreichold if you are interested in miri, there are instructions here

bluss (approved):

PR #1151, opened 2022-01-22
Make ArrayBase::get_ptr(_mut) public to enable indexing into raw views.

adamreichold:
Noticed this as we would like to provide raw views into NumPy arrays when it is difficult to enforce the required aliasing guarantees for reference-based views, c.f. https://github.com/PyO3/rust-numpy/pull/260#discussion_r790126849

 I do wonder whether there is a reason these cannot be made public but could not find anything using Git archaeology.

### DIFF
--- a/src/impl_methods.rs
+++ b/src/impl_methods.rs
@@ -768,7 +768,7 @@ where
         S: DataMut,
         I: NdIndex<D>,
     {
-        unsafe { self.get_ptr_mut(index).map(|ptr| &mut *ptr) }
+        unsafe { self.get_mut_ptr(index).map(|ptr| &mut *ptr) }
     }
 
     /// Return a raw pointer to the element at `index`, or return `None`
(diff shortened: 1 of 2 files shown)


jturner314:
This looks good to me, although I'd prefer to change the name `get_ptr_mut` to `get_mut_ptr` to be more consistent with the existing `as_mut_ptr` method.

adamreichold:
> I'd prefer to change the name `get_ptr_mut` to `get_mut_ptr` to be more consistent with the existing `as_mut_ptr` method.

Changed as suggested.

jturner314 (approved):
This looks good to me. I just noticed a minor unnecessary line. I've added this PR to the 0.15.x milestone.

src/impl_methods.rs
```diff
@@ -732,14 +732,28 @@ where
...
+    /// let a = arr2(&[[1., 2.], [3., 4.]]);
+    ///
+    /// let v = a.raw_view();
+    /// let p = a.get_ptr((0, 1)).unwrap();
+    ///
+    /// assert_eq!(unsafe { *p }, 2.);
+    /// ```
+    pub fn get_ptr<I>(&self, index: I) -> Option<*const A>
     where
+        S: RawData,
```
jturner314, line 756: Huh, that's weird. I don't know why this didn't show up on the review.

This line isn't necessary, since the `impl` block already has a `S: RawData` bound.

```suggestion
```

adamreichold: Fixed.

PR #1164, opened 2022-05-17
Add link to Slicing docs on libary documentation page

Rikorose:
This commit adds a convenient short cut to directly access the slicing docs.
I use these docs quite often and since the usage is slightly different from what we are used to from numpy, I think this is a good addition.


jturner314:
Thanks!

PR #1171, opened 2022-06-16
Fix CI failures (mostly linting with clippy)

aganders3:
Hello - I noticed CI has been consistently failing for some time, so this PR is meant to bring it back to green. I think it was mostly (all?) clippy lints that were blocking it. The way the GitHub Actions are set up now this could break again (mostly for more clippy lints) with new versions of Rust. I think it's still good to test like this against new/current versions, and I'm happy to keep an eye out for similar issues in the future.

The most frequent was a lint related to `#[doc(hidden)]` which (according to clippy) does not work within Trait impl blocks. I basically just deleted all of them and also combined the `private_decl` and `private_impl` macros - it seems to me the `private_decl` can serve the full purpose of making the Trait private and providing a default implementation. Please let me know if I got this wrong or you'd like to take a different track.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,6 +3,7 @@
 name = "ndarray"
 version = "0.15.5"
 edition = "2018"
+rust-version = "1.51"
 authors = [
   "Ulrik Sverdrup \"bluss\"",
   "Jim Turner"
--- a/src/arrayformat.rs
+++ b/src/arrayformat.rs
@@ -217,7 +217,7 @@ where
             ", shape={:?}, strides={:?}, layout={:?}",
             self.shape(),
             self.strides(),
-            layout = self.view().layout()
+            self.view().layout(),
         )?;
         match D::NDIM {
             Some(ndim) => write!(f, ", const ndim={}", ndim)?,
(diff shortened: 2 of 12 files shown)


aganders3:
Arg, just saw #1165 that already does a good chunk of this!

jturner314:
Thanks for the fixes! I merged #1165 into `master`, rebased this PR off the latest `master`, moved the MSRV addition into a separate commit, and added a few more fixes. I dropped the changes which combined the `private_decl!` and `private_impl!` macros, since they made it possible for users to implement the corresponding traits. (The purpose of those macros is to prevent some of `ndarray`'s traits from being implementable by downstream crates.)

I haven't investigated the remaining `cross_test` failures; based on the error message, it may be an issue with GitHub CI which will resolve itself. If anyone has any ideas how to fix it, please create a PR.

Note on the `breaking-change` label: This PR is a breaking change from the 0.15.x series, since it requires a Rust compiler newer than 1.49.

aganders3:
Oh wow - thanks for getting this into shape and accepting it! I really appreciate your work on this crate.

> I dropped the changes which combined the private_decl! and private_impl! macros, since they made it possible for users to implement the corresponding traits. (The purpose of those macros is to prevent some of ndarray's traits from being implementable by downstream crates.)

Ah, okay. I'll have to look over exactly how this works but that makes sense.

Thanks again!

PR #1195, opened 2022-08-13
Simplify `s![]`

### LINKED ISSUE #1194: Use of `$crate` in macro patterns is questionable
See https://github.com/rust-lang/rust/pull/99447#issuecomment-1204019943

The fact that `$crate` in the macro_rules LHS is accepted and glued into `kw::DollarCrate` is likely an accidental bug. The single use is an internal macro helper arm,

https://github.com/rust-ndarray/ndarray/blob/f8522cfeb60b64067ffb1886b0509092a3bacd1f/src/slice.rs#L845

which looking at the macro, should probably be written to take `$in_dim:expr, $out_dim:expr,` like the other arms anyway.

CAD97:
Fixes #1194 with a very slight generalization to make this arm more similar to the other arms.

### DIFF
--- a/src/slice.rs
+++ b/src/slice.rs
@@ -844,9 +844,11 @@ macro_rules! s(
     // empty call, i.e. `s![]`
     (@parse $in_dim:expr, $out_dim:expr, []) => {
         {
+            let in_dim = $in_dim;
+            let out_dim = $out_dim;
             #[allow(unsafe_code)]
             unsafe {
-                $crate::SliceInfo::new_unchecked([], $in_dim, $out_dim)
+                $crate::SliceInfo::new_unchecked([], in_dim, out_dim)
             }
         }
     };
--- a/src/slice.rs
+++ b/src/slice.rs
@@ -842,15 +842,11 @@ macro_rules! s(
         }
     };
     // empty call, i.e. `s![]`
-    (@parse ::core::marker::PhantomData::<$crate::Ix0>, ::core::marker::PhantomData::<$crate::Ix0>, []) => {
+    (@parse $in_dim:expr, $out_dim:expr, []) => {
         {
             #[allow(unsafe_code)]
             unsafe {
-                $crate::SliceInfo::new_unchecked(
-                    [],
-                    ::core::marker::PhantomData::<$crate::Ix0>,
-                    ::core::marker::PhantomData::<$crate::Ix0>,
-                )
+                $crate::SliceInfo::new_unchecked([], $in_dim, $out_dim)
             }
         }
     };


src/slice.rs
```diff
@@ -842,15 +842,11 @@ macro_rules! s(
...
+    (@parse $in_dim:expr, $out_dim:expr, []) => {
         {
             #[allow(unsafe_code)]
             unsafe {
-                $crate::SliceInfo::new_unchecked(
-                    [],
-                    ::core::marker::PhantomData::<$crate::Ix0>,
-                    ::core::marker::PhantomData::<$crate::Ix0>,
-                )
+                $crate::SliceInfo::new_unchecked([], $in_dim, $out_dim)
```
bluss, line 849: what's the impact on the `unsafe` code block here?

CAD97: It's still necessary, as `SliceInfo::new_unchecked` is unsafe.

Technically, this means that someone using the macro could write

```rust
s! { @parse
{
    // statements in an unsafe context
    ::core::marker::PhantomData::<::ndarray::Ix0>
},
::core::marker::PhantomData<::ndaray::Ix0>,
[]
}
```

to get an `unsafe` context without writing `unsafe`, but this is an internal rule, so that's not too worrisome. However, it is reasonable to evaluate the arguments outside of the `unsafe` block just for thoroughness.

CAD97:
Check failures are

- [unrelated](https://github.com/rust-ndarray/ndarray/runs/7817747958?check_suite_focus=true#step:4:57) (new clippy warning)
- [spurious](https://github.com/rust-ndarray/ndarray/runs/7817747878?check_suite_focus=true#step:6:1279) docker error?
  > docker: Error response from daemon: Duplicate mount point: /home/runner/work/ndarray/ndarray.

src/slice.rs
jturner314: Thanks for bringing this up. Some of the other `@parse` arms had safety issues if called directly. I've created #1196 to fix that. With that change, we can change this PR to modify the "empty case" arm to the following:

```rust
    // empty call, i.e. `s![]`
    (@parse $in_dim:expr, $out_dim:expr, []) => {
        {
            debug_assert_eq!($in_dim, ::core::marker::PhantomData::<$crate::Ix0>);
            debug_assert_eq!($out_dim, ::core::marker::PhantomData::<$crate::Ix0>);
            (
                [],
                ::core::marker::PhantomData::<$crate::Ix0>,
                ::core::marker::PhantomData::<$crate::Ix0>,
            )
        }
    };
```

This is safe, since it always expands to valid values for the `SliceInfo::new_unchecked` call (which is in the `($($t:tt)*)` arm with #1196). It has debug assertions to ensure that `$in_dim` and `$out_dim` are the expected values. (They'll always be the expected values, unless the user calls `s![@parse ...]` directly.

bluss:
I'm sorry this has now been conflicted by other changes, so I'll close this for now. I'm not sure about the simplification, the old version makes sure the dimensions are really Ix0 (empty) for the empty slice, and this one doesn't have the same effect? Not sure.

CAD97:
It doesn't, tbf. The main reason I made the PR is that `$crate` in macro patterns is pretty cursed.

PR #1196, opened 2022-08-13
Fix unsafe blocks in s![] macro

jturner314:
Before, the user could silently violate safety requirements of `SliceInfo::new_unchecked` by directly calling `s![@parse inconsistent_values]`, where `inconsistent_values` represents inconsistent values for the dimensions, etc. Now, the macro calls `SliceInfo::new_unchecked` only in the `($($t:tt)*)` arm, which always constructs the correct values for the call.


jturner314:
CI results:

- The clippy warning is a bug (bad diagnostic) in clippy on the current Rust beta; it's fixed in the latest nightly.
- The `cross_test` error is something wrong with the GitHub action -- it's not an issue with `ndarray`.

bluss (approved):

PR #1201, opened 2022-08-28
Allow references to be used as scalar operands

### LINKED ISSUE #1200: Use references for scalar operands
It would be useful for me to be able to use references to scalars instead of owned instances when performing operations on arrays. 

e.g., this fails: 
```rust
let arr1 = array!([1, 2, 3]);
let four: i32 = 4;
let arr2 = arr1 + &four; //error
```

I naively tried modifying the ops macros to accept references but kept getting a conflicting implementation error. It seems that the compiler is worried that someone will come along an add an implementation to `StaticOperand` for reference types: 

```
    = note: downstream crates may implement trait `impl_ops::ScalarOperand` for type `&_`
```

...which is strange because there's a `'static` type bound on `ScalarOperand`? 

In any case, if anyone has any guidance on how to proceed, I'd greatly appreciate it! I'm happy to open a PR with a nudge in the right direction.

zackangelo:
Fixes #1200 

Relaxes the type constraint on `StaticOperand` so that it may be used for references. 

This example now works: 

```rust
let arr1 = array!([1, 2, 3]);
let four: i32 = 4;
let arr2 = arr1 + &four; //previously errored on lack of an Add<> impl
```

### DIFF
--- a/src/impl_ops.rs
+++ b/src/impl_ops.rs
@@ -31,7 +31,7 @@ use num_complex::Complex;
 /// This trait ***does not*** limit which elements can be stored in an array in general.
 /// Non-`ScalarOperand` types can still participate in arithmetic as array elements in
 /// in array-array operations.
-pub trait ScalarOperand: 'static + Clone {}
+pub trait ScalarOperand: Clone {}
 impl ScalarOperand for bool {}
 impl ScalarOperand for i8 {}
 impl ScalarOperand for u8 {}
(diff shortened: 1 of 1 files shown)


zackangelo:
The CI failure appears to be unrelated to the code change: 
```
docker: Error response from daemon: Duplicate mount point: /home/runner/work/ndarray/ndarray.
```

bluss:
Removing `'static` is a breaking change. `'static` implies that one can use `Any` and ad hoc type based "specialization", which is why we have snuck in `'static` in a few places. We would like to continue to have `'static` in all places where we'd like to insert explicit SIMD or BLAS etc operations.

bluss:
closing for now because keeping 'static is preferred

PR #1210, opened 2022-09-24
Add NumPy examples combining slicing and assignment

### LINKED ISSUE #1209: Add numpy example for assigning to a slice
To assign values to a slice of a numpy array it's possible to do e.g.:
```python
import numpy as np
a = np.zeros((2, 2))
a[:, 0] = 1
```
It would be good to add how to do this in `ndarray` to the ndarray for numpy users docs

jturner314:
Fixes #1209.


dstansby:
👍 this looks great

bluss (approved):

PR #1211, opened 2022-09-24
Add slice_axis_move method

jturner314:


src/impl_methods.rs
```diff
@@ -655,6 +655,16 @@ where
         debug_assert!(self.pointer_is_inbounds());
     }
 
+    /// Slice the array in place along the specified axis, then return the sliced array.
+    ///
+    /// **Panics** if an index is out of bounds or step size is zero.<br>
+    /// **Panics** if `axis` is out of bounds.
+    #[must_use = "slice_axis_move returns an array with the sliced result"]
```
bluss, line 662: oh I was going to ask about the must_use, ownership semantics should mean we don't need it. But ArrayViews can be copy, so it applies well there. Nice catch.

bluss (approved):

Issue #1208, opened 2022-09-19
Lifetime issues with `slice` methods


LeBoucEtMistere:
Hello folks,

I think I found an issue in the way lifetimes are inferred when using `slice` or `slice_axis` on array views.

Here is a minimal example with three functions, only `bar` and `baz` compile, while `foo` doesn't. I would argue that `foo` should be able to compile.

```rust
use ndarray; // 0.15.6

use ndarray::{ArrayBase, ViewRepr, Axis, Dimension, Slice, ArrayView};

fn main() {
    let a = ndarray::array![[1, 2, 3], [4, 5, 6]];
    foo(a.view(), ndarray::Axis(1));
}

fn foo<T, D>(arrayview: ArrayView<T, D>, axis: ndarray::Axis) -> ArrayView<T, D>
where
    D: Dimension,
{
    arrayview.slice_axis(axis, Slice::from(1..2))
}

fn bar<T, D>(arrayview: &mut ArrayView<T, D>, axis: Axis)
where
    D: Dimension,
{
    arrayview.slice_axis_inplace(axis, Slice::from(1..2));
}

fn baz<'a, 'data: 'a, T, D>(
    arrayview: &'a ArrayBase<ViewRepr<&'data T>, D>,
    axis: Axis,
) -> ArrayBase<ViewRepr<&'a T>, D>
where
    D: Dimension,
{
    arrayview.slice_axis(axis, Slice::from(1..2))
}
```
(playground link: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=7d62b20eec45d257e629dae07840544b)

The error is the following:
```
error[[E0311]](https://doc.rust-lang.org/stable/error-index.html#E0311): the parameter type `D` may not live long enough
  --> src/main.rs:14:5
   |
14 |     arrayview.slice_axis(axis, Slice::from(1..2))
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
note: the parameter type `D` must be valid for the anonymous lifetime defined here...
  --> src/main.rs:10:25
   |
10 | fn foo<T, D>(arrayview: ArrayView<T, D>, axis: ndarray::Axis) -> ArrayView<T, D>
   |                         ^^^^^^^^^^^^^^^
note: ...so that the type `D` will meet its required lifetime bounds
  --> src/main.rs:14:5
   |
14 |     arrayview.slice_axis(axis, Slice::from(1..2))
[...]

PR #1219, opened 2022-10-10
Impl a lifetime-relaxed broadcast for ArrayView

roblabla:
ArrayView::broadcast has a lifetime that depends on &self instead of its internal buffer. This prevents writing some types of functions in an allocation-free way. For instance, take the numpy `meshgrid` function: It could be implemented like so:

```rust
fn meshgrid_2d<'a, 'b>(coords_x: ArrayView1<'a, X>, coords_y: ArrayView1<'b, X>) -> (ArrayView2<'a, X>, ArrayView2<'b, X>) {
    let x_len = coords_x.shape()[0];
    let y_len = coords_y.shape()[0];

    let coords_x_s = coords_x.into_shape((1, y_len)).unwrap();
    let coords_x_b = coords_x_s.broadcast((x_len, y_len)).unwrap();
    let coords_y_s = coords_y.into_shape((x_len, 1)).unwrap();
    let coords_y_b = coords_y_s.broadcast((x_len, y_len)).unwrap();

    (coords_x_b, coords_y_b)
}
```

Unfortunately, this doesn't work, because `coords_x_b` is bound to the lifetime of `coord_x_s`, instead of being bound to 'a.

This PR introduces a new function, broadcast_ref, that does just that.


roblabla:
The `broadcast_ref` implementation is copy pasted from `broadcast`, which is kinda bad. The common code (in particular, `upcast`) should probably be moved to a common module, but I'm not sure where it should go.

jturner314:
Thanks for this PR! I agree that this method would be useful.

This is related to issue #1208, which discussed slicing. However, broadcasting is different, because we can't have a broadcasting equivalent of `.slice_move()` for arbitrary storage types. We have to have a method specific to `ArrayView`, as in this PR.

I don't like the name `broadcast_ref` much, but it's fine if we can't come up with anything better.

> The `broadcast_ref` implementation is copy pasted from `broadcast`, which is kinda bad. The common code (in particular, `upcast`) should probably be moved to a common module, but I'm not sure where it should go.

`src/dimension/mod.rs` would be a good place for `upcast`, since there are other similar functions in that file, e.g. `do_slice`. I think just moving `upcast` into that file would be sufficient; the rest of the code in `broadcast` is small enough to just duplicate, IMO.

jreniel:
Hello! 
I was looking for an equivalent of numpy's meshgrid on ndarray. Because I couldn't find it, I came up with my own version, which should work in n-dimensions. It is currently hosted [here](https://github.com/jreniel/meshgridrs). I think this should belong into ndarray, and not in a separate crate, so I'd be more than happy, provided that you find my implementation satisfying, to see this code directly integrated into ndarray, instead of having a separate crate as it currently is.

nilgoyette:
@jreniel  Please open an issue if you want to discuss an implementation of `meshgrid`. It's not related to this PR.

PR #1226, opened 2022-10-30
Fixing CI errors

skewballfox:
I noticed that CI was failing, I figured I'd do a PR to fix issues I'm noticing

the first CI failure is due to use of a deprecated function `itertools::zip` in `test/array.rs`.

switching to `std::iter::zip` requires updating the MSRV in `ci.yml` from 1.51 to 1.59, I went ahead and changed my copy to that rust version, but I don't know if there is any specific reason why 1.51 was set as the minimum. If there is, what would be the best alternative to using either version of zip?

the second error seems to be due to a duplicate mount point for the i686 cross-test, still working on that one.

the third error is for the clippy lints, I fixed two by running `cargo clippy --fix --no-deps`, but the third is less straight forward. 

EDIT: 
the second error I believe is caused by the layout of the project, with ndarray and ndarray-rand both being dependencies with relative imports. because ndarray is in the parent directory, and ndarray-rand is a child directory, this creates duplicate mountpoints. I could fix this by moving ndarray into it's own directory and keeping the top level parent for project specs but I should probably confirm that is the direction you guys would like to go first. There may be a way to fix it by providing some info to cross via an env variable or CLI flag, still looking through the repo for cross to figure out if that's the case

EDIT2: I fixed this by adding xtest-numeric to the list of workspace members, which makes all the CI test pass, but has the downside of making the CI test [run for around 20 minutes](https://github.com/skewballfox/ndarray/actions/runs/3768904426). Before the test from `all-test.sh` were finishing pretty quickly. Is there anything else I need to change when adding `xtest-numeric` to the list of workspace-members?

### DIFF
--- a/src/zip/mod.rs
+++ b/src/zip/mod.rs
@@ -14,16 +14,16 @@ mod ndproducer;
 use std::mem::MaybeUninit;
 
 use crate::imp_prelude::*;
+use crate::partial::Partial;
 use crate::AssignElem;
 use crate::IntoDimension;
 use crate::Layout;
-use crate::partial::Partial;
 
-use crate::indexes::{indices, Indices};
-use crate::split_at::{SplitPreference, SplitAt};
 use crate::dimension;
+use crate::indexes::{indices, Indices};
+use crate::split_at::{SplitAt, SplitPreference};
 
-pub use self::ndproducer::{NdProducer, IntoNdProducer, Offset};
+pub use self::ndproducer::{IntoNdProducer, NdProducer, Offset};
 
 /// Return if the expression is a break value.
 macro_rules! fold_while {
@@ -91,6 +91,7 @@ where
     D: Dimension,
 {
     type Output = ArrayView<'a, A, E::Dim>;
+    #[allow(clippy::needless_borrow, unconditional_recursion)]
     fn broadcast_unwrap(self, shape: E) -> Self::Output {
         let res: ArrayView<'_, A, E::Dim> = (&self).broadcast_unwrap(shape.into_dimension());
         unsafe { ArrayView::new(res.ptr, res.dim, res.strides) }
@@ -196,7 +197,6 @@ pub struct Zip<Parts, D> {
     layout_tendency: i32,
 }
 
-
 impl<P, D> Zip<(P,), D>
 where
     D: Dimension,
@@ -256,7 +256,6 @@ where
     );
 }
 
-
 impl<Parts, D> Zip<Parts, D>
 where
     D: Dimension,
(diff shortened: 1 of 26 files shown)


tests/zst.rs
```diff
@@ -18,6 +18,6 @@ fn test_swap() {
 #[test]
 fn test() {
     let c = ArcArray::<(), _>::default((3, 4));
-    let mut d = c.clone();
+    let mut d = c;
```
bluss, line 21: Cloning an Arc changes the ref count. It has to be assumed to be part of the test.

src/zip/mod.rs
```diff
@@ -92,7 +92,7 @@ where
 {
     type Output = ArrayView<'a, A, E::Dim>;
     fn broadcast_unwrap(self, shape: E) -> Self::Output {
-        let res: ArrayView<'_, A, E::Dim> = (&self).broadcast_unwrap(shape.into_dimension());
+        let res: ArrayView<'_, A, E::Dim> = self.broadcast_unwrap(shape.into_dimension());
```
bluss, line 95: Linter is wrong and it shouldn't be changed, it looks like. local suppression possible?

tests/array.rs
```diff
@@ -2466,74 +2466,48 @@ mod array_cow_tests {
 
 #[test]
 fn test_remove_index() {
-    let mut a = arr2(&[[1, 2, 3],
-                       [4, 5, 6],
-                       [7, 8, 9],
-                       [10,11,12]]);
+    let mut a = arr2(&[[1, 2, 3], [4, 5, 6], [7, 8, 9], [10, 11, 12]]);
```
bluss, line 2469: If we can avoid it, we won't commit changes like this where nice visual formatting is removed

tests/zst.rs
skewballfox: I can revert the changes and then add some stuff to ignore lints on specific lines(I think)

src/zip/mod.rs
skewballfox: yeah, I believe so, at least it seems that way from this [SO post](https://stackoverflow.com/questions/55402812/how-to-disable-a-clippy-lint-for-a-single-line-block)

skewballfox:
hey just letting you know, haven't abandoned this, just hit a rough patch of deadlines, likely won't circle back round until after December 7th. If someone wants to perform a PR on this branch to fix issues before then (rather than opening a duplicate PR), I'm happy to merge them.

tests/array.rs
skewballfox: I reverted the changes to the original and reapplied the fixes.

src/zip/mod.rs
skewballfox: I suppressed this lint for the specified function

tests/zst.rs
skewballfox: reverted and suppressed lints for function

PR #1228, opened 2022-11-08
Add Zip::any

nilgoyette:
I find it strange that `ndarray` offers `Zip::all` but not `Zip::any`. Of course we can replace this function with normal iteration tools from std and itertools, but `Zip` is the by far fastest horse around.

I checked with Python and C++ for the `Returns false if the input arrays are empty` behavior.

### DIFF
--- a/src/zip/mod.rs
+++ b/src/zip/mod.rs
@@ -670,6 +670,33 @@ macro_rules! map_impl {
                 }).is_done()
             }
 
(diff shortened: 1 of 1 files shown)


bluss:
any(p) is the same as !all(!p) and I've previously been against bloating the list of adaptors here, but it seems fine.

bluss (approved):

bluss:
Thanks!

PR #1235, opened 2022-11-29
Remove Div trait bound in LinalgScalar trait

vadym-kl:
A minimal way to partially address #737 and enable the use of ndarray for matrices with integer entries.
Motivated by applications in computational number theory and cryptography, see #737.


vadym-kl:
After feedback in #737, it is clear that this is not a suitable way of enabling matrices with integer entries

PR #1237, opened 2022-11-29
implement DoubleEndedIterator for 1d `LanesIter`

Muthsera:
In a project I wanted to iterate over the rows and columns of a 2d array in reverse but while `rows` and `columns` exist, the resulting `LanesIter` does not implement `DoubleEndedIterator` as of yet. It is possible to use indexing via `row` with reversed indices but that is less ergonomic.

This PR fixes this at least for 1d `LanesIter` and now allows one to do:
```
let a = ArcArray::from_iter(0..8).reshape((4, 2));
for (row, check) in a.rows().into_iter().rev().zip(&[[6, 7], [4, 5], [2, 3], [0, 1]]) {
    assert_equal(row, check);
}
```

The implementation just delegates to the underlying 1d `Baseiter` which implements `DoubleEndedIterator`.
The trait is not implemented for higher dimensions however, which would need some additional work.

### DIFF
--- a/src/iterators/mod.rs
+++ b/src/iterators/mod.rs
@@ -732,6 +732,15 @@ where
     }
 }
 
+impl<'a, A> DoubleEndedIterator for LanesIter<'a, A, Ix1>
+{
+    fn next_back(&mut self) -> Option<Self::Item> {
+        self.iter.next_back().map(|ptr| unsafe {
+            ArrayView::new_(ptr, Ix1(self.inner_len), Ix1(self.inner_stride as Ix))
+        })
+    }
+}
+
 // NOTE: LanesIterMut is a mutable iterator and must not expose aliasing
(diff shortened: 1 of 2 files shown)


src/iterators/mod.rs
```diff
@@ -726,6 +726,18 @@ where
     }
 }
 
+impl<'a, A, D> DoubleEndedIterator for LanesIter<'a, A, D>
+where
+    D: Dimension,
```
bluss, line 731: I expected this to be limited to the case where D=Ix1. Just like how .iter()'s `Iter` works. I think it's better that way, then the Baseiter does not leak out (it's in the where clause, that's a leak into public API.)

bluss: Fixed that, and also applied the same change to LanesIterMut - it and LanesIter must of course change together.

PR #1249, opened 2022-12-27
Added stride support to `Windows`

LazaroHurtado:
Currently the `ArrayBase::windows()` method creates an iterator of windows with a stride of 1 along all axes of the given array's dimensions. This is limiting the functionality of windows on arrays. One use case where having the ability to specify the stride is necessary is in convolutional neural networks, where what ndarray calls a window is basically a kernel.

### Important notes
* Including a specified stride [changes the size](https://github.com/rust-ndarray/ndarray/compare/master...LazaroHurtado:ndarray:windows_stride_feature?expand=1#diff-343ac327e5611c195955b2934ca9115dafde8d800a3fe09a79edc51a27b7f9c3R90) of the `ArrayView` found in the `base` field of  the `Windows` struct. This change is derived from [this](https://pytorch.org/docs/stable/generated/torch.nn.Conv2d.html#torch.nn.Conv2d) formula.
* ~The ordering for the stride is `(inner_most_axis, ..., outer_most_axis)`. This is because the axis that the windows gets pushed along first is the inner most axis...~ This was the original intention but noticing that the formula also follows outermost axis first, similar to indexing an ndarray, I believe it should be similar.

### DIFF
--- a/src/iterators/windows.rs
+++ b/src/iterators/windows.rs
@@ -20,32 +20,15 @@ impl<'a, A, D: Dimension> Windows<'a, A, D> {
         E: IntoDimension<Dim = D>,
     {
         let window = window_size.into_dimension();
-        ndassert!(
-            a.ndim() == window.ndim(),
-            concat!(
-                "Window dimension {} does not match array dimension {} ",
-                "(with array of shape {:?})"
-            ),
-            window.ndim(),
-            a.ndim(),
-            a.shape()
-        );
-        let mut size = a.dim;
-        for (sz, &ws) in size.slice_mut().iter_mut().zip(window.slice()) {
-            assert_ne!(ws, 0, "window-size must not be zero!");
-            // cannot use std::cmp::max(0, ..) since arithmetic underflow panics
-            *sz = if *sz < ws { 0 } else { *sz - ws + 1 };
+        let ndim = window.ndim();
+
+        let mut unit_stride = D::zeros(ndim);
+        let stride_slice = unit_stride.slice_mut();
+        for s in stride_slice.iter_mut() {
+            *s = 1;
         }
 
-        let window_strides = a.strides.clone();
-
-        unsafe {
-            Windows {
-                base: ArrayView::new(a.ptr, size, a.strides),
-                window,
-                strides: window_strides,
-            }
-        }
+        Windows::new_with_stride(a, window, unit_stride)
     }
 
     pub(crate) fn new_with_stride<E>(a: ArrayView<'a, A, D>, window_size: E, strides: E) -> Self
--- a/src/iterators/windows.rs
+++ b/src/iterators/windows.rs
@@ -23,11 +23,8 @@ impl<'a, A, D: Dimension> Windows<'a, A, D> {
         let ndim = window.ndim();
 
         let mut unit_stride = D::zeros(ndim);
-        let stride_slice = unit_stride.slice_mut();
-        for s in stride_slice.iter_mut() {
-            *s = 1;
-        }
-
+        unit_stride.slice_mut().fill(1);
+        
         Windows::new_with_stride(a, window, unit_stride)
     }
 
--- a/src/iterators/windows.rs
+++ b/src/iterators/windows.rs
@@ -47,6 +47,64 @@ impl<'a, A, D: Dimension> Windows<'a, A, D> {
             }
         }
     }
+
+    pub(crate) fn new_with_stride<E>(a: ArrayView<'a, A, D>, window_size: E, strides: E) -> Self
+    where
+        E: IntoDimension<Dim = D>,
+    {
+        let window = window_size.into_dimension();
+        let strides_d = strides.into_dimension();
+        ndassert!(
+            a.ndim() == window.ndim(),
+            concat!(
+                "Window dimension {} does not match array dimension {} ",
+                "(with array of shape {:?})"
+            ),
+            window.ndim(),
+            a.ndim(),
+            a.shape()
+        );
+        ndassert!(
+            a.ndim() == strides_d.ndim(),
+            concat!(
+                "Stride dimension {} does not match array dimension {} ",
+                "(with array of shape {:?})"
+            ),
+            strides_d.ndim(),
+            a.ndim(),
+            a.shape()
+        );
+        let mut size = a.dim;
+        for ((sz, &ws), &stride) in size
+            .slice_mut()
+            .iter_mut()
+            .zip(window.slice())
+            .zip(strides_d.slice())
+        {
+            assert_ne!(ws, 0, "window-size must not be zero!");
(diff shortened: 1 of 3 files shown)


LazaroHurtado:
> Now that you created `Windows::new_with_stride`, shouldn't `Windows::new` simply calls `Windows::new_with_stride` with `(1, ...)`? If such a thing is possible, of course.

Great insight, changes have been applied, if there is a more effective way of creating a unit stride please let me know.

tests/windows.rs
```diff
@@ -95,6 +95,76 @@ fn windows_iterator_3d() {
     );
 }
 
+/// Test that verifies the `Windows` iterator panics when stride has an axis equal to zero.
+#[test]
+#[should_panic]
+fn windows_iterator_stride_axis_zero() {
+    let a = Array::from_iter(10..37).into_shape((3, 3, 3)).unwrap();
+    a.windows_with_stride(Dim((2, 2, 2)), Dim((0,2,2)));
```
nilgoyette, line 103: I'm aware that the original maintainer used `Dim` here, but it's useless. It just makes the code longer and harder to read. You can do like you did in the test below: directly use the tuple `(2, 2, 2)`. Please change in all the new tests.

tests/windows.rs
```diff
@@ -95,6 +95,76 @@ fn windows_iterator_3d() {
...
+        ],
+    );
+}
+
+/// Simple test for iterating 2d-arrays via `Windows` with stride.
+#[test]
+fn windows_iterator_2d_with_stride() {
+    let a = Array::from_iter(10..30).into_shape((5, 4)).unwrap();
+    itertools::assert_equal(
+        a.windows_with_stride(Dim((3, 2)), Dim((2,1))),
```
nilgoyette, line 137: Can you please use the standard rust formatter? (Warning: it will format other lines in this file. I'm not sure what to do about it).

nilgoyette: Lets discuss this problem in #1292 . Meanwhile, @LazaroHurtado, please ignore the formatting.

LazaroHurtado: I used `rustfmt` and restored lines that are not a part of my PR but were affected by the formatter

LazaroHurtado:
@nilgoyette Please let me know if you would like me to change the `windows` doc description to refer to `windows_with_stride` with unit stride, similar to [this](https://docs.rs/ndarray/latest/ndarray/struct.ArrayBase.html#method.multi_slice_move).

nilgoyette:
Yes, their description will be identical except for one detail. Good idea.

LazaroHurtado:
Any more feedback would be highly appreciated! If collaborators would want me to squash all commits into one then would be happy to do so.

src/iterators/windows.rs
```diff
@@ -20,6 +20,23 @@ impl<'a, A, D: Dimension> Windows<'a, A, D> {
         E: IntoDimension<Dim = D>,
     {
         let window = window_size.into_dimension();
+        let ndim = window.ndim();
+
+        let mut unit_stride = D::zeros(ndim);
+        let stride_slice = unit_stride.slice_mut();
```
bluss, line 26: Since you asked, I think `unit_stride.slice_mut().fill(1);` would work here.

LazaroHurtado: Ahh was totally unaware this was possible, thank you!

nilgoyette:
Thank you for your contribution @LazaroHurtado

src/iterators/windows.rs
```diff
@@ -30,18 +44,42 @@ impl<'a, A, D: Dimension> Windows<'a, A, D> {
...
+            } else {
+                ((*sz - (ws - 1) - 1) / stride) + 1
+            };
         }
-
         let window_strides = a.strides.clone();
 
+        let mut array_strides = a.strides.clone();
+        for (arr_stride, ix_stride) in array_strides.slice_mut().iter_mut().zip(strides_d.slice()) {
+            *arr_stride *= ix_stride;
```
bluss, line 77: For some inputs this multiplication will wrap around. Check on input is needed so that the resulting array view is always valid in those and other cases, otherwise it's a "back door" into creating an invalid array view and UB.

LazaroHurtado: Good catch! I can release a follow-up PR using `checked_mul` and `assert_ne!` on the checked result with `None` ensuring the array view is valid, would you consider this sufficient?

bluss:
@nilgoyette first of all I'm apologizing to you. I want to contribute net positive here (and not net negative) and chiming in like I did with stuff to second-guess anyone's review is not what I want to do. I want ndarray to live without me, and then it needs to merge stuff without my view of things have to be (which I repeat because that's what I've long accepted, but it's harder to do than say). :slightly_smiling_face: 

I see you're doing reviews and getting stuff merged and that's great :slightly_smiling_face:. Be direct in what you need from me, because I might not be around to do it otherwise.

bluss:
@LazaroHurtado I've rewritten my comment to be shorter and more clear (and deleting the comments github managed to duplicate)

- array strides are signed quantities, stored in an usize field, and that's why we should avoid doing arithmetic on them explicitly. Here we'll have panics (debug mode) due to wraparound immediately if the user passes in any array with inverted axis (negative stride)

- Please use a higher level call to do the same operation, I found the slice axis inplace methods fit well for this

We want to compute `base` by using a number of `a.slice_axis_inplace(...)` calls.
This is how to do it for the old code (before this PR, just needs expansion to the new features:)

```rust
let window_strides = a.strides.clone();

let mut base = a;
base.slice_each_axis_inplace(|ax_desc| {
    let len = ax_desc.len;
    let wsz = window[ax_desc.axis.index()];
    if len < wsz {
        Slice::new(0, Some(0), 1)
    } else {
        Slice::new(0, Some((len - wsz + 1) as isize), 1)
    }
});
Windows {
    base,
    window,
    strides: window_strides,
}
```

LazaroHurtado:
Thank you @bluss for providing your feedback on this new feature, I highly appreciate it. I will take responsibility for making the proper changes to close out any potential bugs or improve code readability related to my work. Please be patient with me since I am still not 100% familiar with this crate and I have limited free time, nonetheless, I will deliver and keep you and @nilgoyette in the loop.

nilgoyette:
Thank you @bluss for "refactoring" your comments. I don't know about@LazaroHurtado , but I was somewhat confused :|

As for the personnal comment

- I understand the apology, but it is not needed (!) You saw a bug that I didn't have the knowledge to see and you explained it. This is perfect. Honestly, I prefer having random educated comments from you than nothing.
- "_Be direct in what you need from me_" Well, my goal was to disturb you as less as possible (and even less than that) but you're the only person that knows the whole codebase and its dark corners so I guess I'll need to use the "@bluss" from time to time. When I do, I'll try to keep it short and clear.

PR #1268, opened 2023-03-11
Add reserve method for owned arrays

ssande7:
This PR adds an `array.reserve(axis, additional)` method for use when appending multiple times to an array.
It abstracts over the code in `append` that was calling `OwnedRepr::reserve` in order to expose it.

Benchmark results (appending 100 times to an empty array):
```
test push_no_reserve ... bench:         900 ns/iter (+/- 52)                                                                                                                          
test push_reserve    ... bench:         739 ns/iter (+/- 50) 
```

Note I'm new to the crate so I may have missed something, but happy to iterate if there's more to do.

### DIFF
--- a/src/impl_owned_array.rs
+++ b/src/impl_owned_array.rs
@@ -764,9 +764,6 @@ where D: Dimension
     pub fn reserve(&mut self, axis: Axis, additional: usize)
     where D: RemoveAxis
     {
-        if additional == 0 {
-            return;
-        }
         let self_dim = self.raw_dim();
         let remaining_shape = self_dim.remove_axis(axis);
         let len_to_append = remaining_shape.size() * additional;
--- a/src/impl_owned_array.rs
+++ b/src/impl_owned_array.rs
@@ -660,14 +660,10 @@ where D: Dimension
             self.strides.clone()
         };
 
-        unsafe {
-            // grow backing storage and update head ptr
-            let offset_from_alloc_to_logical = self.offset_from_alloc_to_logical_ptr().unwrap_or(0);
-            self.ptr = self
-                .data
-                .reserve(len_to_append)
-                .add(offset_from_alloc_to_logical);
+        // grow backing storage and update head ptr
+        self.reserve(axis, array_dim[axis.index()]);
 
+        unsafe {
             // clone elements from view to the array now
             //
             // To be robust for panics and drop the right elements, we want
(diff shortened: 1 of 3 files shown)


src/impl_owned_array.rs
```diff
@@ -657,6 +687,43 @@ impl<A, D> Array<A, D>
...
+    /// use ndarray::{Array3, Axis};
+    /// let mut a = Array3::<i32>::zeros((0,2,4));
+    /// a.reserve(Axis(0), 1000);
+    /// assert!(a.into_raw_vec().capacity() >= 2*4*1000);
+    /// ```
+    pub fn reserve(&mut self, axis: Axis, additional: usize)
+    where
+        D: RemoveAxis,
+    {
+        let self_dim = self.raw_dim();
```
bluss, line 709: I think this method should have an assert that axis is in bounds, that len_to_append didn't overflow (and if data.reserve doesn't check it, also if the calculation in there doesn't overflow)

Then document the panic on invalid argument as is usual (see other methods that can panic)

ssande7: Good point - I think the latest commits should address this. `data.reserve` does check whether the new capacity would be larger than `isize::MAX`, but I've added some bounds checking anyway that returns a ShapeError in that case in keeping with other functions (in addition to the panic if `len_to_append` would overflow `usize::MAX` or if adding it to the current len causes it to wrap).

src/impl_owned_array.rs
```diff
@@ -658,6 +700,69 @@ impl<A, D> Array<A, D>
...
+        D: RemoveAxis,
+    {
+        debug_assert!(axis.index() < self.ndim());
+        let self_dim = self.raw_dim();
+        let remaining_shape = self_dim.remove_axis(axis);
+
+        // Make sure added capacity doesn't overflow
+        debug_assert!(remaining_shape.size().checked_mul(additional).is_some());
+        let len_to_append = remaining_shape.size() * additional;
+
```
bluss, line 739: I'm sorry for the late response here, but properties that are required for soundness (protects rules we necessarily uphold in our code and allows using `unsafe`) they **must be protected by assertions**, not debug assertions. The function must be a safe abstraction and have a sound result for any input parameter, for any regular compilation mode.

I haven't analyzed the code further, sometimes our preconditions are protected by other properties guaranteed by valid arrays and so on, so further thinking is needed but today I'm doing quick scans over PRs and won't delve into the details.

ssande7: Ah I see. Looking back since it's been a while, I think you're right that this check should be stronger. The call to `dimension::size_of_shape_checked` would already be catching cases where the overall size overflows but the new size of the growing dimension is still < `usize::MAX`, but it would miss the case of an overflow when calculating the new length of that growing dimension.

Since it would need to be evaluated anyway in an assert, I think just assigning from the `checked_mul` or returning a `ShapeError::Overflow` makes the most sense here (see most recent commit). Technically the addition to the growing dimension (impl_owned_array.rs:740 in that commit) could still overflow if another dimension had length 0, but in that case we would be reserving 0 bytes so I don't think that's a problem.

bluss:
Rebased, resolved conflict, added test for nonzero alloc to first element pointer

bluss:
Thanks a lot for this

bluss (approved):

src/impl_owned_array.rs
```diff
@@ -751,6 +793,70 @@ where D: Dimension
...
+    /// use ndarray::{Array3, Axis};
+    /// let mut a = Array3::<i32>::zeros((0,2,4));
+    /// a.reserve(Axis(0), 1000).unwrap();
+    /// assert!(a.into_raw_vec().capacity() >= 2*4*1000);
+    /// ```
+    ///
+    pub fn reserve(&mut self, axis: Axis, additional: usize) -> Result<(), ShapeError>
+    where D: RemoveAxis
+    {
+        debug_assert!(axis.index() < self.ndim());
```
bluss, line 824: Why was the branch that returned early on additional == 0 removed?

src/impl_owned_array.rs
bluss: thanks

src/impl_owned_array.rs
bluss: No reason to hold up the PR on this question I think :slightly_smiling_face:

ssande7: > Why was the branch that returned early on additional == 0 removed?

I think that was because it was faster (at least on my machine) without it, but it's been a while so could be worth re-testing with it in.

PR #1289, opened 2023-06-02
Add both approx features to the readme

### LINKED ISSUE #789: Documentation clarification for approx feature
I finally got around to updating some of my crates from ndarray v0.11.* to ndarray v0.13. I noticed that ```allclose``` was deprecated in favor of ```abs_diff_eq```. However, I noticed in order to take advantage of it you have to make use of the ```approx``` crate within your own and then use one of their exposed macros. I'm not really a fan of this. I'd rather see the traits within approx_array module exposed to users if they have the ```approx``` feature added and available within the prelude behind a ```[cfg(feature = "approx")]``` flag.  The ```numpy array doc``` also makes it seem like the ```abs_diff_eq``` is available to outside crates but as far as I can tell that's not the case.

nilgoyette:
`approx` and `approx-0_5` are missing from the readme. They were already present in the [doc](https://docs.rs/ndarray/latest/ndarray/#crate-feature-flags).

This being said, I wonder if the readme should simply link to the documentation instead of having the information in 2 places?

Fixes #789

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -83,6 +83,14 @@ your `Cargo.toml`.
   - Enables parallel iterators, parallelized methods and ``par_azip!``.
   - Implies std
 
+- ``approx``
+
+  - Implementations of traits from version 0.4 of the [`approx`] crate.
+
+- ``approx-0_5``
+
+  - Implementations of traits from version 0.5 of the [`approx`] crate.
+
 - ``blas``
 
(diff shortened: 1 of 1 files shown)


adamreichold:
> This being said, I wonder if the readme should simply link to the documentation instead of having the information in 2 places?

Personally, I think the redundancy is fine as it enables people to find this information more easily.

PR #1297, opened 2023-06-17
Updated Windows `base` Computations to be Safer

LazaroHurtado:
Please refer to #1249 for context and discussions, specifically [this comment](https://github.com/rust-ndarray/ndarray/pull/1249#issuecomment-1595677103).

The assertion verifying a window's size dimension or axis' stride is non-zero has been removed and outsourced to `Slice`, which will error when:
1. `end` > `ax_desc.len`, which can only happen when `wsz == 0`
2. `step == 0`, which can only happen when an axis' stride is zero

Thus, we maintain the same functionality as before

### DIFF
--- a/src/iterators/windows.rs
+++ b/src/iterators/windows.rs
@@ -3,6 +3,7 @@ use crate::imp_prelude::*;
 use crate::IntoDimension;
 use crate::Layout;
 use crate::NdProducer;
+use crate::Slice;
 
 /// Window producer and iterable
 ///
@@ -24,16 +25,19 @@ impl<'a, A, D: Dimension> Windows<'a, A, D> {
 
         let mut unit_stride = D::zeros(ndim);
         unit_stride.slice_mut().fill(1);
-        
+
         Windows::new_with_stride(a, window, unit_stride)
     }
 
-    pub(crate) fn new_with_stride<E>(a: ArrayView<'a, A, D>, window_size: E, strides: E) -> Self
+    pub(crate) fn new_with_stride<E>(a: ArrayView<'a, A, D>, window_size: E, axis_strides: E) -> Self
     where
         E: IntoDimension<Dim = D>,
     {
         let window = window_size.into_dimension();
-        let strides_d = strides.into_dimension();
+
+        let strides = axis_strides.into_dimension();
+        let window_strides = a.strides.clone();
+
         ndassert!(
             a.ndim() == window.ndim(),
             concat!(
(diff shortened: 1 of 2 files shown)


nilgoyette:
> Here we'll have panics (debug mode) due to wraparound immediately if the user passes in any array with inverted axis (negative stride)

I don't want to ask too much of a generous contributor, but if you can, could you please add a test that would have failed before the fix in this PR? (a test with negative strides)

LazaroHurtado:
> could you please add a test that would have failed before the fix in this PR? (a test with negative strides)

No worries! But I am not totally sure if this is possible, `windows_with_stride`  requires `stride` to implement `IntoDimension<Dim = D>` which already restricts each axis to be non-negative (`type Ix = usize;`). If I'm overlooking something please let me know

nilgoyette:
If it's impossible to pass an array with negative strides, then I don't understand the point of this PR.

Can't you call `windows` on a transposed matrix?

    m.t().windows(...)

If I understand bluss comment correctly, it should create negative strides.

LazaroHurtado:
Ah sorry there are two strides here, the parameter and the offset, I was thinking about the parameter. I believe there's already a `windows` test that creates a matrix with negative strides by inverting it [here](https://github.com/rust-ndarray/ndarray/blob/17a8d25294d33e0f562af948434dbb284647872b/tests/windows.rs#L271) but I can create a new test for both `windows` and `windows_with_stride` that acts on a transposed matrix as you suggested

bluss:
> If I understand bluss comment correctly, it should create negative strides.

Transpose just changes order of axes, it doesn't modify their lengths or strides. "Axis 0 and 1 become axis 1 and 0".

Use instead ndarray's method [`.invert_axis(Axis(0))`](https://docs.rs/ndarray/latest/ndarray/struct.ArrayBase.html#method.invert_axis), that will reverse the matrix along that axis

Example on matrix with dimensions (n1, n2) and strides (s1, s2):

- (n1, n2) and (s1, s2) -- transpose -> (n2, n1) and (s2, s1)
- (n1, n2) and (s1, s2) -- invert axis 0 -> (n1, n2)  and (-s1, s2)

bluss:
> I don't want to ask too much of a generous contributor, but if you can, could you please add a test that would have failed before the fix in this PR? (a test with negative strides)

I agree with this one.

@LazaroHurtado  We need a negative axis input array. The window strides (your feature) still just needs to be anything that's not just (1, 1, 1), for example something with a stride of 2.

LazaroHurtado:
Thank you for adding some more context! I included a test that asserts the proper strided (the parameter) windows are returned when a single and all axis are inverted. I also verified that this test failed before the work introduced in this PR.

nilgoyette:
Excellent, thank you @LazaroHurtado for the test, and @bluss for the technical details. Now I feel much more confident merging this branch!

I tested with numpy and I get the same results.

Issue #1304, opened 2023-07-12
`NdProducer::Dim` of `axis_windows()` should be `Ix1`


jonasBoss:
Please correct me if I am wrong, but it is my understanding that calling`axis_windows` will return a Producer that traverses along the provided axis, not in any other axis. Therefore the associated type `NdProducer::Dim` should be `Ix1`, and the `NdProducer::Item = ArrayView<'_, _, D>`. Unfortunately they are both `D`.

I have a use-case where I need to zip some n-D data with a 1-D vector to calculate some result:

```rust
let data = array![
    [1,15],
    [2,14],
    [3,13],
    [4,12],
    [5,11],
    [6,10],
];
let x = array![10,20,30,40,50,60];
let mut result: Array<i32,_> = Array::zeros(data.raw_dim());

// everything except the first and last result we need to 
// calculate them seperately, because there is not sufficient data
let mut inner_result = result.slice_axis_mut(Axis(0), Slice::from(1..-1));

Zip::from(inner_result.axis_iter_mut(Axis(0)))
    .and(x.windows(3))
    .and(data.axis_windows(Axis(0), 3))
    .for_each(|y, x, data|{
        // calculate some result y_n with values from data_n-1, data_n and data_n+1
        // as well as x_n-1, x_n and x_n+1
        todo!()
    });
```

<details> 
<summary>Error</summary>

```
error[E0271]: type mismatch resolving `<Windows<'_, {integer}, Dim<[usize; 2]>> as IntoNdProducer>::Dim == Dim<[usize; 1]>`
   --> src/interp1d.rs:679:18
    |
679 |             .and(data.axis_windows(Axis(0), 3))
    |              --- ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected an array with a fixed size of 1 element, found one with 2 elements
    |              |
    |              required by a bound introduced by this call
    |
    = note: expected struct `Dim<[usize; 1]>`
               found struct `Dim<[usize; 2]>`
note: the method call chain might not have had the expected associated types
   --> src/interp1d.rs:679:23
    |
662 |           let data = array![
    |  ____________________-
663 | |             [1,15],
664 | |             [2,14],
665 | |             [3,13],
...   |
668 | |             [6,10],
669 | |         ];
    | |_________- this expression has type `ArrayBase<OwnedRepr<{integer}>, Dim<[usize; 2]>>`
...
679 |               .and(data.axis_windows(Axis(0), 3))
[...]

PR #1305, opened 2023-07-13
Change `NdProducer::Dim` of `axis_windows()` to `Ix1`

jonasBoss:
See #1304 

This breaks the public API because the method `axis_windows()` now returns a new `AxisWindows` struct. And the dimension of the `NdProducer::Dim` will always be `Ix1`.

If the old behavior is needed the method `windows()` can be used.

I was thinking about adding a method `AxisWindows::into_windows(self) -> Windows` but I don't think that this will see much use. Let me know if i should implement it.

### DIFF
--- a/src/iterators/windows.rs
+++ b/src/iterators/windows.rs
@@ -41,41 +41,7 @@ impl<'a, A, D: Dimension> Windows<'a, A, D>
         let strides = axis_strides.into_dimension();
         let window_strides = a.strides.clone();
 
-        ndassert!(
-            a.ndim() == window.ndim(),
-            concat!(
-                "Window dimension {} does not match array dimension {} ",
-                "(with array of shape {:?})"
-            ),
-            window.ndim(),
-            a.ndim(),
-            a.shape()
-        );
-
-        ndassert!(
-            a.ndim() == strides.ndim(),
-            concat!(
-                "Stride dimension {} does not match array dimension {} ",
-                "(with array of shape {:?})"
-            ),
-            strides.ndim(),
-            a.ndim(),
-            a.shape()
-        );
-
-        let mut base = a;
-        base.slice_each_axis_inplace(|ax_desc| {
-            let len = ax_desc.len;
-            let wsz = window[ax_desc.axis.index()];
-            let stride = strides[ax_desc.axis.index()];
-
-            if len < wsz {
-                Slice::new(0, Some(0), 1)
-            } else {
-                Slice::new(0, Some((len - wsz + 1) as isize), stride as isize)
-            }
-        });
-
+        let base = build_base(a, window.clone(), strides);
         Windows {
             base: base.into_raw_view(),
             life: PhantomData,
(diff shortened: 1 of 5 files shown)


src/iterators/windows.rs
```diff
@@ -147,3 +147,119 @@ impl_iterator! {
...
+impl<'a, A, D: Dimension> AxisWindows<'a, A, D> {
+    pub(crate) fn new(a: ArrayView<'a, A, D>, axis: Axis, window_size: usize) -> Self
+    {   
+        let strides = a.strides.clone();
+        let mut base = a;
+        let axis_idx = axis.index();
+        let mut window = base.raw_dim();
+        window[axis_idx] = window_size;
+
+        base.slice_each_axis_inplace(|ax_desc| {
```
adamreichold, line 171: I am not really happy about the code duplication between this method and `Windows::new_with_stride`. Is it reasonably possible to have re-use without making this too hard to follow? (`Windows::new_with_stride` also appears to have additional assertions/consistency checks?)

src/iterators/windows.rs
```diff
@@ -147,3 +147,119 @@ impl_iterator! {
...
+            base,
+            axis_idx,
+            window,
+            strides,
+        }
+    }
+}
+
+
+impl<'a, A, D: Dimension> NdProducer for AxisWindows<'a, A, D> {
```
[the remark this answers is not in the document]
jonasBoss: almost, the methods `raw_dim`, `uget_ptr`, `stride_of` and `split_at` need a different implementation.

src/iterators/windows.rs
```diff
@@ -147,3 +147,119 @@ impl_iterator! {
...
+    }
+
+    fn stride_of(&self, axis: Axis) -> isize {
+        assert_eq!(axis, Axis(0));
+        self.base.stride_of(Axis(self.axis_idx))
+    }
+
+    fn split_at(self, axis: Axis, index: usize) -> (Self, Self) {
+        assert_eq!(axis, Axis(0));
+        let (a, b) = self.base.split_at(Axis(self.axis_idx), index);
```
jonasBoss, line 232: Here I added assertions to check that this is only called with `Axis(0)`. Do we need this? I could just ignore the axis argument, but that would be confusing when a user tries to call this with a different axis.

---
In general I am not super confident that this trait is correctly implemented, as there is a severe lack of documentation for a lot of the methods.

src/iterators/windows.rs
adamreichold: > The struct GeneralWindow and AxisWindow need to be public, but they can be inaccessible from outside of the crate.

This does sound reasonable to me

I would suggest making `Windows` and `AxisWindows` public and have both wrap an `WindowsImpl` or something like that. (Meaning `AxisWindows` contains two fields, `index: usize` and `impl: WindowsImpl`.)

I would also suggest replacing the term "version" by "variant" to avoid any confusion as "version" often has a temporal connotation.

> I don't know if it is more or less clean.

Of course, if this just increases complexity for very little gain in the end, it might not be worth it, you just have to give it a try.

As an alternative, a simple helper function encapsulating the slice-in-place approach using references to the constituent fields of `Windows` and `AxisWindows` might work as well.

jonasBoss: I just updated to my suggested structure, It actually does reduce code duplication a lot.

adamreichold: I am not a fan of exposing these implementation detail generics in the API, hence the suggestion to expose `Windows` and `AxisWindows` with the same generic parameters as before which contain a `WindowsImpl/Inner/Whatever` as a private field. We should not leak this sharing of implementation in the public API.

jonasBoss: I thought that _This does sound reasonable to [you]_ 🙁

We need to expose that there is a difference between the return type of `windows()` and `axis_windows()` there is no way around that.

This generic is not an implementation detail. It makes a real difference to the public API. Specifically it changes the behavior of `<Windows as NdProducer>` and the Type of `<Windows as NdProducer>::Dim`. Otherwise this PR would be pointless.

Now we have this solution which is as minimal as it gets. 
Or the solution I had before. Sure I could try and faff around with some helper methods and structs to reduce code duplication for the cost of clarity and more boilerplate.
Having a `WindowsInner/Impl/...` will break the `impl_ndproducer!` macro, because `base` needs to be a field of `Windows` not of some wrapped type.

adamreichold: > We need to expose that there is a difference between the return type of windows() and axis_windows() there is no way around that.

Distinct return types is not the issue. It is return types which only differ in a type parameter which I have a problem with. I want the relation between these two return types to be fully opaque to the caller, so that e.g. we could also not have code reuse in their implementation without that affecting our API.

> This generic is not an implementation detail. It makes a real difference to the public API. Specifically it changes the behavior of <Windows as NdProducer> and the Type of <Windows as NdProducer>::Dim. Otherwise this PR would be pointless.

`<Windows as NdProducer>::Dim` and `<AxisWindows as NdProducer>::Dim` without the outside world seeing any relation between those two types.

> Having a WindowsInner/Impl/... will break the impl_ndproducer! macro, because base needs to be a field of Windows not of some wrapped type.

This is a real issue and maybe manually implementing `NdProducer` as you did for `AxisWindows` for both case is simple enough, maybe this actually reduces reuse. This is why I also suggest just using an external helper method with completely disjoint struct types.

jonasBoss: Thank you for clearing that up. This makes much more sense.

I just extracted the main logic of creating the base into a separate function, and call that from both constructors `Windwos::new_with_stride()` and `AxisWindows::new()`

adamreichold:
> Is there anything holding this back that I can do do address?
> 
> The CI errors don't seem to be related to this PR.

No. Sorry this is waiting on me finding time to look into it again.

bluss:
Rebased and updated with rustfmt. Maintainer pushed to do that.

src/iterators/windows.rs
bluss: I totally see what you mean. Assertions are a good ides.

bluss:
Thank you @jonasBoss and @adamreichold for the work on this one. It looks good to me, let's delay no further :slightly_smiling_face:, I'll be ready to push the merge button soon.

bluss (approved):

PR #1306, opened 2023-07-14
Draft: add dlpack support

SunDoge:


nilgoyette:
I don't know much about dlpack, but I'm quite sure this won't be merged if
- it's not hidden behind a feature gate
- it doesn't use a real crates.io version (not git)

SunDoge:
This is still a draft. The api of dlpark is still evolving and I'll release v0.3.0 this week. I want to utilize the ownership system to make sure dlpack can only bu consumed once, otherwise there will be double free error. And thanks for your advice, I will add the feature gate.

src/dlpack.rs
```diff
@@ -0,0 +1,94 @@
...
+    pub fn as_ptr(&self) -> *const A {
+        self.managed_tensor.data_ptr() as *const A
+    }
+}
+
+unsafe impl<A> Sync for ManagedRepr<A> where A: Sync {}
+unsafe impl<A> Send for ManagedRepr<A> where A: Send {}
+
+impl<A> FromDLPack for ManagedArray<A, IxDyn> {
+    fn from_dlpack(dlpack: NonNull<dlpark::ffi::DLManagedTensor>) -> Self {
```
bluss, line 69: this function takes a raw pointer (wrapped in NonNull) and it must be an unsafe function, otherwise we can trivially violate memory safety unfortunately.

The only way to remove this requirement would be if you have a "magical" function that can take an arbitrary pointer and say whether it's a valid, live, non-mutably aliased pointer to a tensor.

Here's how to create a dangling bad pointer:  `NonNull::new(1 as *mut u8 as *mut dlpark::ffi::DLManagedTensor)`  does this code crash if we run with this pointer? I think it would..

src/dlpack.rs
```diff
@@ -0,0 +1,94 @@
...
+        let managed_tensor = ManagedTensor::new(dlpack);
+        let shape: Vec<usize> = managed_tensor
+            .shape()
+            .into_iter()
+            .map(|x| *x as _)
+            .collect();
+
+        let strides: Vec<usize> = match (managed_tensor.strides(), managed_tensor.is_contiguous()) {
+            (Some(s), _) => s.into_iter().map(|&x| x as _).collect(),
+            (None, true) => managed_tensor
```
bluss, line 79: Later work, check compatibility of dlpack and ndarray strides, how they work, their domains etc.

src/dlpack.rs
SunDoge: I agree with you. `from_dlpack` should be unsafe, and users should use it at their own risk.

bluss: I would say, we normally don't commit to public dependencies that are not stable (yes, not a great policy since ndarray itself is not so stable.), and dlpark is a public dependency here because it becomes part of our API. It could mean it takes a long time between version bumps.

PR #1310, opened 2023-07-25
Better shape: Deprecate reshape, into_shape

bluss:
into_shape has some slightly questionable default behaviour w.r.t how it handles memory layout. The newer to_shape() fixes those problems! You should use to_shape() if you can.

Into shape doesn't do it /wrong/ but when it has a successful result, it preserves memory layout. C in gives C out. F in gives F out. But if you don't know what memory layout you have going in, into_shape can surprise you.

----

In this PR:

- `into_shape()`  *deprecated*
- `into_shape_with_order()` *new*
- `reshape()`  *deprecated*
- `into_shape_clone()` *new*

It's not a breaking change, but it introduces deprecations that will be quite loud.

Examples:

- `into_shape` unchanged but deprecated
- `.into_shape_with_order((2, 3))`  same as `Order::RowMajor` below
- `.into_shape_with_order(((2, 3), Order::RowMajor))`  reshapes row major arrays, errors on other.
- `.into_shape_with_order(((2, 3), Order::ColMajor))`  reshapes col major arrays, errors on other.

----

Into shape tries to be too great method:

- User doesn't have to specify index ordering, it just guesses
- It doesn't fall back to copy or clone
  - Useful for: Owned arrays with non-clonable elements
  - Useful for views: We can't fall back to cloning when reshaping a view

Into shape has a niche. Especially the no-copy shape transformation of views is very useful.
But the interface is not great about communicating the magic, in some ways.

This PR is a draft because it's not yet decided exactly what the into shape interface should be.
If we could, the order argument should always have been present. But we'd like to be backwards compatible as far as possible as we can, there must be many thousands of into shape calls out there.

### DIFF
--- a/README-quick-start.md
+++ b/README-quick-start.md
@@ -295,8 +295,7 @@ row: [[100, 101, 102],
 ## Shape Manipulation
 
 ### Changing the shape of an array
-The shape of an array can be changed with `into_shape_with_order` method.
-(One can also use `to_shape` for this.)
+The shape of an array can be changed with the `into_shape_with_order` or `to_shape` method.
 
 ````rust
 use ndarray::prelude::*;
--- a/README-quick-start.md
+++ b/README-quick-start.md
@@ -169,7 +169,7 @@ fn main() {
     println!("a shape {:?}", &a.shape());
     println!("b shape {:?}", &b.shape());
     
-    let b = b.into_shape((4,1)).unwrap(); // reshape b to shape [4, 1]
+    let b = b.into_shape_with_order((4,1)).unwrap(); // reshape b to shape [4, 1]
     println!("b shape after reshape {:?}", &b.shape());
     
     println!("{}", a.dot(&b));            // [1, 4] x [4, 1] -> [1, 1] 
@@ -295,7 +295,8 @@ row: [[100, 101, 102],
 ## Shape Manipulation
 
 ### Changing the shape of an array
-The shape of an array can be changed with `into_shape` method.
+The shape of an array can be changed with `into_shape_with_order` method.
+(One can also use `to_shape` for this.)
 
 ````rust
 use ndarray::prelude::*;
@@ -319,7 +320,7 @@ fn main() {
     let b = Array::from_iter(a.iter());
     println!("b = \n{:?}\n", b);
     
-    let c = b.into_shape([6, 2]).unwrap(); // consume b and generate c with new shape
+    let c = b.into_shape_with_order([6, 2]).unwrap(); // consume b and generate c with new shape
     println!("c = \n{:?}", c);
 }
 ````
@@ -459,7 +460,7 @@ use ndarray::{Array, Axis};
 
 fn main() {
 
-    let mut a = Array::range(0., 12., 1.).into_shape([3 ,4]).unwrap();
+    let mut a = Array::range(0., 12., 1.).into_shape_with_order([3 ,4]).unwrap();
     println!("a = \n{}\n", a);
     
     {
@@ -519,7 +520,7 @@ use ndarray::Array;
 
 fn main() {
 
-    let mut a = Array::range(0., 4., 1.).into_shape([2 ,2]).unwrap();
+    let mut a = Array::range(0., 4., 1.).into_shape_with_order([2 ,2]).unwrap();
     let b = a.clone();
     
     println!("a = \n{}\n", a);
--- a/src/impl_methods.rs
+++ b/src/impl_methods.rs
@@ -1922,6 +1922,74 @@ where
         }
     }
 
+    /// Transform the array into `shape`; any shape with the same number of
+    /// elements is accepted. Array elements are reordered in place if
+    /// possible, otherwise they are copied to create a new array.
+    ///
+    /// If an index ordering is not specified, the default is `RowMajor`.
+    /// The operation will only succeed if the array's memory layout is compatible with
+    /// the index ordering, so that the array elements can be rearranged in place.
+    ///
+    /// # `.to_shape` vs `.into_shape_clone`
+    ///
+    /// - `to_shape` supports views and outputting views
+    /// - `to_shape` borrows the original array, `into_shape_clone` consumes the original
+    /// - `into_shape_clone` preserves array type (Array vs ArcArray), but does not support views.
+    ///
+    /// **Errors** if the shapes don't have the same number of elements.<br>
+    pub fn into_shape_clone<E>(self, shape: E) -> Result<ArrayBase<S, E::Dim>, ShapeError>
+    where
+        S: DataOwned,
+        A: Clone,
+        E: ShapeArg,
+    {
+        let (shape, order) = shape.into_shape_and_order();
+        let order = order.unwrap_or(Order::RowMajor);
+        self.into_shape_clone_order(shape, order)
+    }
+
+    pub fn into_shape_clone_order<E>(self, shape: E, order: Order)
+        -> Result<ArrayBase<S, E>, ShapeError>
+    where
+        S: DataOwned,
+        A: Clone,
+        E: Dimension,
+    {
+        let len = self.dim.size();
+        if size_of_shape_checked(&shape) != Ok(len) {
+            return Err(error::incompatible_shapes(&self.dim, &shape));
+        }
+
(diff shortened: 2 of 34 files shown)


src/impl_methods.rs
```diff
@@ -1911,13 +1933,68 @@ where
...
+                }
+                None | Some(Order::ColumnMajor) if self.raw_view().reversed_axes().is_standard_layout() => {
+                    Ok(self.with_strides_dim(shape.fortran_strides(), shape))
+                }
+                _otherwise => Err(error::from_kind(error::ErrorKind::IncompatibleLayout))
+            }
+        }
+    }
+
+    pub fn into_shape_clone<E>(self, shape: E) -> Result<ArrayBase<S, E::Dim>, ShapeError>
```
bluss, line 1948: Into shape with clone fallback has better usability. But it doesn't work for the view usecase

bluss:
to_shape assumes RowMajor when nothing is specified by the caller, while into_shape picks whichever seems to fit with the array that is given (true for into_shape both before this PR and after it in its current shape).

to_shape is great because its only precondition for success (not error) is that the requested shape has the same number of elements. It can handle either ordering request. Simple math for the user :)

into_shape is great because it can reshape views. But it has a precondition that the input array is exactly C-contig or F-contig, unfortunately.

Reshaping by following RowMajor or ColumnMajor index ordering has different effects, it has a different result. This is why it's quite dangerous to do what `into_shape` does, which is to just pick whatever seems to fit with the given array. The distinction might have been invisible to the user.

It's a documentation and API issue, how to, for `into_shape` and reshaping in general

- Continue providing easy to use APIs
- Make the preconditions clear
- Find the right level of "Do What I Mean"

bluss:
If I could, I would deprecate and remove into_shape (because it has the "wrong" ordering defaults.). Then add it back with the right defaults.

But I think that causes too much churn in user code, so I'm looking for the second best option.

nilgoyette:
> If I could, I would deprecate and remove into_shape (because it has the "wrong" ordering defaults.). Then add it back with the right defaults.

I hesitate to give feedback on subjects that I don't feel confident in, but in this particular case... IMO, you should do exactly what you wrote
- having a "sleeping footgun" is **much** worse than deprecating a method.
- you're not at v1 so a breaking change is not unexpected for the users
- "now" is the best time to do something that could/should have been done ealier :)

bluss:
For me, into_shape is a very commonly used method so it feels like it would break every program out there. Hence I'm looking for a smoother transition than this.

Even contemplating doing something like 

- `into_shape()`  *deprecated*
- `into_shape_with_order()` *new*

And after the deprecation period, `into_shape_with_order` is renamed to `into_shape`. Maybe the same dance for `reshape`.

bluss:
@adamreichold if you have time, would be cool if you had a look. I think it's ready to go

adamreichold:
I plan to but I am not sure when this weekend I will be able to look into it. Are you on a schedule here or would merging it the beginning of next still be fine?

bluss:
@adamreichold no stress. It's not on any schedule. I've been away for months and so on, would be weird and rude if I asked that anything would suddenly move at a faster pace(!) so I'm not going to do that :slightly_smiling_face:  As one of the main maintainers I will continue to sometimes just decide to merge some changes on my own, which is a way I want all the maintainers to work, merge if they think it doesn't require more feedback.

README-quick-start.md
```diff
@@ -295,7 +295,8 @@ row: [[100, 101, 102],
 ## Shape Manipulation
 
 ### Changing the shape of an array
-The shape of an array can be changed with `into_shape` method.
+The shape of an array can be changed with `into_shape_with_order` method.
+(One can also use `to_shape` for this.)
```
adamreichold, line 299:
```suggestion
The shape of an array can be changed with the `into_shape_with_order` or `to_shape` method.
```

src/impl_methods.rs
```diff
@@ -1885,22 +1896,108 @@ where
...
+        }
+
+        // Check if contiguous, then we can change shape
+        unsafe {
+            // safe because arrays are contiguous and len is unchanged
+            match order {
+                Order::RowMajor if self.is_standard_layout() => {
+                    Ok(self.with_strides_dim(shape.default_strides(), shape))
+                }
+                Order::ColumnMajor if self.raw_view().reversed_axes().is_standard_layout() => {
```
adamreichold, line 1969: Isn't `raw_view().reversed_axes().is_standard_layout()` quite common. Maybe a short-hand/helper method is appropriate.

src/impl_methods.rs
```diff
@@ -1885,22 +1896,108 @@ where
...
+    ///
     /// **Errors** if the shapes don't have the same number of elements.<br>
     /// **Errors** if the input array is not c- or f-contiguous.
     ///
     /// ```
     /// use ndarray::{aview1, aview2};
     ///
     /// assert!(
-    ///     aview1(&[1., 2., 3., 4.]).into_shape((2, 2)).unwrap()
+    ///     aview1(&[1., 2., 3., 4.]).into_shape_with_order((2, 2)).unwrap()
```
adamreichold, line 1995:
```suggestion
    ///     aview1(&[1., 2., 3., 4.]).into_shape((2, 2)).unwrap()
```

I think the example for `into_shape` should continue to use `into_shape` and rather explicitly allow usage of the deprecated method in the example.

src/impl_methods.rs
```diff
@@ -1885,22 +1896,108 @@ where
...
+                _otherwise => Err(error::from_kind(error::ErrorKind::IncompatibleLayout))
+            }
+        }
+    }
+
     /// Transform the array into `shape`; any shape with the same number of
     /// elements is accepted, but the source array or view must be in standard
     /// or column-major (Fortran) layout.
     ///
+    /// **Note** that `.into_shape_with_order()` "moves" elements differently depending on if the input array
```
adamreichold, line 1981:
```suggestion
    /// **Note** that `.into_shape()` "moves" elements differently depending on if the input array
```

Search & replace'o

src/impl_methods.rs
```diff
@@ -1942,12 +2110,14 @@ where
...
     /// );
     /// ```
+    #[deprecated(note="Obsolete, use `to_shape` or `into_shape_with_order` instead.", since="0.15.2")]
     pub fn reshape<E>(&self, shape: E) -> ArrayBase<S, E::Dim>
     where
         S: DataShared + DataOwned,
         A: Clone,
         E: IntoDimension,
     {
+        //return self.clone().into_shape_clone(shape).unwrap();
```
adamreichold, line 2120:
```suggestion
```

adamreichold (approved):
Some nits, but nothing functional.

I agree with the change, but the diff itself speaks volumes as to the downstream breakage, so I think this should also get a dedicated entry in `RELEASES.md`.

src/impl_methods.rs
bluss: thanks

src/impl_methods.rs
bluss: Yes, good idea, can be separate change

src/impl_methods.rs
bluss: heh, good, yes

bluss:
Every PR gets an entry in RELEASE.md, but normally, I've been writing that at release time. (Would be great to have a merge conflict free way of doing it in the PR, like the CPython project does.). As you say, this requires a bigger notice.

adamreichold:
> (Would be great to have a merge conflict free way of doing it in the PR, like the CPython project does.)

Over at PyO3, we use [Towncrier](https://towncrier.readthedocs.io/en/stable/) to semi-automatically produce a changelog from separate "newsfragement" Markdown files. The affinity to Python tooling in general made the choice a bit easier, but I suspect there exist similar tools written in Rust if required.

It does entail quite a bit of setup work up front, but once done writing the changelog entries because a breeze (and we even check their presence via the CI via a label to suppress that if the PR does not require an entry).

bluss:
Thank you. I have learned a bit (doubtful, but maybe) from other maintainers and trying to be more relaxed about the maintainership style. This is good enough, let's go.

With that said, I'll still try my best to say no to code that breaks Rust's rules (relevant for safety), which we can often spot in code reviews, even if it takes some effort..

PR #1318, opened 2023-08-04
Fix comparison of slicing with negative step in `ndarray` with that i…

venkat0791:
…n NumPy

The existing text implies that `ndarray` and NumPy behave the same way when `step = -1`. However this is not the case. The behavior of NumPy is illustrated below:

In [1]: import numpy as np
In [2]: x = np.arange(10)
In [3]: print(x[1:5:-1])
[]
In [4]: print(x[5:1:-1])
[5 4 3 2]

The analogous results in `ndarray` would be `array![4, 3, 2, 1]` and `array![]`, respectively.

### DIFF
--- a/src/doc/ndarray_for_numpy_users/mod.rs
+++ b/src/doc/ndarray_for_numpy_users/mod.rs
@@ -111,8 +111,8 @@
 //! When slicing in `ndarray`, the axis is first sliced with `start..end`. Then if
 //! `step` is positive, the first index is the front of the slice; if `step` is
 //! negative, the first index is the back of the slice. This means that the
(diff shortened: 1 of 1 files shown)


venkat0791:
Sorry, I realized there are further differences in slicing with respect to NumPy (having to do with the defaults for `start` and `end` when `step` is negative). I'd like to clarify further.

bluss:
Don't need to close the PR, update your branch, force push and update the PR, update PR description. (For next time.)

venkat0791:
Thanks for the tip, @bluss. I'm still learning the ropes here (this was my first ever contribution of any kind to a GitHub project). Thanks also for all the great work on this crate!

PR #1324, opened 2023-09-09
Add arg?{max,min}

erip:
This is a very rough implementation of max and min with associated friends. The `Copy` bound may be unnecessarily expensive. I'm also not positive whether the arg* variants have a sufficiently clean API since it may be a bit unexpected to return both the position and values, so I'd be happy to change this if desired.

I'd love to get the `_axis` variants added, but some of the ergonomics are a bit clumsy since naively mapping along the axis yields an `ArrayBase<Option<...>>` which is a little ugly to ["traverse"](https://twitter.com/volpegabriel87/status/1296836812275621888?s=20). I'll leave this as a draft to await some initial reviews since this is my first rust PR and I'm surely doing something wrong. 😄 

I'd also be remiss to not note that this doesn't work for floats because they don't implement `Ord`. 😢 We could create our own or pull in some common impls.


nilgoyette:
Thank you for your contribution @erip.

However, I doubt that this will be merged  even if it was feature-complete because an implementation already exists in the [ndarray-stats](https://docs.rs/ndarray-stats/latest/ndarray_stats/trait.QuantileExt.html) sister project. See `min`, `min_skipnan`, `argmin`, etc. I strongly suggest you to use them.

Issue #1323, opened 2023-08-17
Implement numpy's `triu` and `tril` methods


geetmankar:
Hi, an implementation of `numpy.triu` and `numpy.tril` methods ([link here](https://numpy.org/doc/stable/reference/generated/numpy.triu.html)) for arrays that are atleast 2 dimensions.

```python3
import numpy as np

np.triu([[1,2,3],[4,5,6],[7,8,9],[10,11,12]], -1)
```
returns:
```shell
array([[ 1,  2,  3],
       [ 4,  5,  6],
       [ 0,  8,  9],
       [ 0,  0, 12]])
```

I implemented a basic version of it for 2D square owned ndarrays (please forgive if the code is wasteful, I implemented it quickly since just needed it for a physics project)

```rust
pub trait Tri {
    fn triu(&self, k: usize) -> Self;
    fn tril(&self, k: usize) -> Self;
}

impl Tri for ArrayBase<OwnedRepr<f64>, Ix2> {
    fn triu(&self, k: usize) -> Self {
        let cols = self.len_of(Axis(1));
        let copied_arr = self.clone();

        let mut result_arr = Array2::<f64>::zeros((cols, cols).f());

        for (i, row) in copied_arr.axis_iter(Axis(0)).enumerate() {
            let slice_main = (i as usize + k)..self.len_of(Axis(0));
            row.to_owned()
                .slice(s![slice_main.clone()])
                .assign_to(result_arr.slice_mut(s![i, slice_main.clone()]));
        }

        return result_arr;
    }

    fn tril(&self, k: usize) -> Self {
        return self.clone().triu(k).t().to_owned();
    }
}
```

nilgoyette:
ndarray should probably have those functions. 

Making it generic, and with less `to_owned` and `clone` will require some work. Thank you for writing the proposition.

geetmankar:
Sweet, I'll be waiting for the upgrade.

nilgoyette:
> I'll be waiting for the upgrade.

Just to be clear, we're mostly offering bug fixes and support at the moment because all maintainers are busy. If you are actually waiting for it, I advise you to be patient :)

geetmankar:
No worries, the code snippet I wrote is currently good enough for my pet project, so I won't need to hold my breath. Thanks for the heads-up though!

akern40:
[...]

PR #1331, opened 2023-10-01
Add matlab-like tril method for two-dimensional arrays

fedemagnani:
Adding `trill` method for two-dimensional arrays in order to create a lower triangular matrix from a given matrix. This function reproduces the behavior of the [Matlab function with same name](https://www.mathworks.com/help/matlab/ref/tril.html): it returns the elements on and below the kth diagonal of A

### DIFF
--- a/src/impl_2d.rs
+++ b/src/impl_2d.rs
@@ -141,4 +141,78 @@ where
         let (m, n) = self.dim();
         m == n
     }
+
+    /// Return the elements on and below the kth diagonal of A. 
+    /// If k=-1, it returns the lower triangular part of A excluding the main diagonal. 
+    /// If k=0 it returns the lower triangular part of A including the main diagonal. 
+    /// If k=1 it returns the lower triangular part of A including the main diagonal and the diagonal above it, and so on
+    /// 
+    /// # Examples
+    /// ```
+    /// 
+    /// use ndarray::{array, Array2};
+    /// let A = Array2::<f64>::ones((3, 3));
+
+    /// let L_1 = A.tril(-1);
+    /// let L_2 = A.tril(0);
+    /// let L_3 = A.tril(1);
+
+    /// assert_eq!(
+    ///     L_1,
+    ///     array![
+    ///         [0.0, 0.0, 0.0], 
+    ///         [1.0, 0.0, 0.0], 
+    ///         [1.0, 1.0, 0.0]
+    ///     ]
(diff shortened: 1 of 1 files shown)


nilgoyette:
I'm not against adding this feature to ndarray and I **think** the other maintainers would agree. This being said
- I'm not a fan of the name. `tril` seems to be the standard (Matlab and Numpy), but it means nothing (maybe it's just me?)
- I didn't take the time to think about it, but the implementation looks like it could be "better".  For example, numpy does it in 3 lines (I know we can't and shouldn't do it like Numpy)

adamreichold:
> I'm not a fan of the name. tril seems to be the standard (Matlab and Numpy), but it means nothing (maybe it's just me?)

I guess we could just spell it out, e.g. `triangular_lower`?

fedemagnani:
> I'm not against adding this feature to ndarray and I **think** the other maintainers would agree. This being said
> - I'm not a fan of the name. `tril` seems to be the standard (Matlab and Numpy), but it means nothing (maybe it's just me?)
> - I didn't take the time to think about it, but the implementation looks like it could be "better".  For example, numpy does it in 3 lines (I know we can't and shouldn't do it like Numpy)

I wrote this method trying to reproduce the behaviour of a matlab script which was invoking this function. 

Like you I found the name a bit confusing at first glance, but since it generates a lower triangular from a given matrix I thought that "tri" was standing for "triangular" and "l" as "lower". I decided to keep the same name for consistency with the matlab function thinking that it could have been useful for people with my similiar use case.

Regarding the length of the code, this is the way I wrote it and I think that adding a method which expands the functionalities of the package is better than not doing it :) 

I don't think that this method will be used that much but if it will surge the need of optimizing it I bet that it is going to be optimized in the future

nilgoyette:
> I don't think that this method will be used that much but if it will surge the need of optimizing it I bet that it is going to be optimized in the future

There's no need to procrastinate, we can do it properly right now. If you're still interested, of course.
- Direct indexing is "slow" in ndarray and that's what you use
- You could test using a `Zip::indexed(m.rows_mut())`, then something like `row.slice(s![?..]).fill(0.0)`. If I'm right, this could be done in 4-5 lines.
- Once this is done, you might as well code `triu`. It doesn't make much sense to have one without the other!

akern40:
This has been resolved by #1323

PR #1335, opened 2023-11-24
Add optional support for borsh serialisation

Fuuzetsu:
Behind a feature flag.


xtest-serialization/tests/serialize.rs
```diff
@@ -218,3 +220,83 @@ fn serial_many_dim_ron() {
         assert_eq!(a, a_de);
     }
 }
+
+#[test]
+fn serial_ixdyn_borsh() {
+    {
+        let a = arr0::<f32>(2.72).into_dyn();
+        let serial = borsh::to_vec(&a).unwrap();
+        println!("Serde encode {:?} => {:?}", a, serial);
```
nilgoyette, line 229: borsh, here and below

xtest-serialization/tests/serialize.rs
```diff
@@ -218,3 +220,83 @@ fn serial_many_dim_ron() {
...
+    use borsh::to_vec as borsh_serialize;
+
+    {
+        let a = arr0::<f32>(2.72);
+
+        let a_s = borsh_serialize(&a).unwrap();
+
+        let a_de: ArcArray<f32, _> = borsh_deserialize(&a_s).unwrap();
+
+        assert_eq!(a, a_de);
```
nilgoyette, line 268: I know that you respected the style of the previous block (serde), but please don't use double interline. Those 4 lines can use 4 lines.

nilgoyette (commented):
I don't know the opinion of the other maintainers, but I think we don't have much to lose since it's behind a feature gate.

xtest-serialization/tests/serialize.rs
Fuuzetsu: Right, this is just copy+paste artifact.... I'll try to fix this on Monday.

Fuuzetsu:
> I don't know the opinion of the other maintainers, but I think we don't have much to lose since it's behind a feature gate.

Well, there _is_ additional maintenance burden as presumably people that rely on this will expect it to work going forward.

If there's something I can do to try to proactively alleviate the concerns, let me know.

An alternative might be to implement this directly in `borsh` and push the burden in there but it gets much more convoluted this way due to versions and such. It might end up this way for chrono at least though: https://github.com/chronotope/chrono/pull/1366

Fuuzetsu:
I addressed the couple of comments.

nilgoyette:
> Well, there is additional maintenance burden as presumably people that rely on this will expect it to work going forward.

Hum, right. I'm not the usual maintainer here and I hadn't thought about that :) The problem of ndarray in the last 2-3 years is the lack of maintainer. Because of this, I'm in no position to accept this MR. I'm not saying this will never get merged, but probably not by me.

src/array_borsh.rs
```diff
@@ -0,0 +1,101 @@
...
+    }
+}
+
+/// **Requires crate feature `"borsh"`**
+impl BorshDeserialize for IxDyn {
+    fn deserialize_reader<R: borsh::io::Read>(reader: &mut R) -> borsh::io::Result<Self> {
+        // Deserialize the length.
+        let len = <usize as BorshDeserialize>::deserialize_reader(reader)?;
+        // Deserialize the given number of elements. We assume the source is
+        // trusted so we use a capacity hint...
```
bluss, line 47: Why trusted?

src/array_borsh.rs
```diff
@@ -0,0 +1,101 @@
...
+    A: BorshSerialize,
+    D: Dimension + BorshSerialize,
+    S: Data<Elem = A>,
+{
+    fn serialize<W: borsh::io::Write>(&self, writer: &mut W) -> borsh::io::Result<()> {
+        // Dimensions
+        <D as BorshSerialize>::serialize(&self.raw_dim(), writer)?;
+        // Followed by length of data
+        let iter = self.iter();
+        <usize as BorshSerialize>::serialize(&iter.len(), writer)?;
```
bluss, line 68: Why both length and dimensions?

src/array_borsh.rs
```diff
@@ -0,0 +1,101 @@
...
+    }
+}
+
+/// **Requires crate feature `"borsh"`**
+impl<I> BorshDeserialize for Dim<I>
+where
+    I: BorshDeserialize,
+{
+    fn deserialize_reader<R: borsh::io::Read>(reader: &mut R) -> borsh::io::Result<Self> {
+        <I as BorshDeserialize>::deserialize_reader(reader).map(Dim::new)
```
bluss, line 23: I would be worried that we have to break compatibility here if we change how dimensions are stored?

src/array_borsh.rs
```diff
@@ -0,0 +1,101 @@
...
+where
+    A: BorshDeserialize,
+    D: BorshDeserialize + Dimension,
+    S: DataOwned<Elem = A>,
+{
+    fn deserialize_reader<R: borsh::io::Read>(reader: &mut R) -> borsh::io::Result<Self> {
+        // Dimensions
+        let dim = <D as BorshDeserialize>::deserialize_reader(reader)?;
+        // Followed by length of data
+        let len = <usize as BorshDeserialize>::deserialize_reader(reader)?;
```
bluss, line 88: How do you handle integer size here? What if it was serialized on a 64-bit usize platform but you deserialize on a 32-bit?

bluss: I would look at https://github.com/indexmap-rs/indexmap/pull/313 for inspiration - fixed size integers and check error cases

Fuuzetsu: borsh always serialises usize as u64: https://docs.rs/borsh/latest/src/borsh/ser/mod.rs.html#128-132

bluss: borsh says that container sizes should be stored as u32. Unsure how it interacts with ndarray. For me, developing this PR is not my priority unfortunately. Help from others welcome.

PR #1337, opened 2023-12-02
Fix new rustc lints to make the CI pass.

adamreichold:


nilgoyette:
I was about to merge this to be able to merge the other MR, but maybe you wanted to add something else, so I'll wait.

Thank you for the fix!

PR #1344, opened 2023-12-22
Update README-quick-start.md

joelchen:
Fix missing words in example.


adamreichold:
The MSRV breakage is unrelated, hence I will merge this ignoring the CI failure and try to fix that in #1345

Issue #1204, opened 2022-09-04
ndarray::iter::Iter doesn't implement Debug


liam-clink:
The title says it. What I was trying to do when I discovered this was print a zip of a range and an ndarray.

Like so:
```
let foo = ndarray::Array1::<f64>::zeros(100);
println!("{:?}", std::iter::zip(0..foo.len(), &foo);
```

nilgoyette:
I never thought of printing an iterator before! Anyway, as per the official documentation:

> [fmt::Debug](https://doc.rust-lang.org/std/fmt/trait.Debug.html) implementations should be implemented for all public types.

So, yeah, I guess we should do it.

biskwikman:
I'm not very experienced with Rust projects, but I'd like to try to implement this. Is Iter found in the iterators/mod.rs file?

nilgoyette:
Yes, probably. Just clone the project and test, you will see.

bluss:
Fixed in #1353, thanks and thanks @nilgoyette for mentoring

PR #1353, opened 2024-01-29
derived Debug for Iter and IterMut

biskwikman:
My attempt to implement Debug for Iter, I also did so for IterMut. 
#1204 

Example:

input:
```rust
fn main() {
    let foo = ndarray::Array1::<f64>::zeros(10);
    println!("{:?}", std::iter::zip(0..foo.len(), &foo));
}
```

output:
```
Zip { a: 0..10, b: Iter { inner: Slice(Iter([0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0])) } }
```

### DIFF
--- a/src/iterators/mod.rs
+++ b/src/iterators/mod.rs
@@ -14,11 +14,11 @@ pub mod iter;
 mod lanes;
 mod windows;
 
+#[cfg(not(feature = "std"))]
+use alloc::vec::Vec;
 use std::iter::FromIterator;
 use std::marker::PhantomData;
 use std::ptr;
-#[cfg(not(feature = "std"))]
-use alloc::vec::Vec;
 
 use crate::Ix1;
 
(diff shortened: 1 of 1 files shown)


nilgoyette:
Update the description with one output example. Also, you left several lines of commented code.

nilgoyette:
I see 2 problems:

- Please also add the output of a 2D arrays so we know how it will be printed
- `Iter { inner: Slice(Iter([...])) } }` is uselessly descriptive. The user doesn't want to know the technical architecture of ndarray, he simply wants to print the values in the iterator. I think you will be forced to code the function instead of using `#[derive(Debug)]` in order to control what you print.

bluss:
Rebased. It can be prettier and more logical but it's just a Debug fmt, so it can change later. Thanks, this is a good start for this.

PR #1360, opened 2024-02-16
ndarray_for_numpy_users some example to code not pointed out to clippy

higumachan:
I have made a small change to some examples of ndarray_for_numpy_users.

The reason is that if I write the code according to this example, it will be pointed out to clippy in the default configuration.


nilgoyette:
Thank you!

PR #1362, opened 2024-02-26
Fix contig check for single element arrays

### LINKED ISSUE #1359: `ArrayBase::as_slice_memory_order` returns `None` when slice has a single element
For some reason, `as_slice_memory_order` returns `None` when called on a 1-element slice. Minimal example:

```rust
use ndarray::{s, Array1};
let arr = Array1::from_vec(vec![1.0, 2.0, 3.0]);
println!("{:?}", arr.slice(s![0..1]).as_slice_memory_order());
// Prints None
```

bluss:
When an array has 0 or 1 elements, strides don't matter anymore. The general case of this function handled this correctly, but the special case for ndim == 1 did not.

Fixes #1359

### DIFF
--- a/src/dimension/dimension_trait.rs
+++ b/src/dimension/dimension_trait.rs
@@ -9,6 +9,7 @@
 use std::fmt::Debug;
 use std::ops::{Add, AddAssign, Mul, MulAssign, Sub, SubAssign};
 use std::ops::{Index, IndexMut};
+#[cfg(not(feature = "std"))]
 use alloc::vec::Vec;
 
 use super::axes_of;
--- a/src/dimension/dimension_trait.rs
+++ b/src/dimension/dimension_trait.rs
@@ -285,21 +285,25 @@ pub trait Dimension:
             return true;
         }
         if dim.ndim() == 1 {
-            return strides[0] as isize == -1;
-        }
-        let order = strides._fastest_varying_stride_order();
-        let strides = strides.slice();
+            // fast case for ndim == 1:
+            // Either we have length <= 1, then stride is arbitrary,
+            // or we have stride == 1 or stride == -1, but +1 case is already handled above.
+            dim[0] <= 1 || strides[0] as isize == -1
+        } else {
+            let order = strides._fastest_varying_stride_order();
+            let strides = strides.slice();
 
-        let dim_slice = dim.slice();
-        let mut cstride = 1;
-        for &i in order.slice() {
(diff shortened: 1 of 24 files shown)


src/split_at.rs
```diff
@@ -7,6 +7,7 @@ pub(crate) trait SplitAt  {
 }
 
 pub(crate) trait SplitPreference : SplitAt {
+    #[allow(dead_code)]
```
adamreichold, line 10:
```suggestion
    #[allow(dead_code)] // used only when Rayon support is enabled
```

src/dimension/dimension_trait.rs
```diff
@@ -284,21 +284,25 @@ pub trait Dimension:
             return true;
         }
         if dim.ndim() == 1 {
-            return strides[0] as isize == -1;
```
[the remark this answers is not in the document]
bluss: I was wondering why we even have a special case anyway when the general case seems to work, but, the motivation should have been this:

This is the Dimension trait, used everywhere, and every implementation has a well defined dimensionality. We want the code size to be small for the most common dimension (ndim == 1), so we want the code for the else branch (in the new code) to be removed. Writing the code this way just makes the code correspond better to that thought, even if it probably optimizes the same anyway.

tests/array.rs
```diff
@@ -1957,6 +1957,22 @@ fn test_contiguous() {
...
+
+    let arr1 = array![1, 2, 3];
+    assert_matches!(arr1.slice(s![0..1]).as_slice_memory_order(), Some(&[1]));
+    assert_matches!(arr1.slice(s![1..2]).as_slice_memory_order(), Some(&[2]));
+    assert_matches!(arr1.slice(s![2..3]).as_slice_memory_order(), Some(&[3]));
+    assert_matches!(arr1.slice(s![0..0]).as_slice_memory_order(), Some(&[]));
+
+    let arr2 = array![[1, 2, 3], [4, 5, 6]];
+    assert_matches!(arr2.slice(s![.., 2..3]).as_slice_memory_order(), None);
+    assert_matches!(arr2.slice(s![1, 2..3]).as_slice_memory_order(), Some(&[6]));
```
bluss, line 1973: The 2D case was working correctly before this fix too.

adamreichold (approved):

bluss:
thanks for checking. The lint fixes are helpful here too, best to get them in.

PR #1366, opened 2024-03-09
Update ci - run cross tests only on master

bluss:
Cross actions are very slow. They give us coverage on 32-bit and big endian, but are rather less important. Also update the checkout action because the old one has deprecation warnings.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,7 +24,7 @@ jobs:
           - 1.51.0  # MSRV
 
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - uses: dtolnay/rust-toolchain@master
         with:
           toolchain: ${{ matrix.rust }}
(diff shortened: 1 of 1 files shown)


adamreichold:
I think, ideally we'd use merge queues to run the cross asynchronously but before merging into the main branch. The PR workflow it self could then be simplify down even further, e.g. just tests on stable and Clippy.

bluss:
sounds, good, so like bors or whatever the new version would be

adamreichold:
> sounds good, so like bors or whatever the new version would be

Yes, but integrated into GitHub. You'll need to enable it in the project settings and then:
* We don't run CI on the main branch at all any more because all commits are tested before being put there.
* Which tests are run is not differentiated by branch name but by event, either pull requests or merge queues.
* PR run should small and fast for quick feedback on proposed changes.
* Pre-merge checks should be complete so that the main branch is always is a known good state.

PR #1368, opened 2024-03-09
Setup ci so that most checks run in merge queue only

bluss:
Adopt conclusion job that summarizes ci status.
Originally contributed by @messense in pyo3, copied their solution with thanks.


bluss:
The merge queue now working on the right things I think: https://github.com/rust-ndarray/ndarray/actions/runs/8215462578

I listed the required checks in branch protection. I could list tests, cross_tests as those entries (not each test matrix entry, that resolves that problem for me).

And mark those as being skipped for regular PR check (see the yaml file). Too early for bors-ng to be deprecated seeing how this works(!) :slightly_smiling_face:

adamreichold:
> I listed the required checks in branch protection. I could list tests, cross_tests as those entries (not each test matrix entry, that resolves that problem for me).

IIRC, in PyO3 we solved this by having a single job collect the state of all the other jobs (which I think is also the usual approach applied with bors).

bluss:
Oh, it doesn't work. It autocompleted to tests and cross_tests for the required tests in the UI, but in CI it expands to the matrix of tests so the names are **tests (stable)** and so on. Everything's finished but it is still waiting.

PR #1371, opened 2024-03-09
Test using cargo-careful

bluss:
Suggested by @adamreichold.

I think we could test with miri too, the problem is that it's very slow.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -70,11 +70,27 @@ jobs:
         run: cargo install cross
       - run: ./scripts/cross-tests.sh "docs" ${{ matrix.rust }} ${{ matrix.target }}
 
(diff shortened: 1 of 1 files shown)


bluss:
it ran correctly here: https://github.com/rust-ndarray/ndarray/actions/runs/8216936059/job/22472060567

adamreichold:
> I think we could test with miri too, the problem is that it's very slow.

My understanding was that Miri would not handling calling into BLAS et al.

adamreichold (approved):

bluss:
I see. It's been attempted before, such as by jturner in commit 784a9a842b42313eb3ce71b793bf80c777a1fef0 apparently. He added a miri runner in matrixmultiply at least, but we don't have it here.

PR #1379, opened 2024-04-01
Add .into_raw_vec_with_offset() and deprecate .into_raw_vec()

bluss:
Without the offset from start of allocation to the logically first element of the array,
correctly reinterpreting the results of .into_raw_vec() as an n-D array is tricky.

Provide a smoother transition by deprecating the old method and
introducing a new one.

Return Option<usize>, this makes the empty vector case stand out
as a separate case that needs to be handled.

Closes #994 by superseding it.

### DIFF
--- a/src/dimension/mod.rs
+++ b/src/dimension/mod.rs
@@ -410,7 +410,7 @@ fn to_abs_slice(axis_len: usize, slice: Slice) -> (usize, usize, isize)
 }
 
 /// This function computes the offset from the lowest address element to the
-/// logically first element. The result is always >= 0.
+/// logically first element.
 pub fn offset_from_low_addr_ptr_to_logical_ptr<D: Dimension>(dim: &D, strides: &D) -> usize
 {
     let offset = izip!(dim.slice(), strides.slice()).fold(0, |_offset, (&d, &s)| {
--- a/src/dimension/mod.rs
+++ b/src/dimension/mod.rs
@@ -409,8 +409,8 @@ fn to_abs_slice(axis_len: usize, slice: Slice) -> (usize, usize, isize)
     (start, end, step)
 }
 
-/// Returns the offset from the lowest-address element to the logically first
-/// element.
+/// This function computes the offset from the lowest address element to the
+/// logically first element. The result is always >= 0.
 pub fn offset_from_low_addr_ptr_to_logical_ptr<D: Dimension>(dim: &D, strides: &D) -> usize
 {
     let offset = izip!(dim.slice(), strides.slice()).fold(0, |_offset, (&d, &s)| {
--- a/src/impl_owned_array.rs
+++ b/src/impl_owned_array.rs
@@ -77,7 +77,7 @@ where D: Dimension
 
     /// Return a vector of the elements in the array, in the way they are
     /// stored internally, and the index in the vector corresponding to the
-    /// logically first element of the array (or `None` if the array is empty).
+    /// logically first element of the array (or 0 if the array is empty).
     ///
     /// If the array is in standard memory layout, the logical element order
     /// of the array (`.iter()` order) and of the returned vector will be the same.
@@ -92,7 +92,7 @@ where D: Dimension
     ///
     /// let shape = arr.shape().to_owned();
     /// let strides = arr.strides().to_owned();
-    /// let (v, offset) = arr.into_raw_vec();
+    /// let (v, offset) = arr.into_raw_vec_and_offset();
     ///
     /// assert_eq!(v, &[1., 2., 3., 4., 5., 6.]);
     /// assert_eq!(offset, Some(2));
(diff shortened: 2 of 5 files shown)


src/dimension/mod.rs
```diff
@@ -409,8 +409,8 @@ fn to_abs_slice(axis_len: usize, slice: Slice) -> (usize, usize, isize)
     (start, end, step)
 }
 
-/// Returns the offset from the lowest-address element to the logically first
-/// element.
+/// This function computes the offset from the lowest address element to the
+/// logically first element. The result is always >= 0.
```
adamreichold, line 413: I am not sure what to make of the comment with the return type being `usize`?

src/impl_owned_array.rs
```diff
@@ -59,14 +59,101 @@ impl<A> Array<A, Ix0>
 impl<A, D> Array<A, D>
 where D: Dimension
 {
+    /// Returns the offset (in units of `A`) from the start of the allocation
+    /// to the first element, or `None` if the array is empty.
+    fn offset_from_alloc_to_logical_ptr(&self) -> Option<usize>
+    {
+        if self.is_empty() {
+            return None;
```
adamreichold, line 67: Does the differentiating the `None` case really buy anything if it is defaulted to zero at all call sites in any case?

tests/array.rs
```diff
@@ -1173,7 +1176,7 @@ fn array_view0_into_scalar()
     // With this kind of setup, the `Array`'s pointer is not the same as the
     // underlying `Vec`'s pointer.
     let a: Array0<i32> = array![4, 5, 6, 7].index_axis_move(Axis(0), 2);
-    assert_ne!(a.as_ptr(), a.into_raw_vec().as_ptr());
+    assert_ne!(a.as_ptr(), a.into_raw_vec_and_offset().0.as_ptr());
```
adamreichold, line 1179: I think it be nice to check the offset here and below as well, even if just for being zero as implied by the comment.

src/impl_owned_array.rs
bluss: I've been puzzling over that, not sure

src/impl_owned_array.rs
```diff
@@ -59,14 +59,101 @@ impl<A> Array<A, Ix0>
...
+    /// use ndarray::{array, Array2, Axis};
+    ///
+    /// let mut arr: Array2<()> = array![[(), ()], [(), ()], [(), ()]];
+    /// arr.slice_axis_inplace(Axis(0), (1..).into());
+    ///
+    /// let shape = arr.shape().to_owned();
+    /// let strides = arr.strides().to_owned();
+    /// let (v, offset) = arr.into_raw_vec_and_offset();
+    ///
+    /// assert_eq!(v, &[(), (), (), (), (), ()]);
```
adamreichold, line 129: Why not assert on `offset` here as well? This is the only test case for the zero-size case for the new code, isn't it?

bluss: I will maybe look into it.

I'll say that in my personal taste, we should not use examples to do tests, they should be in the test directory. (Examples are tested to make sure the examples do what they say, not that the library is correct.)

adamreichold: > I'll say that in my personal taste, we should not use examples to do tests, they should be in the test directory - so I'm not looking to expand the testing done inside the examples. (Examples are tested to make sure the examples do what they say, not that the library is correct.)

This is a reasonable approach, but then doesn't this miss a test case for the zero-sized case?

src/impl_owned_array.rs
adamreichold: > `(Vec<A>, Option<NonZeroUsize>)`

`NonZeroUsize` is nonsense, I was thinking of the length and not the index. But I still think either `Option<usize>` or `*mut A` would be preferable.

bluss: Ok, why is option good? I think keeping this as option here is ok

adamreichold: Because it avoids returning a value (zero) which cannot actually be used to index into the vector (which is empty) even tough it can be used to offset the pointer. Handling this case by a bit of additional wording in the doc comment is failure prone.

`Option<usize>` means calling code needs to acknowledge and handle this case (yielding `None` instead of an invalid index zero in the empty case). `*mut A` at least avoids this by always be useful as a valid pointer even if dangling.

bluss: This is for into_raw_vec_with_offset right (the public method) and not for offset_from_alloc_to_logical_ptr? just checking

adamreichold: Yes, as said in the beginning. I think keeping the `Option` here makes sense if it is also exposed at the public API. And it should be exposed at the public API, just replacing it by zero for the empty case is a foot gun because the invariants get unnecessarily complicated.

bluss: I've changed it around to use `Option<usize>` again, it seems best.

adamreichold (approved):

bluss:
thanks for looking at this

PR #1063, opened 2021-09-02
Bumped approx's version number

obsgolem:
Closes #1062

PR #1380, opened 2024-04-06
Update to Approx 0.5

bluss:
Old approx-0_5 feature is removed, the approx feature points to approx 0.5

Uses and supersedes #1063. Closes #1063

### DIFF
--- a/src/array_approx.rs
+++ b/src/array_approx.rs
@@ -193,6 +193,3 @@ macro_rules! impl_approx_traits {
 
 #[cfg(feature = "approx")]
 impl_approx_traits!(approx, "**Requires crate feature `\"approx\"`.**");
-
-#[cfg(feature = "approx-0_5")]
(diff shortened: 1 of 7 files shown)


src/array_approx.rs
```diff
@@ -193,6 +193,3 @@ macro_rules! impl_approx_traits {
 
 #[cfg(feature = "approx")]
 impl_approx_traits!(approx, "**Requires crate feature `\"approx\"`.**");
```
adamreichold, line 195: I guess you want to keep the macro around for the time when approx 0.6 is relevant?

From a code maintenance perspective, getting the code out of the macro would be preferable IMHO.

bluss: I'm just adapting necessary changes with as low overhead as possible, unfortunately. I'm going for less than perfect for this release :slightly_smiling_face: 

Keeping the macro doesn't sound so useful no, approx doesn't look like it's going to have a release soon. If you want to fix this, it would be welcome too.

PR #1384, opened 2024-04-28
Make compatible with thumbv6m-none-eabi

BjornTheProgrammer:
### Changes
Add the crates `portable-atomic` and `portable-atomic-util` to make the Arc used in the code compatible with the `thumbv6m-none-eabi` target. Also note that this change will probably make `ndarray` work on more targets than just the `thumbv6m-none-eabi` target.

### Considerations
Those two crates are imported from a GitHub revision. It might be best to wait for an official release to crates.io, but I'm not sure if that's important.

This has also been tested and confirmed to work on a Raspberry Pi Pico.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,6 +44,7 @@ matrixmultiply = { version = "0.3.2", default-features = false, features=["cgemm
 serde = { version = "1.0", optional = true, default-features = false, features = ["alloc"] }
 rawpointer = { version = "0.2" }
 
+
 [dev-dependencies]
 defmac = "0.2"
 quickcheck = { version = "1.0", default-features = false }
@@ -70,7 +71,13 @@ docs = ["approx", "serde", "rayon"]
 std = ["num-traits/std", "matrixmultiply/std"]
 rayon = ["dep:rayon", "std"]
 
+portable-atomic-critical-section = ["portable-atomic/critical-section"]
+
 matrixmultiply-threading = ["matrixmultiply/threading"]
+
+[target.'cfg(not(target_has_atomic = "ptr"))'.dependencies]
+portable-atomic = { version = "1.6.0" }
+portable-atomic-util = { version = "0.2.0", features = [ "alloc" ] }
 
 [profile.bench]
 debug = true
--- a/README.rst
+++ b/README.rst
@@ -97,6 +97,10 @@ your `Cargo.toml`.
 
   - Enable the ``threading`` feature in the matrixmultiply package
 
+- ``portable-atomic-critical-section``
+
+  - Whether ``portable-atomic`` should use ``critical-section``
+
 How to use with cargo
 ---------------------
 
--- a/src/data_traits.rs
+++ b/src/data_traits.rs
@@ -11,7 +11,12 @@
 #[allow(unused_imports)]
 use rawpointer::PointerExt;
 
+#[cfg(target_has_atomic = "ptr")]
 use alloc::sync::Arc;
+
+#[cfg(not(target_has_atomic = "ptr"))]
+use portable_atomic_util::Arc;
+
 #[cfg(not(feature = "std"))]
 use alloc::vec::Vec;
 use std::mem::MaybeUninit;
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -42,6 +42,27 @@ jobs:
           components: rustfmt
       - run: cargo fmt --all --check
 
+  nostd-build:
+    runs-on: ubuntu-latest
+    continue-on-error: ${{ matrix.experimental }}
+    strategy:
+      matrix:
+        include:
+          - rust: stable
+            experimental: false
+            target: thumbv6m-none-eabi
+
+    name: nostd-build/${{ matrix.target }}/${{ matrix.rust }}
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@stable
+        with:
+          toolchain: ${{ matrix.rust }}
+          targets: ${{ matrix.target }}
+      - name: Tests
+        run: |
+          cargo rustc "--target=${{ matrix.target }}" --no-default-features --features portable-atomic-critical-section
+
   tests:
     runs-on: ubuntu-latest
     strategy:
--- a/src/impl_arc_array.rs
+++ b/src/impl_arc_array.rs
@@ -7,7 +7,12 @@
 // except according to those terms.
 
 use crate::imp_prelude::*;
+
+#[cfg(target_has_atomic = "ptr")]
 use alloc::sync::Arc;
+
+#[cfg(not(target_has_atomic = "ptr"))]
+use portable_atomic_util::Arc;
 
 /// Methods specific to `ArcArray`.
 ///
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -123,7 +123,12 @@ extern crate cblas_sys;
 #[cfg(feature = "docs")]
 pub mod doc;
 
+#[cfg(target_has_atomic = "ptr")]
 use alloc::sync::Arc;
+
+#[cfg(not(target_has_atomic = "ptr"))]
+use portable_atomic_util::Arc;
+
 use std::marker::PhantomData;
 
 pub use crate::dimension::dim::*;


Cargo.toml
```diff
@@ -45,6 +45,9 @@ matrixmultiply = { version = "0.3.2", default-features = false, features=["cgemm
 serde = { version = "1.0", optional = true, default-features = false, features = ["alloc"] }
 rawpointer = { version = "0.2" }
 
+portable-atomic = { git = "https://github.com/taiki-e/portable-atomic.git", rev = "4e66bd3", optional = true }
+portable-atomic-util = { git = "https://github.com/taiki-e/portable-atomic.git", rev = "4e66bd3", features = ["alloc"], optional = true}
```
bluss, line 49: crates with git dependencies can't be published to crates.io, so we can't merge this PR for that reason.

BjornTheProgrammer: @bluss This should be fixed now! Thank you for the feedback!

antimora:
Burn project's Ndarray backend would very much appreciate if this PR is merged. We have a long outstanding issue: https://github.com/tracel-ai/burn/issues/302

BjornTheProgrammer:
@bluss, do you have any more concerns about this pull request? If so, I would love to see if I can address them!

bluss:
Clippy is fixed on current master. I just approved the CI run.

Not sure if I can review sensibly.

BjornTheProgrammer:
What issues make it so you might be unable to review sensibly? I can help with hardware if you need it. Additionally, I could write a test to check that it compiles correctly. 

Additionally, I did forget to mention, but here are the commands needed to compile for `thumbv6m-none-eabi`.
1. `rustup target add thumbv6m-none-eabi`
2. `cargo build --target thumbv6m-none-eabi --no-default-features --features portable-atomic --features critical-section`

src/data_traits.rs
```diff
@@ -11,7 +11,12 @@
 #[allow(unused_imports)]
 use rawpointer::PointerExt;
 
+#[cfg(not(feature = "portable-atomic"))]
 use alloc::sync::Arc;
+
+#[cfg(feature = "portable-atomic")]
+use portable_atomic_util::Arc;
```
bluss, line 18: Does this affect user experience at all, does it change the API?

It's encapsulated right, so almost nothing can change - fortunately - but if I didn't check this then something like Send- or Sync-ness of some types could change or something like that.

Cargo.toml
```diff
@@ -45,6 +45,9 @@ matrixmultiply = { version = "0.3.2", default-features = false, features=["cgemm
 serde = { version = "1.0", optional = true, default-features = false, features = ["alloc"] }
 rawpointer = { version = "0.2" }
 
+portable-atomic = { version = "1.6.0", optional = true }
+portable-atomic-util = { version = "0.2.0", features = [ "alloc" ], optional = true }
+
```
bluss, line 50: Can these dependencies be automatically enabled for the relevant platform? I would feel a lot better about this, also lower complexity, if we just managed it all that way with no user visible impact, no feature enablement needed as well.

It's rare that it can be done that way, but here I think the one Arc is a drop-in for the other, and it should work well (?). Or do you see a problem with that?

And yes, it would be good to test this in CI, at least test building it.

bluss:
@BjornTheProgrammer Main reason is I'm a quite bad maintainer, as you can see, so it takes some time. So always wary to take on more complexity especially in an area that is not something I know about.

Cargo.toml
bluss: See https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#platform-specific-dependencies

BjornTheProgrammer: The issue is that the `critical-section` feature is recommended to be passed by the library, as mentioned by `portable-atomic`.

> It is very strongly discouraged to enable this feature in libraries that depend on portable-atomic. The recommended approach for libraries is to leave it up to the end user whether or not to enable this feature. (However, it may make sense to enable this feature by default for libraries specific to a platform where it is guaranteed to always be sound, for example in a hardware abstraction layer targeting a single-core chip.)

According to the Arc [docs](https://doc.rust-lang.org/std/sync/struct.Arc.html), we can detect if the target has access to Arc from the configuration flag `target_has_atomic = "ptr"`

> Note: This type is only available on platforms that support atomic loads and stores of pointers, which includes all platforms that support the std crate but not all those which only support [alloc](https://doc.rust-lang.org/1.80.0/alloc/index.html). This may be detected at compile time using #[cfg(target_has_atomic = "ptr")].

src/data_traits.rs
BjornTheProgrammer: I don't think Send or Sync should be modified, as far as I can tell here is the only possible public APIs that might be impacted

1. ArcArray
2. OwnedArcRepr (which according to the code comments should never be used by the user)

```Rust
/// ArcArray's representation.
///
/// *Don’t use this type directly—use the type alias
/// [`ArcArray`] for the array type!*
#[derive(Debug)]
pub struct OwnedArcRepr<A>(Arc<OwnedRepr<A>>);
```

`ArcArray` as far as I can see, never/rarely gives unfettered access to the `OwnedRepr` which has the `Arc`, as such I really don't think there is any affect to user experience, other than idiosyncratic different behavior from portable-atomic-util, which is unlikely, but possible.

BjornTheProgrammer:
@bluss thank you so much for the review! Don't worry about taking time, we are all busy 😅. I added the GitHub tests, but I don't have an environment to test that these are properly working, so a once-over would be much advised. Additionally I added a cfg flag to enable the dependency as suggested. I made an analysis of the user impact, and there honestly doesn't seem to be much, if any at all. However further testing should be done, just to verify that really is true.

Should I run these tests, or do you have any other suggestions/concerns?

bluss:
I suggest copying this snippet from matrixmultiply and testing it like that. Just using stable, don't need any older version.

https://github.com/bluss/matrixmultiply/blob/bb3dd0be9d8a9f592e567e4402950546bf710eba/.github/workflows/ci.yml#L94-L117

BjornTheProgrammer:
Changed to that! Should be fixed now.

bluss:
MSRV changed to 1.64 on master. Maybe that can solve some problems (but it wasn't changed because of this PR).

BjornTheProgrammer:
Updated to reflect the master branch with 1.64

antimora:
Burn is looking forward to this feature. It seems it's close =D

bluss:
Best would be if you could squash together the changes into one commit. The merge queue preserves commits, but here we don't want to save the history. (The merge queue doesn't allow us to choose merge vs squash depending on the PR, that makes me think it's a big drawback of it.)

Cargo.toml
```diff
@@ -70,8 +71,24 @@ docs = ["approx", "serde", "rayon"]
...
 matrixmultiply-threading = ["matrixmultiply/threading"]
 
+[target.'cfg(not(target_has_atomic = "ptr"))'.dependencies]
+portable-atomic = { version = "1.6.0" }
+portable-atomic-util = { version = "0.2.0", features = [ "alloc" ] }
+
+# [target.'cfg(unix)'.dependencies]
+# openssl = "1.0.1"
+
+# [target.'cfg(target_arch = "x86")'.dependencies]
```
bluss, line 85: please remove these comments

README.rst
```diff
@@ -97,6 +97,14 @@ your `Cargo.toml`.
 
   - Enable the ``threading`` feature in the matrixmultiply package
 
+- ``portable-atomic``
+
+  - Enables ``portable_atomic`` package for compatability with microcontrollers
```
bluss, line 102: This is not up to date

Cargo.toml
```diff
@@ -70,8 +71,24 @@ docs = ["approx", "serde", "rayon"]
 std = ["num-traits/std", "matrixmultiply/std"]
 rayon = ["dep:rayon", "std"]
 
+critical-section = ["portable-atomic/critical-section"]
```
bluss, line 74: is critical section the right feature name for us? Maybe not

BjornTheProgrammer: What would recommend the name to be? Something like `no-std-enable-critical-section`?

[Heepless](https://github.com/rust-embedded/heapless), a very popular embedded rust project names the feature `portable-atomic-critical-section`
```toml
# Enable polyfilling of atomics via portable-atomic, using critical section for locking
portable-atomic-critical-section = ["dep:portable-atomic", "portable-atomic", "portable-atomic?/critical-section"]
```

BjornTheProgrammer:
There I think I've fixed the history, just need to rename `critical-section` and squash, and then we should be good. What name would you like?

Cargo.toml
bluss: portable-atomic-critical-section sounds good then

BjornTheProgrammer:
I've tested on actual hardware, and it seems to be working properly!

BjornTheProgrammer:
@bluss I would just like to thank you very much for the time and effort you put into this. Your contributions are extremely valued. Thanks once again!

bluss:
Great, I'm working on the release  #1401 which is overdue (to say the least)

PR #1386, opened 2024-05-16
Adds `triu` and `tril` methods directly to ArrayBase

### LINKED ISSUE #1323: Implement numpy's `triu` and `tril` methods

akern40:
The implementation seems to work, but unsure about whether this should be a free function vs bound directly to ArrayBase vs a trait. Closes #1323.

### DIFF
--- a/src/tri.rs
+++ b/src/tri.rs
@@ -100,8 +100,6 @@ where
                 Zip::indexed(self.rows())
                     .and(res.rows_mut())
                     .for_each(|i, src, mut dst| {
-                        // This ncols must go inside the loop to avoid panic on 1D arrays.
-                        // Statistically-neglible difference in performance vs defining ncols at top.
                         let row_num = i.into_dimension().last_elem();
                         let upper = min(row_num as isize + k, ncols) + 1;
                         dst.slice_mut(s![..upper]).assign(&src.slice(s![..upper]));
--- a/src/tri.rs
+++ b/src/tri.rs
@@ -1,0 +1,288 @@
+// Copyright 2014-2024 bluss and ndarray developers.
+//
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+use core::cmp::{max, min};
+
+use num_traits::Zero;
+
+use crate::{dimension::is_layout_f, Array, ArrayBase, Axis, Data, Dimension, IntoDimension, Zip};
+
+impl<S, A, D> ArrayBase<S, D>
+where
+    S: Data<Elem = A>,
+    D: Dimension,
+    A: Clone + Zero,
+    D::Smaller: Copy,
+{
+    /// Upper triangular of an array.
+    ///
+    /// Return a copy of the array with elements below the *k*-th diagonal zeroed.
+    /// For arrays with `ndim` exceeding 2, `triu` will apply to the final two axes.
+    /// For 0D and 1D arrays, `triu` will return an unchanged clone.
+    ///
+    /// See also [`ArrayBase::tril`]
+    ///
+    /// ```
+    /// use ndarray::array;
+    ///
+    /// let arr = array![[1, 2, 3], [4, 5, 6], [7, 8, 9]];
+    /// let res = arr.triu(0);
+    /// assert_eq!(res, array![[1, 2, 3], [0, 5, 6], [0, 0, 9]]);
+    /// ```
+    pub fn triu(&self, k: isize) -> Array<A, D>
+    {
+        match self.ndim() > 1 && is_layout_f(&self.dim, &self.strides) {
+            true => {
+                let n = self.ndim();
+                let mut x = self.view();
+                x.swap_axes(n - 2, n - 1);
+                let mut tril = x.tril(-k);
+                tril.swap_axes(n - 2, n - 1);
+
+                tril
+            }
+            false => {
+                let mut res = Array::zeros(self.raw_dim());
+                Zip::indexed(self.rows())
+                    .and(res.rows_mut())
+                    .for_each(|i, src, mut dst| {
+                        let row_num = i.into_dimension().last_elem();
+                        let lower = max(row_num as isize + k, 0);
+                        dst.slice_mut(s![lower..]).assign(&src.slice(s![lower..]));
+                    });
+
+                res
+            }
+        }
+    }
+
+    /// Lower triangular of an array.
+    ///
+    /// Return a copy of the array with elements above the *k*-th diagonal zeroed.
+    /// For arrays with `ndim` exceeding 2, `tril` will apply to the final two axes.
+    /// For 0D and 1D arrays, `tril` will return an unchanged clone.
+    ///
+    /// See also [`ArrayBase::triu`]
+    ///
+    /// ```
+    /// use ndarray::array;
+    ///
+    /// let arr = array![[1, 2, 3], [4, 5, 6], [7, 8, 9]];
+    /// let res = arr.tril(0);
+    /// assert_eq!(res, array![[1, 0, 0], [4, 5, 0], [7, 8, 9]]);
+    /// ```
+    pub fn tril(&self, k: isize) -> Array<A, D>
+    {
+        match self.ndim() > 1 && is_layout_f(&self.dim, &self.strides) {
+            true => {
+                let n = self.ndim();
+                let mut x = self.view();
+                x.swap_axes(n - 2, n - 1);
+                let mut tril = x.triu(-k);
+                tril.swap_axes(n - 2, n - 1);
+
+                tril
+            }
+            false => {
+                let mut res = Array::zeros(self.raw_dim());
+                Zip::indexed(self.rows())
+                    .and(res.rows_mut())
+                    .for_each(|i, src, mut dst| {
+                        // This ncols must go inside the loop to avoid panic on 1D arrays.
+                        // Statistically-neglible difference in performance vs defining ncols at top.
+                        let ncols = src.len_of(Axis(src.ndim() - 1)) as isize;
+                        let row_num = i.into_dimension().last_elem();
+                        let upper = min(row_num as isize + k, ncols) + 1;
+                        dst.slice_mut(s![..upper]).assign(&src.slice(s![..upper]));
+                    });
+
+                res
+            }
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests
+{
+    use crate::{array, dimension, Array0, Array1, Array2, Array3, ShapeBuilder};
+    use std::vec;
+
+    #[test]
+    fn test_keep_order()
+    {
+        let x = Array2::<f64>::ones((3, 3).f());
+        let res = x.triu(0);
+        assert!(dimension::is_layout_f(&res.dim, &res.strides));
+
+        let res = x.tril(0);
+        assert!(dimension::is_layout_f(&res.dim, &res.strides));
+    }
+
+    #[test]
+    fn test_0d()
+    {
+        let x = Array0::<f64>::ones(());
+        let res = x.triu(0);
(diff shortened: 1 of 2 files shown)


src/tri.rs
```diff
@@ -0,0 +1,165 @@
...
+use num_traits::Zero;
+
+use crate::{Array, ArrayBase, Axis, Data, Dimension, IntoDimension, RemoveAxis, Zip};
+
+impl<S, A, D> ArrayBase<S, D>
+where
+    S: Data<Elem = A>,
+    D: Dimension + RemoveAxis,
+    A: Clone + Zero,
+    D::Smaller: Copy,
```
nilgoyette, line 20: Is this required? I don't see any smaller dimension here.

src/tri.rs
```diff
@@ -0,0 +1,165 @@
...
+    /// let res = arr.triu(0);
+    /// assert_eq!(res, array![[1, 2, 3], [0, 5, 6], [0, 0, 9]]);
+    /// ```
+    pub fn triu(&self, k: isize) -> Array<A, D> {
+        let mut res = Array::zeros(self.raw_dim());
+        Zip::indexed(self.rows())
+            .and(res.rows_mut())
+            .for_each(|i, src, mut dst| {
+                let row_num = i.into_dimension().last_elem();
+                let lower: isize = max(row_num as isize + k, 0);
```
nilgoyette, line 42: You can remote the type annotation, no? Or directly use `usize::max`. Idem for `tril`.

src/tri.rs
```diff
@@ -0,0 +1,165 @@
...
+        res
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::array;
+
+    #[test]
+    fn test_2d() {
```
nilgoyette, line 83: So, what happens on 1D array?

src/tri.rs
```diff
@@ -0,0 +1,165 @@
...
+    ///
+    /// ```
+    /// use ndarray::array;
+    ///
+    /// let arr = array![[1, 2, 3], [4, 5, 6], [7, 8, 9]];
+    /// let res = arr.tril(0);
+    /// assert_eq!(res, array![[1, 0, 0], [4, 5, 0], [7, 8, 9]]);
+    /// ```
+    pub fn tril(&self, k: isize) -> Array<A, D> {
+        let mut res = Array::zeros(self.raw_dim());
```
nilgoyette, line 64: This will create a c-order array even if the input is a f-order array. I think calling `clone` is better here.

src/tri.rs
akern40: Ah ya that's left over from an intermediate implementation I was trying before this design, I think we can get rid of it.

src/tri.rs
akern40: Yes, absolutely - more crud accidentally left in. Will remove.

src/tri.rs
akern40: Excellent question, and I think it comes down to design - NumPy broadcasts the 1D array into a square matrix, then does the triangular, like this:
```python
>>> np.triu([1, 2, 3])
array([[1, 2, 3],
       [0, 2, 3],
       [0, 0, 3]])
```
However, it seems from their code that this is an implementation result rather than a design decision? Since they use a mask rather than my direct iteration, causing broadcasting.

I think we have three options:
1. Copy NumPy
2. Use a `Result`
3. If possible, use some sort of trait bound, or only implement it for arrays that are 2D or more.
I like (3) the most, but would love some input.

src/tri.rs
akern40: Ah actually this is something I was unclear about - is there a way to create an array of zeros with the same shape *and* order? Or are you saying I should `.clone` the input and fill it with zeros, instead?

src/tri.rs
akern40: Ok coming back - this is required due to a trait bound on the `indexed` call.

src/tri.rs
akern40: I think `Array::zeros(self.raw_dim().set_f(is_layout_f(&self.dim, &self.strides)))` is doing what I want, just don't know if there's another way.

src/tri.rs
akern40: Ok, the behavior as it is here just... doesn't change 1D arrays. And I think that's a reasonable result that I hadn't really thought of.

src/tri.rs
nilgoyette: No, you're right. `clone()` is sexier but it would force you to fill the array with 0 after cloning, which is wasteful.

I don't know if the maintainers usually try making optimized version for f-order arrays, but one could make a fast version for f-order arrays by calling `columns` instead of `rows`. Feels like code duplication though. I'm not asking you to do it, I'm just wondering.

src/tri.rs
nilgoyette: I just tested numpy's behavior and... wow, it transform the 1D array into a 2D array, something that we can't do in ndarray. I think what you did is great.

akern40:
Hold off on any merge for a moment; `tril` panics when given a 0D array, working on a fix.

akern40:
Ok, fixed.

Not to move conversations, but two thoughts on the (now resolved) discussion about order of the input array:
1. Would there be appetite for `{zeros|ones|full}_like` public functions? That way users don't have to do the rather-convoluted `self.raw_dim().set_f(is_layout_f(&self.dim, &self.strides))` that I'm doing. Happy to make a tracking issue and put in the PRs for that.
2. I did a quick benchmark of `tri[ul]`'s performance on c-order vs f-order, and it's not great; results are below. The slowdown is quite apparent at larger array sizes, as expected. The `via_t` lines are referring to the quick and dirty solution I could think of, which is `x_tril = x.t().triu(0).t().to_owned()`.
```shell
running 3 tests (64 x 64)
test triu_c       ... bench:       1,681 ns/iter (+/- 45)
test triu_f       ... bench:       2,011 ns/iter (+/- 32)
test triu_f_via_t ... bench:       2,133 ns/iter (+/- 29)

running 3 tests (2048 x 2048)
test triu_c       ... bench:   2,418,649 ns/iter (+/- 320,729)
test triu_f       ... bench:  22,184,991 ns/iter (+/- 4,532,682)
test triu_f_via_t ... bench:   4,259,493 ns/iter (+/- 218,849)
```

nilgoyette:
>     1. Would there be appetite for `{zeros|ones|full}_like` public functions? That way users don't have to do the rather-convoluted `self.raw_dim().set_f(is_layout_f(&self.dim, &self.strides))` that I'm doing. Happy to make a tracking issue and put in the PRs for that.

Well, I know that I'm interested in those functions, but it may not be what @bluss had in mind. The more methods we have, the harder it is to find them.

As for the benchmarks, it's kinda normal because those larger arrays don't fit in the cache and you're iterating in the wrong order. Seeing those results, I think adding a condition on the memory order is worth it. Some people (like me) often work in f-order :)

akern40:
> Well, I know that I'm interested in those functions, but it may not be what @bluss had in mind. The more methods we have, the harder it is to find them.

This is true; I don't want to bloat the API. That being said, I think some (difficult and time-consuming and I don't mean to imply it's trivial) documentation work could make the docs easier to navigate and alleviate concerns about more functions.

> Some people (like me) often work in f-order :)

Me too :) too many column-vector algorithms / math. I was interested in seeing how well NumPy does on the transposed version and I guess they didn't write one that dispatches on order:
```python
In [1]: import numpy as np

In [2]: x = np.ones((2048, 2048))

In [3]: %timeit np.triu(x)
4.34 ms ± 53.8 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [4]: x = np.ones((2048, 2048), order="F")

In [5]: %timeit np.triu(x)
19 ms ± 557 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```
(Fun to see that `ndarray` is faster than NumPy here, though! Although that extra bit of time could just be Python overhead, rather than raw NumPy performance).

If we want to create a specialization, I think we can do it via the transpose-transpose method I mentioned above, with a slight tweak to avoid the copy:
```rust
pub fn triu(&self, k: isize) -> Array<A, D> {
    match is_layout_c(&self.dim, &self.strides) {
        false => self.t().tril(-k).into_shared().reversed_axes().into_owned(),
        true => {
            let mut res = Array::zeros(self.raw_dim());
            Zip::indexed(self.rows())
                .and(res.rows_mut())
                .for_each(|i, src, mut dst| {
                    let row_num = i.into_dimension().last_elem();
                    let lower = max(row_num as isize + k, 0);
                    dst.slice_mut(s![lower..]).assign(&src.slice(s![lower..]));
                });

            res
        }
    }
}
```
which gets us the following performance:
```shell
running 3 tests (64 x 64)
test triu_c       ... bench:       1,604 ns/iter (+/- 37)
test triu_f       ... bench:       1,635 ns/iter (+/- 33)

running 3 tests  (2048 x 2048)
test triu_c       ... bench:   2,486,411 ns/iter (+/- 129,116)
test triu_f       ... bench:   2,529,041 ns/iter (+/- 144,131)
```
If this gets some thumbs up I'll add it to the PR (and maybe squash at this point to reduce commits?)

Glad I got to explore the `_owned` methods - seems like it should be possible to do non-copy `_owned` on mutable views, but that's an improvement for another day / PR.

src/tri.rs
```diff
@@ -0,0 +1,288 @@
...
+                tril
+            }
+            false => {
+                let mut res = Array::zeros(self.raw_dim());
+                Zip::indexed(self.rows())
+                    .and(res.rows_mut())
+                    .for_each(|i, src, mut dst| {
+                        // This ncols must go inside the loop to avoid panic on 1D arrays.
+                        // Statistically-neglible difference in performance vs defining ncols at top.
+                        let ncols = src.len_of(Axis(src.ndim() - 1)) as isize;
```
nilgoyette, line 98: I wonder, isn't it simpler to start with

    if self.ndim() <= 1
        return clone or to_owned

remove the first condition in the `match` and put the `ncols` line before the `Zip`? 0D and 1D are simply copied anyway. This would make that rules clear.

Both ways are fine, it's only a suggestion.

nilgoyette:
Thank you for your contribution @akern40

zackmdavis:
Thanks @akern40!!

To anyone waiting to use this before 0.16 is released, remember that [you can run master by changing the version specification in your Cargo.toml to](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-git-repositories) `{git = "https://github.com/rust-ndarray/ndarray"}` . While it doesn't work with the ndarray-rand 0.14 from crates.io, you can just change the ndarray-rand dependency to also be `{git = "https://github.com/rust-ndarray/ndarray"}`, and then it works. (ndarray-rand lives in the same repo; Cargo is smart enough to find it.)

PR #1389, opened 2024-05-20
Expands the `array!` macro.

### LINKED ISSUE #1252: The `array!` macro's documentation should clarify that if we attempt to create arrays with more than 3 dimensions they are ignored
[The documentation](https://docs.rs/ndarray/latest/ndarray/macro.array.html) says 

> Create an [Array](https://docs.rs/ndarray/latest/ndarray/type.Array.html) with one, two or three dimensions.

This is fine. 

However, I wonder if the following code should be able to run silently (not even a warning).

```
use ndarray::array;

fn main() {
    let a = array![
        [
                [[[2, 2]]], 
                [[[3, 1]]], 
                [[[5, 3]]], 
                [[[2, 2]]]
        ]
        ];
    println!("An int array with shape {:?}: {:?}", a.shape(), a.ndim()); // Shape is [1, 4, 1]
}

```

Basically, what's happening is that some dimensions or parentheses are automatically ignored. I think this should at least be documented or maybe the macro should fail in those cases? I am happy to contribute to the documentation, if you think that's the way to go in this case. I am also opening this issue because, if we attempt to create a similar thing in numpy with e.g. the following code

```
import numpy as np

a = np.array(
        [
                [[[2, 2]]], 
                [[[3, 1]]], 
                [[[5, 3]]], 
                [[[2, 2]]]
        ]
        )

print(a.shape)
```

We get a different shape: `(4, 1, 1, 2)`.

Btw, I suppose there's no macro that allows us to create arrays from literals with more than 3 dimensions. Is that correct? Is there any reason why the macro only supports up to 3 dimensions? In machine learning, for example, it it's not rare to have multi-dimensional arrays with more than 3 dimensions, so I think that supporting the creation of arrays from literals with more than 3 dimensions would not be a bad idea.

akern40:
Allows the creation of arrays up to and including 6 dimensions, and causes a compiler error on 7 or more dimensions. Would close #1252.

### DIFF
--- a/src/dimension/ndindex.rs
+++ b/src/dimension/ndindex.rs
@@ -113,6 +113,22 @@ unsafe impl NdIndex<Ix5> for (Ix, Ix, Ix, Ix, Ix)
     }
 }
 
+unsafe impl NdIndex<Ix6> for (Ix, Ix, Ix, Ix, Ix, Ix)
+{
+    #[inline]
+    fn index_checked(&self, dim: &Ix6, strides: &Ix6) -> Option<isize>
+    {
+        dim.stride_offset_checked(strides, &self.into_dimension())
+    }
+    #[inline]
+    fn index_unchecked(&self, strides: &Ix6) -> isize
+    {
+        zip(strides.ix(), self.into_dimension().ix())
+            .map(|(&s, &i)| stride_offset(i, s))
+            .sum()
+    }
+}
+
 unsafe impl NdIndex<Ix1> for Ix
 {
     #[inline]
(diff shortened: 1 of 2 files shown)


bluss:
current status is that I'm trying to write up release notes for ndarray 0.16 so we get a release. It would be good to include this because it's technically a breaking change. (Type changes for identical source input).

akern40:
Sounds good - if there are any changes needed I can get to them fairly quickly.

Also, I've been putting in PRs to help out and get familiar with the code base. If there's anything else I can do to help (write docs for the next release, run rustfmt on all the code, handle specific PRs you want in the next release, etc) please let me know.

PR #1396, opened 2024-07-03
Add `squeeze()` to dynamic dimension arrays

barakugav:


src/impl_dyn.rs
```diff
@@ -52,10 +52,62 @@ where S: Data<Elem = A>
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::{arr1, arr2, arr3};
+
+    #[test]
+    fn test_squeeze() {
+        let a = arr3(&[[[1, 2, 3]], [[4, 5, 6]]]).into_dyn();
```
nilgoyette, line 99: Why are you calling `into_dyn` each time? Surely you can compare owned arrays?

nilgoyette (commented):
Moreover, can you please merge master in your branch?

barakugav:
> Moreover, can you please merge master in your branch?

Rebased

src/impl_dyn.rs
nilgoyette: Oh, that's right, sorry. It's impossible to create a `squeeze` function for fixed arrays in Rust because we don't know how many (if any) axes will be removed. My bad.

src/impl_dyn.rs
```diff
@@ -52,10 +52,62 @@ where S: Data<Elem = A>
...
+    /// let a = arr3(&[[[1, 2, 3]], [[4, 5, 6]]]).into_dyn();
+    /// assert_eq!(a.shape(), &[2, 1, 3]);
+    /// let b = a.squeeze();
+    /// assert_eq!(b, arr2(&[[1, 2, 3], [4, 5, 6]]).into_dyn());
+    /// assert_eq!(b.shape(), &[2, 3]);
+    ///
+    /// let c = arr2(&[[1]]).into_dyn();
+    /// assert_eq!(c.shape(), &[1, 1]);
+    /// let d = c.squeeze();
+    /// assert_eq!(d, arr1(&[1]).into_dyn());
```
nilgoyette, line 78: You use `arr1` but you do not import it.

barakugav: right, fixed

nilgoyette:
Thank you for your contribution.

nilgoyette (approved):

nilgoyette:
@adamreichold I don't have the right to merge this branch. There are several errors when I run `cargo clippy --features docs` on beta. Some of them are important and would necessitate a their own PR, like

    warning: current MSRV (Minimum Supported Rust Version) is `1.57.0` but this item is stable since `1.80.0`
      --> src/data_repr.rs:75:27
       |
    75 |         unsafe { self.ptr.add(self.len) }

There's no error when I use stable. I'm not sure why we run clippy on beta...?

barakugav:
@nilgoyette can we merge this PR?

bluss:
@nilgoyette that sounds like a false positive from clippy (we use a trait that adds that .sub method).

nilgoyette:
@bluss Ok, thanks for the information. I can ask for an auto-merge but I get this message from github

    Required statuses must pass before merging

so the only option I have is to click on `Disable auto-merge`. You probably have more options than me!

bluss:
Please fix rustfmt here if you can.
Then we need to fix clippy in a separate PR, rebase this one and merge. That's the sound way to inclusion.

Clippy doesn't always help us and here it's slowing us down instead. Stable or beta I don't know if it matters (what do you think?), but at least that means it's only a few times per Rust cycle that new lints get published, so it's a bit more predictable.

nilgoyette:
> Stable or beta I don't know if it matters (what do you think?), but at least that means it's only a few times per Rust cycle that new lints get published, so it's a bit more predictable.

Personally, I would never use beta or nightly on a mandatory cicd check (on a project that must run on stable). That's why I was asking above why you were doing it. I think it can only lead to bad surprises like what we have now. Since it doesn't bother you either ways, I'll create a MR to use stable for that check.

bluss:
I guess it's a little bit less likely to have surprises in clippy lints on stable, but I figure most of it will be the same between beta and stable (just a few weeks delayed?).

nilgoyette:
Yes, you're right about the delay, but there's also another reason. When someone pushes a MR, he probably ran the tests and clippy before (I'm not the only one doing this?), and he's likely to be running stable. There's less chance of surprise if we run the cicd with the config that the contributor probably used.

nilgoyette:
> Please fix rustfmt here if you can.

@bluss I was about to merge this, but then I saw this comment. Do you mean the first 3 changes that look like an auto-format? If not, I'm not sure what you meant.

nilgoyette:
@bluss I'm not sure what I'm supposed to fix about the formatting. The three auto-format changes in this MR seem right to me. It's the CI that is wrong. Surely it's not nightly that produces this strange formatting?

bluss:
I was thinking @barakugav (the author) would fix rustfmt usually.

Ndarray uses nightly for rustfmt since we setup some formatting options, and they are only supported on nightly. https://github.com/rust-ndarray/ndarray/pull/1375

bluss:

PR #1397, opened 2024-07-09
Add `flatten`, `flatten_with_order` and `into_flat` to arrays

barakugav:


src/impl_methods.rs
```diff
@@ -2131,6 +2131,20 @@ where
...
+    /// ```
+    /// use ndarray::{arr1, arr3};
+    ///
+    /// let array = arr3(&[[[1, 2], [3, 4]], [[5, 6], [7, 8]]]);
+    /// let flattened = array.flatten();
+    /// assert_eq!(flattened, arr1(&[1, 2, 3, 4, 5, 6, 7, 8]));
+    /// ```
+    pub fn flatten(self) -> ArrayBase<S, Ix1> {
+        let len = self.len();
+        self.into_shape_with_order((len,)).unwrap()
```
bluss, line 2145: will panic on some arrays that are not contiguous. Another design is needed unfortunately.

bluss: Suggestion: base it on to_shape and return the same kind of array that it does. Have both flatten and flatten with order to handle the optional order argument.

barakugav: @bluss patched, what do you think?

src/impl_methods.rs
```diff
@@ -2131,6 +2131,73 @@ where
...
+    /// Otherwise, a new array is allocated and the elements are copied.
+    ///
+    /// ```
+    /// use ndarray::{arr1, arr3};
+    ///
+    /// let array = arr3(&[[[1, 2], [3, 4]], [[5, 6], [7, 8]]]);
+    /// let flattened = array.into_flat();
+    /// assert_eq!(flattened, arr1(&[1, 2, 3, 4, 5, 6, 7, 8]));
+    /// ```
+    pub fn into_flat(self) -> ArrayBase<S, Ix1>
```
bluss, line 2188: I think the changes are good but would like to remove this method. Users can use .flatten().into_owned() here I think

barakugav: But than you force a copy, as `flatten` accept `&self`... Don't you want to provide an `into_` variant?

bluss: right, I somehow missed that.

src/impl_methods.rs
```diff
@@ -2131,6 +2131,73 @@ where
...
+    where
+        A: Clone,
+        S: DataOwned,
+    {
+        let len = self.len();
+        if self.is_standard_layout() {
+            self.into_shape_with_order(len).unwrap()
+        } else {
+            ArrayBase::from_shape_vec(len, self.iter().cloned().collect()).unwrap()
```
bluss, line 2197: I want to avoid this code here, seems like yet another implementation of logic that must exist somewhere else, including in into_shape_clone.

barakugav: @bluss done

bluss (approved):

bluss:
Maintainer pushed to fix to always use into_shape_clone (it only clones when necessary)

bluss:
clippy lints we'll fix in another PR

bluss:

PR #1399, opened 2024-07-21
`is_unique` for `ArcArray`

daniellga:
Implement this function to know if the inner `Arc` is shared.

### DIFF
--- a/src/impl_arc_array.rs
+++ b/src/impl_arc_array.rs
@@ -13,12 +13,12 @@ use alloc::sync::Arc;
 ///
 /// ***See also all methods for [`ArrayBase`]***
 impl<A, D> ArcArray<A, D>
-where
-    D: Dimension,
+where D: Dimension
 {
     /// Returns `true` iff the inner `Arc` is not shared.
     /// If you want to ensure the `Arc` is not concurrently cloned, you need to provide a `&mut self` to this function.
-    pub fn is_unique(&self) -> bool {
+    pub fn is_unique(&self) -> bool
+    {
         // Only strong pointers are used in this crate.
         Arc::strong_count(&self.data.0) == 1
     }
--- a/src/impl_arc_array.rs
+++ b/src/impl_arc_array.rs
@@ -17,7 +17,7 @@ where
     D: Dimension,
 {
     /// Returns `true` iff the inner `Arc` is not shared.
-    /// If you want to ensure the Arc is not concurrently cloned, you need to provide a `&mut self` to this function.
+    /// If you want to ensure the `Arc` is not concurrently cloned, you need to provide a `&mut self` to this function.
     pub fn is_unique(&self) -> bool {
         // Only strong pointers are used in this crate.
         Arc::strong_count(&self.data.0) == 1
--- a/src/impl_arc_array.rs
+++ b/src/impl_arc_array.rs
@@ -6,8 +6,8 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
-use alloc::sync::Arc;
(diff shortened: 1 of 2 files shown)


src/impl_arc_array.rs
```diff
@@ -0,0 +1,23 @@
...
+
+/// Methods specific to `ArcArray`.
+///
+/// ***See also all methods for [`ArrayBase`]***
+impl<A, D> ArcArray<A, D>
+where
+    D: Dimension,
+{
+    /// Returns `true` iff the inner `Arc` is not shared.
+    pub fn is_unique(&mut self) -> bool {
```
bluss, line 20: Is it possible to make this a non-mut method?

daniellga: Wouldn't it risk creating a race condition? I kinda got confused on this point... The alternative would be just to sum strong and weak counts:
```
fn is_unique(&self) -> bool {
    Arc::strong_count(&self.data.0) + Arc::weak_count(&self.data.0) == 1
}
```

bluss: There are race conditions anyway, because the method only returns a boolean. When is_unique has finished and returns its value, it could already be shared again. There's no solving that without creating something entirely different.

What's the use case for this?

daniellga: In my case I am using an `ArcArray` as the core of a struct that I call from other language. I want the user of that language to be able to know when a certain method, lets say for example fft calculation, has created a copy of that inner `ArcArray`.

bluss: It is confusing, and if there would be weak refs you're right that it's racy. ArcArray doesn't support weak refs at all. That means there are two options - (1) the way of just checking strong count and deciding we can never support Weak refs or (2) your get_mut implementation which is more sure to always be safe.

Any "is_unique -> bool" method is racy unless we can control the current Arc, i.e we are sure nobody's cloning the exact Arc handle we are holding. For this reason, ownership or &mut of the current Arc is needed to be sure of that.

I would say both (1) in combination with holding or having the option to hold a &mut Arc would be just as safe as (2) for ArcArray, but which choice to use doesn't matter so much to me.


If there is code that first .clone() and then mutate (unsharing) your ArcArray, you won't notice this way, because the final effect is two independent copies of the data.

daniellga: You are saying something like this for (1)?
```
fn is_unique(&mut self) -> bool {
    Arc::strong_count(&self.0) == 1
}
```

I am ok with both choices, I just used the second one because of extensibility, in case you choose to use weak references in the future, for whatever reason. Let me know if you have any better suggestion than this.

bluss: I think it's clear it's possible to make it a non-mut method - using strong_count.
That seems to be compatible with the use case as well (your use case - if you want to ensure the Arc is not concurrently cloned, you can hold a &mut to the ArcArray, the method doesn't need to use &mut for that). I can't imagine too many other use cases, other than some user wanting to do something they shouldn't w.r.t mutability and sharing :laughing:.

Any other ideas about use cases?
Maybe get_mut is the safest choice for unknown use cases?

daniellga: I will modify it then to count strong count and accept self. Another approach would be to allow access (a getter) to the inner arc, then the user could do whatever he wanted with it.
Just for curiosity, why do you want to make it non-mut?

bluss: For non-mut, just the instinct that it's an interrogating method that doesn't modify.
Inner fields won't be exposed :slightly_smiling_face:

bluss:

bluss:
Thanks

PR #1403, opened 2024-07-26
WIP: Cross builder failure

bluss:
New failure in cross builder observed [here](https://github.com/rust-ndarray/ndarray/actions/runs/10115462037/job/27976277486) probably new due to Rust 1.80.0
specific to s390x platform

2024-07-26T17:49:57.6253443Z thread 'test_mat_mul' panicked at library/core/src/panicking.rs:219:5:
2024-07-26T17:49:57.6255056Z unsafe precondition(s) violated: ptr::replace requires that the pointer argument is aligned and non-null
2024-07-26T17:49:57.6256043Z thread caused non-unwinding panic. aborting.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -9,6 +9,7 @@ env:
   HOST: x86_64-unknown-linux-gnu
   FEATURES: "test docs"
   RUSTFLAGS: "-D warnings"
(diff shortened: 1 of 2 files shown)


bluss:
Fixed in matrixmultiply, cross builder now passes, no change needed here.

PR #1418, opened 2024-08-07
Fix infinite recursion and off-by-one error in triu/tril

### LINKED ISSUE #1415: tril and triu fail on 1×1 matrices
I just [upgraded to my project to use `ndarray` master](https://github.com/zackmdavis/Radiance/commit/c38bbbea14e6cc11fbf14fbee3ad70125889d402) in order to be able to use #1386, but discovered that these are failing with `thread [test name] has overflowed its stack`. (It's a two-dimensional array where both dimensions are 1, perhaps an easy edge-case to overlook.)

I was about to say, "I guess I'm glad I caught this before the eventual 0.16 release", but it looks like 0.16 came out four hours ago?

```rust
    #[test]
    fn test_tril() {
        let a: Array2<f32> = Array2::ones((1, 1));
        let b = a.tril(0);
    }

    #[test]
    fn test_triu() {
        let a: Array2<f32> = Array2::ones((1, 1));
        let b = a.triu(0);
    }
```

akern40:
Also makes the documentation a little nicer to read, since these are such visual methods.

Closes #1415

### DIFF
--- a/src/tri.rs
+++ b/src/tri.rs
@@ -10,7 +10,16 @@ use core::cmp::{max, min};
 
 use num_traits::Zero;
 
-use crate::{dimension::is_layout_f, Array, ArrayBase, Axis, Data, Dimension, IntoDimension, Zip};
+use crate::{
+    dimension::{is_layout_c, is_layout_f},
+    Array,
+    ArrayBase,
+    Axis,
+    Data,
+    Dimension,
+    IntoDimension,
+    Zip,
+};
 
 impl<S, A, D> ArrayBase<S, D>
 where
@@ -30,38 +39,44 @@ where
     /// ```
     /// use ndarray::array;
     ///
-    /// let arr = array![[1, 2, 3], [4, 5, 6], [7, 8, 9]];
-    /// let res = arr.triu(0);
-    /// assert_eq!(res, array![[1, 2, 3], [0, 5, 6], [0, 0, 9]]);
+    /// let arr = array![
+    ///     [1, 2, 3],
+    ///     [4, 5, 6],
+    ///     [7, 8, 9]
+    /// ];
+    /// assert_eq!(
+    ///     arr.triu(0),
+    ///     array![
+    ///         [1, 2, 3],
+    ///         [0, 5, 6],
+    ///         [0, 0, 9]
+    ///     ]
+    /// );
     /// ```
     pub fn triu(&self, k: isize) -> Array<A, D>
     {
         if self.ndim() <= 1 {
             return self.to_owned();
         }
-        match is_layout_f(&self.dim, &self.strides) {
-            true => {
-                let n = self.ndim();
-                let mut x = self.view();
-                x.swap_axes(n - 2, n - 1);
-                let mut tril = x.tril(-k);
-                tril.swap_axes(n - 2, n - 1);
+        if is_layout_f(&self.dim, &self.strides) && !is_layout_c(&self.dim, &self.strides) {
+            let n = self.ndim();
+            let mut x = self.view();
+            x.swap_axes(n - 2, n - 1);
+            let mut tril = x.tril(-k);
+            tril.swap_axes(n - 2, n - 1);
 
-                tril
-            }
-            false => {
-                let mut res = Array::zeros(self.raw_dim());
-                Zip::indexed(self.rows())
-                    .and(res.rows_mut())
-                    .for_each(|i, src, mut dst| {
-                        let row_num = i.into_dimension().last_elem();
-                        let lower = max(row_num as isize + k, 0);
-                        dst.slice_mut(s![lower..]).assign(&src.slice(s![lower..]));
-                    });
-
-                res
-            }
+            return tril;
         }
+        let mut res = Array::zeros(self.raw_dim());
+        Zip::indexed(self.rows())
+            .and(res.rows_mut())
+            .for_each(|i, src, mut dst| {
+                let row_num = i.into_dimension().last_elem();
+                let lower = max(row_num as isize + k, 0);
+                dst.slice_mut(s![lower..]).assign(&src.slice(s![lower..]));
+            });
+
+        res
     }
 
     /// Lower triangular of an array.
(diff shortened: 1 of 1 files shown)


src/tri.rs
```diff
@@ -30,38 +39,44 @@ where
...
+
+            return tril;
         }
+        let mut res = Array::zeros(self.raw_dim());
+        Zip::indexed(self.rows())
+            .and(res.rows_mut())
+            .for_each(|i, src, mut dst| {
+                let row_num = i.into_dimension().last_elem();
+                let lower = max(row_num as isize + k, 0);
+                dst.slice_mut(s![lower..]).assign(&src.slice(s![lower..]));
```
bluss, line 76: I think this code often panics if `lower` is too large, and it depends on k and the array size. Maybe best to check this upfront with an assertion (better error message.)
I don't know if it's intentional or if all inputs `k` should be ok. By some interpretations, k=isize::MAX should work and create an all zeros array? (I'm not saying that's how it has to be.)

src/tri.rs
```diff
@@ -30,38 +39,44 @@ where
...
-                    .and(res.rows_mut())
-                    .for_each(|i, src, mut dst| {
-                        let row_num = i.into_dimension().last_elem();
-                        let lower = max(row_num as isize + k, 0);
-                        dst.slice_mut(s![lower..]).assign(&src.slice(s![lower..]));
-                    });
-
-                res
-            }
+        if is_layout_f(&self.dim, &self.strides) && !is_layout_c(&self.dim, &self.strides) {
```
bluss, line 61: Thanks for the work on this.

Why is there a conditional on memory layout? This purely a performance optimization right? Great place to have a comment to say what it's doing.

src/tri.rs
```diff
@@ -30,38 +39,44 @@ where
...
+    ///     [7, 8, 9]
+    /// ];
+    /// assert_eq!(
+    ///     arr.triu(0),
+    ///     array![
+    ///         [1, 2, 3],
+    ///         [0, 5, 6],
+    ///         [0, 0, 9]
+    ///     ]
+    /// );
```
bluss, line 54: Ndarray functions try to document if they panic.

Here we need something like `**Panics** if k > something` (See other comments why, See other functions for the common format.)

src/tri.rs
```diff
@@ -30,38 +39,44 @@ where
     /// ```
```
bluss, line 39: Could the D::Smaller: Copy bound above be removed (in a future PR, if needed?) It rules out IxDyn for now, is that intentional?

src/tri.rs
```diff
@@ -30,38 +39,44 @@ where
...
-                        dst.slice_mut(s![lower..]).assign(&src.slice(s![lower..]));
-                    });
-
-                res
-            }
+        if is_layout_f(&self.dim, &self.strides) && !is_layout_c(&self.dim, &self.strides) {
+            let n = self.ndim();
+            let mut x = self.view();
+            x.swap_axes(n - 2, n - 1);
+            let mut tril = x.tril(-k);
```
bluss, line 65: If all `k` values are accepted (maybe in the future), then k=`isize::MIN` has a problem here.

src/tri.rs
akern40: Yes it is - I've added a comment detailing the purpose and the point of the two other `&&` conditions.

src/tri.rs
akern40: Aiming for all `k` to be allowed, I've added a check for this optimization that skips it if `k == isize::MIN`.

src/tri.rs
akern40: This was tricky, but I believe the use of saturating arithmetic means we can handle all possible cases of `k`; I've also fixed some dubious uses of `usize as isize`, so now everything is `usize` arithmetic.

src/tri.rs
akern40: No longer panicking for any values of `k`, so I think we should be ok without a comment.

src/tri.rs
akern40: This is tricky, and it's more an artifact of the implementation than a fundamental limit of these methods. The use of `Zip::indexed(self.rows())` is where the `D::Smaler: Copy` bound comes from. Fixing this should be possible, but I just have to wrap my head around the arithmetic for how to calculate row number for arbitrarily-strided arrays. It is interesting to note that this is maybe something that would be better fixed by trying to make the underlying `::indexed` call possible on `IxDyn` and then removing the bound here, rather than building out custom iteration logic in this method. Open to suggestions!

bluss: Oh, I didn't know it came from there. It kind of makes sense now, it's because of indexed. Let's not work around that here.

src/tri.rs
```diff
@@ -285,8 +340,25 @@ mod tests
...
         let res = x.triu(0);
         assert_eq!(res, array![[1, 2], [0, 4], [0, 0]]);
+
+        let res = x.tril(0);
+        assert_eq!(res, array![[1, 0], [3, 4], [5, 6]]);
+    }
+
+    #[test]
+    fn test_odd_k()
+    {
```
bluss, line 356: :+1:

src/tri.rs
```diff
@@ -30,38 +39,54 @@ where
...
-
-                res
-            }
+
+        // Performance optimization for F-order arrays.
+        // C-order array check prevents infinite recursion in edge cases like [[1]].
+        // k-size check prevents underflow when k == isize::MIN
+        let n = self.ndim();
+        if is_layout_f(&self.dim, &self.strides) && !is_layout_c(&self.dim, &self.strides) && k > isize::MIN {
+            let mut x = self.view();
```
bluss, line 67: this branch requires n >= 2 but the F && !C conditional actually ensures that in normal cases. I can't think of any exceptions right now, so it seems good.

src/tri.rs
bluss: Here is one way to work around it. It's kind of nice since it makes the index logic simpler in the inner loop.

Take index sequence like `[1, 2, 3]` and broadcast up to the space of the rows (like `[[1, 2, 3], [1, 2, 3]]` or something like that.

```diff
         let ncols = self.len_of(Axis(n - 1));
-        Zip::indexed(self.rows())
+        let index_length = self.len_of(Axis(n - 2));
+        let indexes = Array::from_iter(0..index_length);
+        let full_indexes = indexes.broadcast(self.dim.try_remove_axis(Axis(n - 1))).unwrap();
+        Zip::from(full_indexes)
+            .and(self.rows())
             .and(res.rows_mut())
-            .for_each(|i, src, mut dst| {
-                let row_num = i.into_dimension().last_elem();
+            .for_each(|&row_num, src, mut dst| {
                 let mut upper = match k >= 0 {
```

bluss (approved):
Nice improvements and fixes, lgtm if you don't want to make further changes.

Would you like to squash it down into separate clean commits? Otherwise I can integrate it with a squash merge, same thing, and that also works.

src/tri.rs
akern40: I like it and it works! Added it to the PR and removed the trait bound.

akern40:
Give me one minute to rebase so my commits are signed (switched computers and forgot to turn signing on) and then you can go ahead and squash merge.

bluss:
I'm not sure why it said it was out of sync with master, maybe because of ci.yaml changes in master. Rebased.

PR #1420, opened 2024-08-08
Update CI settings: Use PR check instead of Merge Queue, and check rustdoc

bluss:
We've used Merge Queue well but now I want to try going back to using PR check instead.
The reason is that PR check allows maintainers to choose to integrate PRs using either squash, rebase or merge.

IMO squash should be preferred when the PR is one coherent change. Merge commit or rebase should be used when the PR is well split into different commits.

I find that we want to have both options available, and that's not possible with the merge queue.


bluss:
I'm wondering what you think @adamreichold - if you have time.

bluss:
Apologies for sending a question and then merging, but I'm moving quickly with these changes (when I actually do something).  I still want to hear thoughts, even if this is already merged, it's easy to update.

PR #1424, opened 2024-08-18
Add `.shrink_to_fit()` for `Array`

bluss:
Add `.shrink_to_fit()` for owned storage arrays.
This initial version does not change array strides, this is to avoid the most tricky part of the implementation (and maybe take it step by step).

Care must be taken since `self.ptr` points inside the allocation - which for `Array` is still `Vec`-allocated, and several Vec methods including its `shrink_to_fit` can reallocate the array.


bluss:
This is a new take on #1141. I don't think this is an easy topic, lots of details to work through. That's also why this change (in current state) does not change the Array strides.

PR #1429, opened 2024-08-21
Fix uniqueness in last_mut()

### LINKED ISSUE #1427: Unsoundness in `last_mut`for `Arc`- and `CowArray`s
I believe that `last_mut` may be unsound for `ArcArray` and `CowArray`, as it neither checks nor enforces the uniqueness guarantee necessary for mutating `ArcArray`s and `CowArray`s. `last_mut`'s body on 0.16.1 is as follows:
```rust
pub fn last_mut(&mut self) -> Option<&mut A>
where S: DataMut
{
    if self.is_empty() {
        None
    } else {
        let mut index = self.raw_dim();
        for ax in 0..index.ndim() {
            index[ax] -= 1;
        }
        Some(unsafe { self.uget_mut(index) })
    }
}
```
Compare that to `first_mut`'s:
```rust
pub fn first_mut(&mut self) -> Option<&mut A>
where S: DataMut
{
    if self.is_empty() {
        None
    } else {
        Some(unsafe { &mut *self.as_mut_ptr() })
    }
}
```
They both seem to use some unsafety, which is fine, but the key is that `first_mut`'s call to `.as_mut_ptr` contains a call to `.try_ensure_unique`, while the call to `.uget_mut` does not. In fact, `uget_mut` explicitly states in its safety section that the data must be uniquely held by the array, which `last_mut` doesn't guarantee. As a result, I believe that the mutable reference of `last_mut`, if obtained from a shared `ArcArray` and actually mutated, is unsound.

bluss:
Last mut did not ensure the array was unique before calling uget_mut. The required properties were protected by a debug assertion, but a clear bug in release mode.

Adding tests that would have caught this.

Fixes #1427

### DIFF
--- a/src/impl_methods.rs
+++ b/src/impl_methods.rs
@@ -406,6 +406,7 @@ where
         if self.is_empty() {
             None
         } else {
+            self.ensure_unique();
             let mut index = self.raw_dim();
             for ax in 0..index.ndim() {
                 index[ax] -= 1;
@@ -3081,6 +3082,7 @@ mod tests
 {
     use super::*;
     use crate::arr3;
+    use defmac::defmac;
 
     #[test]
     fn test_flatten()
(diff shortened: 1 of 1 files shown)


akern40 (approved):
LGTM! Thanks for fixing this, I was going to get to it later this week but couldn't when I found the bug.

bluss:
Well found, I'm sorry for stealing it :slightly_smiling_face:

PR #1430, opened 2024-08-22
Adds vscode editor settings to gitignore

akern40:
Necessary for allowing format-on-save to use nightly for this repo only.


bluss (approved):

bluss:
PR titles will be in the changelog by default now, so I'm editing them with that in mind

PR #1432, opened 2024-09-06
Fix documentation for into_raw_vec_and_offset()

benliepert:
offset_from_alloc_to_logical_ptr returns None, not 0 when the array is empty.


akern40:
Thanks for the PR! I'm not sure this documentation change is correct; if you look at the implementation of `offset_from_alloc_to_logical_ptr` above that, it checks for emptiness and returns `None`. Did you run into a bug here? It might be in `.is_empty`, if so. Could you provide a minimal example of where this documentation is wrong or misleading?

akern40 (approved):
LGTM! Thanks for catching this.

Issue #787, opened 2020-02-26
Equivalent of numpy.diff function


espdev:
In engineering tasks there is often a need for `diff` function (calculate the n-th discrete difference along the given axis). numpy [contains](https://docs.scipy.org/doc/numpy/reference/generated/numpy.diff.html) the function that works with n-d arrays.

I have implemented such function (for n1 difference case and nd-arrays) in my experimental code:
```rust
fn diff<'a, T: 'a, D, V>(data: V, axis: Option<Axis>) -> Array<T, D>
    where T: NumOps + ScalarOperand, D: Dimension, V: AsArray<'a, T, D>
{
    let data_view = data.into();
    let axis = axis.unwrap_or(Axis(data_view.ndim() - 1));

    let head = data_view.slice_axis(axis, Slice::from(..-1));
    let tail = data_view.slice_axis(axis, Slice::from(1..));

    &tail - &head
}
```

It would be nice to have such universal function out of the box (perhaps with `n`/`prepend`/`append` parameters also).

PR #1437, opened 2024-09-25
Add `diff` method as an equivalent to `numpy.diff`

johann-cm:
I've added a `diff` method to the `ArrayBase` struct, as the issue #787 requests.
I didn't include the numpy args `prepend` and `append`, as I think it's fine to let the user do the padding by them self.


src/numeric/impl_numeric.rs
```diff
@@ -437,4 +438,42 @@ where
...
+            self.shape()[axis.0]
+        );
+
+        let mut inp = Array::zeros(self.raw_dim());
+        let mut out = self.to_owned();
+        for _ in 0..n {
+            std::mem::swap(&mut inp, &mut out);
+            let head = inp.slice_axis(axis, Slice::from(..-1));
+            let tail = inp.slice_axis(axis, Slice::from(1..));
+            out = &tail - &head;
```
nilgoyette, line 475: Can't we avoid an allocation by copying with `Zip::from(&mut out).and(&tail)...`?

tests/numeric.rs
```diff
@@ -336,3 +336,78 @@ fn std_axis_empty_axis()
     assert_eq!(v.shape(), &[2]);
     v.mapv(|x| assert!(x.is_nan()));
 }
+
+#[test]
+#[cfg(feature = "std")]
```
nilgoyette, line 341: I know you copied the tests above, but isn't this the default?  What happends when you remove it?

src/numeric/impl_numeric.rs
johann-cm: Good idea.
I have changed the layout of the method. Now I only allocate 2 arrays, `inp` and `out`,
and then shrink them gradually by slicing in-place:
```rust
        let mut inp = self.to_owned();
        let mut out = Array::zeros({
            let mut inp_dim = self.raw_dim();
            // inp_dim[axis.0] >= 1 as per the 2nd assertion.
            inp_dim[axis.0] -= 1;
            inp_dim
        });
        for _ in 0..n {
            let head = inp.slice_axis(axis, Slice::from(..-1));
            let tail = inp.slice_axis(axis, Slice::from(1..));

            azip!((o in &mut out, h in head, t in tail) *o = t.clone() - h.clone());

            // feed the output as the input to the next iteration
            std::mem::swap(&mut inp, &mut out);

            // adjust the new output array width along `axis`.
            // Current situation: width of `inp`: k, `out`: k+1
            // needed width:               `inp`: k, `out`: k-1
            // slice is possible, since k >= 1.
            out.slice_axis_inplace(axis, Slice::from(..-2));
        }

```

This of course implies, that the method will return an owned array, which owns more memory than actually needed (because of the in-place slices), but that may be worth it. What do you think?

tests/numeric.rs
johann-cm: seems to be fine without `#[cfg(feature = "std")]`

src/numeric/impl_numeric.rs
nilgoyette: I think it's acceptable. It's either that or we allocate several times (instead of 2).

It's harder to read now though :) but lets leave it like it is. Thank you!

nilgoyette:
Thank you for your contribution @johann-cm.

PR #1440, opened 2024-10-04
Adds an array reference type

### LINKED ISSUE #879: RFC: Array Reference Type
Please post your thoughts, or let me know if something is unclear/confusing!

# Summary

This RFC proposes the addition of a new storage type `RefRepr` and associated type aliases `ArrayRef<A, D>`, `ArrayRef1<A>`, `ArrayRef2<A>`, etc., such that we can implement the following traits:

- `Borrow<ArrayRef<A, D>>` for `ArrayBase<S, D> where S: Data`
- `Deref<Target = ArrayRef<A, D>>` for `ArrayBase<S, D> where S: Data`
- `BorrowMut<ArrayRef<A, D>>` for `ArrayBase<S, D> where S: DataMut`
- `DerefMut<Target = ArrayRef<A, D>>` for `ArrayBase<S, D> where S: DataMut`
- `ToOwned<Owned = Array<A, D>>` for `ArrayRef<A, D>`

`&'a ArrayRef<A, D>` would be analogous to `ArrayView<'a, A, D>`, and `&'a mut ArrayRef<A, D>` would be analogous to `ArrayViewMut<'a, A, D>`.

Many of the methods on `ArrayBase` would be moved to be implemented only on `ArrayRef` instead, and be available through deref coercion, similar to how much of the functionality of `Vec<T>` is actually implemented on `[T]`.

# Motivation

## Good way to express function parameters

Consider a function that needs a view of a 2-D array of `f64` elements. With current `ndarray`, we have the following ways to express this:

```rust
// Option 1
fn function<S>(arr: &ArrayBase<S, Ix2>)
where
    S: Data<Elem = f64>,
{}

// Option 2
fn function<'a, V: AsArray<'a, f64, Ix2>>(arr: V) {}

// Option 3
fn function<'a>(arr: impl AsArray<'a, f64, Ix2>) {}

// Option 4
fn function(arr: ArrayView2<'_, f64>) {}
```

Each of these has disadvantages:

- Options 1 and 2 are verbose, and become particularly bad for functions taking many array parameters.

- Option 4 is unnatural for callers of the function. (In most cases, they have to call `.view()` on the argument, which is verbose and is different from most Rust code, which typically uses references for this use-case.)

- Options 1, 2, and 3 are generic, so they suffer from monomorphization bloat, as discussed in the next section. For functions taking multiple array parameters, it's even worse—the number of possible type combinations is exponential in the number of array parameters. (To minimize the bloat, it's necessary to create an inner function like option 4, and make the outer function convert the parameter(s) to `ArrayView` and call the inner function.)

With this RFC, it will be possible to express the function like this:

```rust
fn function(arr: &ArrayRef2<f64>) {}
```

This is more concise than the other options, and it's not generic, so there is no monomorphization bloat.

Mutable views would be handled similarly, just with `&mut` instead of `&`.

## Substantially reduce monomorphization bloat

Most of the functionality of `ndarray` is currently implemented as methods on `ArrayBase<S, D>`. As a result, the methods are monomorphized for each new combination of `S` and `D` types, which means that there's a lot of nearly-identical generated code, i.e. "monomorphization bloat". The same issue occurs for crates which provide extension traits, such as `ndarray-linalg`, `ndarray-stats`, and `ndarray-npy`. (It's possible to minimize monomorphization bloat for a method by changing it to convert everything to `ArrayView/Mut` and then call an inner function which takes `ArrayView/Mut` parameters. However, that's very verbose and tedious.)

If, instead, we move most functionality to be implemented only on `ArrayRef<A, D>`, analogous to how much of the functionality of `Vec<T>` is actually implemented on `[T]`, then the monomorphization bloat due to differing storage types would be eliminated. (Callers would rely on deref coercion, like they do for methods actually implemented on `[T]` for `Vec<T>`.)

## Enable in-place arithmetic operator syntax for the result of a slicing operation

Currently, it's not possible to use in-place arithmetic operator syntax for the results of slicing operations; instead, it's necessary to call the appropriate method:

```rust
let mut a = array![0, 1, 2, 3, 4];

// Compiles:
use std::ops::AddAssign;
a.slice_mut(s![1..3]).add_assign(1);

// Doesn't compile:
*a.slice_mut(s![1..3]) += 1;
```

The error message is

```
error[E0614]: type `ArrayBase<ViewRepr<&mut {integer}>, Dim<[usize; 1]>>` cannot be dereferenced
  --> src/main.rs:11:5
   |
11 |     *a.slice_mut(s![1..3]) += 1;
   |    
```

By making `ArrayBase` dereference to `ArrayRef` and implementing the in-place arithmetic operators on `ArrayRef`, this will compile without errors.

## Better integration into the Rust ecosystem

The existing Rust ecosystem provides functionality dependent on traits like `Borrow`. Implementing these traits will allow `ndarray` to interact more cleanly with the rest of the ecosystem. See #878 for one example.

# Guide-level explanation

In many ways, an owned array `Array<A, D>` is similar to `Vec<A>`. It is an allocated container for a collection of elements.

The new `ArrayRef<A, D>` type for arrays is analogous to the `[A]` (i.e. slice) type for vectors. Arrays dereference to `ArrayRef<A, D>`, like `Vec<A>` derefs to `[A]`, so functionality available for `ArrayRef<A, D>` is available for other array types through deref coercion, like how functionality available for `[A]` is available for `Vec<A>` through deref coercion.

When writing a function, you should use `&ArrayRef<A, D>` or `&mut ArrayRef<A, D>` as the parameter types when possible, similar to how you'd use `&[A]` or `&mut [A]` instead of `&Vec<A>` or `&mut Vec<A>`.

In general, handle `ArrayRef<A, D>` analogously to how you'd handle `[A]`. The primary ways in which they are not analogous are the following:

- `ArrayRef<A, D>` is `Sized`, unlike `[A]`. As a result, `&ArrayRef<A, D>` is a thin pointer, while `&[A]` is a thick pointer.

- Due to technical limitations, given an array it's not possible to *directly* create an `&ArrayRef` that is a view of a portion of the array. In other words, you can create a slice (i.e. `&[A]`) of a portion of a `&[A]`/`Vec<A>` with indexing, e.g. `&var[3..12]`, but this is not possible for `&ArrayRef`. In order to create an `&ArrayRef` that references a portion of an array, you have to first create an `ArrayView` (by slicing, etc.), and then coerce that `ArrayView` to `ArrayRef`.

Also note that `&'a ArrayRef<A, D>` is similar to `ArrayView<'a, A, D>`, and `&'a mut ArrayRef<A, D>` is similar to `ArrayViewMut<'a, A, D>`, with the following exceptions:

- You don't have call a method like `.reborrow()` (for `ArrayView/Mut`) to manually shorten the lifetime of the borrow. The lifetimes of references to `ArrayRef` will be automatically adjusted as necessary, just like other references in Rust.

- The shape, strides, and pointer of an `&ArrayRef<A, D>` or `&mut ArrayRef<A, D>` cannot be modified. For example, you can't call `.slice_axis_inplace()`. To get an array with different shape/strides/pointer, you have to create a view. For example, you could call `.slice_axis()`, or you could create a view with `.view()` and then call `.slice_axis_inplace()` on the view.

# Implementation

## Implementing `Deref` and `DerefMut` with `Target = ArrayRef<A, D>`

The new `RefRepr` storage type will be:

```rust
#[repr(C)]
pub struct RefRepr<A> {
    elem: PhantomData<A>,
}
```

The `ArrayBase` struct will be changed to the following:

```rust
#[repr(C)]
pub struct ArrayBase<S, D>
where
    S: RawData,
{
    dim: D,
    strides: D,
    ptr: NonNull<S::Elem>,
    data: S,
}
```

The important parts of this change are:

- The `dim`, `strides`, and `ptr` fields are the first fields in the struct.

- The struct is declared `repr(C)`.

These changes are necessary so that it's possible to implement `Deref` and `DerefMut` by casting references to `ArrayBase` into references to `ArrayRef`.

@bluss pointed out below that it's necessary to avoid infinite recursion in deref coercion. To do this, a `DataDeref` trait will be added:

```rust
pub unsafe trait DataDeref: Data {}
```

This trait will be implemented by all storage types except for `RefRepr` and will be used as a bound for the `Deref` and `DerefMut` implementations.

`Deref` and `DerefMut` will be implemented like this:

```rust
impl<A, S, D> Deref for ArrayBase<S, D>
where
    S: Data<Elem = A> + DataDeref,
{
    type Target = ArrayRef<A, D>;

    fn deref(&self) -> &ArrayRef<A, D> {
        unsafe {
            &*(self as *const Self).cast::<ArrayRef<A, D>>()
        }
    }
}

impl<A, S, D> DerefMut for ArrayBase<S, D>
where
    S: DataMut<Elem = A> + DataDeref,
{
    fn deref_mut(&mut self) -> &mut ArrayRef<A, D> {
        self.ensure_unique();
        unsafe {
            &mut *(self as *mut Self).cast::<ArrayRef<A, D>>()
        }
    }
}
```

## Moving most existing methods to be implemented only for `ArrayRef<A, D>`

Most of the existing methods which don't modify the shape/strides/pointer (such as `slice`, `mapv`, etc.) will no longer be implemented for all `ArrayBase<S, D>`, but rather will be implemented only on `ArrayRef<A, D>`. This change isn't necessary in order to introduce the `ArrayRef` type, but it is necessary in order to reduce the existing monomorphization bloat, which is one of the primary benefits of the `ArrayRef` type.

## Data traits for `RefRepr`

The `RefRepr` type will implement the following traits:

- `RawData`
- `RawDataMut`
- `Data` (but make the `into_owned` implementation panic with `unreachable!()`)
- `DataMut`
- `RawDataSubs`

In other words, the `&'a` or `&'a mut` in a reference to an `ArrayRef` will control the lifetime and mutability of the data.

We need to audit all of the available methods to make sure that:

- it's possible to create an `&'a ArrayRef<A, D>` only for a base array with `S: Data` that lives at least as long as `'a`,
- it's possible to create an `&'a mut ArrayRef<A, D>` only for a base array with `S: DataMut` that lives at least as long as `'a`, and
- it's not possible to create an `ArrayRef<A, D>`

(I believe this is already true if we just add the `Deref`, `DerefMut`, `Borrow`, and `BorrowMut` implementations, but we should check.)

## Preventing modification of shape/strides/pointer via `&mut ArrayRef<A, D>`

To minimize confusion and behave analogously to `&mut [A]`, it should not be possible to modify the underlying shape/strides/pointer through a `&mut ArrayRef<A, D>`. For example,

```rust
let mut owned: Array2<f32> = Array::<f32>::zeros((4, 5));
let mut borrowed: &mut ArrayRef<A, D> = &mut owned;
// should not compile since it would modify the shape/strides/pointer:
borrowed.slice_axis_inplace(Axis(0), ..);
```

To achieve this:

- A new `MetaMut` trait will be added to control whether the shape/strides/pointer can be modified.

- `MetaMut` will be implemented for all of the existing storage types (`OwnedRepr`, `ViewRepr`, etc.), but not the new `RefRepr` type.

- A `S: MetaMut` bound will be added to all methods which modify the shape/strides/pointer in-place (such as `slice_collapse`, `slice_axis_inplace`, `collapse_axis`, `swap_axes`, `invert_axis`, `merge_axes`, `insert_axis_inplace`, and `index_axis_inplace`) .

# Drawbacks

This proposal introduces yet another storage type, when `ndarray` already has quite a few.

The similarity of `&'a ArrayRef<A, D>` to `ArrayView<'a, A, D>` and `&'a mut ArrayRef<A, D>` to `ArrayViewMut<'a, A, D>` could be confusing to new users.

The requirement that it must be possible to create `ArrayRef` references by reinterpreting the bits in existing `ArrayBase` instances limits the possible field layouts of `ArrayBase`.

# Rationale and alternatives

## Allow mutation of the shape/strides/pointer for `&mut ArrayRef<A, D>`

It would be possible to avoid the new `MetaMut` trait by allowing modification of the shape/strides/pointer through `&mut ArrayRef<A, D>` instances. However, this would likely be confusing to users who are familiar with how `&mut [T]`/`Vec<T>` behave, and it would increase the potential for bugs.

## Keep most method implementations on all `ArrayBase<S, D>`

It would be possible to keep the existing method implementations for all `ArrayBase<S, D>` instead of moving them to be implemented only on `ArrayRef<A, D>`. However, this would not bring the reduced monomorphization bloat which is one of the advantages of the new `ArrayRef` type.

## Make `ArrayRef` be a separate struct from `ArrayBase`

It would be possible for `ArrayRef` to be a separate struct from `ArrayBase`, instead of an alias of `ArrayBase`. This would have the following advantages:

- `Deref`, `DerefMut`, `Borrow`, and `BorrowMut` could be implemented for `ArrayBase` without `unsafe` code.
- The `MetaMut` and `DataDeref` traits wouldn't be necessary.
- There's less potential for unwanted overlapping or recursive trait impls.
- Keeping the types separate may be less confusing for new users.

The disadvantages of a separate struct type are:

- The transition to `ArrayRef` would be more disruptive. Until libraries update their functions to take `&ArrayRef` or `&mut ArrayRef` parameters instead of `&ArrayBase` or `&mut ArrayBase` parameters, `ArrayRef` would not be compatible with those functions. (Users of those libraries would likely have to call `.view()`/`.view_mut()` in some places.)

- It may be less convenient for generic code which wants to operate on all array types.
  
  For example, if `ArrayRef` is *not* a separate struct, then these impls are sufficient for `AddAssign`:

  ```rust
  impl<'b, A, B, S2, D1, D2> AddAssign<&'b ArrayBase<S2, D2>> for ArrayRef<A, D1>
  where
      A: AddAssign<&'b B>,
      B: 'b,
      S2: Data<Elem = B>,
      D1: Dimension,
      D2: Dimension,
  {}

  impl<A, B, D> AddAssign<B> for ArrayRef<A, D>
  where
      A: AddAssign<B>,
      B: ScalarOperand,
      D: Dimension,
  {}
  ```

  (Note that, unfortunately, we can't replace `impl<'b, A, B, S2, D1, D2> AddAssign<&'b ArrayBase<S2, D2>> for ArrayRef<A, D1>` with `impl<'b, A, B, D1, D2> AddAssign<&'b ArrayRef<B, D2>> for ArrayRef<A, D1>`. For some reason, the compiler has trouble figuring out which impl to use when `impl<A, B, D> AddAssign<B> for ArrayRef<A, D>` also exists.)

  However, if `ArrayRef` is a separate struct, then it would be necessary to add another impl:

  ```rust
  impl<'b, A, B, D1, D2> AddAssign<&'b ArrayRef<B, D2>> for ArrayRef<A, D1>
  where
      A: AddAssign<&'b B>,
      B: 'b,
      D1: Dimension,
      D2: Dimension,
  {}
  ```

## Custom DST (i.e. custom thick pointer types)

Instead of using thin pointers to `ArrayRef`, we could wait for custom DSTs (e.g. [RFC 2594](https://github.com/rust-lang/rfcs/pull/2594)) to be added to Rust, and then create a custom DST similar to `ArrayRef`. This would have the following advantages:

- Since the shape/strides/pointer would be stored within the thick pointer, it would be possible to modify the shape/strides/pointer in the thick pointer in-place, unlike for a reference to an `ArrayRef`.

- Slicing methods could return a thick pointer instead of an `ArrayView/Mut` except in the `IxDyn` case.

- *Raw* thick pointers (i.e. `*const`/`*mut`) would be more useful. In other words, since the shape/strides would be stored within the thick pointer itself, it should be possible to access that information without dereferencing the pointer. So, except in the `IxDyn` case, thick pointers could replace `RawArrayView/Mut`. In contrast, given a `*const ArrayRef<A, D>` or `*mut ArrayRef<A, D>`, you have to dereference the pointer to access the shape/strides, so `ArrayRef` cannot replace most uses of `RawArrayView/Mut`.

However, I'm not aware of any Rust RFC for custom DSTs which would make it possible to create a custom DST with support for the dynamic-dimensional (`IxDyn`) case. So, there would be this weird dichotomy between the convenience of thick pointers for the const-dimensional case and the complexity of the dynamic-dimensional case. In contrast, the `ArrayRef` type proposed in this RFC would support the const-dimensional and dynamic-dimensional cases in a uniform way.

# Future possibilities

One thing I wanted to make sure when writing this proposal was that it would still allow replacing the `ArrayBase` struct with a trait-based API in the future. In other words, if we had

```rust
struct Array<A, D> { ... }

struct ArrayView<'a, A, D> { ... }

// ...

trait NdArray { ... }

trait NdArrayMut { ... }

// ...
```

instead of having all arrays be instances of `ArrayBase`, it would be nice for this proposal to still apply. This proposal would, in fact, be compatible with a trait-based API, and the implementation could be somewhat cleaner because it wouldn't require any `unsafe` code. I'd suggest implementing the types like this:

```rust
struct Array<A, D> {
    meta: ArrayRef<A, D>,
    data: Vec<A>,
}

struct ArrayView<'a, A, D> {
    meta: ArrayRef<A, D>,
    life: PhantomData<&'a A>,
}

struct ArrayRef<A, D> {
    shape: D,
    strides: D,
    ptr: NonNull<A>,
}

impl<A, D> Deref for Array<A, D> {
    type Target = ArrayRef<A, D>;

    fn deref(&self) -> &ArrayRef<A, D> {
        &self.meta
    }
}

// ...
```

akern40:
This PR would add an array reference type into `ndarray`. This type's relationship to `ArrayBase` is analogous to the relationship between `[T]` and `Vec<T>`. Closes #879; look there for more information on the motivation of this type.

EDITED: See below

~This implementation is chosen to make as few breaking changes as possible, introducing only one: array views are no longer `Copy`. This is a break that the compiler will immediately error on, and can be fixed relatively easily with calls to `.clone()`. As a result, the behavior is functionally equivalent, but the calling convention has to change. This change must happen because the concept of a reference type cannot implement `Clone` (since reference types must only be available via the `Deref` traits as `&RawRef` or `&ArrRef`); as a result, it cannot be `Copy`, and neither can any variant of `ArrayBase`.~

~I had originally tried to implement `RawRef` and `ArrRef` as two different types, and it's worth discussing why I didn't. The goal there was to further reduce monomorphization bloat (the original motivation for the RFC mentioned above, although I don't think it's the most important reason) and to improve ergonomics of the reference type. However, this would fundamentally require breaking `ArrayBase` into two variants: raw views and everything else, since a single type cannot implement `Deref` twice (trust me, I tried). If [`generic_const_exprs`](https://github.com/rust-lang/rust/issues/76560) or [`implement_specialization`](https://rust-lang.github.io/rfcs/1210-impl-specialization.html) land, then this approach is possible and perhaps preferred.~

~Right now, this PR just introduces the type and some documentation. It does not do the logical next step, which is moving most of `ndarray`'s functionality into the `RefBase` type. This would accomplish the goal of creating a single way for users to write functions that accept arrays of all kinds. Neither does it add the functionality necessary for `AddAssign` and its variants, which is also an important motivator for this change.~


bluss:
Copyable-ArrayViews :open_mouth: nice!

akern40:
Early on I'd been trying to avoid the "layout-compatible structs" approach because I was worried about maintainability. But once I realized we could have 0 breaking changes at the cost of a maintenance concern, I went back and changed it. It helps that as long as nobody touches the internals of `ArrayBase` or `RefBase`, we should be good.

akern40:
Ok, for real this time: add array types (plural!) and change the entire library to use them.

There is a lot to write about this PR; frankly, I could make a blog post about it, and maybe we should. Certainly, the changelog is going to need to be extensive. I guess this is the first cut at that.

This PR still needs significant documentation work before it is ready to merge; however, the vast majority of the implementation is now complete.

# What's Changed?
This PR introduces three new types: `ArrayRef`, `RawRef`, and `LayoutRef`. Most functionality has been moved to those types, and users are encouraged to switch to using them as soon as possible.

## `ArrayRef`
As the original RFC suggested, this type is the `Deref` target of `ArrayBase` when `S: Data`; in other words, it represents a look into an array whose data is safe to read. In addition, the `DerefMut` implementation guarantees that any user with `&mut ArrayRef<...>` holds unique, mutable access to the underlying data.

Unlike the original RFC, this implementation allows users to modify the shape and strides (although not dimensionality) of an array by using `ArrayRef`; this capability actually extends all the way down to `LayoutRef`. Practically, separating out data mutation from shape/stride mutation leads to a tricky problem for users: they'd have to juggle between functions that take `ArrayRef` and functions that take `ArrayBase`. Because you can't call `ArrayBase` functions from `ArrayRef` functions, users will be forced to frequently use `views` and write functions with two different paradigms.

## `RawRef`
This type represents a look into an array whose data is not safe to read. It is the `Deref(Mut)` target of `ArrayRef`.

Due to limitations of Rust's compiler, a `RawRef` is *not* the `Deref(Mut)` target of `RawArrayView(Mut)`. Those types have no `Deref` implementation. As a result, getting a `RawRef` from a `RawArrayView(Mut)` must be done using `view.as_ref()`.

## `LayoutRef`
This type represents read and write access to an array's *layout*, but never its data. It is the `Deref(Mut)` target of `RawRef`. Due to the aforementioned compiler limitations, `LayoutRef`s can only be obtained from `RawArrayView(Mut)`s via `view.as_ref()`.

This type is the home to very basic methods such as `is_empty()` as well as in-place slicing methods such as `slice_axis_inplace`. **Maintainers of `ndarray` must guarantee going forward that `LayoutRef` never leaks its underlying data**. Although `RawRef` and `ArrayRef` are neither `Clone` nor `Copy`, `LayoutRef` implements both; as long as data isn't leaked, this is safe.

## When to Use Each Type
When writing a function or trait implementation, users can use the following table to decide what type they want to use:

| Element Deref Safety| Read elements | Mutate elements | Read shape / strides | Mutate shape / strides / pointer, but not data |
| -- | -- | -- | -- | -- |
| Safe | `&ArrRef` | `&mut ArrRef` | `&LayoutRef` | `&mut LayoutRef` |
| Unsafe | `&RawRef` | `&mut RawRef` | `&LayoutRef` | `&mut LayoutRef` |

If users need to a) alter the underlying buffer size of an array (e.g., shrink, reserve, append, etc), b) split an array into multiple pieces, or c) move data, they will need to take `ArrayBase` or `ArrayView`.

## `Deref` and `AsRef` Implementations
To connect these types, we use a series of `Deref` and `AsRef` implementations, captured as follows:
```
                 ┌─────────┐                   
              ┌──┼ArrayBase│                   
              │  └────┬────┘                   
              │       │Deref when S: Data      
AsRef         │       │DerefMut when S: DataMut
when          │       │                        
S: RawData    │  ┌────▼────┐                   
              │  │ArrayRef ┼───────┐           
AsMut         │  └────┬────┘       │           
when          │       │Deref       │AsRef      
S: RawDataMut │       │DerefMut    │AsMut      
              │       │            │           
              │  ┌────▼────┐       │           
              ├──►RawRef   ◄───────┤           
              │  └────┬────┘       │           
        AsRef │       │ Deref      │AsRef      
        AsMut │       │ DerefMut   │AsMut      
              │       │            │           
              │  ┌────▼────┐       │           
              └──►LayoutRef◄───────┘           
                 └─────────┘                   
```

The result of this is that `ArrayBase<S: RawData, D>`, `RawArrayView`, or `RawArrayViewMut` can only access methods on `RawRef` and `LayoutRef` via the `AsRef` trait.

## Writing Functions on `RawRef` and `LayoutRef`
Users who want to write functions on `RawRef` and `LayoutRef` must make a choice that affects how callers with (possibly) raw arrays will interact with their functions. They can either
1. Write their functions or traits directly on `RawRef` or `LayoutRef`, letting users call `.as_ref()` as needed or
2. Write their functions or traits on a generic `T where T: AsRef<RawRef<A, D>>` (and analogously for `LayoutRef`). It's a slightly more cumbersome way to write functions, but is more ergonomic for users of their new functionality.

This tradeoff is not relevant for callers who hold `ArrayBase<S: Data, D>` or any functionality written for `ArrayRef`.

# Future Designs
This limitations imposed on interactions with raw views, etc, can be fixed if the [Trait specialization RFC](https://rust-lang.github.io/rfcs/1210-impl-specialization.html) lands.

bluss:
Hm, what does this mean "Maintainers of ndarray must guarantee going forward that LayoutRef never leaks its underlying data."

Avoiding memory leaks in general is hard and not part of memory safety, but I'm not sure what this means, looking around in the PR to try to find out.

You're getting very far ahead - that's exciting, but this is a lot to take in one bite or one PR! Plan for new types - AsRef vs Deref - backwards compat situation - should this be discussed?

Why is AsRef needed? From the diagram it looks like deref coercion should work. I.e &ArrayRef will coerce to &LayoutRef, I thought.

akern40:
I know this is a big PR, and I'd be happy to break it up into smaller chunks that can be more easily reviewed and discussed! I did all of it because I kept running into design problems when I went to actually implement functionality. I think I've settled onto a very defensible design, but we should discuss and potentially break up the PR now that I feel more confident about the choices.

Leak is the wrong word here, that's my bad. I meant "leak" in the sense of an API boundary: a `LayoutRef` contains a pointer to data, which it must have to support slicing (the pointer may move into the interior of the array or view). But that pointer should never be *exposed* to the user.

On the note of backwards compatibility: I've tried very hard to design this such that it is fully backwards compatible. I think I achieved about 99% of that. Breaking this up into smaller PRs would let us identify where any incompatibilities may have snuck in.

On the note of new types, while I'm fairly confident about the need for `LayoutRef`, I'm actually on the fence about keeping `RawRef` (another reason I implemented everything: I wanted to figure out what use `RawRef` really had). Turns out, not a lot! These are the methods that `ndarray` provides that previously operated on `ArrayBase<S: RawData(Mut), D>` (i.e., not including those operations explicitly on raw *views*):
- `as_mut_ptr`
- `as_ptr`
- `get_mut_ptr`
- `get_ptr`
- `raw_view`
- `raw_view_mut`

Removing `RawRef` would simplify the types, and we could simply implement the above functionality directly on `ArrayRef` and `ArrayBase` (you'd have to do it both places for backwards compatibility because these functions don't guarantee the unique hold of their underlying data, which always happens when you have `ArrayRef`). The question is whether we want a type that can unsafely access data without guaranteeing that data is a) safe to read or b) uniquely held.

The `AsRef` is there for two reasons. The first is performance on shared arrays. Functionality implemented on `LayoutRef` does not need to guarantee the uniqueness of the underlying data, a potentially expensive operation. Relying on the deref chain from shared arrays down to `LayoutRef` will go through `ArrayRef`, thereby triggering un-sharing. In contrast, the `AsRef` implementation just exposes the underlying `layout` without guaranteeing data uniqueness. When users write functions with `T: AsRef<LayoutRef>>`, they allow callers to pass in shared arrays without triggering that deref path.
Unfortunately, this one performance necessity triggers a cascade of `AsRef`: once we ask users to write `LayoutRef` functionality using `AsRef`, you have to implement `AsRef` everywhere. [This playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=e91dd96aab51bef161b2ef6792bce432) illustrates why: Rust won't auto-deref to satisfy a generic trait bound.

 The second reason is to imbue raw views with the functionality implemented on `LayoutRef`. They don't get to be part of the `Deref` chain (I tried so many ways, I can't figure out how), so they're essentially "cut off" from the main stream of functionality. By using `AsRef`, you also imbue raw arrays with the capability that the rest of the arrays get through the `Deref` chain.

akern40:
I believe this PR is now fully ready for review, with the exception of documenting `RawRef` (since we should decide whether to keep it or get rid of it). @bluss do you want me to break it up into smaller PRs? I think I could do the following:
1. PR that provides the types, the documentation, and their various `Deref` and `AsRef` ties
2. PR that moves relevant methods to `LayoutRef`
3. PR that moves relevant methods to `RawRef` (if we decide to keep it)
4. Various PRs that move the remaining methods to `ArrayRef`

Or we can just leave it all as one. Let me know.

Since this is such a big change, I'll also ping @nilgoyette, @adamreichold, @jturner314, and @termoshtt; I know everyone is busy, but I really want to avoid (even the perception of) unilaterally changing the library's internals without a larger maintainer consensus. Feel free to look, skim, criticize, question, etc.

bluss:
I'm impressed by this work.

No tests even have to change, best possible backwards compatibility. Surely something must be a breaking change, some methods have been given new homes under different types (?)

The diff is big but most of the changes are repetitive and can be ignored, this is probably ok to review as it is. That's my initial feeling

akern40:
Thanks! While it's true that methods have new homes, the deref structure means that most things are callable from `ArrayBase`. The exception there is methods that have been re-homed to `RawRef` or `LayoutRef`. For those, the file `alias_asref.rs` provides shadowing functions on `ArrayBase` that use the `AsRef` / `AsMut` to skip the uniqueness guarantee, thereby maintaining the functional backwards-compatibility. If there's a breaking change, it's probably there: I should double-check that I've added all of those aliases.

akern40:
I should update the `serde` serialization. We can fold it in here or just do a follow-up PR, but either way we should make sure to include it before the next release.

akern40:
> I should double-check that I've added all of those aliases

Done! I was missing a few.

Also added documentation to `RawRef`, deciding to keep it. By keeping it, we duplicate the method signatures and docs twice (in `RawRef` and in `ArrayBase`) and the body once (in `RawRef`). If we got rid of it, we'd have to duplicate the body as well.

tests/array.rs
```diff
@@ -2819,3 +2819,11 @@ fn test_split_complex_invert_axis()
     assert_eq!(cmplx.re, a.mapv(|z| z.re));
     assert_eq!(cmplx.im, a.mapv(|z| z.im));
 }
+
+#[test]
+fn test_slice_assing()
+{
+    let mut a = array![0, 1, 2, 3, 4];
+    *a.slice_mut(s![1..3]) += 1;
```
nilgoyette, line 2827: Wasn't this forbidden before?

examples/functions_and_traits.rs
```diff
@@ -0,0 +1,178 @@
+//! Examples of how to write functions and traits that operate on `ndarray` types.
```
nilgoyette, line 1: This runs only when we run specifically ask for it with `cargo run --example ...`? Shouldn't it always run when we run the tests? Or it's already safe because the rest of the codebase uses everything here and more?

crates/blas-tests/tests/oper.rs
```diff
@@ -280,7 +280,7 @@ fn gen_mat_mul()
                         cv = c.view_mut();
                     }
 
-                    let answer_part = alpha * reference_mat_mul(&av, &bv) + beta * &cv;
+                    let answer_part: Array<f64, _> = alpha * reference_mat_mul(&av, &bv) + beta * &cv;
```
nilgoyette, line 283: Is this change actually needed?

nilgoyette (commented):
A big "Thank you" for taking the time and effort to do this! This is kinda crazy that it will be invisible for the users.

As I wrote earlier, I don't know much about the internals, so my review is mainly questions, not criticism.

tests/array.rs
akern40: Impossible, in fact! That's one of the big motivators for the change. `slice_mut` provides an `ArrayViewMut`, which couldn't be dereferenced. Now that `ArrayRef` exists, this operation is valid.

examples/functions_and_traits.rs
akern40: I think it's fine - I originally had it in tests, but I realized it was more helpful to act as an example. Our CI/CD still runs the examples, so it'll be "checked" in that sense even if it's not a test.

crates/blas-tests/tests/oper.rs
akern40: Let me double check this later, it might not be. In terms of breaking changes, this is one kinda-uncontrollable issue: Rust's type inference may run into some issues. But perhaps this is unnecessary, or otherwise something we can fix.

akern40: Ah fascinating; yes, it does need it. If you don't include it, you get the following compiler error:
```shell
error[E0271]: type mismatch resolving `<ArrayBase<OwnedRepr<f64>, Dim<[usize; 2]>> as Add>::Output == ArrayRef<f64, _>`
   --> crates/blas-tests/tests/oper.rs:283:75
    |
283 |                     let answer_part = alpha * reference_mat_mul(&av, &bv) + beta * &cv;
    |                                                                           ^ expected `ArrayBase<OwnedRepr<f64>, Dim<[usize; 2]>>`, found `ArrayRef<f64, _>`
    |
    = note: expected struct `ndarray::ArrayBase<OwnedRepr<f64>, Dim<[usize; 2]>>`
               found struct `ArrayRef<f64, _>`
```
which I was initially confused by. The key is that the following line:
```rust
answer.slice_mut(s![..;s1, ..;s2]).assign(&answer_part);
```
calls `.assign`, which now expects `&ArrayRef<_, _>`. So the compiler tries to infer the type of `answer_part` as `ArrayRef<_, _>`, which isn't possible. Fascinating look into Rust's type inference. I feel like it should be able to infer the `Array` type since it knows that both both `cv` and the output of `reference_mat_mul` are both `Array`, but I guess not.

tests/array.rs
```diff
@@ -2819,3 +2819,11 @@ fn test_split_complex_invert_axis()
     assert_eq!(cmplx.re, a.mapv(|z| z.re));
     assert_eq!(cmplx.im, a.mapv(|z| z.im));
 }
+
+#[test]
+fn test_slice_assing()
```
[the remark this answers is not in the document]
akern40: thanks for catchign this 😁

src/alias_asref.rs
```diff
@@ -0,0 +1,310 @@
...
+    /// Return true if the array is known to be contiguous.
+    pub(crate) fn is_contiguous(&self) -> bool
+    {
+        <Self as AsRef<LayoutRef<_, _>>>::as_ref(self).is_contiguous()
+    }
+
+    /// Return an iterator over the length and stride of each axis.
+    pub fn axes(&self) -> Axes<'_, D>
+    {
+        <Self as AsRef<LayoutRef<_, _>>>::as_ref(self).axes()
```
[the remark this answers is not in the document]
akern40: The `AsRef` / `AsMut` aliasing functions serve two purposes. First off, a function like `axes` that got moved to `LayoutRef` would not longer be directly available on a raw view - i.e., `raw_view.axes()` would be an error - because raw views do not dereference. (They would still exist on any array with `S: Data`, though). So these aliases restore that capability to raw views.

The second is a performance consideration, only relevant for mutating functions on shared arrays. Functions defined on `LayoutRef` and `RawRef` *are* directly available to shared arrays, even without these aliases. However, they do that by going through `&mut ArrayRef`, which un-shares the data. But we don't want that in these cases; `LayoutRef` and `RawRef` don't require unique data. So these aliases shadow these mutating functions and create a method that doesn't un-share data.

src/alias_asref.rs
```diff
@@ -0,0 +1,310 @@
...
+    /// - if an index is out of bounds
+    /// - if a step size is zero
+    /// - if [`NewAxis`](`crate::SliceInfoElem::NewAxis`) is in `info`, e.g. if `NewAxis` was
+    ///   used in the [`s!`] macro
+    /// - if `D` is `IxDyn` and `info` does not match the number of array axes
+    #[track_caller]
+    pub fn slice_collapse<I>(&mut self, info: I)
+    where I: SliceArg<D>
+    {
+        self.as_mut().slice_collapse(info);
```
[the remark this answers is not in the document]
akern40: I think this is a great call. I've added the functions publicly (I think our users will want that stability and unambiguity as well) and replaced our usage in the crate.

akern40:
> I think that we should treat this change as worthy of the 'breaking-changes' label, unless we have very good evidence it's almost entirely free of existing code breakage.

Ya, as much as I've done my best to avoid breaking changes, I think Rust's type inference will introduce errors, if nothing else. I agree we should likely treat this as a bump to 0.17.

However, I think we could hold off on the release for a bit. We've got some other PRs we're trying to clean up, so we can roll those into a 0.17. As I've said on Zulip, I want to take a crack at the shape / stride problem, too. If I can make some headway, would that be good to roll into 0.17 as well? Or should we just aim for one major crate redesign per version?
- If we wait, users only have to deal with a version bump once, not twice
- If we do it in two releases, users don't have to deal with learning two new major redesigns at once

akern40:
@bluss did you have any final follow-up on this PR? Because this changes the paradigm of writing functions in ndarray, I have been building any additional PRs on top of it (otherwise, I would have to write a PR for `ArrayBase` and then alter this PR to provide the same functionality for `ArrayRef`). I've accumulated a bit of a backlog. If you're good merging this into `master` then I can bang out quite a few PRs that are lying around.

akern40:
We should reconsider the naming of `LayoutRef` - as I work through a shape/stride/dimensionality redesign, I think that the term `LayoutRef` to refer to a joint package of shape, strides, and pointer is not going to be ideal.

Issue #1455, opened 2024-11-22
RFC: MSRV Policy


akern40:
A [recent change to `url`](https://github.com/servo/rust-url/issues/937) has caused `openblas-src`'s MSRV to bump to 1.67 (when using most recent dependencies). While this doesn't affect `ndarray` directly, it does affect the clarity and interoperability of `ndarray` with a major BLAS provider and raises an important question: should `ndarray` have an MSRV policy and, if yes, what should that policy be?

This question breaks down into three parts:
- Should `ndarray` _promise_ a certain MSRV? (This means listing a `rust-version` in the `Cargo.toml` _and_ pinning any necessary dependencies in patch versions to ensure that `rust-version` is maintained)
- If `ndarray` does promise an MSRV, what should be the policy for which MSRV is guaranteed? For example, two years (the approximate MSRV right now) vs two versions.
- When `ndarray` bumps its MSRV, is that a patch update, or a minor version update?

I would like to solicit feedback on this topic from users of `ndarray`. If you use `ndarray`, what is your usual Rust configuration? Are you running on the most recent version, or something older?

MSRV and its association with versioning is a much, much larger conversation. If possible, I would like to keep any conversation to just what `ndarray`'s policy should be; not what the wider ecosystem does.

My personal preference is that `ndarray` _does_ promise a _recent_ MSRV, and that updates to that MSRV occur only in minor version updates.

bluss:
Current docs say 'MSRV: Requires Rust 1.64 or later', which I thought has to be mentioned.

akern40:
Absolutely - when I talk about a "policy", I'm just asking about how and when we decide to bump the MSRV. How have you made this decision in the past?

PR #1456, opened 2024-11-29
Change CI to handle openblas MSRV compatibility

akern40:
Changes to BLIS for testing MSRV, since openblas is currently MSRV 1.67 and incompatible with ndarray's MSRV of 1.64


README.rst
```diff
@@ -156,6 +156,14 @@ there is no tight coupling to the ``blas-src`` version, so version selection is
 0.13        0.2.0        0.6.0
 =========== ============ ================ ==============
 
+------------
+BLAS on MSRV
+------------
+
+Although ``ndarray`` currently maintains an MSRV of 1.64.0, this is separate from the MSRV (either stated or real) of the various BLAS providers.
+As of the time of writing, ``openblas`` currently supports MSRV of 1.67.0.
+So, while ``ndarray`` and ``openblas-src`` are compatible, they can only work together with toolchains 1.67.0 or above.
```
bluss, line 165: I wonder if ndarray can just use an older blas-src and older openblas-src for earlier Rust. I think that's the case, but I'm not sure we need to use it in CI. Thanks for fixing this

bluss (approved):

README.rst
akern40: No problem - I'll try seeing if there's an older openblas-src that we can use and add it to the documentation. But this all was the motivation for #1455, since I was wondering whether we want to bump to 1.67 (in order to make it easier for users to use BLAS)

akern40:
Merged as part of #1458

PR #1460, opened 2024-12-04
add axis_windows_with_stride() method

goertzenator:
This adds a variation of `axis_windows()` that accepts a stride parameter.  It is analogous to the `windows_with_stride()` variant of `windows()`.  This was already 95% implemented thanks to the work of @jonasBoss in commit 2aec819.

Ref issue #1304.

### DIFF
--- a/src/iterators/windows.rs
+++ b/src/iterators/windows.rs
@@ -141,7 +141,7 @@ pub struct AxisWindows<'a, A, D>
 
 impl<'a, A, D: Dimension> AxisWindows<'a, A, D>
 {
-    pub(crate) fn new(a: ArrayView<'a, A, D>, axis: Axis, window_size: usize) -> Self
+    pub(crate) fn new_with_stride(a: ArrayView<'a, A, D>, axis: Axis, window_size: usize, stride_size: usize) -> Self
     {
         let window_strides = a.strides.clone();
         let axis_idx = axis.index();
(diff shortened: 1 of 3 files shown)


src/iterators/windows.rs
```diff
@@ -141,7 +141,7 @@ pub struct AxisWindows<'a, A, D>
 
 impl<'a, A, D: Dimension> AxisWindows<'a, A, D>
 {
-    pub(crate) fn new(a: ArrayView<'a, A, D>, axis: Axis, window_size: usize) -> Self
+    pub(crate) fn new_with_stride(a: ArrayView<'a, A, D>, axis: Axis, window_size: usize, stride_size: usize) -> Self
```
akern40, line 144: Any reason not to keep the name `new`?

akern40 (commented):
Thanks for the PR! This is looking good, and great testing. Just the one question about keeping the name `new`. If you could run `rustfmt` (which must be run on `+nightly`) that would also be good.

src/iterators/windows.rs
goertzenator: I was following the naming precedent set by `Window` which has both `new()` and `new_with_stride()`.  I know calling a sole constructor function something other than `new()` is odd, but in this case making an association with what happens in `Window` is useful.

goertzenator:
`cargo fmt` from nightly has been applied to the PR.

src/iterators/windows.rs
akern40: Sure, I guess that's fair. It's an internal function anyway, so I'm not picky about it. Thanks for the explanation!

akern40:
Let me fix our CI/CD and then this should be good to merge

akern40:
@goertzenator sorry I dropped the ball on this! Would you mind please running `git commit --amend --no-edit` then `git push --force-with-lease` on your branch? That should reset the CI which should go through and we can merge!

PR #1468, opened 2024-12-23
Generalize ref-to-ref ops

akern40:
The current reference-to-reference operations defined in `impl_ops.rs` only allow for outputs where the output type is the same as the left-hand operand. This makes sense for reference-to-owned, where we can reuse an allocation, but not for reference-to-reference.

This is a follow on to #1440, and so cannot be merged until then.


akern40:
This is also blocked until the inclusion of the next-gen solver, see rust-lang/rust#80542

Issue #1462, opened 2024-12-05
BaseArray::abs(&self) isn't implemented for integers (isize, i64, etc)


Sh3mm:
The `abs(&self)` method in `BaseArray` only works for float arrays since it's defined as part of the "Element-wise methods for float arrays" section.  

However, in my opinion, we should also be able to get the absolute version of negative integers arrays.

akern40:
So I think the general approach up to now has been that this can be accomplished via `arr.mapv(num_traits::abs)` or `arr.mapv(i64::abs)` if you want have / want more specificity. I would maybe like to see a sort of parallel implementation of `num`-like traits for arrays, but this gets back to the problem of how many methods are in `ndarray`.

Sh3mm:
If I'm not mistaken, I believe  that's the exact way it's being done for the floats using the `unary_ops` macro. Maybe some of the methods such as `abs`and `signum` could be moved to a different section/macro that would treat both cases?

```rust
#[cfg(feature = "std")]
macro_rules! unary_ops {
    ($($(#[$meta:meta])* fn $id:ident)+) => {
        $($(#[$meta])*
        #[must_use = "method returns a new array and does not mutate the original value"]
        pub fn $id(&self) -> Array<A, D> {
            self.mapv(A::$id)
        })+
    };
}
```

What's bugging me is the issue of partial functionality. Why should this method only work with floats when it affects negative numbers as a whole?

If it the functionality is not added to integers, we should maybe update the documentation to make it clearer since, although the `abs` method is in the _Element-wise methods for float arrays_ section, it is far enough into it that the section name does not appear on screen when navigating through the side menu. This makes it unclear that it only applies to floats when quickly browsing the docs.

akern40:
Honestly I agree, and would welcome a PR that documents that. If you're up for it, what I'd love to see is a PR that modifies the implementation of element-wise mathematics to mimic `num`, so that if a type is `num_traits::Signed` it will have `abs` and `signum`.

I have some code laying around that starts that more comprehensive PR, but feel that I've got a few items I've got to check off first.

akern40:
@Sh3mm you've inspired me to go back and take a look at that code. I think it's realistic for me to implement the more comprehensive PR. However, I'd like for the methods with additional arguments to accept both scalars and arrays, and I've got four ways to do it. Which of these do you think you'd prefer?
[...]

Issue #339, opened 2017-08-13
`NdArray` trait


bluss:
We need to traitify the array types, so that we shift from using `ArrayBase` for generic programming to something more flexible.

Imagine a sketch:

```rust
pub fn std_axis<T>(array: T) -> Array<T::Elem, T::Dim::Smaller>
    where T: NdArray,
          T::Elem: Num,
{
     // implementation here.
}
```

Foreseen problems: How do we conditionalize methods?

- Can mutate the axes or view
- Can mutate elements
- etc

termoshtt:
I agree to introduce such trait. My current generic code typically has the following form:

```rust
pub fn func<A, S1, S2>(a: &mut ArrayBase<S1, Ix2>, b: &ArrayBase<S2, Ix2>)
where
  A: LinalgScalar,
  S1: DataMut<Elem=A>,
  S2: Data<Elem=A>,
{
   ...
}
```

In my understanding, `NdArray` trait would rewrite it into following:

```rust
pub fn func<K, A1, A2>(a: &mut A1, b: &A2)
where
  K: LinalgScalar,
  A1: NdArrayMut<A, Ix2>,
  A2: NdArray<A, Ix2>,
{
   ...
}
```

I think `Data*` traits should be hidden from end users.

bluss:
Logically they are associated types:

A specific type like Array f32, Ix2 has anot asociale type for the element because it is given from the type uniquely.

bluss:
Ok, can't edit on mobile. Thanks github.

bluss:
Here's are some tantalizing examples of what `impl NdArray` can look like. This kind of interface is very flexible!

Instead of dynamic typing, we get abstract types, and we can pick and choose which of the associated types of the array should be abstract for each interface. It makes sense when you look at these examples.

Written for playground which as of this writing uses ndarray 0.12. [(playground link)](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=27f22a52880ff81ec9bfc10c370736df)

```rust
use ndarray::prelude::*;
use ndarray::Data;
use ndarray::Zip;

// Usage examples

pub fn example_longest_axis(array: impl NdArray) -> Option<Axis> {
    array.as_array().axes().max_by_key(|ax| ax.len()).map(|ax| ax.axis())
}

pub fn example_sum(array: impl NdArray<Item=f64>) -> f64 {
    array.as_array().sum()
}

[...]

PR #1469, opened 2024-12-26
Introduce an ArrayLike trait

akern40:
This is one of two PRs I'm working on that start to "traitify" ndarray. This trait is like the `Ndarray` trait suggested by @bluss in #339, but is slightly broader: it's meant to encapsulate not just actual `ndarray` types, but anything that could act *like* a multidimensional array. This includes Vecs, slices, arrays, and (critically) scalars that implement `ScalarOperand`.

The motivation for me was trying to design a set of methods expanded numerical methods, addressing #1462 (and others). My goal was to have these methods
1. Work for types that implement the associated `num_traits` traits, i.e., `abs` for `T: Signed`.
2. Have both array-creating and array-mutating variants
3. Broadcast amongst all arguments
4. Work for arrays and scalars interchangeably

I was able to accomplish 1-3 with existing infrastructure, but you can see in #1462 that I was struggling with (4). This trait would allow me to accomplish all of these tasks.

Feedback is greatly welcome. Keep an eye out for two more upcoming PRs:
1. An implementation of the math stuff above that uses `ArrayLike`, to close #14632
2. A PR with the `Ndarray` trait, following the design of #339

### DIFF
--- a/src/arraylike.rs
+++ b/src/arraylike.rs
@@ -22,46 +22,161 @@ use crate::{
 /// and other types that conceptually act like arrays. It's designed to make your functions
 /// more flexible by letting them handle a wide range of types without extra boilerplate.
 ///
-/// ## More Details
-/// The key idea of `ArrayLike` is to bridge the gap between native ndarray types
-/// (e.g., [`Array2`](crate::Array2), [`ArrayView`](crate::ArrayView)) and other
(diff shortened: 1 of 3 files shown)


akern40:
I'm not sure this is the right approach. Either way, I think this needs #1506 to be stitched up first. Closing for now, at least the code is here for future reference.

PR #1481, opened 2025-02-09
Add `softmax` function

barakugav:


akern40:
Hey @barakugav, thanks for the PR. I have two things I'm worried about here: first, `softmax` is a pretty ML-specific function that feels like it belongs in a different crate, not as part of `ndarray` itself. You'll see around the repo that the maintainers (cc @nilgoyette) try to be judicious about adding functions in order to not bloat the API. My second concern is that this is a fairly complex implementation when a simpler one also exists. Do you happen to have benchmarks on how this implementation compares to a simpler sum-then-divide approach?

barakugav:
the implementation is more complicate than the simpler sum-then-divide not due to performance, but due to float overflow., subtracting the max value is standard in softmax implementation and the regular implementation is not practical. Anyway, thanks for the feedback, ill look for a more appropriate crate for this implementation

PR #1482, opened 2025-03-10
add method to produce view with swapped axes

jonaspleyer:


nilgoyette:
Maybe @akern40 has a different opinion, but I don't think we want to add that kind of function. Having too much functions hurts discoverability.

akern40:
Ya this is a recurring topic in our PRs, and one that I think needs addressing more thoroughly. I think @nilgoyette is right - this function turns a two-liner into a one-liner, and I'm not sure it's worth the impact on the discoverability. _However_, I also think we need to come up with some sort of reasoning for which methods to accept so that contributors can understand what kinds of functions we would accept, or what the criteria are for acceptance.

jonaspleyer:
Thanks for the response. I can very much relate to your viewpoint. My proposal would be to add a note (by using templates) when creating a new PR or Issue in Github. By using such templates, contributors will get notified. If they ignore said notes, it is easy to point to them.

akern40:
Thanks for understanding, and for the suggestion - I'd love to have a template we could use, eventually. First, we need to work with the community to articulate which functions we want and which we don't.

jonaspleyer:
> Thanks for understanding, and for the suggestion - I'd love to have a template we could use, eventually. First, we need to work with the community to articulate which functions we want and which we don't.

I have opened a new discussion to come up with possible guidelines. However I think we could still add a note which points to said discussion which basically says:
"We are currently reviewing our guidelines as to what should be included in this library and what is considered a shallow addition and should be omitted. Think carefully if the proposed functionality should be added. (see discussion https://github.com/rust-ndarray/ndarray/discussions/1493)"

PR #1483, opened 2025-03-15
Add dot product support for ArrayD

### LINKED ISSUE #1439: Dot product is only a valid call for ndarray::Array2
Dot product is only a valid call for ndarray::Array2 -- even if the ArrayD is of the same dimensions.
```rust
use ndarray::prelude::*;
fn main() {
    let mat1 = Array2::from_shape_vec((3, 2), vec![3.0; 6]).unwrap();
    let mat2 = Array2::from_shape_vec((2, 3), vec![1.0; 6]).unwrap();
    let dot = mat1.dot(&mat2);
}
```
works, while
```rust
use ndarray::prelude::*;
fn main() {
    let mat1 = ArrayD::from_shape_vec(vec![3, 2], vec![3.0; 6]).unwrap();
    let mat2 = ArrayD::from_shape_vec(vec![2, 3], vec![1.0; 6]).unwrap();
    let dot = mat1.dot(&mat2);
}
```
does not
This results in a very opaque compiler error:
```rust
overflow evaluating the requirement `&ArrayBase<_, _>: Not`
consider increasing the recursion limit by adding a `#![recursion_limit = "256"]` attribute to your crate (`ndarray-issue`)
required for `&ArrayBase<_, _>` to implement `Not`
127 redundant requirements hidden
required for `&ArrayBase<OwnedRepr<{float}>, Dim<IxDynImpl>>` to implement `Not`
```
In my opinion this should either give a clearer error message or be a valid call. I am not experienced enough with rust at the moment to suggest a fix myself.

NewBornRustacean:
# Add dot product support for ArrayD
Hello, awesome ndarray community! 
- This is my very first PR and implements the `Dot` trait for `ArrayD` (dynamic-dimensional arrays), addressing issue #1439. 
- for now, the tests are in same location `ndarray/src/linalg/impl_linalg.rs` 
- thanks to detailed description for the issue @JulianDicken and @JamboChen, I'm not sure my implementation match with your original discussion. it would be very helpful to give me any comments/feedback! 
 

## Changes

- Added implementation of `Dot<ArrayBase<S2, IxDyn>>` for `ArrayBase<S, IxDyn>`
- Supports the following operations:
  - Vector-vector dot product (1D × 1D)
  - Matrix-matrix multiplication (2D × 2D)
  - Matrix-vector multiplication (2D × 1D)
  - Vector-matrix multiplication (1D × 2D)
- Added comprehensive test cases covering all supported operations and error cases

## Details

converts dynamic-dimensional arrays to their fixed-dimensional counterparts when needed, leveraging the existing implementations for `Array1` and `Array2`. This ensures consistent behavior with the fixed-dimensional array implementations.

## Example Usage

```rust
use ndarray::ArrayD;

// Matrix-matrix multiplication
let mat1 = ArrayD::from_shape_vec(vec![3, 2], vec![3.0; 6]).unwrap();
let mat2 = ArrayD::from_shape_vec(vec![2, 3], vec![1.0; 6]).unwrap();
let result = mat1.dot(&mat2);

// Vector-vector dot product
let vec1 = ArrayD::from_shape_vec(vec![3], vec![1.0, 2.0, 3.0]).unwrap();
let vec2 = ArrayD::from_shape_vec(vec![3], vec![4.0, 5.0, 6.0]).unwrap();
let scalar = vec1.dot(&vec2);
```

## Testing

Added test cases for:
- Matrix-matrix multiplication
- Vector-vector dot product
- Matrix-vector multiplication
- Error cases (3D arrays, incompatible dimensions)

All tests pass and maintain consistency with the existing fixed-dimensional array implementations.

Fixes #1439


src/linalg/impl_linalg.rs
```diff
@@ -1067,3 +1067,129 @@ mod blas_tests
...
+{
+    type Output = Array<A, IxDyn>;
+
+    fn dot(&self, rhs: &ArrayBase<S2, IxDyn>) -> Self::Output {
+        match (self.ndim(), rhs.ndim()) {
+            (1, 1) => {
+                // Vector-vector dot product
+                if self.len() != rhs.len() {
+                    panic!("Vector lengths must match for dot product");
+                }
```
nilgoyette, line 1085: Shouldn't this test appear in the other 3 cases too? In fact, no, can you please check in the actual `dot` implementation if the lengths are already checked? If they are, then you don't need to do it here.

src/linalg/impl_linalg.rs
```diff
@@ -1067,3 +1067,129 @@ mod blas_tests
...
+                let a = self.view().into_dimensionality::<Ix1>().unwrap();
+                let b = rhs.view().into_dimensionality::<Ix1>().unwrap();
+                let result = a.dot(&b);
+                ArrayD::from_elem(vec![], result)
+            }
+            (2, 2) => {
+                // Matrix-matrix multiplication
+                let a = self.view().into_dimensionality::<Ix2>().unwrap();
```
nilgoyette, line 1093: IIUC, this (and the others around) will never allocate any memory. Right?

nilgoyette (commented):
Thank you for this really clean PR. Offering `dot` for the dynamic arrays is clearly a good idea.

src/linalg/impl_linalg.rs
NewBornRustacean: thanks to review this quickly!
you're right, I found the same logic from `dot_generic` fucntion. 
gonna remove redundant check 👍 

``` rust
fn dot_generic<S2>(&self, rhs: &ArrayBase<S2, Ix1>) -> A
where
    S2: Data<Elem = A>,
    A: LinalgScalar,
{
    debug_assert_eq!(self.len(), rhs.len());
    assert!(self.len() == rhs.len());  
}
```

src/linalg/impl_linalg.rs
NewBornRustacean: yes! as far as I know, these are zero-cost to conversion. both of `view()` and `into_dimensionality()`

NewBornRustacean:
Thanks for the quick and kind review! @nilgoyette I just added docstring and removed length check which is redundant.

nilgoyette:
I don't know if the other maintainers reached a decision concerning the "formatting problem", so I'm unsure if your last commit is acceptable or not. Personally, I would run a `cargo fmt` on the whole project and never talk about this problem again, but it's a little more complex than that :)  You can wait for their input, or you can revert your last commit (if you want the PR to be merged faster). Up to you.

NewBornRustacean:
> I don't know if the other maintainers reached a decision concerning the "formatting problem", so I'm unsure if your last commit is acceptable or not. Personally, I would run a `cargo fmt` on the whole project and never talk about this problem again, but it's a little more complex than that :) You can wait for their input, or you can revert your last commit (if you want the PR to be merged faster). Up to you.

oh, I got it! thanks to kind explanation! I just undo the last commit(cargo fmt) 👍

akern40:
Unfortunately our current formatting settings require `cargo +nightly fmt`, since many of the formatting features we use are unstable. I am starting to think that the barrier to new PRs presented by a nightly-only `cargo fmt` is not worth the pain of the single formatting commit that would be necessary to move us to a stable-only `cargo fmt`.

Btw this PR looks great! Let me fix the CI on our master branch and retrigger this CI run. We can discuss the whole `cargo fmt` issue, but perhaps you could just run `cargo +nightly fmt` and we can get this merged in?

NewBornRustacean:
@akern40 Good morning(or evening)! `cargo +nightly fmt` commit included! iirc, there's guide to use _nightly_ cargo for `ndarray`, I think I missed it.
I know many cool features are in cargo nightly, so it would be nice to stick to `cargo +nightly fmt`, imo.

NewBornRustacean:
@akern40 Sorry to bother you, but some of the integrated tests still went to fail. I read your PR which is merged(#1485 ), and some of the packages need higher rust version, I guess.

for example, 

``` bash
error: package `zerofrom v0.1.6` cannot be built because it requires rustc 1.81 or newer, while the currently active rustc version is 1.71.1
```

image

hope this is helpful to you!

akern40:
Huh, that's very odd! Can you make sure that you have rebased / pulled the latest commits from main and run `cargo update --locked` to make sure you are on the same dependencies? Also, please remove the `no_run` from the example, we do want examples run in the CI.

akern40:
(Also you're not bothering me! This is what being a maintainer is all about - making sure others can successfully commit to the library)

NewBornRustacean:
> Huh, that's very odd! Can you make sure that you have rebased / pulled the latest commits from main and run `cargo update --locked` to make sure you are on the same dependencies? Also, please remove the `no_run` from the example, we do want examples run in the CI.

@akern40 thanks! I forgot pull master before I push the changes  😅

NewBornRustacean:
@akern40  ping you again because some tests went fail due to examples in docstring 😥 fixed it, now it's fine with `cargo test --doc`

akern40:
Ok I think I understand what's happening here, after a bit of testing / reading: the tests you have written work fine without `blas`, as expected. However, the way `blas` works with `ndarray` right now is that two things must occur: `ndarray` must be built with the `blas` feature, and the binary (in our case, the test binary) must be built by selecting a `bias-src` and using `extern crate blas`. So your tests - which _always_ run - have prerequisite #1 completed, but not prerequisite #2. I believe the solution is to move these tests into `crates/blas-tests/tests/oper.rs` (or, maybe better, a new file: `crates/blas-tests/tests/dyn.rs`). This crate is set up to do both prerequisite steps.

NewBornRustacean:
> Ok I think I understand what's happening here, after a bit of testing / reading: the tests you have written work fine without `blas`, as expected. However, the way `blas` works with `ndarray` right now is that two things must occur: `ndarray` must be built with the `blas` feature, and the binary (in our case, the test binary) must be built by selecting a `bias-src` and using `extern crate blas`. So your tests - which _always_ run - have prerequisite #1 completed, but not prerequisite #2. I believe the solution is to move these tests into `crates/blas-tests/tests/oper.rs` (or, maybe better, a new file: `crates/blas-tests/tests/dyn.rs`). This crate is set up to do both prerequisite steps.

wow, thanks! I didn't know that. I just made new commit to follow your suggestion.
btw, before push `crates/blas-tests/tests/dyn.rs`, what if I use this `#[cfg(feature = "blas")]` on my test? other tests in `impl_linalg.rs` are using it. (if this is intended to run `blas` backend. )

for example,

``` rust
#[cfg(test)]
#[cfg(feature = "blas")]
fn blas_column_major_2d<A, S>(a: &ArrayBase<S, Ix2>) -> bool
where
    S: Data,
    A: 'static,
    S::Elem: 'static,
{
    if !same_type::<A, S::Elem>() {
        return false;
    }
    is_blas_2d(&a.dim, &a.strides, BlasOrder::F)
}
```

akern40:
It's a little misleading, but if you take a close look at the tests that are currently in `impl_linalg`, you'll notice that they're testing the _surrounding machinery_ of the BLAS setup, but _not the BLAS code itself_. So when `blas` is enabled, this code tests the non-BLAS-calling code we write to support it, but it doesn't depend on the `extern blas_src` that is required for actually executing BLAS operations.

So, in short, your code would still have to go into `dyn.rs`, sorry!

Btw, thank you for this - it's a good opportunity for me to really wrap my head around the BLAS situation here. I'm starting in on some ideas to make this all easier, hopefully. We'll see if any of the ideas work 😄

NewBornRustacean:
@akern40  moved tests `crates/blas-tests/tests/dyn.rs`.
I checked it with my own machine with blas integration, hope it works fine with machine on CI pipeline!

Thanks to kind and clear guide to contribute!

akern40:
@NewBornRustacean thank you for your pull request, your patience, and your collaboration, and congratulations on your first PR to `ndarray`! I hope it will be the first of many.

NewBornRustacean:
@akern40 @nilgoyette thank you all!  want to  contribute more obviously😁 will be back with next pr 🫡

PR #1491, opened 2025-03-18
Add cumprod

NewBornRustacean:
# Add cumprod() to match NumPy behavior

Hello ndarray community!

This is implementation of the `cumprod()` function to match NumPy's behavior regarding axis handling.

## Related links
- #1354 
- [numpy doc](https://numpy.org/doc/stable/reference/generated/numpy.cumulative_prod.html)

## Details
1. ~~check `if self.ndim() > 1 && axis.is_none() { panic!(...) }`~~- removed since we now use `Axis` directly
2. ~~`Some(axis)` branch to use input array values instead of result array values~~ - removed since we now use `Axis` directly
3. tests to match NumPy's behavior

## Example Usage
```rust
use ndarray::{array, Axis};

// 1D array - Axis(0) works
let a = array![1, 2, 3, 4];
assert_eq!(a.cumprod(Axis(0)), array![1, 2, 6, 24]);

// 2D array - must specify axis
let b = array![[1, 2], [3, 4]];
assert_eq!(b.cumprod(Axis(0)), array![[1, 2], [3, 8]]);
assert_eq!(b.cumprod(Axis(1)), array![[1, 2], [3, 12]]);
```

## Testing
All existing tests pass,
* 1D array behavior with `Axis(0)`
* N-D array behavior requiring axis specification
* Error cases for invalid axis usage

The implementation maintains consistency with NumPy's behavior, but there might be more elegant way to implement this(especially for the nested loop). please give me any comment/feedback! 

@akern40 what's up! I love the automated CI pipeline, thanks 👍


src/numeric/impl_numeric.rs
```diff
@@ -99,6 +100,73 @@ where
...
+    {
+        // First check dimensionality
+        if self.ndim() > 1 && axis.is_none() {
+            panic!("axis parameter is required for arrays with more than one dimension");
+        }
+
+        match axis {
+            None => {
+                // This case now only happens for 1D arrays
+                let mut res = Array::ones(self.raw_dim());
```
nilgoyette, line 143: Can't you move this declaration our of the match?

src/numeric/impl_numeric.rs
```diff
@@ -99,6 +100,73 @@ where
...
+                let mut acc = A::one();
+
+                for (r, x) in res.iter_mut().zip(self.iter()) {
```
nilgoyette, line 146: Standard iteration are slower than using `Zip` in ndarray. Can you please refactor? Idem for the loops below.

src/numeric/impl_numeric.rs
```diff
@@ -99,6 +100,73 @@ where
...
+                }
+
+                res
+            }
+            Some(axis) => {
+                let mut res: Array<A, D> = Array::ones(self.raw_dim());
+
+                // Process each lane independently
+                for (mut out_lane, in_lane) in res.lanes_mut(axis).into_iter().zip(self.lanes(axis)) {
+                    let mut acc = A::one();
```
nilgoyette, line 158: You're allocating this array N times. It should be declared only once and re-used.

nilgoyette (changes requested):

NewBornRustacean:
@nilgoyette thanks for your quick review! I was seeking for what would be a better approach, your comment was very helpful! I tried to vectorize it a bit more with `fold_axis()`

src/numeric/impl_numeric.rs
```diff
@@ -99,6 +101,76 @@ where
...
+                // For nD arrays, use fold_axis approach
+                // Create accumulator array with one less dimension
+                let mut acc = Array::ones(self.raw_dim().remove_axis(axis));
+
+                for i in 0..self.len_of(axis) {
+                    // Get view of current slice along axis, and update accumulator element-wise multiplication
+                    let view = self.index_axis(axis, i);
+                    acc = acc * &view;
+                    res.index_axis_mut(axis, i).assign(&acc);
+                }
```
nilgoyette, line 168: Would something like this work? (untested, you probably need to modify something)

    Zip::from(self.axis_iter(axis)).and(res.axis_iter_mut(axis)).for_each(|view, res| {
        acc = acc * &view;
        res.assign(&acc);
    })

src/numeric/impl_numeric.rs
```diff
@@ -99,6 +101,76 @@ where
         sum
     }
 
+    /// Return the cumulative product of elements along a given axis.
+    ///
+    /// If `axis` is None, the array is flattened before taking the cumulative product.
```
akern40, line 106: This is the right behavior, as defined by the NumPy API, but you're limiting it to 1D arrays on Line 137.

akern40: Oh hold on I see the issue - we can't dynamically decide we're going to return a 1D array based on the argument, can we. So this part of the documentation should be removed.

src/numeric/impl_numeric.rs
```diff
@@ -99,6 +101,76 @@ where
...
+    /// assert_eq!(
+    ///     a.cumprod(Some(Axis(1))),
+    ///     arr2(&[[1., 2., 6.],
+    ///           [4., 20., 120.]])
+    /// );
+    /// ```
+    ///
+    /// **Panics** if `axis` is out of bounds.
+    #[track_caller]
+    pub fn cumprod(&self, axis: Option<Axis>) -> Array<A, D>
```
akern40, line 131: Since we can't dynamically return 1D arrays (and therefore can't use `axis: None` unless the array is already 1D), I'd suggest we just accept the argument as `axis: Axis`.

akern40 (changes requested):
@NewBornRustacean great to see you back so soon!

src/numeric/impl_numeric.rs
```diff
@@ -99,6 +101,76 @@ where
...
+            panic!("axis parameter is required for arrays with more than one dimension");
+        }
+
+        let mut res = Array::ones(self.raw_dim());
+
+        match axis {
+            None => {
+                // For 1D arrays, use simple iteration
+                let mut acc = A::one();
+                Zip::from(&mut res).and(self).for_each(|r, x| {
```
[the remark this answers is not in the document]
NewBornRustacean: ah ha, I think it's possible. but as @akern40 mentioned below, this branch should be removed since `Axis` is always specified.

src/numeric/impl_numeric.rs
NewBornRustacean: oh I know what you mean, thanks! I'll try 👍

src/numeric/impl_numeric.rs
NewBornRustacean: you're right. it's more clear, thanks!  👍

src/numeric/impl_numeric.rs
NewBornRustacean: since we have to modify `res` and update `acc` to calculate accumulated product, `acc` should be cloned, I guess.
is it ok to `clone()` in `for_each`?

``` rust
Zip::from(self.axis_iter(axis))
            .and(res.axis_iter_mut(axis))
            .for_each(|view, mut res| {
                acc = acc.clone() * &view;
                res.assign(&acc);
            });
```

NewBornRustacean:
@nilgoyette @akern40  sorry to belated reply! thanks to suggestions 🥇  most of the suggestions are included and passes all the tests.

src/numeric/impl_numeric.rs
NewBornRustacean: sorry to make you confuse, this one would be better

```rust 
        Zip::from(self.axis_iter(axis))
            .and(res.axis_iter_mut(axis))
            .for_each(|view, mut res| {
                running_product = &running_product * &view;
                res.assign(&running_product);
            });

```

nilgoyette: Thanks for removing the clone. The fact is, this clone allocates (and copies) memory. Since we can easily imagine this algorithm working without N allocations, we shouldn't have to call clone. But you removed it! Job done :confetti_ball: 

In fact, no ;) You replaced a visible clone with an invisible clone. Please read this section of the [documentation](https://docs.rs/ndarray/latest/ndarray/struct.ArrayBase.html#arithmetic-operations). It's easy to miss. Those ampersands have a important meaning in ndarray.

So, what can we do? I might be wrong, but I **think** you can write

    acc *= &view;

Hopefully it works :)  If it still doesn't work, we will find something. Allocating here is unacceptable.

nilgoyette: Wait, I'm not even sure myself. This line

    let sum4 = owned2 + &view2;   // Consumes `owned2`, updates it, and returns it.

seems to indicates that it has the exact same behavior as what I'm suggesting. @akern40 can you please confirm?

NewBornRustacean: however, in my previous implementation, it's possible to write something like this:
```rust
let sum4 = owned2 + &view2;   // Consumes `owned2`, updates it, and returns it.
```

``` rust
for i in 0..self.len_of(axis) {
                    // Get view of current slice along axis, and update accumulator element-wise multiplication
                    let view = self.index_axis(axis, i);
                    acc = acc * &view;
                    res.index_axis_mut(axis, i).assign(&acc);
                }
```

NewBornRustacean: how about this?
```rust
 Zip::from(self.axis_iter(axis))
            .and(res.axis_iter_mut(axis))
            .fold(running_product, |mut running_product, view, mut res| {
                running_product = running_product * &view;
                res.assign(&running_product);
                running_product
            });

```

NewBornRustacean: how about this one? consume `running_product` and pass ownership.

```rust
        Zip::from(self.axis_iter(axis))
            .and(res.axis_iter_mut(axis))
            .fold(running_product, |mut running_product, view, mut res| {
                running_product = running_product * &view;
                res.assign(&running_product);
                running_product
            });
```

akern40: Just seeing this conversation! Let me think it over and get back to you when I'm in front of a computer, but there should be a way to do this with just a single additional allocation, the output array.

akern40: Ok! To answer @nilgoyette's question, you're correct - when doing operators, if one of them is an _owned_ array, we eagerly consume its memory for use as the return. So `running_product = running_product * &view` should not perform additional allocation.

On a separate note, @NewBornRustacean please take a look at [`accumulate_axis_inplace`](https://docs.rs/ndarray/latest/ndarray/struct.ArrayBase.html#method.accumulate_axis_inplace). This is the method that you likely want to use for implementing `cumprod`, and should allow you to write it with only a single allocation (`let result = self.clone()`).

nilgoyette (approved):
The code is really nice with `accumulate_axis_inplace`! Sorry for the detour. I guess we all learned something ;)

NewBornRustacean:
> The code is really nice with `accumulate_axis_inplace`! Sorry for the detour. I guess we all learned something ;)

yes indeed! thank you so much 😃

akern40:
Thanks for your persistence! Last thing: I don't think `Copy` should be needed. I am assuming it prompted you for it when you wrote `*curr = *curr * prev`? If so, I think you can either a) use `*curr = curr.clone() * prev`, in which case Rust should elide the clone; or b) call `*curr *= prev`, which would require a `MulAssign` trait bound. I think I'm partial to option (a), since Rust _should_ still just do a multiply-assign operation when possible, and you have fewer trait bounds. But I'm open to ideas.

NewBornRustacean:
> Thanks for your persistence! Last thing: I don't think `Copy` should be needed. I am assuming it prompted you for it when you wrote `*curr = *curr * prev`? If so, I think you can either a) use `*curr = curr.clone() * prev`, in which case Rust should elide the clone; or b) call `*curr *= prev`, which would require a `MulAssign` trait bound. I think I'm partial to option (a), since Rust _should_ still just do a multiply-assign operation when possible, and you have fewer trait bounds. But I'm open to ideas.

thanks for the suggestion! 😺 how about this one?  
```rust
    #[track_caller]
    pub fn cumprod(&self, axis: Axis) -> Array<A, D>
    where
        A: Clone + Mul<Output = A> + MulAssign,
        D: Dimension + RemoveAxis,
    {
        if axis.0 >= self.ndim() {
            panic!("axis is out of bounds for array of dimension");
        }

        let mut result = self.to_owned();
        result.accumulate_axis_inplace(axis, |prev, curr| *curr *= prev.clone());
        result
    }
```

akern40:
Looks good to me! Thanks again for your work 😄

Issue #1497, opened 2025-03-27
add numpy.partition like functionality


NewBornRustacean:
# add numpy.partition like functionality

## Overview
What's up awesome ndarray community! I searched through old issues and found the related issues below.
There were couple of discussions and PRs, but currently, it seems like ndarray and ndarray-stats lack several key sorting and searching operations that are available in NumPy. 

## Start with partitioning: API is gonna be like..

```rust
impl<A, D> ArrayBase<S, D>
where
    A: Ord,
    D: Dimension,
{
    /// Partially sorts the array around the k-th element.
    /// 
    /// Similar to NumPy's `partition`. Returns a new array where:
    /// - All elements smaller than the k-th element are to its left
    /// - All elements equal or greater than the k-th element are to its right
    /// - The ordering within each partition is undefined
    pub fn partition(&self, kth: usize) -> Array<A, D> { ... }

    /// Returns the indices that would partition the array.
    /// 
    /// Similar to NumPy's `argpartition`. Returns indices that would partition
    /// the array around the k-th element.
    pub fn argpartition(&self, kth: usize) -> Array<usize, D> { ... }
}
```
## Note
- I searched the codebase thoroughly and started implementing this feature, but there might already be an existing implementation that I missed. If that's the case, please let me know!
- will open draft PR soon! 

## Related Issues
- #416  
- #649  
- #1145  
- (might be more)

## References
- [NumPy documentation for argsort](https://numpy.org/doc/stable/reference/generated/numpy.argsort.html)
- [NumPy documentation for partition](https://numpy.org/doc/stable/reference/generated/numpy.partition.html)
- [NumPy documentation for argwhere](https://numpy.org/doc/stable/reference/generated/numpy.argwhere.html)
- [NumPy documentation for where](https://numpy.org/doc/stable/reference/generated/numpy.where.html)

NewBornRustacean:
now `partition` is added. will be back with `argpartition` or `partition_multiple` 😃

PR #1498, opened 2025-03-27
Add partition(similar to numpy.partition)

NewBornRustacean:
## 1. Overview

This is about the `partition` function to handle both contiguous and non-contiguous memory more efficiently. The function now uses a simpler approach for handling non-contiguous memory, reducing overhead in the fallback path.

## 2. Motivation

The `partition` function is an important operation that partially sorts an array around the k-th element along a specified axis. This is useful for:

- Efficiently finding the k-th smallest/largest element without fully sorting
- Dividing data into smaller-than-k and larger-than-k groups (e.g., for quick select algorithms)
- Implementing median-finding algorithms

## 3. Note
- The implementation relies on [select_nth_unstable](https://doc.rust-lang.org/std/primitive.slice.html#method.select_nth_unstable)
  - It might be possible to write our own functions to implement `quickselect` algorithm rather than to use existing one. 
  - let me know if that's the case😃 
- The current implementation of partition() creates a temporary array for each lane when memory is non-contiguous, which can be inefficient. 

## 4. We can go further
- in `numpy.partition`, it is possible to pass sequence of ints as a `kth` parameter. It would be nice to have one in our case, maybe seperated function like `partition_multiple()`
- there is also interesting and useful function `numpy.argpartition`. I think I could implement that also as a follow-up. 

## Related issue
- #1497

### DIFF
--- a/src/impl_methods.rs
+++ b/src/impl_methods.rs
@@ -3184,6 +3184,92 @@ impl<A, D: Dimension> ArrayRef<A, D>
             f(&*prev, &mut *curr)
         });
     }
+
+    /// Return a partitioned copy of the array.
+    ///
+    /// Creates a copy of the array and partially sorts it around the k-th element along the given axis.
+    /// The k-th element will be in its sorted position, with:
+    /// - All elements smaller than the k-th element to its left
+    /// - All elements equal or greater than the k-th element to its right
+    /// - The ordering within each partition is undefined
+    ///
+    /// # Parameters
+    ///
+    /// * `kth` - Index to partition by. The k-th element will be in its sorted position.
+    /// * `axis` - Axis along which to partition. Default is the last axis (`Axis(ndim-1)`).
+    ///
+    /// # Returns
+    ///
+    /// A new array of the same shape and type as the input array, with elements partitioned.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use ndarray::prelude::*;
+    ///
+    /// let a = array![7, 1, 5, 2, 6, 0, 3, 4];
+    /// let p = a.partition(3, Axis(0));
+    ///
+    /// // The element at position 3 is now 3, with smaller elements to the left
+    /// // and greater elements to the right
+    /// assert_eq!(p[3], 3);
+    /// assert!(p.slice(s![..3]).iter().all(|&x| x <= 3));
+    /// assert!(p.slice(s![4..]).iter().all(|&x| x >= 3));
+    /// ```
+    pub fn partition(&self, kth: usize, axis: Axis) -> Array<A, D>
+    where A: Clone + Ord
+    {
+        // Check if axis is valid
+        if axis.index() >= self.ndim() {
+            panic!("axis {} is out of bounds for array of dimension {}", axis.index(), self.ndim());
+        }
+
+        // Check if kth is valid
+        if kth >= self.len_of(axis) {
+            panic!("kth {} is out of bounds for axis {} with length {}", kth, axis.index(), self.len_of(axis));
+        }
+
+        // If the array is empty, return a copy
+        if self.is_empty() {
+            return self.to_owned();
+        }
+
+        // If the array is 1D, handle as a special case
+        if self.ndim() == 1 {
+            let mut result = self.to_owned();
+            if let Some(slice) = result.as_slice_mut() {
+                slice.select_nth_unstable(kth);
+            }
+            return result;
+        }
+
+        // For multi-dimensional arrays, partition along the specified axis
+        let mut result = self.to_owned();
+
+        // Use Zip to efficiently iterate over the lanes
+        Zip::from(result.lanes_mut(axis)).for_each(|mut lane| {
+            // For each lane, perform the partitioning operation
+            if let Some(slice) = lane.as_slice_mut() {
+                // If the lane's memory is contiguous, use select_nth_unstable directly
+                slice.select_nth_unstable(kth);
+            } else {
+                // For non-contiguous memory, create a temporary array with contiguous memory
+                let mut temp_arr = Array::from_iter(lane.iter().cloned());
+
+                // Partition the temporary array
+                if let Some(slice) = temp_arr.as_slice_mut() {
+                    slice.select_nth_unstable(kth);
+                }
+
+                // Copy values back to original lane
+                Zip::from(&mut lane).and(&temp_arr).for_each(|dest, src| {
+                    *dest = src.clone();
+                });
+            }
+        });
+
+        result
+    }
 }
 
 /// Transmute from A to B.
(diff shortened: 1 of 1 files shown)


src/impl_methods.rs
```diff
@@ -3184,6 +3184,90 @@ impl<A, D: Dimension> ArrayRef<A, D>
...
+
+        // Process each lane with partitioning
+        Zip::from(result.lanes_mut(axis)).for_each(|mut lane| {
+            // For each lane, we need a contiguous slice to partition
+            if let Some(slice) = lane.as_slice_mut() {
+                // If the lane's memory is contiguous, use select_nth_unstable directly
+                slice.select_nth_unstable(kth);
+            } else {
+                // For non-contiguous memory, create a temporary vector
+                let mut values = lane.iter().cloned().collect::<Vec<_>>();
```
nilgoyette, line 3257: A single vector should be declared, not N vectors. I'm pretty sure it's possible to know in advance (before the `Zip`) if the lanes will be contiguous or not. I **think** we can safely say that if the first lane is contiguous, then all other lanes will be too.

nilgoyette (changes requested):
I wonder if this function should be in `ndarray` or in another crate. I don't have the answer so I did a normal review :)

akern40:
@nilgoyette here's that question again about what belongs in the library - it's pesky! It keeps coming! I think I'm ok with this? I suppose its home _could_ be in a sort of `ndarray-sort` crate? Certainly its implementation is complex enough that we shouldn't expect users to write it themselves. And we could always include it in some sort of future, large "crate split".

On the other hand, devil's advocate: this function doesn't have an issue from a user, but rather is driven by parity with NumPy. If we decide to keep accepting NumPy functions, our method-discoverability problem is gonna get worse, right?

@NewBornRustacean you're welcome to join the discussion on this. In the meantime: I believe that `to_owned` will always return a fully-contiguous array (in fact, I believe that `Array` must always be contiguous?). If that's true, then the non-contiguous part of your code will never be called. You may be able to get rid of it!

NewBornRustacean:
> @nilgoyette here's that question again about what belongs in the library - it's pesky! It keeps coming! I think I'm ok with this? I suppose its home _could_ be in a sort of `ndarray-sort` crate? Certainly its implementation is complex enough that we shouldn't expect users to write it themselves. And we could always include it in some sort of future, large "crate split".
> 
> On the other hand, devil's advocate: this function doesn't have an issue from a user, but rather is driven by parity with NumPy. If we decide to keep accepting NumPy functions, our method-discoverability problem is gonna get worse, right?
> 
> @NewBornRustacean you're welcome to join the discussion on this. In the meantime: I believe that `to_owned` will always return a fully-contiguous array (in fact, I believe that `Array` must always be contiguous?). If that's true, then the non-contiguous part of your code will never be called. You may be able to get rid of it!

@akern40 @nilgoyette what's up! thanks to kind review and suggestions! 

**about `contiguous` part** :  working on it! removing part to handle non-contiguous. thanks!


I largely agree with you akern40 ! If there were a dedicated crate like `ndarray-sort`, that would be ideal. Expanding functionality gradually and then splitting it into a separate crate at some point sounds like a good approach.  

Before submitting this PR, I reviewed the maintainer table discussions and past issues, and it seems like this "what belongs in the library" question is closely tied to the future direction of `ndarray` development.  

Thinking about why users rely on NumPy, I’d say (at least in my case) the two main reasons are:  
1. Rapid prototyping  
2. Data processing for data analysis  

For `ndarray` users, a major motivation seems to be removing Python from the ML ecosystem. But in that case, most people probably wouldn't start developing directly with `ndarray` from scratch. Instead, they’re likely trying to migrate something that already exists in a Python-based environment (at least, that was my experience—I'm curious if others feel the same!). If that’s the case, wouldn’t these users be particularly interested in reason 1: "rapid prototyping"?  

I don’t think `ndarray` needs to be fully functionally equivalent to NumPy! However, in a perspective of "rapid prototyping", I do believe that supporting at least some of NumPy’s functionality is essential for attracting those users who are trying to remove Python from their ML workflows.

nilgoyette:
Wait, I didn't even realize there was a `to_owned` just before!
- Why do we need it? The comment above doesn't explain. In fact, that comment is not related to this line.
- Even if the array is owned, it will be contiguous **only** for one axis, no? If you iterate on the "wrong" axis, won't you still need to non-contiguous code?

NewBornRustacean:
> Wait, I didn't even realize there was a `to_owned` just before!
> - Why do we need it? The comment above doesn't explain. In fact, that comment is not related to this line.
> - Even if the array is owned, it will be contiguous **only** for one axis, no? If you iterate on the "wrong" axis, won't you still need to non-contiguous code?

Yeah, I'm facing the issue. When I call the  function with Axis(1), for example, each of the lanes could be non contigous(I think this is because of the stride, but I'm not sure).

Or we can implement our own select function so don't rely on select_nth_unstable() that  require contiguous memory.

akern40:
@nilgoyette yes you're right!! I am being silly - you can definitely still be non-contiguous depending on the axis. Sorry everyone.

nilgoyette:
> When I call the function with Axis(1), for example, each of the lanes could be non contigous(I think this is because of the stride, but I'm not sure).

Any contiguous array (2D or more) will be non-contiguous on at least one axis. And I can't say which axis because it can use 'c' or 'f' order (please read about those terms if you never heard them before). Imagine a matrix (2D array)

     -->
    1 2 3
    4 5 6

It is contiguous if you read it in its natural order (1 2 3 4 5 6), but it's not if you read it in the other axis (1 4 2 5 3 6).

Knowing this, it's a good idea to provide code that works in both cases (non-contiguous version of `select_nth_unstable`), or to check and optimize, like you did. Of course, you're free to choose what you want to code (we're doing open source after all). Both versions work. If you're really motivated, you can do both and create a small benchmark to know which one is the best/fastest.

Btw, I honestly don't have the answer :) Allocating is slow, copying is not slow but we try to avoid it if we can. However, working on a big matrix on a non-contiguous array (can't fit in the cache) is slow too. You might discover that creating a temp array is faster only on huge matrices, but not in general.

NewBornRustacean:
@nilgoyette  Sorry for the late update! I was thinking about ways to avoid copying in the non-contiguous case. I was considering whether it might be possible by flattening the input array and then obtaining slices corresponding to each lane, but it didn’t work out as expected.

(I referred to [NumPy’s internal implementation](https://github.com/numpy/numpy/blob/ea965e4cd328fbcf76b03ff749ef06d2aa38c28b/numpy/core/src/multiarray/item_selection.c#L1232), and it seems that NumPy also performs a copy in the non-contiguous case.)

For now, I’ve only applied the changes you mentioned in the last review!

[stackoverflow: how can I find the actual implementation of np.partition?](https://stackoverflow.com/questions/57573453/how-can-i-find-the-actual-implementation-of-np-partition)

src/impl_methods.rs
```diff
@@ -3184,6 +3184,80 @@ impl<A, D: Dimension> ArrayRef<A, D>
...
+
+        let mut result = self.to_owned();
+
+        // Check if the first lane is contiguous
+        let is_contiguous = result
+            .lanes_mut(axis)
+            .into_iter()
+            .next()
+            .map(|lane| lane.is_contiguous())
+            .unwrap_or(false);
```
nilgoyette, line 3238: With the "Bounds checking" above, this `_or(false)` will never be called. I'm not sure about ndarray stance on `unwrap`... I think it's "use it if you can prove it's safe". @akern40 ?

If I'm right, the `unwrap()` should go after `next()`. If not, don't change anything.

src/impl_methods.rs
```diff
@@ -3184,6 +3184,80 @@ impl<A, D: Dimension> ArrayRef<A, D>
...
+        if is_contiguous {
+            Zip::from(result.lanes_mut(axis)).for_each(|mut lane| {
+                lane.as_slice_mut().unwrap().select_nth_unstable(kth);
+            });
+        } else {
+            let mut temp_vec = Vec::with_capacity(axis_len);
+
+            Zip::from(result.lanes_mut(axis)).for_each(|mut lane| {
+                temp_vec.clear();
+                temp_vec.extend(lane.iter().cloned());
```
nilgoyette, line 3249: Clear and extend is fine, but we try to avoid `iter` in ndarray (because `Zip` is faster). You can simply reverse the `Zip` you wrote below.

src/impl_methods.rs
NewBornRustacean: hey nilgoyette! thanks for the quick review 😄 
like you said, with the bounds checking we do earlier we're guaranteed to have at least one lane, so next() will never return None. `unwrap()` moved after `next()` !

```rust
        // Bounds checking
        let axis_len = self.len_of(axis);
        if kth >= axis_len {
            panic!("partition index {} is out of bounds for axis of length {}", kth, axis_len);
        }

        let mut result = self.to_owned();

        // Check if the first lane is contiguous
        let is_contiguous = result
        .lanes_mut(axis)
        .into_iter()
        .next()
        .unwrap() // Safe to unwrap because we've checked axis_len > 0
        .is_contiguous();
```

src/impl_methods.rs
NewBornRustacean: @nilgoyette thanks! just added to new commit😃

src/impl_methods.rs
```diff
@@ -3184,6 +3184,84 @@ impl<A, D: Dimension> ArrayRef<A, D>
...
+        } else {
+            let mut temp_vec = Vec::with_capacity(axis_len);
+
+            Zip::from(result.lanes_mut(axis)).for_each(|mut lane| {
+                temp_vec.clear();
+                temp_vec.resize(axis_len, A::zero());
+
+                Zip::from(&mut temp_vec).and(&lane).for_each(|dest, src| {
+                    *dest = src.clone();
+                });
```
nilgoyette, line 3253: Sorry for the back and forth, but I see a problem with this block.

Can you please tell me what happens if you remove the 2 first lines (clear and resize)?

NewBornRustacean: @nilgoyette oh that's quite interesting question!, I can confirm rightaway. as far as I know, when we iterate over the lanes with `zip`, we need to fill up the tmp vec to make it good to go. 

if we remove the 2 first lines,
1. The first lane would be processed correctly because we created the `temp_vec` with capacity, but it has zero length initially.
2. When we do `Zip::from(&mut temp_vec).and(&lane)` on the first lane, nothing would happen because `temp_vec` is empty!
3. Then select_nth_unstable would operate on an empty vector, which would likely cause a panic.
4. Even if we somehow got pass this, Zip would only process matching indices, so none of the lane values would be copied back.

So without clearing and resizing, the vector doesn't have elements for Zip to operate on, and we'd get panics.
The `clear()` ensures we start fresh for each lane, while the `resize()` ensures the vector has the right number of elements (with zero values) that Zip can then modify when iterating.

because of these, we got failed on test2d, 3d with mismatched dims.

image

nilgoyette: Exactly. Thank you for checking! Now, what happens if you **also** change your vector initialization for

    let mut temp_vec = vec![A::zero(); axis_len];

? (Sorry for asking you these strange questions. I just want you to understand something without me giving you the answer.)

NewBornRustacean: ah ha! thanks, I know what you mean. I was silly  😭 @nilgoyette

nilgoyette: You already know what I'm going to tell you, but ...  never do something in a loop if you can do it outside! :)

nilgoyette (approved):

src/impl_methods.rs
akern40: @nilgoyette I think that's the right approach for unwrap, as long as you leave a comment about why it's safe.

nilgoyette:
I decided to merge this PR, leaving the question "should we merge this in ndarray?" open. We will probably have to split ndarray into several smaller crates someday. In the meantime, I think we should avoid driving contributor away.

PR #1502, opened 2025-04-10
Fix partition on empty arrays

### LINKED ISSUE #1501: Partition needs a check for empty arrays
Quick fix needed for #1498, since the `unwrap` on empty arrays will panic.

akern40:
Closes #1501

### DIFF
--- a/src/impl_methods.rs
+++ b/src/impl_methods.rs
@@ -3229,16 +3229,22 @@ impl<A, D: Dimension> ArrayRef<A, D>
 
         let mut result = self.to_owned();
 
+        // Must guarantee that the array isn't empty before checking for contiguity
+        if result.shape().iter().any(|s| *s == 0) {
+            return result;
+        }
+
         // Check if the first lane is contiguous
         let is_contiguous = result
             .lanes_mut(axis)
             .into_iter()
             .next()
+            // This unwrap shouldn't cause panics because the array isn't empty
             .unwrap()
             .is_contiguous();
 
         if is_contiguous {
-            Zip::from(result.lanes_mut(axis)).for_each(|mut lane| {
+            result.lanes_mut(axis).into_iter().for_each(|mut lane| {
                 lane.as_slice_mut().unwrap().select_nth_unstable(kth);
             });
         } else {


src/impl_methods.rs
```diff
@@ -3229,16 +3229,22 @@ impl<A, D: Dimension> ArrayRef<A, D>
...
             .lanes_mut(axis)
             .into_iter()
             .next()
+            // This unwrap shouldn't cause panics because the array isn't empty
             .unwrap()
             .is_contiguous();
 
         if is_contiguous {
-            Zip::from(result.lanes_mut(axis)).for_each(|mut lane| {
+            result.lanes_mut(axis).into_iter().for_each(|mut lane| {
```
nilgoyette, line 3247: @akern40 I know the PR is already merged, but may I ask you why you changed this? Afaik, it does the exact same thing, but it might be slower.

akern40: Good question, and it was maybe a code style choice that I should have just left alone. My feeling is that `Zip` should be reserved for iterating over multiple arrays, not over a single one. I'm not sure which is faster or slower - we could maybe do a benchmark, it's an interesting question. If I have time today I'll try to set one up.

Thoughts on this coding style? Also, thoughts on my edit? Even if using `Zip` isn't preferred, should I have just left it alone?

akern40: Ok so two quickly-written benchmarks (sum over 1D array, sum rows over 2D array) show that iter and zip are identical in performance for 2D arrays. Zip has a _very_ slight advantage in summing 1D arrays that is most noticeable when the arrays have 32 or fewer elements.

On reflection, I realized why I even noticed the `Zip` and thought to change it: I saw `Zip` on that line, and thought to myself "uh oh, that operation doesn't require lockstep iteration over more than one array! Why is `Zip` there?" So to me, seeing `Zip` is an indication of a particular kind of behavior.

akern40: Update: I also benchmarked the actual partition line as edited, and the difference is negligible between the two!

PR #1503, opened 2025-04-10
ci: Add Dependabot

sjackman:
Enable [Dependabot](https://docs.github.com/en/code-security/dependabot) to open PRs to update dependencies.

Related PR
- https://github.com/rust-ndarray/ndarray-stats/pull/106

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: cargo
+    directory: /
+    schedule:
+      interval: daily


akern40:
@sjackman would you mind making a related issue that explains what you see to be the benefits of Dependabot? I'm familiar with its security scanning, but I'm not sure what you mean by "open PRs to update dependencies".

sjackman:
> I'm not sure what you mean by "open PRs to update dependencies".

https://docs.github.com/en/code-security/dependabot/dependabot-version-updates

> Keeping your dependencies updated automatically with Dependabot version updates
> You can use Dependabot to automatically keep the dependencies and packages used in your repository updated to the latest version, even when they don’t have any known vulnerabilities.

Dependabot will open PRs against your repo to keep your dependencies up to date. See for example this PR that @dependabot opened in `rust-bio` to update `petgraph`:
- https://github.com/rust-bio/rust-bio/pull/632

akern40:
Thank you for bringing this to my attention! I've looked into this and learned a few things. First, that we do have Dependabot set up. Second, that I should be making sure to pay closer attention to it. And third, that Dependabot PRs would probably be more noise than signal. I say that because they take the approach of directly pinning the offending dependencies. But for `ndarray`, our security vulnerability alerts are a) mostly in dependencies of dependencies or b) in our workspace test crates, which I'm not sure that Dependabot will handle correctly.

My opinion here is that pinning every offending dependency directly in `ndarray` is a recipe for a laundry list of pins that aren't true dependencies. Personally, I think it would be much better for the downstream dependencies to do the pinning of their direct dependencies - as we should (and will) do for our direct dependencies.

Does this seem like a reasonable policy?

Issue #1443, opened 2024-10-20
Reversing and permuting axes do not require consuming an array


akern40:
Currently, `reversed_axes` and `permuted_axes` take `mut self` and `self`, respectively. However, neither of them "consume" the array, since they are just modifying the shape and strides. Can these be changed to alter the arrays in-place? For a transition, perhaps we could provide `reverse_axes` and `permute_axes` as inplace variants, and either keep or deprecate these two methods?

akern40:
Closed by #1505

PR #1505, opened 2025-04-20
Axis permutation to correctly handle cycles using bitmask approach

NewBornRustacean:
## Overview
What's up! 😃 Currently, there're two functions to deal with permuting and reversing arrays(`permuted_axes` and `reversed_axes`).
As mentioned from related issue, they're not `in-place`, so I added `permute_axes` and `reverse_axes`. The new implementation for `permute_axes` uses a bitmask-based approach to efficiently track and process cycles in axis permutations!

## Related issue
- #1443 

## Changes
1. **permute_axes (in-place)**: 
   - A new function with a cycle detection algorithm based on bitmasking. 
   - try to maintain the zero-allocation, in-place modification approach
  
2. **reverse_axes (in-place)**:
    - simply use `&mut self`

### How it works
The new implementation uses a bitmask to track visited axes during permutation cycles:

1. **Cycle Detection**: When permuting axes, we identify cycles of positions that need to be updated.
2. **Bitmask Tracking**: A single `usize` is used as a bitmask to efficiently track which axes have been processed.
4. **In-place Updates**: Each cycle is processed in-place by following the permutation pattern.

### Example

Consider permuting a 2×2 array from `[[1, 2], [3, 4]]` with axes permutation `[1, 0]` (transpose):

```rust
// Original array
let mut arr = array![[1, 2], [3, 4]];
// Original: dims = [2, 2], strides = [2, 1]

// Permute axes [1, 0]
arr.permute_axes([1, 0]);
// Result: dims = [2, 2], strides = [1, 2]
// Array is now [[1, 3], [2, 4]]
```

Here's how the algorithm processes this:

1. Start with axis 1 (new_axis=0):
   - Store initial values: dim[1]=2, stride[1]=1
   - Follow cycle: 1 → 0 → 1
   - Update dim[1] = dim[0] = 2
   - Update stride[1] = stride[0] = 2
   - Mark axes 0 and 1 as visited

2. Process axis 0 (new_axis=1):
   - Already visited in the previous cycle, skip

The bitmask approach ensures each axis is processed exactly once, even in complex permutations with multiple cycles. The key operations are:
- `visited |= 1 << axis`: Mark axis as visited
- `(visited & (1 << axis)) != 0`: Check if axis is visited

@akern40 @nilgoyette added test cases similar to the original functions(`premuted_axes` and `reversed_axes`), but there might be missing cases. please take a look when you guys have some time! 🚀

### DIFF
--- a/src/impl_methods.rs
+++ b/src/impl_methods.rs
@@ -2596,11 +2596,11 @@ where
         for count in usage_counts.slice() {
             assert_eq!(*count, 1, "each axis must be listed exactly once");
         }
-        
+
         // Create temporary arrays for the new dimensions and strides
         let mut new_dim = D::zeros(self.ndim());
         let mut new_strides = D::zeros(self.ndim());
-        
+
         {
             let dim = self.layout.dim.slice();
             let strides = self.layout.strides.slice();
@@ -2609,10 +2609,13 @@ where
                 new_strides[new_axis] = strides[axis];
             }
         }
-        
+
         // Update the dimensions and strides in place
         self.layout.dim.slice_mut().copy_from_slice(new_dim.slice());
-        self.layout.strides.slice_mut().copy_from_slice(new_strides.slice());
+        self.layout
+            .strides
+            .slice_mut()
+            .copy_from_slice(new_strides.slice());
     }
 }
 
--- a/src/impl_methods.rs
+++ b/src/impl_methods.rs
@@ -2548,6 +2548,72 @@ where
         self.layout.strides.slice_mut().reverse();
         self
     }
+
+    /// Reverse the axes of the array in-place.
+    ///
+    /// This does not move any data, it just adjusts the array's dimensions
+    /// and strides.
+    pub fn reverse_axes(&mut self)
+    {
+        self.layout.dim.slice_mut().reverse();
+        self.layout.strides.slice_mut().reverse();
+    }
+
+    /// Permute the axes in-place.
+    ///
+    /// This does not move any data, it just adjusts the array's dimensions
+    /// and strides.
+    ///
+    /// *i* in the *j*-th place in the axes sequence means `self`'s *i*-th axis
+    /// becomes `self`'s *j*-th axis
+    ///
+    /// **Panics** if any of the axes are out of bounds, if an axis is missing,
+    /// or if an axis is repeated more than once.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// use ndarray::{arr2, Array3};
+    ///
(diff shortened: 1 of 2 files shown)


src/impl_methods.rs
```diff
@@ -2538,6 +2538,91 @@ where
...
+    /// This does not move any data, it just adjusts the array's dimensions
+    /// and strides.
+    ///
+    /// *i* in the *j*-th place in the axes sequence means `self`'s *i*-th axis
+    /// becomes `self`'s *j*-th axis
+    ///
+    /// **Panics** if any of the axes are out of bounds, if an axis is missing,
+    /// or if an axis is repeated more than once.    
+    ///
+    /// # About the Cycle Detection
```
nilgoyette, line 2552: I think this section belongs in a comment, not in the function documentation. IMO, the doc should explain **what** the function does, not **how** it's done.

src/impl_methods.rs
```diff
@@ -2538,6 +2538,91 @@ where
...
+    {
+        let axes = axes.into_dimension();
+        // Ensure that each axis is used exactly once.
+        let mut usage_counts = D::zeros(self.ndim());
+        for axis in axes.slice() {
+            usage_counts[*axis] += 1;
+        }
+        for count in usage_counts.slice() {
+            assert_eq!(*count, 1, "each axis must be listed exactly once");
+        }
```
nilgoyette, line 2607: What happens for empty arrays?

src/impl_methods.rs
NewBornRustacean: thank for the comment! I agree that _the doc should describe what the function does, not how it' done._ 
this block moved into in-line comment 😃

src/impl_methods.rs
NewBornRustacean: if I understand correctly, the validation is checking that each axis index is used exactly once in the permutation.
It doesn't care about the length of each dimension. The number of dimensions (ndim()) determines how many axes there are to permute, not the number of elements.

So even for empty arrays, the validation works normally and doesn't panic unless the permutation itself is invalid (like using an axis twice or missing an axis).

(and this block(2601 to 2067) is same as before: `permuted_axes` (2517 to 2525))
```rust
        let axes = axes.into_dimension();
        // Ensure that each axis is used exactly once.
        let mut usage_counts = D::zeros(self.ndim());
        for axis in axes.slice() {
            usage_counts[*axis] += 1;
        }
        for count in usage_counts.slice() {
            assert_eq!(*count, 1, "each axis must be listed exactly once");
        }
```

PR #1507, opened 2025-05-09
Add more element-wise math functions for floats

Waterdragen:
Extends the functionality of #1042

### Exponential and Logarithmic 
* exp_m1
* ln_1p

### Trig functions
* asin
* acos
* atan
* sinh
* cosh
* tanh
* asinh
* acosh
* atanh

### Other
* hypot

PR #1516, opened 2025-05-28
Implement float math functions for two arrays

Waterdragen:
### Rationale
Since we have something like `array.binary_fn(float)`, we could also have `array.binary_fn(array)`.
Extends functionality of #1042 and #1507

### New binary methods for RHS = float
- atan2

### New binary methods for RHS = arrays
- powi_all
- powf_all
- log_all
- abs_sub_all
- hypot_all
- atan2_all

### DIFF
--- a/src/numeric/impl_float_maths.rs
+++ b/src/numeric/impl_float_maths.rs
@@ -175,7 +175,9 @@ where
         fn atan2(A)
     }
     binary_ops_array! {
-        /// Integer power of each element.
+        /// Integer power of each element
+        ///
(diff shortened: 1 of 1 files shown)


nilgoyette:
Can you please add one test for at least one of those functions?

akern40:
I'm not sure this is the right approach here. If we grant that we want this (and I think we do?) I'd rather we take advantage of generics and traits to make it happen. I had a working example at #1469 of what this might look like.

We also should consider whether a method like this should perform broadcasting for the user or not.

PR #1532, opened 2025-10-18
Makes the reference types sound via a DST

### LINKED ISSUE #1531: Yanking 0.17.0 due to use-after-free unsoundness
The excellent @sarah-quinones has pointed out to me that the current design of `ArrayRef` is unsound and can cause use-after-free without the use of any `unsafe`. She has also provided a feasible path forward for a redesign that should be relatively small. However, I feel obliged to yank 0.17.0 to avoid leaving use-after-free inside the core of the library.

My apologies to everyone, especially those contributors whose contributions will be temporarily unavailable. I will work as fast as I can to get a new version up, which I will be submitting to some rigorous review before its release (0.17.1). I appreciate everybody's patience as I work through this.

akern40:
The current implementation of `ArrayRef` and its cousins has them as sized types, which turns out to be a critical and unsound mistake. This PR is large, but its heart is small: change the `ArrayRef` implementation to be unsized.

The approach this PR takes is to package the core array "metadata" - the pointer, dim, and strides - into a struct that can either be sized or unsized. This is done by appending a generic "`_dst_control`" field. For the "sized" version of the metadata, that field is a 0-length array. For the "unsized" version of the metadata, that sized field is a struct. This core type is private, so users cannot construct any other variants other than these two.

We then put the sized version into the `ArrayBase` types, put the unsized version into the reference types, and perform an unsizing coercion to convert from one to the other. Because Rust has no (safe, supported) "resizing" coercion, this switch is irreversible. Sized types cannot be recovered from the unsized reference types.

Although this PR contains considerable code changes, most are renames and minor changes from altering the internal structure of the array types. To make this PR easier to review, I'll give a list of major changes in the order I'd review them. Then, in comments below, I will highlight those changes in the same order:
- `ArrayParts` definition
- Making `LayoutRef` a newtype
- The mechanism for actually dereferencing the array types into the reference types

Closes #1531.

---

An alternative method would be to use a trait object to create the DST. I'd originally used this approach, but was worried that embedding a vtable into the heart of the library's type could harm performance. One way around this would be to cast the trait object's fat pointer to a thin pointer, discarding the vtable. However, I could not find any guarantees in the [Rust type layout](https://doc.rust-lang.org/beta/reference/type-layout.html) documentation nor in the [Rustonomicon on DSTs](https://doc.rust-lang.org/nomicon/exotic-sizes.html) about the layout of DSTs. Relying on behavior that not even the Rustonomicon will guarantee gives me the heebie-jeebies, and the current method appears equally viable.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1437,7 +1437,8 @@ impl<A, D> ArrayPartsSized<A, D>
 // which alter the layout / shape / strides of an array must also
 // alter the offset of the pointer. This is allowed, as it does not
 // cause a pointer deref.
-pub struct LayoutRef<A, D>(ArrayParts<A, D, [usize]>);
+#[repr(transparent)]
+pub struct LayoutRef<A, D>(ArrayPartsUnsized<A, D>);
 
 impl<A, D> LayoutRef<A, D>
 {


src/lib.rs
```diff
@@ -1299,7 +1301,41 @@ where S: RawData<Elem = A>
...
+    ptr: NonNull<A>,
+    /// The lengths of the axes.
+    dim: D,
+    /// The element count stride per axis. To be parsed as `isize`.
+    strides: D,
+    _dst_control: T,
+}
+
+type ArrayPartsSized<A, D> = ArrayParts<A, D, [usize; 0]>;
+type ArrayPartsUnsized<A, D> = ArrayParts<A, D, [usize]>;
```
akern40: This is the first major review point: defining a possibly-sized type to hold the array parts.

Thanks to the use of an explicitly 0-length array, the `ArrayPartsSized` type is no larger than just the `ptr`, `dim`, and `strides` packaged together.

src/impl_ref_types.rs
```diff
@@ -50,7 +63,9 @@ where S: Data
         // - It is "dereferencable" because it comes from a reference
         // - For the same reason, it is initialized
         // - The cast is valid because ArrayRef uses #[repr(transparent)]
-        unsafe { &*(&self.layout as *const LayoutRef<S::Elem, D>).cast::<ArrayRef<S::Elem, D>>() }
+        let parts: &ArrayPartsUnsized<S::Elem, D> = &self.parts;
+        let ptr = (parts as *const ArrayPartsUnsized<S::Elem, D>) as *const ArrayRef<S::Elem, D>;
+        unsafe { &*ptr }
```
akern40: The third, and final, major review point. This particular set of lines is the dereference from `ArrayBase` to `ArrayRef`; but the core idea is repeated many times across this file for all of the `Deref`, `DerefMut`, `AsRef`, and `AsMut` implementations.

Our first line is now an unsizing coercion - going from `Sized` to a DST. The next line converts a reference to a pointer, then converts a pointer to `ArrayPartsUnsized` to `ArrayRef`. Since `ArrayRef` is transparent to `LayoutRef` which is transparent to `ArrayPartsUnsized`, this is valid. Finally, we reborrow the pointer.

src/impl_ref_types.rs
```diff
@@ -368,3 +394,17 @@ where S: RawData<Elem = A>
...
+/// let mut x = Array1::from_vec(vec![0, 1, 2]);
+/// {
+///     let mut y = Array1::from_vec(vec![4, 5, 6]);
+///     let x_ref = x.as_layout_ref_mut();
+///     let y_ref = y.as_layout_ref_mut();
+///     core::mem::swap(x_ref, y_ref);
+/// }
+/// ```
+#[allow(dead_code)]
+fn test_no_swap_via_doctests() {}
```
akern40: And one bonus section: a doctest to assert that the original use-after-free scenario no longer compiles.

src/lib.rs
bluss: Here we could add `+ ?Sized` too maybe?

akern40: Definitely - I'll add `+ ?Sized` across the board in another documentation PR. I'm also going to write up an explainer for the release notes about how to write functions for `LayoutRef` and `RawRef`.

src/lib.rs
akern40: I thought about this more, and realized don't require the `Sized` and `Unsized` versions of array types to be transmutable, we just rely on [Unsized Coercions](https://doc.rust-lang.org/reference/type-coercions.html#r-coerce.unsize). The only memory-layout guarantee we rely on is on the `#[repr(transparent)]` of the reference types.

I _do_ think that this would be a cleaner internal API; however, it increases the verbosity of a good number of internal implementations. Since these are private types, I'm going to defer this change until there's a need to decouple the parts from the sizing control.

PR #1541, opened 2025-11-03
Add release log for ndarray-rand 0.16.0

akern40:


akern40:
@bluss looks like you'll need to add the `release` GitHub group as an owner on crates.io for `ndarray-rand` before I can release the next version.

Issue #1517, opened 2025-06-03
any equivalent to np.tensordot?


hugolgst:
https://numpy.org/doc/2.1/reference/generated/numpy.tensordot.html

hugolgst:
there's https://crates.io/crates/ndarray_einsum_beta

jmg049:
I'm preparing a PR for this right now. I have the implementation ready and just need to focus on some tests, docs and where exactly to put the functionality. 

I think in my fork I have it under linalg and impl_linalg.rs

Generally speaking I think there's probably a few more functions like tensordot and also the phase angle functionality I have in another PR.

I think they're a low hanging fruit of sorts since we have solid reference implementations (Numpy, etc) and they are mostly self contained and do not interfere with broader functionality beyond making the codebase a bit bigger and running some more tests.

But adding them provides some benefit to users who find themselves reaching for them.

PR #1545, opened 2025-11-08
Tensordot implementation

jmg049:
Tensordot implementation [Issue #1517](github.com/rust-ndarray/ndarray/issues/1517)

Adds a full implementation of `tensordot` for n-dimensional arrays,
supporting both numeric and paired axis specifications via the new
`AxisSpec` enum. The design mirrors NumPy’s `tensordot` behaviour while
integrating cleanly with ndarray’s trait-based approach for the ``dot``
product and existing shape and stride logic.

All internal reshaping and permutation operations use `unwrap` and
`expect` with explicit safety reasoning: each call is guarded by
dimension and axis validation, ensuring panics can only occur under
invalid `AxisSpec` input. Documentation and inline comments
describe these invariants and the exact failure conditions.

Includes tests verifying correct contraction for
both paired and integer axis modes.

### DIFF
--- a/src/linalg/impl_linalg.rs
+++ b/src/linalg/impl_linalg.rs
@@ -1252,14 +1252,14 @@ pub trait Tensordot<Rhs: ?Sized>
     /// - The computed product of reshaped dimensions does not equal the
     ///   array’s total element count (which would indicate internal logic error).
     #[track_caller]
-    fn tensordot(&self, rhs: &Rhs, axes: AxisSpec) -> Self::Output;
+    fn tensordot(&self, rhs: &Rhs, axes: &AxisSpec) -> Self::Output;
 }
 
 /// Perform a tensor contraction along specified axes.
 ///
 /// See [`Tensordot::tensordot`] for more details.
 #[track_caller]
-pub fn tensordot<T, Sa, Sb, Da, Db>(a: &ArrayBase<Sa, Da>, b: &ArrayBase<Sb, Db>, axes: AxisSpec) -> ArrayD<T>
+pub fn tensordot<T, Sa, Sb, Da, Db>(a: &ArrayBase<Sa, Da>, b: &ArrayBase<Sb, Db>, axes: &AxisSpec) -> ArrayD<T>
 where
     T: LinalgScalar,
     Sa: Data<Elem = T>,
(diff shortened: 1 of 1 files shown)


src/linalg/impl_linalg.rs
```diff
@@ -1140,3 +1140,379 @@ where A: LinalgScalar
...
+    ///
+    /// Negative indices are supported and count from the end
+    /// (e.g. `-1` refers to the last axis).
+    ///
+    /// # Example
+    /// ```
+    /// # use ndarray::linalg::AxisSpec;
+    /// let axes = AxisSpec::Pair(vec![1, -1], vec![0, 2]);
+    /// ```
+    Pair(Vec<isize>, Vec<isize>),
```
nilgoyette: I know both vectors must be of equal length, but must the number of elements be equal to the dimension?

I always try to reduce the number of memory allocations. I wonder if there's a better way than to accept 2 vectors.

src/linalg/impl_linalg.rs
```diff
@@ -1140,3 +1140,379 @@ where A: LinalgScalar
...
+    let a2 = a_std
+        .into_shape_with_order(Ix2(m, k))
+        .expect("reshaping a to 2D");
+    let b2 = b_std
+        .into_shape_with_order(Ix2(k, n))
+        .expect("reshaping b to 2D");
+
+    let c2 = a2.dot(&b2);
+
+    let mut out_shape: Vec<usize> = notin_a.iter().map(|&ax| a.shape()[ax]).collect();
```
nilgoyette: You're allocating 3 vectors in this function. Can't you re-use one of them for `out_shape`?

src/linalg/impl_linalg.rs
jmg049: In tensordot, the two axis lists do not need to match the full dimension of either tensor, they only list the contracted axes. Their only requirement is that they have the same length and that corresponding axes match in size. Users can contract any subset of axes in any order, so the number of axes is a runtime value, not a compile-time constant.

Because of that, constant-size arrays are not a good fit. The maximum number of contracted axes is not known in advance. In practice these vectors are very small, so the allocation cost is negligible.

There is also no guarantee that one tensor has equal or larger rank than the other, only the selected axes must align.

PR #1546, opened 2025-11-12
Allow cross compilation with BLAS on ARM bare metal (no libc)

unovongalixor:
Replaces libc references with core::ffi, and update cblas-sys to 0.3.0 which makes the same change there.

Following this change, you can add the following to your Cargo.toml

```
ndarray = { version = "0.17.1", default-features = false, features = ["blas"] }
blas-src = { version = "0.14", features = ["openblas"] }
openblas-src = { version = "0.10.13", features = ["cblas", "static"] }

```
and compile a static lib for --target=thumbv7em-none-eabi by providing the following ENV variables:

NO_SHARED=1 
NO_FORTRAN=1 
USE_THREAD=0 
EMBEDDED=1 
ONLY_CBLAS=1
ONLY_LAPACK=1
CROSS=1
OPENBLAS_TARGET=ARMV7 
OPENBLAS_CC=arm-none-eabi-gcc
HOST_CC=gcc

naturally you need the arm-none-eabi-gcc toolchain. Since it's a bit hard to configure, I figured i'd share the details in case you want it for documentation.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -109,12 +109,9 @@ checksum = "1fd0f2584146f6f2ef48085050886acf353beff7305ebd1ae69500e27c67f64b"
 
 [[package]]
 name = "cblas-sys"
-version = "0.1.4"
+version = "0.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b6feecd82cce51b0204cf063f0041d69f24ce83f680d87514b004248e7b0fa65"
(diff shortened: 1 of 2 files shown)


akern40:
Hello! I'm not super familiar with cross-compilation or ARM specifically. You mention that this requires some configuration; was this configuration always necessary? Or is there something about this change that has made it harder to configure?

Issue #1548, opened 2025-11-24
Access to &ArrayRef from multiple threads


gaumut:
I’m currently updating my code to the new ArrayRef type and I have some problem with access to &ArrayRef from multiple threads (see example below).

    use ndarray::Zip;
    use ndarray::prelude::*;
    
    fn test_ref(a: &ArrayRef1<f64>) { // does not compile
        let b = array![0.];
        Zip::from(&b).par_for_each(|_| {
            let _ = a[0];
        });
    }

    fn test_view(a: &ArrayView1<f64>) { // works
        let b = array![0.];
        Zip::from(&b).par_for_each(|_| {
            let _ = a[0];
        });
    }
    
    fn main() {
        let a = array![1.];
        test_view(&a.view());
        test_ref(&a);
    }

Attached is the error I get.

[error.log](https://github.com/user-attachments/files/23723771/error.log)

Is this expected ? Should I stick to ArrayView for those function ?

PR #1552, opened 2025-12-03
Implement Sync for ArrayParts

gaumut:
Closes https://github.com/rust-ndarray/ndarray/issues/1548


gaumut:
Also I saw that Send is implemented on ArrayBase, if this PR is accepted maybe Send should also be implemented on ArrayRef.

PR #1553, opened 2025-12-08
Only include files that are necessary for crates.io

SwishSwushPow:
Hi everyone 👋

While reviewing dependency updates in our project (trying to assure supply chain safety) I noticed that there are a couple of scripts, tests, benchmarks etc. that are not necessarily required to be published to crates.io. They make it harder to review `ndarray` when checking the supply chain and I was wondering if it would be possible to remove these items from the published package. That would remove potential vectors for a security vulnerability in the future and it would also shrink the size of `ndarray` from 309.5KiB to 237.6KiB compressed. :)

The downside of course would be that e.g. the tests couldn't be run from the crate package anymore, but I'm not sure how popular that is.

I've tried to include all the files that are required (licenses) and that make reviewing things a bit easier (readme and release files can give a good context what has changed between versions).

Best regards!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,14 @@ description = "An n-dimensional array for general elements and for numerics. Lig
 keywords = ["array", "data-structure", "multidimensional", "matrix", "blas"]
 categories = ["data-structures", "science"]
 
-exclude = ["docgen/images/*"]
+include = [
+    "/src/**/*.rs",
+    "LICENSE-MIT",
+    "LICENSE-APACHE",
+    "RELEASES.md",
+    "README.rst",
+    "README-quick-start.md"
+]
 resolver = "2"
 
 [lib]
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -19,7 +19,7 @@ description = "An n-dimensional array for general elements and for numerics. Lig
 keywords = ["array", "data-structure", "multidimensional", "matrix", "blas"]
 categories = ["data-structures", "science"]
 
-exclude = ["docgen/images/*"]
+include = ["src", "LICENSE-MIT", "LICENSE-APACHE", "RELEASES.md", "README.rst", "README-quick-start.md"]
```
nilgoyette: I read the documentation and I think what you wrote is ok. However:

-  I wonder about `ndarray-rand`. Shouldn't it be included?
- I'd prefer using `/src` than `src`.

nilgoyette: You're right, `ndarray-rand` is another crate. Thank you for your contribution.

Issue #1175, opened 2022-07-01
How to make zip work with more than 6 elements?


T4varo:
I want to use zip (or azip!) with 7 producers but its only implemented for max 6 producers. Is there a convenient way to also implement it for more producers? 

Minimal Example Code:
`let v1: Array1<i32> = Array1::zeros(3);
let v2: Array1<i32> = Array1::zeros(3);
let v3: Array1<i32> = Array1::zeros(3);
let v4: Array1<i32> = Array1::zeros(3);
let v5: Array1<i32> = Array1::zeros(3);
let v6: Array1<i32> = Array1::zeros(3);
let v7: Array1<i32> = Array1::zeros(3);
Zip::from(&v1)
	.and(&v2)
	.and(&v3)
	.and(&v4)
	.and(&v5)
	.and(&v6)
	.and(&v7)
	.for_each(|v1, v2, v3, v4, v5, v6, v7| {
		println!("{}, {}, {}, {}, {}, {}, {}", v1, v2, v3, v4, v5, v6, v7);
	});`

Error:

> error[E0599]: no method named `and` found for struct `ndarray::Zip<(ArrayBase<ViewRepr<&i32>, Dim<[usize; 1]>>, ArrayBase<ViewRepr<&i32>, Dim<[usize; 1]>>, ArrayBase<ViewRepr<&i32>, Dim<[usize; 1]>>, ArrayBase<ViewRepr<&i32>, Dim<[usize; 1]>>, ArrayBase<ViewRepr<&i32>, Dim<[usize; 1]>>, ArrayBase<ViewRepr<&i32>, Dim<[usize; 1]>>), Dim<[usize; 1]>>` in the current scope
  --> src\main.rs:47:14
   |
47 |             .and(&v7)
   |              ^^^ method not found in `ndarray::Zip<(ArrayBase<ViewRepr<&i32>, Dim<[usize; 1]>>, ArrayBase<ViewRepr<&i32>, Dim<[usize; 1]>>, ArrayBase<ViewRepr<&i32>, Dim<[usize; 1]>>, ArrayBase<ViewRepr<&i32>, Dim<[usize; 1]>>, ArrayBase<ViewRepr<&i32>, Dim<[usize; 1]>>, ArrayBase<ViewRepr<&i32>, Dim<[usize; 1]>>), Dim<[usize; 1]>>`
   |
   = note: the method was found for
           - `ndarray::Zip<(P1,), D>`
[...]

PR #1559, opened 2025-12-16
Expand Zip to up to 9 producers

felsenhower:
As has been discussed in #1175, I expanded `Zip` for up to 9 producers (including `azip!`, `par_azip!`). Previously, it was only working for up to 6 producers.

I also added two simple tests with 9 producers.


felsenhower:
> You still have a tiny fmt error.

Apologies again. That should've fixed it now.

PR #1561, opened 2025-12-24
Add type aliases for higher-dimensional ArcArrays

varchasgopalaswamy:
Added type aliases for 3-6D ArcArrays


akern40:
Could you also add `ArcArrayD`?

akern40:
Merged! Thanks for your contribution!

Issue #1506, opened 2025-04-24
RFC: Comprehensive Plan for Restructuring Dimension Types


akern40:
# Summary

This RFC proposes a long-term plan for restructuring the library's core types and approach for handling shapes, strides, and dimensionalities.
A successful implementation of new dimension types should accomplish the following goals:

1. Make the library friendlier to newcomers and more ergonomic for power users by reducing confusion around dimension types
2. Store strides as `isize`, not `usize`
3. Maintain the ability to capture the dimensionality of an array at the type level
4. Allow for capturing axis lengths at the type level, and therefore allow compile-time constant optimizations
5. Support the move towards a "trait-based" library
6. Reduce the number of "sealed" traits, thereby making the library more extensible
7. Incorporate, as best as possible, const generics for dimensionality and fixed-length axes
8. Provide a strong basis for other feature enhancements, such as optimized iteration orders and named axes

This RFC is _not_ meant to propose all the changes at once. Rather, this RFC lays out the overall vision and how we will build a smooth transition path. Each step in that path will require its own set of issues and pull requests.

An experimental implementation of this RFC (but not its integration into `ndarray`) can be found [here](https://github.com/akern40/ndarray-design/blob/dimensions/src/core/layout.rs) (for the `Layout` trait and module definition) and [here](https://github.com/akern40/ndarray-design/tree/dimensions/src/core/layout) for the rest of the traits and types.
It includes types that implement the traits, additional design notes and rationale, and examples of advanced usage (such as a completely-constant two-by-two matrix shape).

cc @bluss @jturner314 @NewBornRustacean @lucascolley 

# Motivation

## Clarity

The current handling of shape and strides in the library is a point of confusion for users (#489, #367, [this comment](https://github.com/rust-ndarray/ndarray/issues/1272#issuecomment-2240875720)) and a point of complexity for maintainers (#283) since at least 2017.
The dimension traits and types in `ndarray` are, at least for me, the most difficult part of the library to understand.
I managed to write an entire RFC to revamp the array types themselves without truly understanding the current setup for dimension types.
This is a testament to both the complexity of the implementation details and to the power of the current solution.

## Features

In addition to complexity, the current approach limits some very promising and important optimizations and extensions.
As @grothesque points out in [this comment](https://github.com/rust-ndarray/ndarray/issues/1272#issuecomment-2241818760) on #1272, fixed-length axes are a critical step for many powerful optimizations.
[...]

PR #1565, opened 2026-01-04
Configure docs for feature gates globally

akern40:
In #1479 we used `cfg_attr(docsrs, doc(cfg(feature = "...")))` to add feature-gating information to the documentation. However, this required us to "remember" this at every call site. It turns out that there is a `doc(auto_cfg)` setting that we can use globally to do this for us automatically.


akern40:
Thanks! I ran across it by accident while trying to figure out how we might start marking some new code as "unstable" in our documentation (for some work on #1506 that I'm hoping to start merging...).

PR #1571, opened 2026-01-09
Allow usage and borrow cleanups

RPG-Alex:
I noticed several duplicated `clippy::many_single_char_names` allows across files, removed a few unnecessary uses of the `vec!` macro for literal values, and cleaned up some redundant `clone()` calls where the type implements `Copy`. I also elided one unnecessary lifetime annotation and removed a few cases of unnecessary borrowing. Also ran `cargo fmt --all` for good measure.

### DIFF
--- a/tests/oper.rs
+++ b/tests/oper.rs
@@ -146,8 +146,6 @@ fn dot_product()
         assert_abs_diff_eq!(a2.dot(&b2), reference_dot(&a2, &b2), epsilon = 1e-5);
     }
 
-    let a = a.map(|f| *f);
-    let b = b.map(|f| *f);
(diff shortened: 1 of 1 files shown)


RPG-Alex:
> Thanks for the cleanup. What were you using to catch all of these? Did you do it by hand? If there's a way to automate it, we should add it to the CI

Hey, I actually run:
`cargo clippy --all-targets --all-features -- -D warnings` to get the most pedantic version of clippy (that I am aware of). That was the entry point for most of the small changes I made. 

I didn't push it in the github workflow because it may be overly strict but I can add it if its desired

PR #1575, opened 2026-01-11
Add a trait to tie together compile-time and runtime number of dimensions

akern40:
This is the next PR to address #1506. In the previous PR (#1568) we established a trait for capturing the "dimensionality" or "rank" of an array at the type level. This PR is focused on providing a bridge from type-level dimensionality to runtime dimensionality, which we do through the new `Ranked` trait.

First, a note on language: we now have three words / phrases to refer to the same thing: number of dimensions, dimensionality, and rank. That's ok, they're all accurate, but I'm worried it will be confusing as time goes on. Similarly, in #1568, we decided on using `Rank` as the name of the associated type that tells you the compile-time number of dimensions. However, the _function_ that has traditionally provided the number of dimensions is `ndim`. So we have a slight vocabulary problem. I would really love some input and feedback about what language we should be using; especially from people for whom English is not their first / primary language.

The `Ranked` trait is pretty simple: require an associated type (`Rank`) that carries compile-time dimensionality, and a function `fn rank(&self) -> usize` that tells the user runtime dimensionality. The `src/layout/ranked.rs` file contains the definition, implementations, and blanket implementations. There are likely some/many types in the library or outside of it that I forgot to implement `Ranked` for, but I'm sure we'll find them as time goes on.

I've also removed the `Rank` associated type on `Dimension` and made `Ranked` a supertrait of `Dimension`. I think this is the best way to do this, but it does mean that I've had to remove the `"unstable"` feature flag we had introduced. Since this code isn't merging to main yet, I think that's fine.

### DIFF
--- a/src/layout/ranked.rs
+++ b/src/layout/ranked.rs
@@ -34,7 +34,7 @@ pub trait Ranked
 {
     /// The compile-time rank of the type; can be [`DynRank`][DynRank] if unknown.
     ///
-    /// [DynRank]: crate::layout::dimensionality::DynRank
+    /// [DynRank]: crate::layout::rank::DynRank
     type NDim: Rank;
 
     /// The runtime number of dimensions of the type.


akern40:
Names have been changed; @nilgoyette let me know what you think.

PR #1576, opened 2026-01-17
docs/tests: fix rustdoc links and appease clippy strict lints

RPG-Alex:
Tighten linting; fix rustdoc warnings across feature gates

I continued using stricter linting and found several odd cases in tests and
examples, along with rustdoc warnings triggered by feature-gated items.

This commit cleans up those warnings by:
- adjusting tests to avoid clippy false positives under -D warnings
- removing unused generics and ambiguous empty vec initializers
- fixing rustdoc links via feature-gated doc attributes rather than
  linking to non-exported items

All tests are still passing.

Clippy still warns about the use of `1..-1` as an empty range; this appears
intentional in ndarray slicing semantics, so I’ve left it unchanged for now.

Please review the rustdoc changes that rely on feature-gated documentation.

### DIFF
--- a/examples/sort-axis.rs
+++ b/examples/sort-axis.rs
@@ -20,10 +20,14 @@ pub struct Permutation
 impl Permutation
 {
     /// Checks if the permutation is correct
-    pub fn from_indices(v: Vec<usize>) -> Option<Self>
+    pub fn from_indices(v: Vec<usize>) -> Result<Self, ()>
     {
         let perm = Permutation { indices: v };
-        perm.correct().then_some(perm)
+        if perm.correct() {
+            Ok(perm)
+        } else {
+            Err(())
+        }
     }
 
     fn correct(&self) -> bool


examples/sort-axis.rs
```diff
@@ -20,14 +20,10 @@ pub struct Permutation
 impl Permutation
 {
     /// Checks if the permutation is correct
-    pub fn from_indices(v: Vec<usize>) -> Result<Self, ()>
+    pub fn from_indices(v: Vec<usize>) -> Option<Self>
```
nilgoyette: I understand that you didn't "change" anything, this is mostly reformatting, but this one here is different. This is a public function, so there might be several callers out there who will be surprised by this API change.

akern40: Personally I'm ok with this, but I'm glad you brought it up

akern40: Agreed that it's the better practice, just good for us to keep in mind it will have to be a breaking change release.

akern40:
@RPG-Alex same question as before, is this just another round of `cargo clips --all-features --all-targets -- -D warnings`? Or something else? I do think I'll want to add these to CI

src/linalg/impl_linalg.rs
akern40: Why did a large block of code move in this file?

Cargo.toml
akern40: To my knowledge, there is no Rust-mandated or even Rust-suggested TOML formatter, is there? So these changes could be undone as soon as I hit save on my `Cargo.toml`? If we want to enforce formatting for non-Rust files, we'll need to do it with something more standardized.

src/partial.rs
```diff
@@ -60,6 +62,7 @@ impl<T> Partial<T>
     }
 
     #[cfg(feature = "rayon")]
+    #[allow(dead_code)]
     /// Merge if they are in order (left to right) and contiguous.
     /// Skips merge if T does not need drop.
     pub(crate) fn try_merge(mut left: Self, right: Self) -> Self
```
akern40: Under what check is this dead code? I can clearly see that it is called inside the `zip_impl` macro

akern40: Ah interesting, I hadn't seen that comment. Personally, I'd only use the `allow(dead_code)` directive when Rust explicitly calls it dead code - so I'd remove it from there

src/linalg/impl_linalg.rs
akern40: ah haha! Ok cool, thank you

Cargo.toml
RPG-Alex: Ok. Luckily I did that all in one PR, so just reverted it.

examples/sort-axis.rs
akern40: If it's not too much hassle, I'd appreciate if we put it in another PR. Just for the sake of clarity

PR #1588, opened 2026-03-08
Add optional half crate support for ScalarOperand

emiddleton:
Add `half` feature flag with optional dependency on the `half` crate. When enabled, implements `ScalarOperand` for `half::f16` and `half::bf16`, allowing these types to participate in array arithmetic operations.

This enables the burn-ndarray backend to support bf16/f16 tensor operations.


nilgoyette:
Thanks for your contribution, Edward. However, [this PR](https://github.com/rust-ndarray/ndarray/pull/1551) does the same, and more, so we're gonna continue with that one.

PR #1592, opened 2026-05-01
Accept owned arrays in `concatenate` and `stack`

### LINKED ISSUE #1591: Accept owned arrays in `concatenate` and `stack`
It would be convenient to be able to call `concatenate` or `stack` with a slice of owned arrays or array views, rather than just views currently.

So for `concatenate`, this:
```rust
let arrays = vec![a, b, c]; // more realistically, this would be returned by a function
let arrays = arrays.iter().map(|a| a.view()).collect::<Vec<_>>();
let res = ndarray::concatenate(Axis(2), &arrays);
```
would become:
```rust
let arrays = vec![a, b, c];
let res = ndarray::concatenate(Axis(2), &arrays);
```

This could be done by changing the signature of `concatenate` to the following, which would allow either views or owned arrays:
```diff
-pub fn concatenate<A, D>(axis: Axis, arrays: &[ArrayView<A, D>]) -> Result<Array<A, D>, ShapeError>
+pub fn concatenate<S, D, A>(axis: Axis, arrays: &[ArrayBase<S, D, A>]) -> Result<Array<A, D>, ShapeError>
 where
+    S: Data<Elem = A>,
     A: Clone,
     D: RemoveAxis,
 {
```

This is technically a breaking API change though, at least in the niche case of passing an empty slice as a parameter (which causes a type inference error). Before I open an unsolicited PR, would there be interest in this change?

watsaig:
Closes #1591

The doc comments still sound correct imo, but please let me know if you think they need changes (or feel free to push changes yourself).


src/stacking.rs
```diff
@@ -33,8 +33,9 @@ use crate::imp_prelude::*;
 ///                  [3., 3.]]))
 /// );
 /// ```
-pub fn concatenate<A, D>(axis: Axis, arrays: &[ArrayView<A, D>]) -> Result<Array<A, D>, ShapeError>
+pub fn concatenate<S, D, A>(axis: Axis, arrays: &[ArrayBase<S, D, A>]) -> Result<Array<A, D>, ShapeError>
```
nilgoyette: Can't you use `ArrayBase<S, D>` here and below? I see several examples in other files. The `A` seems optional.

watsaig: Yes, good point. It wasn't technically needed on the return type either so I removed it entirely

akern40:
Would you mind rebasing on master and running the formatter with the nightly toolchain? Would be happy to merge after that

watsaig:
> Would you mind rebasing on master and running the formatter with the nightly toolchain? Would be happy to merge after that

Done! I ran `cargo +nightly fmt` which mostly reformatted files unrelated to my changes, I hope that's ok.

akern40:
Not sure what happened with your formatting command, but I went ahead and redid the formatting, we should be good to go now!

akern40:
(For reference I also just ran `cargo +nightly fmt` from the root)

PR #1593, opened 2026-05-30
Add contains() method to check if an index is within array bounds

### LINKED ISSUE #1247: Array bounds/contains check?
I find myself very frequently having a coordinate and an ndarray Array2/Array3. I want to check whether the coordinate is within the bounds of the Array2/Array3 (IE, gte than 0,0,0, less than the array size). There doesn't appear to be a way to do this. There is "get" to return a value or None but that is not the same thing. Sometimes I want to know if the value is inbounds without, at that time, accessing the value.

It would be helpful to add a within(), contains() etc function to ArrayBase (I *think* adding it to Dim would also serve equivalent purposes).

(If this already exists and I just missed it, sorry)

theoxfaber:
Adds a contains() method to ArrayBase to check whether a given index falls within array dimensions without accessing the element.

Closes #1247

let a = arr2(&[[1., 2.], [3., 4.]]);
assert!(a.contains((0, 1)));
assert!(!a.contains((2, 0)));

### DIFF
--- a/src/impl_methods.rs
+++ b/src/impl_methods.rs
@@ -823,6 +823,24 @@ impl<A, D: Dimension> ArrayRef<A, D>
     {
         unsafe { self.get_ptr(index).map(|ptr| &*ptr) }
     }
+
+    /// Return `true` if the index is within the array bounds.
+    ///
+    /// ```
+    /// use ndarray::arr2;
+    ///
+    /// let a = arr2(&[[1., 2.],
+    ///                [3., 4.]]);
+    ///
+    /// assert!(a.contains((0, 1)));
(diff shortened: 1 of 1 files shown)


akern40:
Could you please add a test or two on the new method? And rebase off of master as well, to fix potential CI issues

theoxfaber:
Rebased on master and added 3 tests (1D, 2D, and empty array). All 194 tests pass locally including the new ones.

akern40:
Sorry I see a force push but it looks like it just contains your changes still

theoxfaber:
yeah rebased on master and added the tests you asked for. thanks for checking.

PR #1594, opened 2026-05-31
feat: derive clone for linspace

skrobchik:
It would be useful for `Linspace` to implement `Clone` when the element type is `Clone`.

For example, iterating over a 2D grid with `itertools`' `cartesian_product` requires the iterators to implement `Clone`.

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=3ff8d5b95c2b6acef66c577dc9205f5d

which currently does not compile:

```
error[E0277]: the trait bound `Linspace<{float}>: Clone` is not satisfied
    --> src/main.rs:5:60
     |
   5 |     for (x, y) in linspace(-1.0, 1.0, 3).cartesian_product(linspace(-1.0, 1.0, 3)) {
     |                                          ----------------- ^^^^^^^^^^^^^^^^^^^^^^ the trait `Clone` is not implemented for `Linspace<{float}>`
     |                                          |
     |                                          required by a bound introduced by this call
```

> For a generic struct, #[derive] implements Clone conditionally by adding bound Clone on generic parameters.
https://doc.rust-lang.org/std/clone/trait.Clone.html


akern40 (approved):

akern40:
Seems sane, happy to accept after CI finishes, thanks!

PR #1599, opened 2026-06-08
refactor: add a trait to describe shapes

akern40:
This is the next PR to address #1506. The aim of this PR is to introduce a minimal trait for representing shapes - an ordered collection of axis lengths for an array. Although this is a starting point for the trait, I suspect I will have to add / modify it as I move through the refactor and see just exactly how/when a `Shape` is used.

A few design quirks of this trait. First of all, there is no `Index` bound on `Shape`, or a default `Index` implementation. Unfortunately, `Index` requires returning `&T`: it requires that the return be borrowed from the indexed type. One of the explicit goals of this rework is to allow constant-sized shapes, including shapes which are entirely defined at compile time and zero-sized. These shapes wouldn't have any value to lend out for the borrow. The same goes for shapes which are computed.

Instead, this design makes two requirements of the shape. First, a function, `axis_len`, that returns the length of an axis. Second, that the shape can be iterated over.

We use these required methods to define several checks and derived methods on `Shape`, which are mostly concerned with the existing safety limit that `ndarray` imposes under the hood: that the number of elements and number of bytes in an array both be less than or equal to `isize::MAX`.

I also provide structs that implement `Shape` for both `const`-dimensionality and dynamic dimensionality.

### DIFF
--- a/src/layout/dyn_repr.rs
+++ b/src/layout/dyn_repr.rs
@@ -1,13 +1,13 @@
 use alloc::boxed::Box;
 use alloc::vec::Vec;
+use core::iter::Cloned;
 use core::ops::{Add, AddAssign, Mul, MulAssign, Sub, SubAssign};
 use core::ops::{Deref, DerefMut, Index, IndexMut};
 use core::slice::Iter;
-use num_traits::Zero;
 
 use crate::layout::rank::DynRank;
 use crate::layout::ranked::Ranked;
-use crate::layout::shape::{IntoShape, Shape};
+use crate::layout::shape::Shape;
 use crate::Axis;
 
 const CAP: usize = 4;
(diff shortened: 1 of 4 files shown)


src/layout/dyn_repr.rs
```diff
@@ -0,0 +1,188 @@
...
+use crate::layout::ranked::Ranked;
+use crate::layout::shape::Shape;
+use crate::Axis;
+
+const CAP: usize = 4;
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum DynAxesRepr<T>
+{
+    Inline(usize, [T; CAP]),
```
nilgoyette, line 18: I don't understand the limit of 4 here.

PR #1600, opened 2026-06-19
Make ArrayViewMut::into_view public to allow lifetime preservation during conversion

### LINKED ISSUE #1595: `ArrayViewMut::into_view()` should be public
There is no public API to convert an owned `ArrayViewMut<'a, T, D>` to `ArrayView<'a, T, D>` while preserving the original lifetime `'a`. The method `into_view()` exists internally but is `pub(crate)`.

`ArrayViewMut` has a `.view()` method, but it cannot be used here because its signature borrows `self`. A simplified example of my use case where I have an `ArrayViewMut` inside a `struct`:

```rust
use ndarray::{ArrayView2, ArrayViewMut2};

struct Foo<'a> {
    data: ArrayViewMut2<'a, f32>,
}

impl<'a> Foo<'a> {
    /// Downgrade to a shared view. This fails to compile.
    fn into_shared(self) -> ArrayView2<'a, f32> {
        self.data.view() // error[E0515]: returns value referencing local `self.data`
    }
}
```

The compiler complains:

```
error[E0515]: cannot return value referencing local data `self.data`
  --> src/main.rs:10:9
   |
10 |         self.data.view()
   |         ^^^^^^^^^------
   |         |        |
   |         |        `self.data` is borrowed here
   |         returns a value referencing data owned by the current function
```

The data pointed to by self.data is valid for 'a, but .view() can only express "valid for the duration of the &self.data borrow", which expires when self is dropped at the end of into_shared. The internal ArrayViewMut::into_view(self) -> ArrayView<'a, A, D> handles this correctly but is pub(crate).

I can work around this with `unsafe { std::mem::transmute(...) }` as a workaround which is valid but gross.

# Request

Make `into_view(self) -> ArrayView<'a, A, D>` public, or add a `From<ArrayViewMut<'a, A, D>> for ArrayView<'a, A, D>` impl.

EbrahimEldesoky:
Fixes #1595.

Currently, ArrayViewMut lacks a public API to convert to ArrayView while preserving the lifetime 'a. This PR makes 

into_view
 public to allow consuming the mutable view into a shared one without lifetime truncation.

src/impl_views/conversions.rs
: Make 

into_view
 public

src/arraytraits.rs
: Implement From<ArrayViewMut> for ArrayView

tests/views.rs
: Add regression tests for struct-wrapped conversions


nilgoyette:
Thanks. Can you please fix the failing cicd checks?

nilgoyette:
Lgtm, thanks again. I prefer to wait for the principal maintainer @akern40  to merge this. Shouldn't take too long.

EbrahimEldesoky:
> Lgtm, thanks again. I prefer to wait for the principal maintainer @akern40  to merge this. Shouldn't take too long.

OK @nilgoyette thanks for review

akern40:
Seems sound to me, thanks for the contribution!

PR #1601, opened 2026-06-25
Add parallel iteration for exact chunks

### LINKED ISSUE #1192: Exact Chunks Parallel Iterator
Hello,

Thank you for this fantastic package.

I would like to split an `MxN` array into `k^2` chunks of size `(M/k)x(N/k)` and then process these in parallel. I'm aware of `exact_chunks`, but it doesn't natively support `par_iter`, right? What is the idiomatic way to do this?

Choudhry18:
Implement Rayon IntoParallelIterator support for ExactChunks and ExactChunksMut by wrapping their NdProducer traversal in an unindexed parallel producer. Add coverage for immutable and mutable exact chunk iteration and document the new API.

Closes #1192

### DIFF
--- a/src/parallel/mod.rs
+++ b/src/parallel/mod.rs
@@ -15,6 +15,7 @@
 //! - [`ArrayViewMut`] `.into_par_iter()`
 //! - [`AxisIter`], [`AxisIterMut`] `.into_par_iter()`
 //! - [`AxisChunksIter`], [`AxisChunksIterMut`] `.into_par_iter()`
+//! - [`ExactChunks`], [`ExactChunksMut`] `.into_par_iter()`
 //! - [`Zip`] `.into_par_iter()`
 //!
 //! The following other parallelized methods exist:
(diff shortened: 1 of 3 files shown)


nilgoyette:
This is not a critic of your code (I haven't checked), but I wonder if we actually want a `_par` version of our functions. I mean, it will double our API. However, most (all?) iterators/producers can be used with `Zip::par_*`. Are you telling me that`ExactChunks` can't?

Sorry @akern40 to disturb you :) Do you have an opinion on this `_par` matter?

Choudhry18:
`ExactChunks` can be used with `Zip::par` the comment on the issue also mentions that. My understanding was that we would want to add the `_par` versions for consistency as `AxisIter` and `AxisChunksIter` both have `_par` versions when they can also be used with `Zip::par`

akern40:
Not a bother! Sorry for the slow response - I see both sides of the argument. I would rather we get rid of all the par helpers, I think? But for now, I think consistency continues to be paramount. I'm in favor of the `_par` methods, and we will need to figure out how to factor out a more minimal API in the future.

PR #1602, opened 2026-07-05
implement dot for higher-dimensional arrays with Ix2

### LINKED ISSUE #1587: [Feature request] `ArrayRef<A, Ix2>.dot()` for axis greater than `Ix2`
In NumPy, the left-hand side of a matrix multiplication can have as many axes as desired, as long as it has more than 2 axes and the last axis's dimension matches that of the 0th axis of the right-hand side, e.g.:

```python
import numpy as np
x = np.random.random((3, 2, 5, 9, 12))
y = np.random.random((12, 13))
(x @ y).shape

# (3, 2, 5, 9, 13)
```

In ndarray, you can't do this directly:

```rust
// Doesn't compile:
use ndarray::prelude::*;
use ndarray_rand::RandomExt;
use ndarray_rand::rand_distr::Uniform;


fn main() {
    let x: Array<f64, Ix3> = Array::random(
        (12, 4, 3),
        Uniform::new(0., 1.).unwrap()
    );
    let y: Array<f64, Ix2> = Array::random(
        (3, 2),
        Uniform::new(0., 1.).unwrap()
    );
    let x_y = x.dot(&y);
    println!("{}", x_y);
}
```

<details> <summary>Compiler Output</summary>

```
$ cargo run
   Compiling playground v0.1.0 (/home/connor/RustroverProjects/playground)
error[E0275]: overflow evaluating the requirement `&ArrayBase<_, _, _>: Not`
  --> src/main.rs:15:17
   |
15 |     let x_y = x.dot(&y);
   |                 ^^^
   |
   = help: consider increasing the recursion limit by adding a `#![recursion_limit = "256"]` attribute to your crate (`playground`)
   = note: required for `&ArrayBase<_, _, _>` to implement `Not`
   = note: 127 redundant requirements hidden
   = note: required for `&ArrayBase<OwnedRepr<f64>, Dim<[usize; 3]>, f64>` to implement `Not`

For more information about this error, try `rustc --explain E0275`.
error: could not compile `playground` (bin "playground") due to 1 previous error
```

</details>

Emulating the behavior in the previous NumPy example requires a non-trivial amount of work, e.g. for `x` with 3 axes:

```rust
use ndarray::prelude::*;
use ndarray_rand::RandomExt;
use ndarray_rand::rand_distr::Uniform;


fn main() {
    let x: Array<f64, Ix3> = Array::random((12, 4, 3), Uniform::new(0., 1.).unwrap());
    let y = Array::random((3, 2), Uniform::new(0., 1.).unwrap());

    let (a, b, c) = (x.len_of(Axis(0)), x.len_of(Axis(1)), x.len_of(Axis(2)));
    let d = y.len_of(Axis(1));

    let x_y: Array3<f64> = x
        .to_shape((a * b, c)).unwrap()
        .dot(&y)
        .to_shape((a, b, d)).unwrap()
        .to_owned();
    println!("{:?}", x_y);
}
```

Therefore, I think it would be nice to have `dot()` be implemented for axis numbers greater than `Ix2`

EbrahimEldesoky:
closes #1587

right now calling `.dot()` on anything above `Ix2` doesn't work, so if you have a 3-D or 5-D array and want to multiply it by a matrix you have to manually reshape, dot, then reshape back.

this adds `Dot<ArrayRef<A, Ix2>>` for `Ix3`, `Ix4`, `Ix5`, `Ix6`, and `IxDyn`. the implementation flattens all axes except the last into a single "rows" dimension, delegates to the existing 2-D dot (which already handles BLAS), then reshapes the result back to the correct output shape.

for example:

```rust
use ndarray::prelude::*;
use ndarray::linalg::Dot;

let x = Array5::<f64>::zeros((3, 2, 5, 9, 12));
let y = Array2::<f64>::zeros((12, 13));
let z = x.dot(&y); // shape: (3, 2, 5, 9, 13)

### DIFF
--- a/src/linalg/impl_linalg.rs
+++ b/src/linalg/impl_linalg.rs
@@ -157,6 +157,22 @@ unsafe fn blas_1d_params<A>(ptr: *const A, len: usize, stride: isize) -> (*const
 ///
 /// For two-dimensional arrays, the dot method computes the matrix
 /// multiplication.
+///
+/// For higher-dimensional arrays (3-D through 6-D, and dynamic-dimensional),
+/// `Dot<Ix2>` contracts the last axis of the left-hand side with the first
+/// axis of the right-hand side, following NumPy semantics.  For example, if
+/// `self` has shape *I* × *J* × *K* and `rhs` has shape *K* × *N*, the
+/// result has shape *I* × *J* × *N*.
+///
+/// ```
+/// use ndarray::{Array3, Array2};
+/// use ndarray::linalg::Dot;
+///
+/// let a = Array3::<f64>::zeros((3, 4, 5));
+/// let b = Array2::<f64>::zeros((5, 6));
+/// let c = a.dot(&b);
+/// assert_eq!(c.shape(), &[3, 4, 6]);
+/// ```
 pub trait Dot<Rhs: ?Sized>
 {
     /// The result of the operation.
@@ -222,13 +238,58 @@ impl_dots!(Ix1, Ix2);
 impl_dots!(Ix2, Ix1);
 impl_dots!(Ix2, Ix2);
 
-fn nd_dot_non_contiguous<A, S1, S2, S3>(
-    lhs: &ArrayBase<S1, IxDyn>, rhs: &ArrayBase<S2, Ix2>, out: &mut ArrayBase<S3, IxDyn>,
+/// Compute the dot product of an N-dimensional LHS array with a 2-D RHS matrix.
+///
+/// Contracts the last axis of `lhs` with the first axis of `rhs`.
+/// If `lhs` has shape *d₀ × d₁ × … × dₙ₋₁ × K* and `rhs` has shape
+/// *K × N*, the result has shape *d₀ × d₁ × … × dₙ₋₁ × N*.
+///
+/// Two paths are used internally:
+/// - **C-contiguous LHS**: the leading axes are flattened into a single
+///   2-D view (zero-copy) and delegated to the optimised 2-D `dot`.
+/// - **Non-contiguous LHS**: a result array is pre-allocated and filled
+///   in-place via recursive `general_mat_mul` calls (no intermediate copies).
+#[track_caller]
+fn nd_dot<A: LinalgScalar>(lhs: &ArrayRef<A, IxDyn>, rhs: &ArrayRef<A, Ix2>) -> Array<A, IxDyn>
+{
+    let ndim = lhs.ndim();
+    let k = lhs.shape()[ndim - 1];
+    let k2 = rhs.shape()[0];
+    let n = rhs.shape()[1];
+    if k != k2 {
+        panic!(
+            "shapes {:?} and {:?} are not compatible for nd dot \
+             (last axis of lhs must equal first axis of rhs)",
+            lhs.shape(),
+            rhs.shape()
+        );
+    }
+
+    let mut out_shape = lhs.shape().to_vec();
+    *out_shape.last_mut().unwrap() = n;
+
+    if lhs.is_standard_layout() {
+        // C-contiguous: to_shape returns a *view* (no copy of LHS data).
+        let rows = lhs.len() / k;
+        // unwrap: rows * k == lhs.len() by construction, so reshape always succeeds.
+        let lhs_2d = lhs.to_shape((rows, k)).unwrap();
+        let result_2d = lhs_2d.dot(rhs);
+        // unwrap: result_2d is a fresh C-contiguous owned array of the correct size.
+        result_2d.into_shape_with_order(IxDyn(&out_shape)).unwrap()
+    } else {
+        // Non-contiguous: pre-allocate and fill in-place to avoid whole-array copying.
+        let mut out = Array::zeros(IxDyn(&out_shape));
+        nd_dot_non_contiguous(&lhs.view(), &rhs.view(), &mut out.view_mut());
+        out
+    }
+}
+
+/// Recursive helper: writes the N-D × 2-D product directly into `out`
+/// using `general_mat_mul` at the 2-D base case.
+fn nd_dot_non_contiguous<A>(
+    lhs: &ArrayRef<A, IxDyn>, rhs: &ArrayRef<A, Ix2>, out: &mut ArrayRef<A, IxDyn>,
 ) where
     A: LinalgScalar,
-    S1: Data<Elem = A>,
-    S2: Data<Elem = A>,
-    S3: DataMut<Elem = A>,
 {
     let ndim = lhs.ndim();
     if ndim == 2 {
@@ -237,8 +298,8 @@ fn nd_dot_non_contiguous<A, S1, S2, S3>(
         let mut out_2d = out.view_mut().into_dimensionality::<Ix2>().unwrap();
         general_mat_mul(A::one(), &lhs_2d, rhs, A::zero(), &mut out_2d);
     } else {
-        Zip::from(lhs.axis_iter(Axis(0)))
-            .and(out.axis_iter_mut(Axis(0)))
+        Zip::from(lhs.view().axis_iter(Axis(0)))
+            .and(out.view_mut().axis_iter_mut(Axis(0)))
             .for_each(|lhs_slice, mut out_slice| {
                 nd_dot_non_contiguous(&lhs_slice, rhs, &mut out_slice);
             });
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by EbrahimEldesoky** [comment]
hi @akern40 please when you have time check this pr

**Comment by nilgoyette** [comment]
One thing you could do before someone checks your PR is remove the formatting changes. It's hard[er] to know what you actually modified. We should probably format the whole project with a specific rule but it hasn't been done yet.

**Comment by EbrahimEldesoky** [comment]
> One thing you could do before someone checks your PR is remove the formatting changes. It's hard[er] to know what you actually modified. We should probably format the whole project with a specific rule but it hasn't been done yet.

> One thing you could do before someone checks your PR is remove the formatting changes. It's hard[er] to know what you actually modified. We should probably format the whole project with a specific rule but it hasn't been done yet.

Thank you @nilgoyette, Please take a look.

**Comment by nilgoyette** [comment]
I don't know how common this operation is used (I never had to use it). If it is common, then yes we should add this feature to ndarray. Is it the same behavior as numpy dot?

**Comment by EbrahimEldesoky** [comment]
> I don't know how common this operation is used (I never had to use it). If it is common, then yes we should add this feature to ndarray. Is it the same behavior as numpy dot?

Yep, this lines up exactly with how numpy's dot works — for an N-D array against a 2-D matrix, it takes the sum-product over the last axis of the first array and the first axis of the second.
It's a pattern that shows up constantly in ML/data science too, like passing batched inputs shaped (Batch, Seq, Features) through a (Features, Outputs) weight matrix. Right now you'd have to flatten, multiply, then reshape everything back manually, which gets tedious fast and is an easy place to introduce bugs

**Comment by nilgoyette** [comment]
I'm not a fan of the `reference_nd_dot` function, but I understand that not using it would force you to hardcode long and unreadable arrays.

Good job. LGTM.

**Comment by EbrahimEldesoky** [comment]
> I'm not a fan of the `reference_nd_dot` function, but I understand that not using it would force you to hardcode long and unreadable arrays.
> 
> Good job. LGTM.

thank you

File: src/linalg/impl_linalg.rs [hunk 1]
```diff
@@ -222,6 +222,85 @@ impl_dots!(Ix1, Ix2);
...
+        let k = self.shape()[ndim - 1];
+        let k2 = rhs.shape()[0];
+        let n = rhs.shape()[1];
+        if k != k2 {
+            dot_shape_error(self.len() / k, k, k2, n);
+        }
+        let rows = self.len() / k;
+        let lhs_2d = self
+            .to_shape((rows, k))
+            .expect("ndarray: to_shape failed in nd dot (IxDyn)");
```
bluss: The code has no comment that explains why `unwrap`/`expect` would be ok here.

src/linalg/impl_linalg.rs
```diff
@@ -222,6 +222,85 @@ impl_dots!(Ix1, Ix2);
...
+
+    #[track_caller]
+    fn dot(&self, rhs: &ArrayRef<A, Ix2>) -> Array<A, IxDyn>
+    {
+        let ndim = self.ndim();
+        let k = self.shape()[ndim - 1];
+        let k2 = rhs.shape()[0];
+        let n = rhs.shape()[1];
+        if k != k2 {
+            dot_shape_error(self.len() / k, k, k2, n);
```
bluss: Can we provide a better error to the user here? Their array has no dimension `self.len() / k` so they won't understand this error message.

src/linalg/impl_linalg.rs
```diff
@@ -222,6 +222,85 @@ impl_dots!(Ix1, Ix2);
...
+                let result_2d = lhs_2d.dot(rhs);
+
+                let mut out_dim = <$dim>::zeros(ndim);
+                for i in 0..ndim - 1 {
+                    out_dim[i] = self.shape()[i];
+                }
+                out_dim[ndim - 1] = n;
+
+                result_2d
+                    .to_shape(out_dim)
```
bluss: to_shape forces a copy, should be `into_shape_with_order`.

bluss (commented):
The relevant numpy documentation for `dot` is here:

https://numpy.org/doc/stable/reference/generated/numpy.dot.html

----

I'm basically not the maintainer here anymore - I think this is a good idea if we think enough about the interface and how to get the best out of performance. We should avoid whole-array-copying traps if possible.

The existing use of `to_shape` is pretty good for this, but will unfortunately cause copying in many cases.

Maybe this is a good enough compromise for now (with some improvements)?

----

In theory we could do it without any extra copying using `matrixmultiply` or blas like this:

Numpy had their example like this:

```
dot(a, b)[i,j,m] = sum(a[i,j,:] * b[:,m])
```

If axes `i`, `j` are contiguous together we can fold them together and compute as:

```
dot(a, b)[ij,m] = sum(a[ij,:] * b[:,m])
```

where `ij` is the combined axis. That's what's happening in this PR, more or less.

If axes `i`, `j` are not contiguous together (we can't reshape away that axis for free), then you could repair that by breaking up that axis in to multiple reductions, like this:

```
dot(a, b)[0,j,m] = sum(a[0,j,:] * b[:,m])  // sgemm i=0
dot(a, b)[1,j,m] = sum(a[1,j,:] * b[:,m])  // sgemm i=1
...         // and so on for the lenght of the axis
```

and then the operation still doesn't need extra copying of the ndarray. (of course both blas and matrixmultiply will copy the elements, efficiently, to their packing buffers anyway, but that's a necessary part of the matrix multiplication implementation, and usually quite optimized.)

----

Also for interest, general tensordot is already implemented for ndarray by the crate https://docs.rs/ndarray-einsum/ but I don't know its implementation. I think it actually might be a bit similar to the solution in this PR.

src/linalg/impl_linalg.rs
```diff
@@ -222,6 +222,114 @@ impl_dots!(Ix1, Ix2);
...
+                        "shapes {:?} and {:?} are not compatible for nd dot \
+                         (last axis of lhs must equal first axis of rhs)",
+                        self.shape(),
+                        rhs.shape()
+                    );
+                }
+
+                if self.is_standard_layout() {
+                    // C-contiguous: to_shape returns a *view* (no copy of LHS data).
+                    // Safety: rows * k == self.len() by construction.
```
bluss: Best to not say "safety" when it's not about `unsafe`. Thanks

src/linalg/impl_linalg.rs
```diff
@@ -222,6 +222,114 @@ impl_dots!(Ix1, Ix2);
...
+                    }
+                    out_dim[ndim - 1] = n;
+
+                    // result_2d is a fresh C-contiguous owned array;
+                    // into_shape_with_order is free.
+                    result_2d.into_shape_with_order(out_dim).unwrap()
+                } else {
+                    // Non-contiguous: iterate over the first axis so no whole-array
+                    // copy is needed. Each sub-array is (ndim-1)-D; the impl for that
+                    // dimension is already compiled (macro invocations are in order).
```
bluss: no macro here

src/linalg/impl_linalg.rs
```diff
@@ -222,6 +222,114 @@ impl_dots!(Ix1, Ix2);
...
+            // result_2d is a fresh C-contiguous owned array;
+            // into_shape_with_order is free.
+            result_2d.into_shape_with_order(IxDyn(&out_shape)).unwrap()
+        } else {
+            // Non-contiguous: iterate over the first axis so no whole-array
+            // copy is needed. Each sub-array is (ndim-1)-D IxDyn; recursion
+            // eventually reaches contiguous 2-D which terminates the recursion.
+            let sub_results: Vec<_> = self.axis_iter(Axis(0)).map(|lane| lane.dot(rhs)).collect();
+            let views: Vec<_> = sub_results.iter().map(|a| a.view()).collect();
+            crate::stack(Axis(0), &views).unwrap()
```
bluss: This line still copies the whole array (the `stack`). With the recursion over dimensions, we end up with even more copying (copying the whole ndarray multiple times).

There is a path to *no extra copy* by allocating the result and using [general_mat_mul](https://docs.rs/ndarray/latest/ndarray/linalg/fn.general_mat_mul.html) repeatedly. I'd recommend using `Zip` for the iteration if it's possible, and avoiding recursion, unless you only use non-copying operations like general_mat_mul in the recursion.

EbrahimEldesoky: Implemented the suggested refactor.

- The output array is now allocated once with `Array::zeros(...)`.
- `Zip` is used to iterate over `Axis(0)` for both the input and output arrays.
- At the 2D base case, `general_mat_mul` writes directly into the pre-allocated output slices, avoiding intermediate allocations.

This improves the `dot_3d_f64_non_contiguous` benchmark from ~72k ns/iter to ~65k ns/iter (~9% faster)

EbrahimEldesoky:
@bluss, thanks for the guidance Please take a look at the updates Is there anything else?

src/linalg/impl_linalg.rs
```diff
@@ -222,6 +222,139 @@ impl_dots!(Ix1, Ix2);
...
+                        rhs.shape()
+                    );
+                }
+
+                if self.is_standard_layout() {
+                    // C-contiguous: to_shape returns a *view* (no copy of LHS data).
+                    let rows = self.len() / k;
+                    // unwrap: rows * k == self.len() by construction, so reshape always succeeds.
+                    let lhs_2d = self.to_shape((rows, k)).unwrap();
+                    let result_2d = lhs_2d.dot(rhs);
```
bluss: This has a dependency on this place in the code https://docs.rs/ndarray/0.17.2/src/ndarray/linalg/impl_linalg.rs.html#327

What does that other code do? It picks the sense of the `result_2d` array based on the sense of the input arrays.

Test coverage will be enough to make sure these don't grow apart, I think.

There is a corner case there we can worry about, where `result_2d` can come out as F-contig from `.dot()`. It takes some more detail to look into, so we should do that together. At first glance `lhs_2d` could trip this case in some very weird circumstance, but it can maybe be ruled out.

src/linalg/impl_linalg.rs
bluss: Nice work! 65k ns/iter would be 65 µs/iter right :slightly_smiling_face: ?

EbrahimEldesoky: > Nice work! 65k ns/iter would be 65 µs/iter right :slightly_smiling_face: ?

Right

src/linalg/impl_linalg.rs
EbrahimEldesoky: I looked at the code at [line 327](https://docs.rs/ndarray/0.17.2/src/ndarray/linalg/impl_linalg.rs.html#327). This corner case is ruled out: since the contiguous path is guarded by is_standard_layout(), lhs_2d from .to_shape((rows, k)) is always C-contiguous, so lhs_s0 = k. For any non-trivial k > 1, the condition lhs_s0 == 1 && rhs_s0 == 1 is always false, meaning result_2d will always be C-contiguous. When k = 1 or rows = 1, both layouts are identical in memory anyway

src/linalg/impl_linalg.rs
```diff
@@ -222,6 +222,139 @@ impl_dots!(Ix1, Ix2);
...
+                let k = self.shape()[ndim - 1];
+                let k2 = rhs.shape()[0];
+                let n = rhs.shape()[1];
+                if k != k2 {
+                    panic!(
+                        "shapes {:?} and {:?} are not compatible for nd dot \
+                         (last axis of lhs must equal first axis of rhs)",
+                        self.shape(),
+                        rhs.shape()
+                    );
```
bluss: thanks for setting up a more helpful error message here!

bluss:
I'm mostly happy, good that we could push this in the right direction towards avoiding unnecessary copying.

Remaining things I would think of are, there is no new documentation, so users won't know that this operation is available. It should probably be documented, in brief.

And if possible I would try to reduce the code duplication, the existance of function `nd_dot_non_contiguous` is already good. I'm not fluent in ArrayRef yet, maybe it can be rewritten using ArrayRef? Maybe the new macro can generate less code and punt off to another common function?

EbrahimEldesoky:
@bluss  when you have time please, follow up here

